Issue #26, opened 2018-07-25
TxPb and Tx struct should be cleaned up


zjshen14:
Transaction is deprecated already. We removed most of the code, but not the protobuf definition as well as Tx struct. We should cleanup the unused code.

zjshen14:
Merged

PR #29, opened 2018-07-27
Extract common fields from TransferPb/VotePb to ActionPb

FishMeat:


zjshen14:
@FishMeat, talked to @dustinxie offline. We think it's better to cleanup Tx and TxPb: https://github.com/iotexproject/iotex-core/issues/26 before this issue. The reason is that after this change, TxPb (which now shares the same action definition) will own these fields too, but the fields don't make sense. Once that code is cleaned up, action will only be shared by transfer and vote

zjshen14:
@FishMeat this PR could be unblocked. Would you mind rebasing it?

FishMeat:
ok

zjshen14:
@FishMeat have you figured out how to extract common fields?

FishMeat:
below is my solution, but when I do this, it rquire too much files changed.
``` protobuf
message TransferPb {
    // used by state-based model
    bytes amount  = 4;
    string sender = 5;
    string recipient = 6;
    bytes payload  = 7;
    bytes senderPubKey = 8;
    bool isCoinbase = 9;
}

message VotePb {
    uint64 timestamp = 4;
    bytes selfPubkey = 5;
    string voterAddress = 6;  // the address of this node
    string voteeAddress = 7;  // the address this node is voting for
}

message ActionPb {
    oneof action {
        TransferPb transfer = 1;
        VotePb vote = 2;
    }

    uint32 version = 3;
    uint64 nonce = 4;
    bytes signature = 5;
}
```

zjshen14:
1. Make TransferPb's and VotePb's fields start from 1.
2. > below is my solution, but when I do this, it rquire too much files changed.
Yes I thin so. If you define an Action struct to have the three fields and make Transfer and Vote having Action struct as its sub struct, see if the code change could be smaller, though i think it will still be a lot

FishMeat:
pls see #37

Issue #31, opened 2018-07-30
We need to retry transit error from boltDB


zjshen14:
We use boltDB to persist the blockchain data. As the external service, we need to our blockchain node to be tolerant to transit failures by talking to it.

PR #39, opened 2018-08-13
add retry when boltDB update error

FishMeat:
#31


db/db.go
```diff
@@ -160,6 +174,13 @@ func (b *boltDB) PutIfNotExists(namespace string, key []byte, value []byte) erro
 		}
 		return ErrAlreadyExist
 	})
+
+	if err != nil && c < quota {
```
zjshen14: If the error is ErrAlreadyExist, i think we don't want to retry, because it's a business logic error. We only should retry on db transit issue.

db/db.go
```diff
@@ -139,18 +142,29 @@ func (b *boltDB) Stop(_ context.Context) error {
...
 		bucket, err := tx.CreateBucketIfNotExists([]byte(namespace))
 		if err != nil {
 			return err
 		}
 		return bucket.Put(key, value)
 	})
+
+	if err != nil && c < quota {
+		c++
+		goto retry
```
zjshen14: Can we use for loop instead of goto?

db/db.go
```diff
@@ -93,7 +93,10 @@ func (m *memKVStore) Batch() KVStoreBatch {
 	return NewMemKVStoreBatch(&m.data)
 }
 
-const fileMode = 0600
+const (
+	fileMode = 0600
+	quota    = 3
```
zjshen14: quota -> numRetries

And can we make it configurable?

zjshen14:
@xiaxiangzhou, please review this PR.

db/db.go
xiaxiangzhou: +1

FishMeat:
@zjshen14 ok, I make numRetries configurable

FishMeat:
@xiaxiangzhou I think when batch update failed, we can retry partial failure data. how about it?

zjshen14:
> @xiaxiangzhou I think when batch update failed, we can retry partial failure data. how about it?

I think if the batch fails, all changes in a batch will be rolled back, so that we still need to resubmit the whole batch.

db/db_batch.go
```diff
@@ -134,44 +134,59 @@ func NewBoltDBBatch(bdb *boltDB) KVStoreBatch {
...
+						errNamespace = write.namespace
+						return bolt.ErrBucketNotFound
+					}
+					if err := bucket.Delete(write.key); err != nil {
+						return errors.Wrapf(err, write.errorFormat, write.errorArgs)
+					}
 				}
 			}
+			// clear queues
+			err := b.Clear()
```
zjshen14: I think this step wouldn't be able to retry. If the batch operations are already corrupted, we should just give up.

@xiaxiangzhou a question to you, why clear throws error?

xiaxiangzhou: hmm, yeah, it should not throw errors.

db/db_batch.go
```diff
@@ -134,44 +134,59 @@ func NewBoltDBBatch(bdb *boltDB) KVStoreBatch {
...
+			if err != nil {
```
xiaxiangzhou: delete 176-178, it should not return errors.

xiaxiangzhou:
overall, LGTM. please fix those minor issues.

FishMeat:
@zjshen14 make some changes.

xiaxiangzhou:
LGTM as well

xiaxiangzhou:
go go go

zjshen14:
@FishMeat would you please resolve the conflict? Thanks

PR #52, opened 2018-09-05
Fix the duration assignment bug

zjshen14:


consensus/scheme/rolldpos/fsm.go
```diff
@@ -686,11 +686,13 @@ func (m *cFSM) produceStartRoundEvt() error {
 	// If we have the cached last block, we get the timestamp from it
```
yutongp:
```golang
var (
  duration time.Duration
  err error
)
```

yutongp: so you don't need to reassign `duration = d` in else block

PR #65, opened 2018-09-07
IOTEX-2 Support time based rotation on proposing a block

zjshen14:
For every block height, the node will check the timespan since last block, and determine if it's whose slot.

In addition, I fixed the clock usage in the consensus module.


consensus/scheme/rolldpos/fsm_test.go
```diff
@@ -385,6 +448,65 @@ func TestHandleProposeBlockEvt(t *testing.T) {
 		assert.Equal(t, ePrevoteTimeout, (<-cfsm.evtq).Type())
 	})
 
+	t.Run("invalid-proposer", func(t *testing.T) {
```
golangcibot: 451-477 lines are duplicate of `consensus/scheme/rolldpos/fsm_test.go:320-346`

PR #73, opened 2018-09-11
IOTEX-27 Let nodes panic if committing state changes fails & Add some error logs in statefactory

lizhefeng:


state/factory.go
```diff
@@ -271,24 +271,29 @@ func (sf *factory) CommitStateChanges(blockHeight uint64, tsf []*action.Transfer
 	if blockHeight > 0 && len(sf.cachedCandidates) == 0 {
 		candidates, err := sf.getCandidates(blockHeight - 1)
 		if err != nil {
+			logger.Error().Err(err).Msgf("Failed to get previous candidates on height %d", blockHeight-1)
```
zjshen14: No need to log the error separately. Because you have already captured and logged error above.

blockchain/blockchain.go
```diff
@@ -696,7 +696,7 @@ func (bc *blockchain) commitBlock(blk *Block) error {
 	if bc.sf != nil {
 		ExecuteContracts(blk, bc)
 		if err := bc.sf.CommitStateChanges(blk.Height(), blk.Transfers, blk.Votes, blk.Executions); err != nil {
-			return err
+			logger.Fatal().Err(err).Msgf("Failed to commit state changes on height %d", blk.Height())
```
zjshen14: Leave a  TODO message to say we will fix the log level in the future, once committing the block and the state become a transaction

dustinxie: we should also Fatal() if bc.dao.putBlock() != nil above?

zjshen14: IMHO, it should be fine for putBlock.If it fails, nothing else will be persisted. The blockchain will stuck at a block height, so that we could easily see.

zjshen14: But putReceipts may need, which is yet another separate batch after block is persisted

PR #79, opened 2018-09-13
IOTEX-38 Separate start lifecycle event from instantiate a blockchain struct

zjshen14:


blockchain/blockchain.go
```diff
@@ -286,19 +243,46 @@ func (bc *blockchain) Start(ctx context.Context) (err error) {
...
 	bc.mu.Lock()
 	defer bc.mu.Unlock()
+	// get blockchain tip height
 	if bc.tipHeight, err = bc.dao.getBlockchainHeight(); err != nil {
 		return err
 	}
 	if bc.tipHeight == 0 {
-		return nil
+		return bc.startEmptyBlockchain()
+	} else {
```
golangcibot: if block ends with a return statement, so drop this else and outdent its block

blockchain/blockchain.go
```diff
@@ -286,19 +243,46 @@ func (bc *blockchain) Start(ctx context.Context) (err error) {
...
 	}
-	// get blockchain tip hash
-	if bc.tipHash, err = bc.dao.getBlockHash(bc.tipHeight); err != nil {
-		return err
+}
+
+func (bc *blockchain) startEmptyBlockchain() error {
+	genesis := NewGenesisBlock(bc.config)
+	if genesis == nil {
+		return errors.New("cannot create genesis block.")
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline

blockchain/blockchain.go
```diff
@@ -729,29 +697,56 @@ func (bc *blockchain) ExecuteContractRead(ex *action.Execution) ([]byte, error)
 //======================================
 // private functions
 //=====================================
+
+func (bc *blockchain) validateBlock(blk *Block) error {
+	if bc.validator == nil {
+		logger.Panic().Msg("no block validator")
+	}
+
+	// replacement logic, used to replace a fake old dummy block
```
dustinxie: we don't have dummy block now right? why we still need this?

blockchain/blockchain.go
```diff
@@ -672,9 +641,8 @@ func (bc *blockchain) MintNewDummyBlock() *Block {
 
 //  CommitBlock validates and appends a block to the chain
 func (bc *blockchain) CommitBlock(blk *Block) error {
-	if err := bc.ValidateBlock(blk); err != nil {
```
dustinxie: i think it's OK to leave the code like this? Commit() = Validate(), then followed by private commit(), the intention/logic is quite clear

blockchain/blockchain.go
```diff
@@ -729,29 +697,56 @@ func (bc *blockchain) ExecuteContractRead(ex *action.Execution) ([]byte, error)
...
+		}
+		return bc.validator.Validate(blk, lastBlock.Height(), lastBlock.HashBlock())
+	}
+
+	return bc.validator.Validate(blk, bc.tipHeight, bc.tipHash)
+}
+
 // commitBlock commits a block to the chain
 func (bc *blockchain) commitBlock(blk *Block) error {
+	if err := bc.validateBlock(blk); err != nil {
```
dustinxie: can keep it as is, see above comment

blockchain/blockchain.go
```diff
@@ -231,51 +233,6 @@ func NewBlockchain(cfg *config.Config, opts ...Option) Blockchain {
 	if chain.sf != nil {
 		chain.lifecycle.Add(chain.sf)
 	}
-	if err := chain.Start(context.Background()); err != nil {
```
dustinxie: good clean-up to separate Start() from New(), this makes it consistent

dustinxie: ps. i had a similar change some time ago to split Start() from New() for SF and Trie, at that time noticed also need to do for Blockchain. Good to have this PR, makes codebase consistent

blockchain/blockchain.go
zjshen14: We still have dummy block logic in the code, but just disabled in production

blockchain/blockchain.go
zjshen14: I intentionally split upper case Commit/Validate and lower case commit/validate, where the former has lock protection. In general we should only lock on public methods, and I do this on purpose. All public methods should of blockchain (Start in this case) should call the lower case to prevent deadlock.

blockchain/blockchain.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/facebookgo/clock"
 	"github.com/pkg/errors"
 
+	"fmt"
```
lizhefeng: Please group imports

actpool/actpool_test.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 
+	"context"
```
lizhefeng: Please group imports

PR #101, opened 2018-09-25
IOTEX-60 Define the new address struct

zjshen14:


address/address.go
```diff
@@ -0,0 +1,178 @@
...
+	return BytesToAddress(payload)
+}
+
+// BytesToAddress converts a byte array into an address struct
+func BytesToAddress(bytes [AddressLength]byte) (Address, error) {
+	var pkHash [hash.PKHashSize]byte
+	copy(pkHash[:], bytes[5:])
+	return Address{
+		chainID: enc.MachineEndian.Uint32(bytes[:4]),
+		version: uint8(bytes[4]),
```
golangcibot: unnecessary conversion

address/address.go
```diff
@@ -0,0 +1,178 @@
...
+func IotxAddressToAddress(iotxRawAddr string) (Address, error) {
+	payload, err := decodeBech32(iotxRawAddr)
+	if err != nil {
+		return Address{}, err
+	}
+	var pkHash [hash.PKHashSize]byte
+	copy(pkHash[:], payload[5:])
+	return Address{
+		chainID: enc.MachineEndian.Uint32(payload[1:5]),
+		version: uint8(payload[0]),
```
golangcibot: unnecessary conversion

address/address.go
```diff
@@ -0,0 +1,178 @@
...
+
+// Address is the new address format to be used on IoTeX blockchain and subchains. It is composed of parts in the
+// following order:
+// 1. uint32: chain ID
+// 2. uint8: version of address
+// 3. 20 bytes: hash derived from the the public key
+type Address struct {
+	chainID uint32
+	version uint8
+	pkHash  [hash.PKHashSize]byte
```
dustinxie: can use hash.AddrHash instead of [hash.PKHashSize]byte

address/address.go
```diff
@@ -0,0 +1,56 @@
...
+
+var (
+	isTestNet bool
+	// ErrInvalidAddr indicates the invalid address error
+	ErrInvalidAddr = errors.New("Invalid address")
+)
+
+// Address defines the interface of the blockchain address
+type Address interface {
+	ChainID() uint32
```
CoderZhi: for ChainID and Version, should we define:
  type ChainIDType uint32
  type VersionType uint8
same for the other common types

address/address.go
zjshen14: Fixed

address/address.go
zjshen14: I think uint32/uint8 is already strong type, we may not need to redefine it into a more intuitive type name

PR #119, opened 2018-09-28
IOTEX-69 Get rid of config.ProducerAddr()

zjshen14:


consensus/sim/consensus_sim.go
```diff
@@ -300,3 +294,23 @@ func CombineMsg(msgType uint32, msgBody []byte) proto.Message {
 	}
 	return protoMsg
 }
+
+func getAddr(cfg *config.Config) *iotxaddress.Address {
+	addr, err := cfg.BlockchainAddress()
+	if err != nil {
+		logger.Panic().Err(err).Msg("Fail to create new consensus")
```
yutongp: "Fail to create new consensus" is not the right msg

PR #126, opened 2018-09-29
reset the boundary of 'no' decision

CoderZhi:
Say "no" number of "no" is more than 1/3 of delegates.

We could remove self notification, and jump from Prepare to eFinishEpoch if Prepare's decision is "no", which I will do in a follow up PR


consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -432,12 +432,12 @@ func (b *Builder) Build() (*RollDPoS, error) {
 		b.clock = clock.New()
 	}
 	ctx := rollDPoSCtx{
-		cfg:                    b.cfg,
-		addr:                   b.addr,
-		chain:                  b.chain,
-		actPool:                b.actPool,
-		p2p:                    b.p2p,
-		clock:                  b.clock,
+		cfg:     b.cfg,
```
golangcibot: File is not `goimports`-ed

PR #150, opened 2018-10-04
More endorse verification

CoderZhi:
1. Add endorser signature verification
2. Add height to each consensus message, and use it to determine when to use or dump an event
3. Move timeout events to Round Start
4. Unify candidates generation in rolldpos
5. Fix unit tests

### DIFF
--- a/consensus/scheme/rolldpos/fsm.go
+++ b/consensus/scheme/rolldpos/fsm.go
@@ -101,22 +101,26 @@ var (
 // iConsensusEvt is the interface of all events for the consensusEvt FSM
 type iConsensusEvt interface {
 	fsm.Event
+	Height() uint64
 	timestamp() time.Time
 	// TODO: we need to add height or some other ctx to identify which consensus round the event is associated to
 }
 
 type consensusEvt struct {
-	t  fsm.EventType
-	ts time.Time
+	height uint64
+	t      fsm.EventType
+	ts     time.Time
 }
 
-func newCEvt(t fsm.EventType, c clock.Clock) *consensusEvt {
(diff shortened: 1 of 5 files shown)


consensus/scheme/rolldpos/fsm.go
```diff
@@ -101,22 +101,26 @@ var (
 // iConsensusEvt is the interface of all events for the consensusEvt FSM
 type iConsensusEvt interface {
 	fsm.Event
+	Height() uint64
```
zjshen14: Lowercase? This method could be package only

consensus/scheme/rolldpos/fsm.go
```diff
@@ -409,17 +426,17 @@ func (m *cFSM) Start(c context.Context) error {
 					logger.Debug().Msg("timeoutEvt is stale")
 					continue
 				}
+				if evt.Height() < m.ctx.round.height {
+					logger.Debug().Msg("message of a previous round")
+					continue
+				}
+				if evt.Height() > m.ctx.round.height {
```
zjshen14: I wonder if we should requeue in this case. When we starts a new round, round is either not set, or is still the last round cache.

consensus/scheme/rolldpos/fsm.go
```diff
@@ -547,6 +564,13 @@ func (m *cFSM) handleStartRoundEvt(_ fsm.Event) (fsm.State, error) {
 		commitEndorses:   make(map[hash.Hash32B]map[string]bool),
 		proposer:         proposer,
 	}
+	// Setup timeout for waiting for proposed block
+	ttl := m.ctx.cfg.AcceptProposeTTL
```
zjshen14: We need to update config default TTL accordingly

PR #162, opened 2018-10-10
Introduce signable, signature, pubkey, verifier, signer, and certific…

CoderZhi:
Also add uint test for these interfaces

### DIFF
--- a/crypto/certificate.go
+++ b/crypto/certificate.go
@@ -1,0 +1,104 @@
+package crypto
+
+import (
+	"errors"
+)
+
+// Signable defines an interface which would generate a unique hash based on its content
+type Signable interface {
+	Hash() []byte
+}
+
+// PublicKey defines an interface of Public Key
+type PublicKey interface {
+	Bytes() []byte
+	Hex() string
+}
+
+// Signature defines an interface of
+type Signature interface {
+	Bytes() []byte
+}
+
+// Verifier defines an interface which could verify a signature
+type Verifier interface {
+	Verify(signable Signable, signature Signature) bool
+	PublicKey() PublicKey
+}
+
+// Signer defines an interface which could sign a signable
+type Signer interface {
+	PublicKey() PublicKey
+	Sign(Signable) (Signature, error)
+}
+
+// Certificate represents a certificate for a signable
+type Certificate struct {
+	signable   Signable
+	endorsers  []PublicKey
+	signatures map[string]Signature
+}
+
+// NewCertificate returns a certificate draft given a signable object
+func NewCertificate(signable Signable) (*Certificate, error) {
+	if signable == nil {
+		return nil, errors.New("Cannot create a certificate with signable as nil")
+	}
+	return &Certificate{
+		signable:   signable,
+		endorsers:  []PublicKey{},
+		signatures: make(map[string]Signature),
+	}, nil
+}
+
+// Content returns the signable content to ceritificate
+func (c *Certificate) Content() Signable {
+	return c.signable
+}
+
+// Endorse adds the endorser's signature to the certificate
+func (c *Certificate) Endorse(endorser Signer) error {
+	signature, err := endorser.Sign(c.signable)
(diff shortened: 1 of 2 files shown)


crypto/certificate.go
```diff
@@ -0,0 +1,114 @@
...
+	if signature := c.Signature(pkey); signature != nil {
+		return verifier.Verify(signature)
+	}
+
+	return false
+}
+
+// VerifiedByAll checks whether the certificate contains the signatures of all these verifiers
+// and only these verifiers
+func (c *Certificate) VerifiedByAll(verifiers []Verifier) bool {
```
yutongp: can we combine `VerifiedBy` and `VerifiedByAll` to `VerifiedBy(verifiers ...Verifier)`?

crypto/certificate.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Signature defines an interface of
+type Signature interface {
+	Bytes() []byte
+}
+
+// Verifier defines an interface which could verify a signature
+type Verifier interface {
+	Verify(signature Signature) bool
```
zjshen14: Why not Verify(sig, pk)?

crypto/certificate.go
```diff
@@ -0,0 +1,114 @@
...
+	PublicKey() PublicKey
+}
+
+// Signer defines an interface which could sign a signable
+type Signer interface {
+	Verifier
+	Sign(Signable) (Signature, error)
```
zjshen14: Sign(signable, sk)

crypto/certificate.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Verifier defines an interface which could verify a signature
+type Verifier interface {
+	Verify(signature Signature) bool
+	PublicKey() PublicKey
+}
+
+// Signer defines an interface which could sign a signable
+type Signer interface {
+	Verifier
```
zjshen14: Decouple?

crypto/certificate.go
```diff
@@ -0,0 +1,114 @@
...
+	Sign(Signable) (Signature, error)
+}
+
+// Certificate represents a certificate for a signable
+type Certificate struct {
+	signable   Signable
```
zjshen14: Instead of signable, how about hash Hash256/160

crypto/certificate.go
```diff
@@ -0,0 +1,114 @@
...
+	return nil
+}
+
+// Endorsers returns the list of endorsers of this certificate
+func (c *Certificate) Endorsers() []PublicKey {
+	return c.endorsers
+}
+
+// Signature returns the signature of the verifier saved in this certificate
+func (c *Certificate) Signature(pk PublicKey) Signature {
```
yutongp: maybe `func (c *Certificate) Signature(pk PublicKey) (Signature, bool)`?

crypto/certificate.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Signer defines an interface which could sign a signable
+type Signer interface {
+	Verifier
+	Sign(Signable) (Signature, error)
+}
+
+// Certificate represents a certificate for a signable
+type Certificate struct {
```
zjshen14: Is it better one sig/pk per Certificate?

crypto/certificate.go
CoderZhi: I assumed that the interface defined here contains the public key. What you suggested above is more like the interface of crypto algorithm

crypto/certificate.go
CoderZhi: same here

crypto/certificate.go
CoderZhi: will do

crypto/certificate.go
CoderZhi: In most cases, we will store a list of signatures for a given signable, thus the signable will be redundant data if we store one sig/pk per certificate.

PR #171, opened 2018-10-12
IOTEX-101 Add nightly build of minicluster into circleci config

lizhefeng:


zjshen14:
@lizhefeng  can you check why there's one ci/circleci check which is blocking? Related to this pr's change?

lizhefeng:
I think we need to change the setting of protected branch.

PR #204, opened 2018-10-19
IOTEX-118 Change the type of token related fields from int to string in explorer

lizhefeng:


action/execution.go
```diff
@@ -147,10 +147,10 @@ func (ex *Execution) ToJSON() (*explorer.Execution, error) {
 		Signature:      hex.EncodeToString(ex.signature),
 	}
 	if ex.amount != nil && len(ex.amount.Bytes()) > 0 {
-		execution.Amount = ex.amount.Int64()
+		execution.Amount = hex.EncodeToString(ex.amount.Bytes())
```
zjshen14: @lizhefeng why do we use hex instead of itoa?

lizhefeng: @zjshen14 Do you mean use itoa to convert int to hex string?

zjshen14: I mean e.g. 1234.5678 -> "1234.5678" instead of "0xA07E..."

PR #210, opened 2018-10-22
IOTEX-99 Delete old state when updating state factory to reduce trie storage size

dustinxie:


CoderZhi:
As we sync offline, we can improve the readability of the trie implementation, which is not urgent.

PR #218, opened 2018-10-23
create pathToRoot type and delete clear

CoderZhi:
toRoot is not supposed to be a class level parameter. Replace it with local parameter which will be returned by function query. With this implementation, we can delete function clear, and change the lock of function Get to RLock

### DIFF
--- a/trie/trie.go
+++ b/trie/trie.go
@@ -62,9 +62,8 @@ type (
 		mutex     sync.RWMutex
 		root      patricia
 		rootHash  hash.Hash32B
-		toRoot    *list.List // stores the path from root to diverging node
-		bucket    string     // bucket name to store the nodes
-		numEntry  uint64     // number of entries added to the trie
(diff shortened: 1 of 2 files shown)


trie/trie.go
```diff
@@ -286,7 +280,7 @@ func (t *trie) upsert(key, value []byte) error {
 				return err
 			}
 		} else {
-			ptr, index = t.popToRoot()
+			ptr, index = path.pop()
```
golangcibot: ineffectual assignment to `index`

trie/trie.go
```diff
@@ -527,26 +521,28 @@ func isLeaf(ptr patricia) bool {
...
-		t.toRoot.Remove(n)
-	}
+type pathToRoot struct {
+	list.List
+}
+
+func newPathToRoot() *pathToRoot {
+	path := new(pathToRoot)
+	path.Init()
+
```
dustinxie: remove this line

PR #219, opened 2018-10-23
create processEndorseEvent function

CoderZhi:
Unify the endorsement event processing with processEndorseEvent

### DIFF
--- a/consensus/scheme/rolldpos/fsm.go
+++ b/consensus/scheme/rolldpos/fsm.go
@@ -79,6 +79,8 @@ var (
 	ErrEvtCast = errors.New("error when casting the event")
 	// ErrEvtConvert indicates the error of converting the event from/to the proto message
 	ErrEvtConvert = errors.New("error when converting the event from/to the proto message")
+	// ErrEvtType represents an unexpected event type error
+	ErrEvtType = errors.New("error when check the event type")
 
 	// consensusStates is a slice consisting of all consensusEvt states
 	consensusStates = []fsm.State{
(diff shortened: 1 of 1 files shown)


consensus/scheme/rolldpos/fsm.go
```diff
@@ -482,13 +482,30 @@ func (m *cFSM) handleProposeBlockTimeout(evt fsm.Event) (fsm.State, error) {
...
+		}
+	}
+	return false
+}
+
+func (m *cFSM) validateEndorse(
+	en *endorsement.Endorsement,
+	expectedConsensusTopics map[endorsement.ConsensusVoteTopic]bool,
+) bool {
 	errorLog := logger.Error().
```
zjshen14: In this way, do we always compute the log even if no error happens?

consensus/scheme/rolldpos/fsm.go
```diff
@@ -539,18 +560,33 @@ func (m *cFSM) handleEndorseProposalEvt(evt fsm.Event) (fsm.State, error) {
...
 		map[endorsement.ConsensusVoteTopic]bool{
 			endorsement.PROPOSAL: true,
 			endorsement.COMMIT:   true, // commit endorse is counted as one proposal endorse
 		},
-		m.ctx.epoch.delegates,
 	)
-	if validNum <= len(m.ctx.epoch.delegates)*2/3 {
-		// Wait for more lock votes to come
-		return sAcceptProposalEndorse, nil
+	if panicErr != nil {
```
zjshen14: sysErr?

zjshen14: And should we define a ErrSys. processEndorseEvent only returns one error. If errors.Cause(err) == ErrSys -> blah blah

PR #220, opened 2018-10-23
Use protobuf to serialize state

dustinxie:
…member


state/account.go
```diff
@@ -9,7 +9,10 @@ package state
 import (
 	"math/big"
 
+	"github.com/golang/protobuf/proto"
+
```
golangcibot: File is not `goimports`-ed

zjshen14:
> As we discussed previously, protobuf is not designed to do serialization. If we do try to replace Gob with ProtoBuf, any concern?

It may be safe here if the serialization is not used to generate the hash, and eventually the key for state

PR #229, opened 2018-10-24
add genesis block when we start service

xiaxiangzhou:
we need to sync genesis block in index service. Because indexservice is start after block start. so indexservice will miss the first block

### DIFF
--- a/indexservice/server.go
+++ b/indexservice/server.go
@@ -62,6 +62,13 @@ func (s *Server) Start(ctx context.Context) error {
 		return errors.Wrap(err, "error when subscribe to block")
 	}
 
+	// sync genesis block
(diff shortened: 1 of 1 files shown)


indexservice/server.go
```diff
@@ -62,6 +62,13 @@ func (s *Server) Start(ctx context.Context) error {
 		return errors.Wrap(err, "error when subscribe to block")
 	}
 
+	// sync genesis block
+	genesisBlk, err := s.bc.GetBlockByHeight(0)
+	if err != nil {
+		return errors.Wrap(err, "error when get genesis block")
+	}
+	s.idx.BuildIndex(genesisBlk)
```
golangcibot: Error return value of `s.idx.BuildIndex` is not checked

PR #230, opened 2018-10-25
special handling when number of delegates is less than 4

CoderZhi:
If the number of delegates is less than 4, we request all delegates reach agreement


CoderZhi:
Will add one

PR #247, opened 2018-10-29
Use protobuf to serialize trie node

dustinxie:
1. Testing shows for Branch node, serialization size reduced 75%, for Leaf node the reduction is 50%
2. The running time of perform 2k Insertion/Deletion reduced 50%


trie/branch.go
```diff
@@ -154,29 +153,42 @@ func (b *branch) hash() hash.Hash32B {
 	return blake2b.Sum256(stream)
 }
 
+func (b *branch) toProto() *iproto.NodePb {
```
zjshen14: I prefer call it `ToProto` as we have widely use throughout the code. Later on, we could extract interface from this.

trie/branch.go
```diff
@@ -154,29 +153,42 @@ func (b *branch) hash() hash.Hash32B {
...
-	}
-	// first byte denotes the type of patricia: 1-branch, 0-ext/leaf
-	return append([]byte{BRANCH}, stream.Bytes()...), nil
+	return proto.Marshal(b.toProto())
 }
 
-// deserialize to branch
-func (b *branch) deserialize(stream []byte) error {
-	// reset variable
+func (b *branch) fromProto(pbBranch *iproto.BranchPb) {
```
zjshen14: Ditto

trie/branch.go
```diff
@@ -154,29 +153,42 @@ func (b *branch) hash() hash.Hash32B {
...
+	}
 }
 
-func (b *branch) print() {
-	for i := 0; i < RADIX; i++ {
-		if len(b.Path[i]) > 0 {
-			logger.Info().Int("k", i).Hex("v", b.Path[i]).Msg("branch")
-		}
+// deserialize to branch
+func (b *branch) deserialize(stream []byte) error {
```
zjshen14: Same as From/ToProto

trie/trie.go
```diff
@@ -73,17 +73,34 @@ type (
...
 	}
-	return newTrieSharedBatch(kvStore, batch, name, root), nil
+	t := &trie{
+		cb:        batch,
+		dao:       kvStore,
+		rootHash:  root,
+		bucket:    name,
+		numEntry:  1,
+		numBranch: 1}
+	t.lifecycle.Add(kvStore)
```
zjshen14: I doubt this is a safe operation:

1. We don't want to start/stop kv store upon each batch/block cycle
2. We don't know kvStore is started/stopped here. It's better for the creator to start/stop it.

zjshen14: In fact, I think it's never been used, because working set doesn't use lifecycle

trie/patricia_test.go
```diff
@@ -51,48 +51,47 @@ func TestPatricia(t *testing.T) {
...
 	assert.NotNil(stream)
-	assert.Equal(byte(1), stream[0])
 	b1 := branch{}
 	err = b1.deserialize(stream)
 	assert.Nil(err)
 	assert.Equal(0, bytes.Compare(root, b1.Path[0]))
 	assert.Equal(0, bytes.Compare(hash1, b1.Path[2]))
 	assert.Equal(0, bytes.Compare(hash2, b1.Path[11]))
-	assert.Equal(430, len(stream))
+	assert.Equal(114, len(stream))
```
dustinxie: size reduce to 1/4 of using Gob

trie/patricia_test.go
```diff
@@ -51,48 +51,47 @@ func TestPatricia(t *testing.T) {
...
 	err = e1.deserialize(stream)
 	assert.Nil(err)
+	assert.Equal(0, e1.Ext)
 	assert.Equal(hash1, e1.Value)
 	assert.Equal(byte(2), e1.Path[0])
 	assert.Equal(byte(3), e1.Path[1])
 	assert.Equal(byte(5), e1.Path[2])
 	assert.Equal(byte(7), e1.Path[3])
-	assert.Equal(91, len(stream))
+	assert.Equal(42, len(stream))
```
dustinxie: size reduce to 1/2 of using Gob

trie/trie_test.go
```diff
@@ -482,11 +489,11 @@ func Test2kEntries(t *testing.T) {
 	require.Nil(tr.Start(context.Background()))
 	root := EmptyRoot
 	seed := time.Now().Nanosecond()
-	// insert 2k entries
+	// insert 4k entries
```
dustinxie: in the same amount of time, can test 4k entries. meaning protobuf is 2x faster than Gob

trie/trie.go
dustinxie: yes, this can be removed, since the kvStore here must have been already added

trie/branch.go
dustinxie: after discussion, this is an internal data struct, will keep as-is, can change to public if we need to expose interface from this

trie/branch.go
```diff
@@ -7,15 +7,14 @@
 package trie
 
 import (
-	"bytes"
-	"encoding/gob"
-
+	"github.com/golang/protobuf/proto"
```
zjshen14: wrong import group

PR #248, opened 2018-10-29
IOTEX-129 Implement protocols for abstract action, transfer, vote, and execution

lizhefeng:
In this PR, action validators and handlers are moved to separate action protocols. Previous action handling logic in WorkingSet is kept in order not to break test. In next PR, this will be fully replaced by the action handlers defined in protocols and the corresponding tests will be fixed.


actpool/actpool.go
```diff
@@ -79,22 +52,31 @@ type actPool struct {
...
 		bc:          bc,
 		accountActs: make(map[string]ActQueue),
 		allActions:  make(map[hash.Hash32B]action.Action),
-		validators:  validators,
 	}
 	return ap, nil
 }
 
+// NewAbstractValidator constructs a new abstractValidator
+func NewAbstractValidator(bc blockchain.Blockchain) *abstractValidator { return &abstractValidator{bc} }
```
golangcibot: exported func NewAbstractValidator returns unexported type *actpool.abstractValidator, which can be annoying to use

action/vote/protocol.go
```diff
@@ -0,0 +1,100 @@
+package vote
+
+import (
+	"math/big"
+
+	"github.com/pkg/errors"
```
golangcibot: File is not `goimports`-ed

PR #250, opened 2018-10-30
IOTEX-87 Fix the bugs of putting sub-chain block proof

zjshen14:
The root cause is not increasing nonce of the producer account, so that the action will never be moved out of action pool. Fix includes:

1. Mutate the producer's nonce
2. Validate put block action in the action pool
3. Make producer the only node to do put block


action/subchain/putblock.go
```diff
@@ -3,7 +3,7 @@ package subchain
 import (
 	"sort"
```
yutongp: remove empty line

yutongp: remove empty line

PR #252, opened 2018-10-30
create endorsement set protobuf

CoderZhi:
the protobuf will be included in the proposed block of a second round if proof of lock is not nil


endorsement/endorsementset.go
```diff
@@ -28,6 +31,22 @@ func NewSet(blkHash hash.Hash32B) *Set {
 	}
 }
 
+// FromProtoBuf converts protobuf to endorsement set
+func (s *Set) FromProtoBuf(sPb *iproto.EndorsementSet) error {
```
dustinxie: use "FromProto" to be consistent with existing func?

endorsement/endorsementset.go
```diff
@@ -65,3 +94,17 @@ func (s *Set) NumOfValidEndorsements(topics map[ConsensusVoteTopic]bool, endorse
 
 	return cnt
 }
+
+// ToProtoBuf convert the endorsement set to protobuf
```
dustinxie: use "ToProto" to be consistent with existing func?

PR #258, opened 2018-10-31
add sendAction explorer API

yutongp:
Add a general purpose send action API which send protobuf defined action payload through JRPC.


explorer/explorer.go
```diff
@@ -1073,6 +1075,41 @@ func (exp *Service) PutSubChainBlock(putBlockJSON explorer.PutSubChainBlockReque
...
+		return explorer.SendActionResponse{}, err
+	}
+
+	var actPb pb.ActionPb
+	if err := proto.Unmarshal(payload, &actPb); err != nil {
+		return explorer.SendActionResponse{}, err
+	}
+	// broadcast to the network
+	if err = exp.p2p.Broadcast(exp.bc.ChainID(), &actPb); err != nil {
+		return explorer.SendActionResponse{}, err
```
CoderZhi: we may still process the message on broadcast failure.

explorer/explorer.go
```diff
@@ -1073,6 +1075,41 @@ func (exp *Service) PutSubChainBlock(putBlockJSON explorer.PutSubChainBlockReque
...
+	logger.Debug().Msg("receive send action request")
+
+	defer func() {
+		succeed := "true"
+		if err != nil {
+			succeed = "false"
+		}
+		requestMtc.WithLabelValues("SendAction", succeed).Inc()
+	}()
+
```
CoderZhi: do we need to check the chain id or any other parameters in the request?

PR #259, opened 2018-10-31
add action builder

yutongp:
add action builder to build an `AbstractAction`.
expose `AbstractAction`


action/builder_test.go
```diff
@@ -0,0 +1,39 @@
...
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package action
+
+import (
+	"math/big"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
golangcibot: File is not `goimports`-ed

action/builder_test.go
```diff
@@ -0,0 +1,39 @@
...
+	act := bd.SetVersion(version.ProtocolVersion).
+		SetNonce(2).
+		SetSourceAddress(srcAddr.RawAddress).
+		SetSourcePublicKey(srcAddr.PublicKey).
+		SetDestinationAddress(dstAddr.RawAddress).
+		SetGasLimit(10003).
+		SetGasPrice(big.NewInt(10004)).
+		Build()
+
+	assert.Equal(t, uint32(version.ProtocolVersion), act.Version())
```
dustinxie: change to require? we agreed to use require, existing assert should be changed to require when we modify existing codes.

PR #264, opened 2018-11-01
IOTEX-136 Create multichain dir and split mainchain and subchain protocols

zjshen14:
Also, removed the hard coded mainchain id


yutongp:
how about using `fedchain` and `fedchain/fedmain`, `fedchain/fedchild`.
so other type of subchain can be `xxxchain` and has sub package `xxxchain/xxxmain` `xxxchain/xxxchild`?

PR #265, opened 2018-11-01
Clean up locks

CoderZhi:
1. Delete ClearAndUnlock function
2. Downgrade a few locks: Lock->RLock, delete RLock
3. Add Locks to DB operations


db/db.go
```diff
@@ -101,7 +101,15 @@ func (m *memKVStore) Delete(namespace string, key []byte) error {
...
+	defer func() {
+		b.Unlock()
+		// clear the batch if commit succeeds
+		if succeed {
+			b.Clear()
+		}
+	}()
 	for i := 0; i < b.Size(); i++ {
 		write, err := b.Entry(i)
 		if err != nil {
```
CoderZhi: This is one bug for MemKVStore that the batch was not unlocked

db/db.go
```diff
@@ -181,6 +186,9 @@ func (b *boltDB) Stop(_ context.Context) error {
 
 // Put inserts a <key, value> record
 func (b *boltDB) Put(namespace string, key, value []byte) error {
+	b.mutex.Lock()
```
CoderZhi: Do not quite follow why Put doesn't have a lock. Let me know if this is on purpose.

db/db.go
```diff
@@ -312,14 +340,11 @@ func (b *boltDB) Commit(batch KVStoreBatch) (err error) {
 			break
```
CoderZhi: Should we clear the batch if the error is ErrAlreadyExist? I would say yes. So this could be a potential problem.

db/db.go
```diff
@@ -101,7 +101,15 @@ func (m *memKVStore) Delete(namespace string, key []byte) error {
 
 // Commit commits a batch
 func (m *memKVStore) Commit(b KVStoreBatch) (e error) {
+	succeed := false
 	b.Lock()
+	defer func() {
```
dustinxie: i think the entire block can be simplified to defer b.Unlock() here, no need to add succeed, see comment below

db/db.go
dustinxie: right, good catch!

db/db.go
```diff
@@ -124,14 +132,11 @@ func (m *memKVStore) Commit(b KVStoreBatch) (e error) {
 			}
 		}
 	}
-	if e != nil {
-		// if commit fails at KVStore, we do not clear the batch
-		b.Unlock()
-		return e
+	if e == nil {
+		succeed = true
```
dustinxie: call b.Clear() here

db/db.go
dustinxie: inside (db *DB) Update(), it calls db.Begin(true) --> db.beginRWTx() or db.beginTx(), inside there there is lock to protect the Update() as a transaction, so my understanding is that we don't need a lock here

db/db.go
```diff
@@ -267,7 +284,18 @@ func (b *boltDB) Delete(namespace string, key []byte) error {
 
 // Commit commits a batch
 func (b *boltDB) Commit(batch KVStoreBatch) (err error) {
+	b.mutex.Lock()
+	defer b.mutex.Unlock()
+
+	succeed := false
 	batch.Lock()
+	defer func() {
```
dustinxie: same as above, can simplify to defer batch.Unlock()

db/db.go
```diff
@@ -312,14 +340,11 @@ func (b *boltDB) Commit(batch KVStoreBatch) (err error) {
 			break
 		}
 	}
-	if err != nil {
-		// if commit fails at KVStore, we do not clear the batch
-		batch.Unlock()
-		return err
+	if err == nil {
+		succeed = true
```
dustinxie: call batch.Clear() here

trie/trie.go
```diff
@@ -120,7 +120,12 @@ func (t *trie) Start(ctx context.Context) error {
 	return t.dao.Commit(t.cb)
 }
 
-func (t *trie) Stop(ctx context.Context) error { return t.lifecycle.OnStop(ctx) }
+func (t *trie) Stop(ctx context.Context) error {
+	t.mutex.Lock()
```
dustinxie: why do we need lock here?

if needed, then we also need to lock in Start()?

db/db.go
dustinxie: and yes, this way is better than the current Unlock() + ClearAndUnlock()

db/db.go
dustinxie: i think it's ok, the purpose is that if any failure (regardless of AlreadyExist or other error) happens, the batch is not cleared for submitter to figure out what's wrong.

in our current code, if error does occur, we did not figure out or fix the batch, nor did we try re-submit. it only submit once, then the batch is never used any more.

db/db.go
CoderZhi: batch is locked at this time

PR #270, opened 2018-11-02
IOTEX-137 Define the deposit action

zjshen14:


action/deposit.go
```diff
@@ -0,0 +1,144 @@
...
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package action
+
+import (
+	"math/big"
+	"reflect"
+
+	"github.com/pkg/errors"
```
golangcibot: File is not `goimports`-ed

yutongp:
Now we can define this under it's protocol package right?

zjshen14:
> Now we can define this under it's protocol package right? With the action builder

Yes, but I'd rather defer this the one shot cleanup

PR #271, opened 2018-11-02
Remove startHeight and refine the logic of Flush function

CoderZhi:
StartHeight will always be confirmedheight+1, and there is no dummy block to replace.

### DIFF
--- a/blocksync/buffer.go
+++ b/blocksync/buffer.go
@@ -21,98 +21,70 @@ const (
 	bCheckinLower
 	bCheckinExisting
 	bCheckinHigher
+	bCheckinSkipNil
 )
 
 // blockBuffer is used to keep in-coming block in order.
 type blockBuffer struct {
-	mu              sync.RWMutex
-	blocks          map[uint64]*blockchain.Block
-	bc              blockchain.Blockchain
-	ap              actpool.ActPool
-	size            uint64
-	startHeight     uint64
-	confirmedHeight uint64
+	mu     sync.RWMutex
+	blocks map[uint64]*blockchain.Block
+	bc     blockchain.Blockchain
+	ap     actpool.ActPool
+	size   uint64
 }
 
 // Flush tries to put given block into buffer and flush buffer into blockchain.
 func (b *blockBuffer) Flush(blk *blockchain.Block) (bool, bCheckinResult) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	l := logger.With().Uint64("recvHeight", blk.Height()).Uint64("startHeight", b.startHeight).Uint64("confirmedHeight", b.confirmedHeight).Str("source", "blockBuffer").Logger()
-
-	var (
-		syncedHeight uint64
-		moved        bool
-	)
-
+	moved := false
+	if blk == nil {
+		return moved, bCheckinSkipNil
+	}
+	confirmedHeight := b.bc.TipHeight()
 	// check
 	h := blk.Height()
-	if h <= b.confirmedHeight {
+	if h <= confirmedHeight {
 		return moved, bCheckinLower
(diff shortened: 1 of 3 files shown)


blocksync/buffer.go
```diff
@@ -30,89 +30,57 @@ type blockBuffer struct {
...
 	defer b.mu.Unlock()
-	l := logger.With().Uint64("recvHeight", blk.Height()).Uint64("startHeight", b.startHeight).Uint64("confirmedHeight", b.confirmedHeight).Str("source", "blockBuffer").Logger()
-
-	var (
-		syncedHeight uint64
-		moved        bool
-	)
-
+	moved := false
 	// check
```
dustinxie: add chk-null 'h' here to make sure it is not empty?

blocksync/buffer.go
```diff
@@ -30,89 +30,57 @@ type blockBuffer struct {
...
-			return moved, bCheckinLower
-		}
-
-		b.confirmedHeight = h
-		b.startHeight = h + 1
-		l.Info().Uint64("confirmedHeight", b.confirmedHeight).Msg("replace old dummy.")
-		return true, bCheckinValid
-	}
-	if b.blocks[h] != nil {
+	if blk, ok := b.blocks[h]; ok && blk != nil {
```
dustinxie: can remove blk != nil since blk==nil wont' be put into the map, if we chk-null received block at beginning

blocksync/buffer.go
```diff
@@ -30,89 +30,57 @@ type blockBuffer struct {
...
-		} else {
-			l.Error().Err(err).Msg("Failed to commit the block")
+	l := logger.With().
+		Uint64("recvHeight", blk.Height()).
+		Uint64("confirmedHeight", b.confirmedHeight).
+		Str("source", "blockBuffer").
+		Logger()
+	for b.size > 0 {
+		next := b.confirmedHeight + 1
+		blk, ok := b.blocks[next]
```
dustinxie: can reduce to one line if blk, ok = ; ok {  b/c blk, ok already defined above

blocksync/buffer.go
```diff
@@ -30,89 +30,57 @@ type blockBuffer struct {
...
+		if !ok || blk == nil {
+			break
```
dustinxie: yes, break is better than continue 👍  if we don't have tip+1, no need to check further

blocksync/buffer.go
```diff
@@ -30,89 +30,57 @@ type blockBuffer struct {
...
 		}
-	}
-	if syncedHeight != 0 {
-		b.startHeight = syncedHeight + 1
 		moved = true
+		b.confirmedHeight = next
+		l.Info().Uint64("syncedHeight", next).Msg("Successfully committed block.")
 	}
 
 	// clean up on memory leak
```
dustinxie: i am wondering we actually don't need this check? since blocks < tip and > tip + b.size are discarded in Flush(), map can only contain blocks from tip to tip + b.size

blocksync/buffer.go
```diff
@@ -135,12 +103,13 @@ func (b *blockBuffer) GetBlocksIntervalsToSync(targetHeight uint64) []syncBlocks
...
 
-	if targetHeight >= b.startHeight+b.size {
-		targetHeight = b.startHeight + b.size - 1
+	if targetHeight > b.confirmedHeight+b.size {
+		targetHeight = b.confirmedHeight + b.size
 	}
 
 	for h := b.confirmedHeight + 1; h <= targetHeight; h++ {
-		if b.blocks[h] == nil {
+		blk, ok := b.blocks[h]
```
dustinxie: can reduce to one line if blk, ok = ; ok {

PR #273, opened 2018-11-02
IOTEX-137 Implement the deposit balance to the sub-chain protocol

zjshen14:


action/protocols/multichain/mainchain/protocol.go
```diff
@@ -11,13 +11,13 @@ import (
 
 	"github.com/pkg/errors"
 
+	"fmt"
```
CoderZhi: group imports

PR #275, opened 2018-11-02
IOTEX-137 Make explorer accept deposit request

zjshen14:
I removed the hardcode address/pk/sk btw.


zjshen14:
cc @puncsky deposit api

PR #280, opened 2018-11-03
IOTEX-137 Make it possible to query deposit into a sub-chain from explorer

zjshen14:


explorer/explorer.go
```diff
@@ -14,8 +14,12 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"fmt"
```
lizhefeng: Group imports

explorer/explorer_test.go
```diff
@@ -20,8 +20,12 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"strings"
```
lizhefeng: Ditto

PR #281, opened 2018-11-03
IOTEX-138 Refactor vote protocol and add unit tests

lizhefeng:


action/protocols/vote/protocol.go
```diff
@@ -7,18 +7,27 @@
...
-const VoteSizeLimit = 278
+const (
+	// VoteSizeLimit is the maximum size of vote allowed
+	VoteSizeLimit = 278
+	// CandidatesPrefix is the prefix of the key of candidateList
+	CandidatesPrefix = "Candidates."
+)
 
 // Protocol defines the protocol of handling votes
 type Protocol struct{ bc blockchain.Blockchain }
```
dustinxie: we don't need bc here. bc.StateByAddr() can be replaced by ws.State()

dustinxie: Also i think we can put map[hash.PKHash]*state.Candidate into this struct, then we don't need pass the map around as input param below

action/protocols/vote/protocol.go
```diff
@@ -104,3 +162,60 @@ func (p *Protocol) Validate(act action.Action) error {
 	}
 	return nil
 }
+
+func (p *Protocol) constructKey(height uint64) hash.PKHash {
+	heightInBytes := byteutil.Uint64ToBytes(height)
+	k := []byte(CandidatesPrefix)
+	k = append(k, heightInBytes...)
+	hashedKey := hash.Hash160b(k)
```
dustinxie: can simplify to byteutil.BytesTo20B(hash.Hash160b(k))

action/protocols/vote/protocol.go
```diff
@@ -104,3 +162,60 @@ func (p *Protocol) Validate(act action.Action) error {
...
+		candidateList, ok := candidateState.(*state.CandidateList)
+		if !ok {
+			return nil, fmt.Errorf("error when casting %T state into candidateList", candidateState)
+		}
+		return state.CandidatesToMap(*candidateList)
+	}
+	return nil, err
+}
+
+func (p *Protocol) addCandidate(candidateMap map[hash.PKHash]*state.Candidate, vote *action.Vote, height uint64) error {
```
dustinxie: add map into Protocol struct{}, then no need to pass in map as first param

action/protocols/vote/protocol.go
```diff
@@ -104,3 +162,60 @@ func (p *Protocol) Validate(act action.Action) error {
...
+			Address:        vote.Voter(),
+			PublicKey:      votePubkey,
+			CreationHeight: height,
+		}
+	}
+	return nil
+}
+
+func (p *Protocol) updateCandidate(
+	candidateMap map[hash.PKHash]*state.Candidate,
```
dustinxie: same as above

action/protocols/vote/protocol.go
```diff
@@ -32,46 +41,95 @@ func (p *Protocol) Handle(act action.Action, ws state.WorkingSet) error {
...
-		voteFrom.IsCandidate = true
-		if err := ws.UpdateCachedCandidates(vote); err != nil {
-			return errors.Wrap(err, "failed to update cached candidates in working set")
-		}
+	sort.Sort(candidateList)
+	candidatesKey := p.constructKey(ws.Height())
+	if err := ws.PutState(candidatesKey, &candidateList); err != nil {
+		return errors.Wrap(err, "failed to put updated candidates to trie")
 	}
 	return nil
```
dustinxie: clear the map before exit

action/protocols/vote/protocol.go
```diff
@@ -32,46 +41,95 @@ func (p *Protocol) Handle(act action.Action, ws state.WorkingSet) error {
 	if !ok {
 		return nil
 	}
-	voteFrom, err := ws.LoadOrCreateAccountState(vote.Voter(), big.NewInt(0))
+	// Get candidateList from trie and convert it to candidates map
```
dustinxie: here should be same as workingset.go code now:

if ws.Height() > 0 {
    candidates, err := ws.getCandidates(blockHeight - 1)
    ....
}

PR #286, opened 2018-11-06
Add config file for starting fullnode

dustinxie:


e2etest/config_local_fullnode.yaml
```diff
@@ -0,0 +1,29 @@
+# go-yaml expects the YAML field corresponding to a struct field to be lowercase. So if your struct field is
+# UpdateInterval, the corresponding field in YAML is updateinterval.
+
+nodeType: "full_node"            # should be one of "delegate", "full_node", and "lightweight"
+
+network:
+    host: "127.0.0.1"
+    port: 4689
+    bootstrapNodes:
+        - "35.230.110.31:30333"
```
zjshen14: 127.0.0.1 instead as 35.230.110.31:30333 may not exist at all

PR #288, opened 2018-11-06
Add Delete function to SortedSlice

CoderZhi:


CoderZhi:
maybe in a follow up pr or derive a new struct

PR #301, opened 2018-11-08
IOTEX-143 Make sub-chain start from an empty genesis block

zjshen14:
1. Clean up the code of handling genesis block state mutation
2. Relax the assumption that candidates always exist
3. The code to check if we should use empty genesis block is not refined. There requires a future code refactor work to make genesis block creation more coherent

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -293,46 +293,39 @@ func (bc *blockchain) Start(ctx context.Context) (err error) {
 }
 
 func (bc *blockchain) startEmptyBlockchain() error {
-	ctx := context.Background()
-	genesis := NewGenesisBlock(bc.config)
-	if genesis == nil {
-		return errors.New("cannot create genesis block")
-	}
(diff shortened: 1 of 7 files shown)


blockchain/blockchain.go
```diff
@@ -293,46 +293,39 @@ func (bc *blockchain) Start(ctx context.Context) (err error) {
...
-	}
-	if err := bc.sf.Commit(ws); err != nil {
-		return errors.Wrap(err, "failed to add Creator into StateFactory")
-	}
-	// run execution and update state trie root hash
-	root, err := bc.runActions(genesis, ws, false)
-	if err != nil {
-		return errors.Wrap(err, "failed to update state changes in Genesis block")
+	var genesis *Block
+	if bc.config.Chain.GenesisActionsPath != "" || !bc.config.Chain.EmptyGenesis {
```
lizhefeng: Can we remove the first condition here?

PR #327, opened 2018-11-16
fix block sync bug

CoderZhi:
Assume the block to sync is h, confirmed block is c. If c+1 is in buffer and there is a block in (c+1, h) not in buffer, target height won't get set.

### DIFF
--- a/blocksync/buffer.go
+++ b/blocksync/buffer.go
@@ -37,21 +37,20 @@ type blockBuffer struct {
 func (b *blockBuffer) Flush(blk *blockchain.Block) (bool, bCheckinResult) {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	moved := false
 	if blk == nil {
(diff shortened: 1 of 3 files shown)


blocksync/buffer.go
```diff
@@ -37,47 +37,45 @@ type blockBuffer struct {
...
 	l := logger.With().
 		Uint64("recvHeight", blk.Height()).
 		Uint64("confirmedHeight", confirmedHeight).
 		Str("source", "blockBuffer").
 		Logger()
-	for b.size > 0 {
-		next := confirmedHeight + 1
-		blk, ok := b.blocks[next]
+	syncedHeight := confirmedHeight
+	for confirmedHeight+b.size > syncedHeight {
```
yutongp: `for syncedHeight := confirmedHeight; confirmedHeight+b.size > syncedHeight;syncedHeight++`

Issue #251, opened 2018-10-30
Annoying put block failure log


zjshen14:
We have two code path that a block could be appended to the blockchain: 1) consensus module commits a block after reaching consensus; 2) block sync receives a block from peers. The race condition could happen between the two path. However, it should not matter if a block has been put successfully already, and the second put fails because the block exists already. We need to catch the error on both side and check the error type. If the error is that block exists already, we shouldn't log an error

PR #341, opened 2018-11-22
remove error log when block already exist blockchain

FishMeat:
#251


blockchain/blockchain.go
```diff
@@ -678,6 +681,9 @@ func (bc *blockchain) MintNewSecretBlock(
 func (bc *blockchain) CommitBlock(blk *Block) error {
 	bc.mu.Lock()
 	defer bc.mu.Unlock()
+	if b, _ := bc.GetBlockByHash(blk.HashBlock()); b != nil {
```
zjshen14: This may not work perfectly. Between L686 and L687, if there's a context switch, the db will still throw error. I don't think we need to define another error. Instead, if we see the cause of the error is `db.ErrAlreadyExist`, we should just ignore it, and not print error message.

consensus/consensus.go
```diff
@@ -8,6 +8,7 @@ package consensus
 
 import (
 	"context"
+	"github.com/iotexproject/iotex-core/db"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

blocksync/buffer.go
```diff
@@ -7,6 +7,7 @@
 package blocksync
 
 import (
+	"github.com/iotexproject/iotex-core/db"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

consensus/scheme/rolldpos/fsm.go
```diff
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	"github.com/iotexproject/iotex-core/db"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

FishMeat:
make some changes

blocksync/buffer.go
```diff
@@ -65,7 +66,7 @@ func (b *blockBuffer) Flush(blk *blockchain.Block) (bool, bCheckinResult) {
 			break
 		}
 		delete(b.blocks, heightToSync)
-		if err := commitBlock(b.bc, b.ap, blk); err != nil {
+		if err := commitBlock(b.bc, b.ap, blk); err != nil && err != db.ErrAlreadyExist {
```
zjshen14: if err == db.ErrAlreadyExist, we should also not say "Successfully committed block". Maybe ether skip the log or say "Already exists"

FishMeat:
If If block already exists,  ignore the err and log info msg

consensus/scheme/rolldpos/fsm.go
```diff
@@ -684,7 +685,12 @@ func (m *cFSM) handleEndorseCommitEvt(evt fsm.Event) (fsm.State, error) {
 		}
 	}
 	// Commit and broadcast the pending block
-	if err := m.ctx.chain.CommitBlock(pendingBlock); err != nil {
+	err := m.ctx.chain.CommitBlock(pendingBlock)
+	if err == db.ErrAlreadyExist {
+		err = nil
+		logger.Info().Uint64("block", pendingBlock.Height()).Msg("Block already exists.")
+	}
+	if err != nil {
```
zjshen14:
```suggestion
	if err := m.ctx.chain.CommitBlock(pendingBlock); err != nil {
		if err == db.ErrAlreadyExist {
```

blocksync/buffer.go
FishMeat: emm, looks better

consensus/scheme/rolldpos/fsm.go
FishMeat: emm, looks better

consensus/scheme/rolldpos/fsm.go
```diff
@@ -685,10 +686,15 @@ func (m *cFSM) handleEndorseCommitEvt(evt fsm.Event) (fsm.State, error) {
 	}
 	// Commit and broadcast the pending block
 	if err := m.ctx.chain.CommitBlock(pendingBlock); err != nil {
-		logger.Error().
-			Err(err).
-			Uint64("block", pendingBlock.Height()).
-			Msg("error when committing a block")
+		if err == db.ErrAlreadyExist {
+			err = nil
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

blocksync/buffer.go
```diff
@@ -66,6 +67,11 @@ func (b *blockBuffer) Flush(blk *blockchain.Block) (bool, bCheckinResult) {
 		}
 		delete(b.blocks, heightToSync)
 		if err := commitBlock(b.bc, b.ap, blk); err != nil {
+			if err == db.ErrAlreadyExist {
```
zjshen14: Sorry, I just double check , and i think the right logic here is
```suggestion
			if err == db.ErrAlreadyExist {
				l.Info().Uint64("syncHeight", heightToSync).Msg("Block already exists.")
			} else {
				l.Error().Err(err).Uint64("syncHeight", heightToSync).Msg("Failed to commit the block.")
				break
			}
```

And then, we don't need L75-L81

PR #342, opened 2018-11-23
IOTEX-157 Make blocksync tentatively sync tip height + 1 block

zjshen14:
If a node has temporary outage, but is unfortunately be chosen as the proposer during this period, there's no chance for the whole network to recover, because no flying broadcasted block exist, and the node wouldn't able to proactively catchup the missing blocks.

This PR makes the node to tentatively request tip height + 1 block every 10 block sync interval, so that it could recover and catch up other delegates to continue consensus. This will not be very efficient if a node miss tons of blocks, but a working solution for short-term outage.

### DIFF
--- a/blocksync/blocksync.go
+++ b/blocksync/blocksync.go
@@ -18,6 +18,7 @@ import (
 	"github.com/iotexproject/iotex-core/network"
 	"github.com/iotexproject/iotex-core/network/node"
 	"github.com/iotexproject/iotex-core/pkg/lifecycle"
+	"github.com/iotexproject/iotex-core/pkg/routine"
 	pb "github.com/iotexproject/iotex-core/proto"
 )
 
@@ -41,6 +42,7 @@ type blockSyncer struct {
 	worker         *syncWorker
 	bc             blockchain.Blockchain
 	p2p            network.Overlay
+	chaser         *routine.RecurringTask
 }
 
 // NewBlockSyncer returns a new block syncer instance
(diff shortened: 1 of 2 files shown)


blocksync/blocksync_test.go
```diff
@@ -443,10 +443,36 @@ func newTestConfig() (config.Config, error) {
...
+	bs, err := NewBlockSyncer(cfg, chain, ap, network.NewOverlay(cfg.Network))
+	require.NoError(err)
+	require.NoError(bs.Start(ctx))
+
+	defer func() {
+		require.NoError(chain.Stop(ctx))
+		require.NoError(bs.Stop(ctx))
+	}()
+
+	testutil.WaitUntil(100*time.Millisecond, 10*time.Second, func() (bool, error) {
```
golangcibot: Error return value of `testutil.WaitUntil` is not checked

blocksync/blocksync.go
```diff
@@ -85,12 +89,18 @@ func (bs *blockSyncer) P2P() network.Overlay {
...
+	if err := bs.chaser.Start(ctx); err != nil {
+		return err
+	}
 	return bs.worker.Start(ctx)
 }
 
 // Stop stops a block syncer
 func (bs *blockSyncer) Stop(ctx context.Context) error {
 	logger.Debug().Msg("Stopping block syncer")
+	if err := bs.chaser.Stop(ctx); err != nil {
```
CoderZhi: do we have any rule of when to the order of start/stop modules? As bs exends lifecycle, we may add it to bs via Add

PR #346, opened 2018-11-29
refactor the options and start function of factory

CoderZhi:
Move the redundant code to Start function

### DIFF
--- a/state/factory/factory.go
+++ b/state/factory/factory.go
@@ -78,17 +78,8 @@ func PrecreatedTrieDBOption(kv db.KVStore) Option {
 		if kv == nil {
 			return errors.New("Invalid empty trie db")
 		}
-		if err = kv.Start(context.Background()); err != nil {
-			return errors.Wrap(err, "failed to start trie db")
-		}
 		sf.dao = kv
-		// get state trie root
-		if sf.rootHash, err = sf.getRoot(trie.AccountKVNameSpace, AccountTrieRootKey); err != nil {
-			return errors.Wrap(err, "failed to get accountTrie's root hash from underlying DB")
-		}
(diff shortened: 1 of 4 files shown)


state/factory/factory.go
```diff
@@ -173,8 +143,19 @@ func NewFactory(cfg config.Config, opts ...Option) (Factory, error) {
 func (sf *factory) Start(ctx context.Context) error {
 	sf.mutex.Lock()
 	defer sf.mutex.Unlock()
+	var err error
+	if err = sf.lifecycle.OnStart(ctx); err != nil {
```
zjshen14: Do you need to remove L126-L128? Otherwise, sf.accountTrie will be started as well by this call?

CoderZhi: addressed

state/factory/factory.go
```diff
@@ -143,19 +143,8 @@ func NewFactory(cfg config.Config, opts ...Option) (Factory, error) {
...
-	// get state trie root
-	if sf.rootHash, err = sf.getRoot(trie.AccountKVNameSpace, AccountTrieRootKey); err != nil {
-		return errors.Wrap(err, "failed to get accountTrie's root hash from underlying DB")
-	}
-	if sf.accountTrie, err = trie.NewTrie(sf.dao, trie.AccountKVNameSpace, sf.rootHash); err != nil {
-		return errors.Wrap(err, "failed to generate accountTrie from config")
-	}
 
-	return sf.accountTrie.Start(ctx)
+	return sf.lifecycle.OnStart(ctx)
```
zjshen14: So the order doesn't matter?

CoderZhi: The order doesn't matter as long as we handle the dependency properly. In our previous implementation, the root hash was stored in both sf and trie, which results in the messy of dependency.

PR #349, opened 2018-12-02
Fix one bug of cached batch

CoderZhi:
Two bugs found in cached batch:
1. If we use CachedBatch and DB at the same time to achieve Get, we may read wrong data from DB
2. PutIfNotExists only check the existence in CachedBatch, but not DB

This PR tries to solve the first bug.

For the second bug, we may need further discussion.


db/db_batch.go
```diff
@@ -228,16 +238,21 @@ func (cb *cachedBatch) Clear() {
...
 func (cb *cachedBatch) Get(namespace string, key []byte) ([]byte, error) {
 	cb.lock.RLock()
 	defer cb.lock.RUnlock()
-	if v, ok := cb.cache[cb.hash(namespace, key)]; ok {
+	h := cb.hash(namespace, key)
+	if v, ok := cb.cache[h]; ok {
 		return v, nil
 	}
+	if _, ok := cb.deleted[h]; ok {
+		return nil, ErrAlreadyDeleted
```
zjshen14: Should we just return ErrNotExist? Make it seamless as no data pollution

CoderZhi: ok

PR #354, opened 2018-12-05
IOTEX-163 Return an empty account if the requested account does not exist in trie

lizhefeng:


state/account.go
```diff
@@ -23,6 +23,12 @@ var (
 	ErrAccountCollision = errors.New("account already exists")
 )
 
+// EmptyAccount indicates an empty account
+var EmptyAccount = &Account{
```
golangcibot: `EmptyAccount` is a global variable (from `gochecknoglobals`)

action/protocol/account/protocol.go
```diff
@@ -54,18 +54,16 @@ func LoadOrCreateAccountState(sm protocol.StateManager, addr string, init *big.I
 		return nil, errors.Wrap(err, "failed to convert address to public key hash")
 	}
 	account, err := LoadAccountState(sm, addrHash)
-	switch {
-	case errors.Cause(err) == state.ErrStateNotExist:
-		account := state.Account{
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get account of %x from account trie", addrHash)
+	} else if account == state.EmptyAccount {
```
dustinxie: no need for else

state/factory/workingset.go
```diff
@@ -92,25 +92,20 @@ func NewWorkingSet(
...
 	if err != nil {
-		return nil, err
-	}
-	s, err := ws.CachedState(addrHash, &state.Account{})
-	switch {
-	case errors.Cause(err) == state.ErrStateNotExist:
-		account := state.Account{
-			Balance:      big.NewInt(0).SetBytes(init.Bytes()),
+		return nil, errors.Wrapf(err, "failed to get account of %s from cached account", addr)
+	} else if account == state.EmptyAccount {
```
dustinxie: no need for else

state/factory/workingset.go
```diff
@@ -92,25 +92,20 @@ func NewWorkingSet(
...
-		ws.cachedStates[addrHash] = &account
-		return &account, nil
-	case err != nil:
-		return nil, errors.Wrapf(err, "failed to get account of %x from cached account", addrHash)
-	}
-	account, err := stateToAccountState(s)
-	if err != nil {
-		return nil, err
+		addrHash, err := iotxaddress.AddressToPKHash(addr)
+		if err != nil {
```
dustinxie: no need to check error here
it is already checked in CachedAccountState() above

state/account.go
dustinxie: can you add a comment that this is a read-only var for comparison
caller should not modify it

dustinxie: also move it to the var() group above

PR #359, opened 2018-12-07
IOTEX-168 Implement Snapshot and Revert for CachedBatch interface

dustinxie:


db/db_batch.go
```diff
@@ -153,6 +165,22 @@ func (b *baseKVStoreBatch) Clear() {
...
+// CloneBatch clones the batch
+func (b *baseKVStoreBatch) CloneBatch() KVStoreBatch {
+	b.mutex.Lock()
+	defer b.mutex.Unlock()
+
+	c := baseKVStoreBatch{
+		writeQueue: make([]writeInfo, b.Size()),
+	}
+	// clone the writeQueue
+	for i, entry := range b.writeQueue {
```
golangcibot: should use copy() instead of a loop (from `megacheck`)

PR #386, opened 2018-12-20
Fix memory thrash after enabling Snapshot()

dustinxie:


zjshen14:
@dustinxie let's follow up on why the references to the block are not fully released

PR #390, opened 2018-12-21
Snapshot using slice to reduce memory

dustinxie:


state/factory/factory_test.go
```diff
@@ -83,7 +83,15 @@ func TestSnapshot(t *testing.T) {
 	require.Equal(big.NewInt(7), s.Balance)
 	s2 := ws.Snapshot()
 	require.Equal(2, s2)
+	s.AddBalance(big.NewInt(6))
```
golangcibot: Error return value of `s.AddBalance` is not checked (from `errcheck`)

CoderZhi: address the comment before merging

PR #369, opened 2018-12-13
implement empty interface of statedb

CoderZhi:

PR #391, opened 2018-12-23
Implement refund interface

CoderZhi:


CoderZhi:
This is based on #369

PR #392, opened 2018-12-25
Revert "Disable Snapshot() function before solving the memory thrash/…

dustinxie:
…leak (#384)"

This reverts commit f07f66cebeaf294b4483f0024c5d27d2f4b78ba9.


dustinxie:
the memory thrash/leak has been fixed last Friday, enable Snapshot() functionality so it can be monitored in nightly run
someone please help merge this PR

dustinxie:
only 7 nodes are calling snapshots is a different issue we observed, it would still be there regardless the feature is enabled or not

PR #417, opened 2019-01-03
fix unit tests race and enable race detection flag

yutongp:


pkg/routine/recurringtask_test.go
```diff
@@ -31,10 +35,9 @@ func TestRecurringTask(t *testing.T) {
 	ck := clock.NewMock()
 	task := routine.NewRecurringTask(h.Do, 100*time.Millisecond, routine.WithClock(ck))
 	task.Start(ctx)
-	defer func() {
-		task.Stop(ctx)
-	}()
-
 	ck.Add(600 * time.Millisecond)
+	task.Stop(ctx)
```
golangcibot: Error return value of `task.Stop` is not checked (from `errcheck`)

PR #421, opened 2019-01-04
embed header in block; move hdr's getters to header struct

yutongp:


blockchain/block/header.go
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) 2018 IoTeX
```
zjshen14: 2019

PR #427, opened 2019-01-06
IOTEX-197 Harden the check of commiting the pending state changes

zjshen14:
1. In production, block's working set is not supposed to be nil, and height shouldn't mismatch. Error will prompt if it happens.
2. Panic on committing state changes now to speed up state corruption debug
3. FIx a bug introduced from refactoring genesis block creation


blockchain/blockchain.go
```diff
@@ -922,13 +922,13 @@ func (bc *blockchain) startEmptyBlockchain() error {
 		PublicKey:  keypair.ZeroPublicKey,
 		RawAddress: addr.IotxAddress(),
 	}
+	if bc.sf == nil {
```
raullenchai: We have unit test for startEmptyBlockchain() IIRC. Need to add test coverage for this.

zjshen14: 2chain e2e tests have covered this regression already

PR #429, opened 2019-01-07
IOTEX-197 Fix the missing src address in the broadcast message

zjshen14:
In addition, fixing the racing of accepting broadcast/unicast message even when the agent is not fully started.

Test plan: alter the unit tests to cover the changes


p2p/agent_test.go
```diff
@@ -66,7 +66,8 @@ func TestBroadcast(t *testing.T) {
 		require.NoError(t, testutil.WaitUntil(100*time.Millisecond, 10*time.Second, func() (bool, error) {
 			mutex.RLock()
 			defer mutex.RUnlock()
-			return counts[uint8(i)] == n, nil
+			// Broadcast message will be skipped by the source node
+			return counts[uint8(i)] == n-1, nil
```
golangcibot: Using the variable on range scope `i` in function literal (from `scopelint`)

p2p/agent_test.go
```diff
@@ -121,7 +124,7 @@ func TestUnicast(t *testing.T) {
 		require.NoError(t, testutil.WaitUntil(100*time.Millisecond, 10*time.Second, func() (bool, error) {
 			mutex.RLock()
 			defer mutex.RUnlock()
-			return counts[uint8(i)] == n-1, nil
+			return counts[uint8(i)] == n-1 && src == agents[i].Self().String(), nil
```
golangcibot: Using the variable on range scope `i` in function literal (from `scopelint`)

p2p/agent.go
```diff
@@ -75,6 +75,7 @@ func NewAgent(cfg config.Network, broadcastHandler HandleBroadcast, unicastHandl
 
 // Start connects into P2P network
 func (p *Agent) Start(ctx context.Context) error {
+	ready := make(chan interface{})
```
raullenchai: the declaration could move to line 94.

p2p/agent.go
```diff
@@ -92,8 +93,15 @@ func (p *Agent) Start(ctx context.Context) error {
 	}
 
 	if err := host.AddBroadcastPubSub(broadcastTopic, func(data []byte) (err error) {
+		// Blocking handling the broadcast message until the agent is started
+		<-ready
```
raullenchai: where is the wring to "ready"?

zjshen14: It is used to block the call back function to move forward. The goroutine will be blocked here if the callback is triggered, until L178 `ready` is closed

p2p/agent.go
zjshen14: I intentionally add it as the first line in this func, in case in the future we added more resources into start lifecycle, and will potentially be used before fully started

PR #430, opened 2019-01-07
IOTEX-197 Prevent skipping validation if the block's working set is nil

zjshen14:


consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -175,8 +175,9 @@ func (ctx *rollDPoSCtx) validateProposeBlock(blk Block, expectedProposer string)
 		errorLog.Msg("error when validating the block signature")
 		return false
 	}
-	if producer == ctx.addr.RawAddress {
-		// If the block is self proposed, skip validation
+	// TODO: in long term, block in process and after process should be represented differently
+	if producer == ctx.addr.RawAddress && block.WorkingSet != nil {
```
raullenchai: Can we log if this actually happens.

PR #438, opened 2019-01-09
fix another data race in dispatcher

yutongp:


dispatcher/dispatcher.go
```diff
@@ -288,13 +288,16 @@ func (d *IotxDispatcher) HandleBroadcast(chainID uint32, message proto.Message)
...
 		logger.Warn().
 			Uint32("chainID", chainID).
 			Msg("chainID has not been registered in dispatcher")
+		d.subscribersMU.RUnlock()
 		return
 	}
+	d.subscribersMU.RUnlock()
 
 	switch msgType {
 	case pb.MsgConsensusType:
```
raullenchai: should RWLock subscriber in L304?

yutongp: no. the lock is on the map

PR #445, opened 2019-01-11
IOTEX-219 Put all receipts in a block as a single payload

zjshen14:
This is to optimize the write of committing a block at the cost of higher reading receipts cost.

Test plan: unit test + minicluster. Will update the real deployment result


blockchain/blockdao.go
```diff
@@ -19,6 +20,7 @@ import (
 	"github.com/iotexproject/iotex-core/pkg/hash"
 	"github.com/iotexproject/iotex-core/pkg/lifecycle"
 	"github.com/iotexproject/iotex-core/pkg/util/byteutil"
+	"github.com/iotexproject/iotex-core/proto"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

PR #453, opened 2019-01-15
IOTEX-205 Full migration from iotxaddress to address V1

lizhefeng:
This PR changes all the dependencies on iotxaddress to address V1. The former raw iotxaddress is replaced by Bech32 encoded string address, and key pair  is no longer a part of address. Iotxaddress package and functions for backward compatibility are cleaned up.


raullenchai:
> _No description provided._

Please provide details in PR description.

crypto/ec283.go
```diff
@@ -22,6 +22,7 @@ import (
 	"encoding/binary"
 	"errors"
 
+	"fmt"
```
raullenchai: remove

PR #457, opened 2019-01-17
mute AlreadyRegisteredError

CoderZhi:
This is a harmless error.

Tested with uint tests: The error won't be returned or printed


pkg/prometheustimer/timer.go
```diff
@@ -44,6 +44,10 @@ func New(name, tip string, labelNames []string, defaultLabels []string) (*TimerF
 		labelNames,
 	)
 	err := prometheus.Register(vect)
+	switch err.(type) {
```
golangcibot: singleCaseSwitch: should rewrite switch statement to if statement (from `gocritic`)

PR #459, opened 2019-01-17
IOTEX-228 Add chainID to createDeposit to remove the dependency of address on chainID

lizhefeng:
In mainchain protocol, when validating a createDeposit action, we need to get the ID of chain where the deposit goes. Since we want to remove chainID from address, the chain ID can be obtained from action message itself instead of recipient address.


zjshen14:
cc @puncsky this will break explorer

PR #462, opened 2019-01-19
IOTEX-230 Add benchmark test for signature verifycation

lizhefeng:


raullenchai:
please add description.

crypto/benchmark_test.go
```diff
@@ -0,0 +1,41 @@
+package crypto
+
+import (
+	"testing"
+
+	"github.com/CoderZhi/go-ethereum/crypto"
```
raullenchai: coderzhi?

lizhefeng: go-ethereum is imported by coderzhi when evm is introduced.

PR #464, opened 2019-01-20
IOTEX-219 Disable writing index data into chain db by default

zjshen14:


zjshen14:
Verified that the commit time was reduced to 0.0x sec in the last two days.

PR #470, opened 2019-01-24
IOTEX-241 Fix the receipt root and verify it during consensus

zjshen14:
1. Fix receipt inconsistency on different nodes
2. Merklize receipts and put it into the block header
3. Verify receipt root

Test plan: unit test + minicluster run

It is supposed to add O(K) where k is # of Tx in a block, but it only does one serialization and hash per TX. Probably it is just fine. Will update the experiment results later


blockchain/blockchain.go
```diff
@@ -1177,3 +1184,15 @@ func (bc *blockchain) emitToSubscribers(blk *block.Block) {
 }
 
 func (bc *blockchain) now() int64 { return bc.clk.Now().Unix() }
+
+func calculateReceiptRoot(receipts []*action.Receipt) hash.Hash32B {
+	var h []hash.Hash32B
```
golangcibot: Consider preallocating `h` (from `prealloc`)

PR #483, opened 2019-01-26
print endorsement details when validation failed

CoderZhi:
1. Print endorsement details on block footer validation failure
2. Don't print error if the block height is lower than requested block sync height, and update unit test


CoderZhi:
merged in #487

PR #486, opened 2019-01-27
IOTEX-232 Make sure action sender is the action owner

lizhefeng:
This PR fixes a security hole: we didn't check whether the action source address is actually the action owner. If an attacker uses other's encoded address as sender address while signing the action using his owner private key, the signature is still valid but the victim will be charged gas fee.


explorer/explorer_test.go
```diff
@@ -193,7 +193,7 @@ func addActsToActPool(ap actpool.ActPool) error {
 	if err != nil {
 		return err
 	}
-	vote1, err := testutil.SignedVote(ta.Addrinfo["producer"].Bech32(), ta.Addrinfo["producer"].Bech32(), ta.Keyinfo["alfa"].PriKey, 3, testutil.TestGasLimit, big.NewInt(testutil.TestGasPrice))
+	vote1, err := testutil.SignedVote(ta.Addrinfo["producer"].Bech32(), ta.Addrinfo["producer"].Bech32(), ta.Keyinfo["producer"].PriKey, 3, testutil.TestGasLimit, big.NewInt(testutil.TestGasPrice))
```
golangcibot: line is 194 characters (from `lll`)

PR #492, opened 2019-01-29
add a prefix to contract trie hash func

CoderZhi:
make test && make minicluster


action/protocol/execution/evm/contract_test.go
```diff
@@ -241,7 +241,12 @@ func TestSnapshot(t *testing.T) {
 	k2 := byteutil.BytesTo32B(hash.Hash160b([]byte("dog")))
 	v2 := byteutil.BytesTo32B(hash.Hash256b([]byte("dog")))
 
-	c1, err := newContract(s, db.NewMemKVStore(), db.NewCachedBatch())
+	c1, err := newContract(
```
raullenchai: Should have a test to delpoying the same bytecode twice and see how they work?

PR #501, opened 2019-01-31
IOTEX-251 Remove coinbase transfer and modify the affected protocols

lizhefeng:
Since the block rewarding mechanism will be implemented differently in producer protocol, there is no need for the existence of coinbase transfer.


explorer/explorer.go
```diff
@@ -1813,7 +1807,7 @@ func convertTsfToExplorerTsf(selp action.SealedEnvelope, isPending bool) (explor
 		Fee:        "", // TODO: we need to get the actual fee.
 		Payload:    hex.EncodeToString(transfer.Payload()),
 		GasLimit:   int64(selp.GasLimit()),
-		IsCoinbase: transfer.IsCoinbase(),
+		IsCoinbase: false,
```
zjshen14: We should remove IsCoinbase from explorer data model definition too

PR #509, opened 2019-02-04
Add cryptokey interface and change action.srcPubkey to bytes to reduc…

dustinxie:
…e dependency

1. added a new interface for public/private key
2. data objects like block/action/header does not need to know the specific type of crypto key (ECDSA, or RSA, etc.), passing the key's serialized bytes makes more sense
3. at the time of signing or verifying, the key is extracted from serialized bytes, and Sign() or Verify() is called

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -9,12 +9,11 @@ package action
 import (
 	"math/big"
 
-	"github.com/iotexproject/go-ethereum/crypto"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/blake2b"
 
+	"github.com/iotexproject/iotex-core/crypto/key"
 	"github.com/iotexproject/iotex-core/pkg/hash"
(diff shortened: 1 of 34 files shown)


action/action.go
```diff
@@ -257,16 +252,20 @@ func (sealed *SealedEnvelope) LoadProto(pbAct *iproto.ActionPb) error {
 }
 
 // Sign signs the action using sender's private key
-func Sign(act Envelope, addr string, sk keypair.PrivateKey) (SealedEnvelope, error) {
+func Sign(act Envelope, addr string, prvk []byte) (SealedEnvelope, error) {
 	sealed := SealedEnvelope{Envelope: act}
 
+	sk, err := key.NewPrivateKeyFromBytes(prvk)
```
dustinxie: extract the key from serialized bytes, then call Sign()

raullenchai:
As discussed offline, we could close this pr.

PR #512, opened 2019-02-05
IOTEX-255 Implement protocol registry service (WIP)

zjshen14:
Protocol registry is a service to register protocols to be installed on the blockchain. Itself is implemented as protocol too.


zjshen14:
After more investigation, I prefer to defer this work as the scope is way bigger than I expected.

PR #513, opened 2019-02-05
Make state recover from empty if the recovery height is lower than the current state height

lizhefeng:
Previously we only support recovering state db to a specific height which is lower than the blockchain height if and only if the existing state db height is lower than the target height. Now the state db can recover to any lower target height by clearing the existing state db and rebuilding it from scratch.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -9,6 +9,7 @@ package blockchain
 import (
 	"context"
 	"math/big"
+	"os"
 	"strconv"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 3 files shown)


blockchain/blockchain.go
```diff
@@ -958,13 +958,16 @@ func (bc *blockchain) startExistingBlockchain(recoveryHeight uint64) error {
 		}
 		startHeight = 1
 	}
-	if recoveryHeight > 0 && startHeight <= recoveryHeight {
+	if recoveryHeight > 0 {
```
raullenchai: pls add unit test to cover this case.

blockchain/blockchain_test.go
```diff
@@ -1126,6 +1126,12 @@ func TestStartExistingBlockchain(t *testing.T) {
 	height, _ = chain.sf.Height()
 	require.Equal(bc.TipHeight(), height)
 	require.True(3 == height)
+
+	// recover to height 2 from an existing state DB with Height 3
+	require.NoError(chain.startExistingBlockchain(2))
+	height, _ = chain.sf.Height()
+	require.Equal(bc.TipHeight(), height)
+	require.True(2 == height)
```
golangcibot: don't use Yoda conditions (from `stylecheck`)

raullenchai:
> _No description provided._

description needed

blockchain/blockchain_test.go
```diff
@@ -1098,7 +1098,7 @@ func TestStartExistingBlockchain(t *testing.T) {
 	}()
 
 	require.NoError(addTestingTsfBlocks(bc))
-	require.True(5 == bc.TipHeight())
+	require.True(bc.TipHeight() == 5)
```
raullenchai: require.Equal

blockchain/blockchain_test.go
```diff
@@ -1125,7 +1125,13 @@ func TestStartExistingBlockchain(t *testing.T) {
 	require.NoError(chain.startExistingBlockchain(3))
 	height, _ = chain.sf.Height()
 	require.Equal(bc.TipHeight(), height)
-	require.True(3 == height)
+	require.True(height == 3)
```
raullenchai: require.Equal

PR #516, opened 2019-02-05
bump libp2p versions

yutongp:
unit tested
not breaking anything


raullenchai:
> _No description provided._

Need test plan, e.g, does this break anything for us?

PR #518, opened 2019-02-05
fix the issue sever cannot be stopped by control-c

yutongp:


yutongp:
> Do we want graceful shutdown?

yes we do

PR #523, opened 2019-02-07
IOTEX-257 Avoid referring the empty account pointer

zjshen14:
Previously we refer to the same empty account struct by pointers. This is so risky to introduce bug accidentally. This PR is trying to replacing it by creating a separate empty account copy every time we try to load it.

Verified with unit test and minicluster test.


zjshen14:
Additionally I added one account API called `Recorded` for execution protocol to use, because in its use case, it want to actually check if an account is persisted into the ledger or not. This is sort of hack, because poor isolation of account and execution code. At long term, execution should have its separate store/load contract methods.

PR #549, opened 2019-02-15
IOTEX-258 Make pending action queue always check against the current confirmed nonce

zjshen14:
Previously we had the race that block's committed, but the actpool is not reset, so that processed actions will be picked again. The root cause is that action queue caches a start nonce, which could be stale after committing a block. If we kept using the current confirmed nonce, it should be able to avoid the problem. By doing this, we don't need to validate actions again when minting the block.


actpool/actqueue.go
```diff
@@ -12,6 +12,9 @@ import (
 	"sort"
 	"time"
 
+	"github.com/iotexproject/iotex-core/pkg/log"
```
lizhefeng: Group imports

actpool/actqueue_test.go
```diff
@@ -12,6 +12,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/mock/gomock"
+	"github.com/iotexproject/iotex-core/test/mock/mock_blockchain"
+
+	"github.com/iotexproject/iotex-core/config"
```
lizhefeng: Ditto

zjshen14:
Checked in a real cluster with 100tps. It seems to work fine.

PR #551, opened 2019-02-15
IOTEX-247 Remove stateRoot and chainID from block header

lizhefeng:
This is an explorer breaking change.


explorer/idl/explorer.idl
```diff
@@ -38,10 +38,8 @@ struct Block {
     executions int
     generateBy BlockGenerator
     amount string
```
raullenchai: should deprecate api v1 after this?

proto/types/blockchain.proto
```diff
@@ -17,17 +17,14 @@ import "google/protobuf/timestamp.proto";
 // header of a block
 message BlockHeader {
   uint32 version = 1;
-  uint32 chainID = 2;
```
raullenchai: Note that this introduce a breaking change to the protocol, meaning this version doesn't work with previous ones. At least we should label this as a protocol-breaking change.

explorer/idl/explorer.idl
lizhefeng: We can label it as a breaking change first. Explorer on master branch is already broken before this PR. We can deprecate it now to avoid further changes on explorer backend.

PR #557, opened 2019-02-17
Remove per action gas limit cap

zjshen14:
Remove per action gas limit cap as we already calculate the actual gas consumption per block, and restrict the total gas consumption by block gas limit. By doing this, there's no need to setup a per action gas limit if we think the block gas limit is affordable. And the gas consumption per action various quite a lot. Some big contract deployment will actually a lot.

### DIFF
--- a/action/protocol/execution/evm/evm.go
+++ b/action/protocol/execution/evm/evm.go
@@ -87,7 +87,7 @@ func NewParams(raCtx protocol.RunActionsCtx, execution *action.Execution, stateD
 		BlockNumber: new(big.Int).SetUint64(raCtx.BlockHeight),
 		Time:        new(big.Int).SetInt64(raCtx.BlockTimeStamp),
 		Difficulty:  new(big.Int).SetUint64(uint64(50)),
-		GasLimit:    raCtx.ActionGasLimit,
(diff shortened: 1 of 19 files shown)


action/protocol/execution/evm/evm.go
```diff
@@ -87,7 +87,7 @@ func NewParams(raCtx protocol.RunActionsCtx, execution *action.Execution, stateD
 		BlockNumber: new(big.Int).SetUint64(raCtx.BlockHeight),
 		Time:        new(big.Int).SetInt64(raCtx.BlockTimeStamp),
 		Difficulty:  new(big.Int).SetUint64(uint64(50)),
-		GasLimit:    raCtx.ActionGasLimit,
+		GasLimit:    execution.GasLimit(),
```
zjshen14: @CoderZhi Please confirm if this change makes sense

raullenchai: +1

zjshen14:
> But for launching the mainnet, is it better to have this limit as a protection to reject crazy actions?

Sounds fair, will crate another PR to fix the evm gas limit setting bug only

PR #558, opened 2019-02-18
add epoch data api

CoderZhi:
This will be used in election. This PR defines the api request and response format, and return a fake data before implementing details.


api/api.go
```diff
@@ -219,6 +219,17 @@ func (api *Server) GetChainMeta(ctx context.Context, in *iotexapi.GetChainMetaRe
 	return &iotexapi.GetChainMetaResponse{ChainMeta: chainMeta}, nil
 }
 
+// GetEpochData is the API to get epoch data
+func (api *Server) GetEpochData(ctx context.Context, in *iotexapi.GetEpochDataRequest) (res *iotexapi.GetEpochDataResponse, err error) {
```
golangcibot: line is 136 characters (from `lll`)

proto/types/blockchain.proto
```diff
@@ -63,12 +63,19 @@ message CandidateList {
   repeated Candidate candidates = 1;
 }
 
+message EpochData {
+  uint64 num = 1;
+  uint64 height = 2;
```
zjshen14: Is it the start height of each block, or current block height?

CoderZhi: it is the start height, same as we defined in epochCtx

PR #560, opened 2019-02-18
IOTEX-247 Remove block hash from run action context

zjshen14:
Easy PR


yutongp:
can we remove this? I think many action rely on the knowledge of block hash. but right now we just don't use the block hash in run action context. instead we are using the block hash in abstract action which should be deleted.

PR #593, opened 2019-02-23
IOTEX-270 Improve rewarding protocol

zjshen14:
3 improvements:

1. Uniform the receipt status code conform with evm's
2. Use snapshot to protect states being polluted
3. Trim the number of candidates to grant epoch reward


action/protocol/rewarding/fund.go
```diff
@@ -118,6 +115,10 @@ func (p *Protocol) assertEnoughBalance(
 
 // DepositGas deposits gas into the rewarding fund
 func DepositGas(ctx context.Context, sm protocol.StateManager, amount *big.Int, registry *protocol.Registry) error {
+	// If the gas fee is 0, return immediately
+	if amount.Cmp(big.NewInt(0)) == 0 {
```
CoderZhi: should we handle negative case?

action/protocol/rewarding/protocol_test.go
```diff
@@ -11,6 +11,10 @@ import (
 	"math/big"
 	"testing"
 
+	"github.com/iotexproject/iotex-core/action"
+
+	"github.com/iotexproject/iotex-core/test/identityset"
+
```
CoderZhi: delete empty lines

action/protocol/rewarding/fund.go
zjshen14: negative amount are asserted in the rewarding protocol

PR #598, opened 2019-02-23
Simplify config and make docker image runnable with default command

zjshen14:
1. Remove the node type config:
a. Delegate and full node shouldn't be different on system setup but only one is elected as delegate and the other isn't.
b. We don't have light node yet, and how it would be different will be different from current arbitrary difference.
2. Make docker image default command runnable by adjusting a few configs, paths and etc.


zjshen14:
Verified in the cluster that it's working

PR #606, opened 2019-02-23
IOTEX-259 Try to make consensus unit tests less likely fail

zjshen14:
I know it makes sense to have debug level for unit tests, but our debug level logging are too verbose to be useful, and make the test run further slower.

Also tune some parameters. Hopefully the tests could be less fragile. Let's see.


zjshen14:
13 straight ci job success. The sporadic test failure pain should be relieved.

PR #609, opened 2019-02-24
IOTEX-282 fix minicluster no-receipt failure

zcchai-iotex:
The fix is to check all the servers are ready before deploying contract, instead of checking only the first server as before fix.
The issue is hard to reproduce in my local PC. So the code needs to be verified on CircleCI environment.

### DIFF
--- a/tools/minicluster/minicluster.go
+++ b/tools/minicluster/minicluster.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"math"
 	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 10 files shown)


tools/minicluster/minicluster.go
```diff
@@ -94,26 +95,31 @@ func main() {
...
 	for i := 0; i < numNodes; i++ {
 		probeSvrs[i] = probe.New(7788 + i)
 	}
-	err = probeSvrs[0].Start(context.Background())
-	if err != nil {
-		log.L().Error("Failed to start probe server")
+	for i := 0; i < numNodes; i++ {
+		err = probeSvrs[i].Start(context.Background())
+		if err != nil {
+			log.L().Error("Failed to start probe server")
```
zjshen14: Panic instead?

zcchai-iotex: Yes, will update with log.L().Panic

PR #621, opened 2019-02-26
Avoid flooding of calibrate events

zjshen14:
First of all, this fix is hacky, and we need to look for a more appropriate strategy of handling block height being updated when running consensus.

Anyway, this PR is trying to prevent calibrate events from flooding, especially when a node are syncing bunch of blocks from peers. When the node is already at prepare stage, there's no sense to calibrate it any more.


zjshen14:
Verified the change wouldn't break consensus by running a cluster over night

PR #629, opened 2019-02-27
IOTEX-278 Implement local transfer test

zcchai-iotex:
Add local transfer test cases:
-Simple 1 to 1 transfer test cases, enumerating combination of balance, nonce, amount, gas
-1 to n: only add a test case to verify transfer will be handled by the order of increasing nonce (a missing nonce will hold the rest transfers in action pool)

To test, naviate to ./e2etest, then:
/e2etest$ go test -run TestLocalTransfer


e2etest/local_transfer_test.go
```diff
@@ -0,0 +1,379 @@
...
+			uint64(200000), big.NewInt(1),
+			TsfFail, "Transfer with nonce 0",
+		},
+		{
+			identityset.PrivateKey(0), big.NewInt(100000),
+			nil, big.NewInt(100000),
+			1, big.NewInt(100),
+			make([]byte, 4),
+			uint64(200000), big.NewInt(1),
+			TsfSuccess, "Transfer with same nouce from a single sender 1",
```
zjshen14: nonce

e2etest/local_transfer_test.go
```diff
@@ -0,0 +1,379 @@
...
+		},
+		{
+			identityset.PrivateKey(0), big.NewInt(100000),
+			nil, big.NewInt(100000),
+			1, big.NewInt(100),
+			make([]byte, 4),
+			uint64(200000), big.NewInt(1),
+			TsfFail, "Transfer with same nouce from a single sender 2",
```
zjshen14: nonce, and check the spell bellow

PR #633, opened 2019-02-27
filter invalid operator address and reward address

CoderZhi:
filter delegates with wrong operator address or reward address


zjshen14:
Shall we stop invalid address right after the source than before the sink?

PR #639, opened 2019-02-28
Remove pk from consensus module

zjshen14:
Test plan: unit tests + minicluster run


raullenchai:
make minicluster failed

zjshen14:
The minicluster failure isn't related

PR #642, opened 2019-02-28
IOTEX-289 Handle actions with block height passed from context

lizhefeng:


xiaxiangzhou:
Thanks! @lizhefeng , LGTM, leave for @zjshen14  to accept

PR #648, opened 2019-03-01
IOTEX-291 Using sha3 instead of blake

zjshen14:
Will test it in the cluster and update here


pkg/hash/hash.go
```diff
@@ -24,17 +26,17 @@ type (
 
 // Hash160b returns 160-bit (20-byte) hash of input
 func Hash160b(input []byte) Hash160 {
-	// use Blake2b algorithm
-	digest := blake2b.Sum256(input)
+	// use sha3 algorithm
+	digest := crypto.Keccak256(input)
 	var hash Hash160
 	copy(hash[:], digest[7:27])
```
dustinxie: i think Ethereum use the latest 20-bytes of hash, that is digest[12:] as the address

action/protocol/execution/protocol_test.go
```diff
@@ -505,7 +504,8 @@ func TestProtocol_Handle(t *testing.T) {
 		require.Equal(0, balance.Cmp(big.NewInt(500000000)))
 
 		log.S().Info("Roll Dice")
-		data, _ = hex.DecodeString("797d9fbd000000000000000000000000d950673630e2286adc157c35f2fd73a1ef49d40e")
+		h := keypair.HashPubKey(testaddress.Keyinfo["alfa"].PubKey)
```
dustinxie: we need to change the HashPubKey() implementation, according to https://kobl.one/blog/create-full-ethereum-keypair-and-address/
1. digest = keccak-256(hexdecimal format of the pubkey, not bytes)
2. ethereum address = digest[12:]
i am checking ethereum code to confirm

zjshen14: good catch

zjshen14:
verified in a real cluster

PR #677, opened 2019-03-06
IOTEX-276 Make polling pending actions get rid of expired actions and refactor actpool code

zjshen14:
Two things:

1. Make polling pending actions get rid of expired actions, so that if there's a bad action, it will eventually be removed and the block production will not be blocked forever.

2. Cleanup unused action pool API and config


zjshen14:
Verified that with 30tps, reset action pool op will take about 0.01sec, which should be fine to put into the block production cycle.

PR #682, opened 2019-03-06
Add BytesToHash256/160()

dustinxie:
small clean-up


action/protocol/multichain/subchain/protocol.go
```diff
@@ -146,11 +144,3 @@ func (p *Protocol) mutateDeposit(ctx context.Context, deposit *action.SettleDepo
 func depositAddress(index uint64) hash.Hash160 {
 	return hash.Hash160b([]byte(fmt.Sprintf("depositToSubChain.%d", index)))
 }
-
```
dustinxie: this func is not used, so remove it

PR #718, opened 2019-03-11
IOTEX-315 Cleanup minicluster DB files after successful completion

zcchai-iotex:


tools/minicluster/minicluster.go
```diff
@@ -12,7 +12,9 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/iotexproject/iotex-core/pkg/util/fileutil"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

zjshen14: +1

PR #732, opened 2019-03-13
IOTEX-331 Bounce up protoc-gen-go to 1.3.0 and regenerate the files

zjshen14:


zjshen14:
> How do we handle breaking changes like this in the future?

We will forbid the upgrade until next big release, which allows incompatible changes.

PR #740, opened 2019-03-14
Remove admin and its operations from rewarding protocol

zjshen14:


action/protocol/rewarding/admin.go
```diff
@@ -187,15 +148,6 @@ func (p *Protocol) BlockReward(
 	return a.blockReward, nil
 }
 
-// SetBlockReward sets the block reward amount for the block rewarding. Only the current admin could make this change
-func (p *Protocol) SetBlockReward(
```
raullenchai: Should we keep this one? My thinking is, when time comes, we will have a logic in minting a new block:
```
if (tip height == 50000000 && rewarding params != new rewarding params) {
  SetBlockReward(new rewarding params)
}
```

zjshen14: SetBlockReward has some logic that depends on admin to exist. It should not be hard when we want to hard fork and overwrite the block/epoch amount

PR #741, opened 2019-03-14
Move rpc message type definitions into proto file

zjshen14:


proto/rpc/rpc.proto
```diff
@@ -29,3 +29,29 @@ message Consensus {
   google.protobuf.Timestamp timestamp = 4;
   bytes data = 5;
 }
+
+enum MessageType {
+  UNKNOWN = 0;
+  Action = 1;
```
yutongp: why only `Action` is first char cap, others are all CAP

PR #746, opened 2019-03-14
IOTEX-307 Use LRU cache to keep some blocks in memory

zjshen14:
Block is immutable data, it should be safe to cache in memory upon the first request. The cache size is set to the number of blocks per epoch


blockchain/blockdao.go
```diff
@@ -46,22 +53,37 @@ var (
 	actionToPrefix   = []byte("to.")
 )
 
-var _ lifecycle.StartStopper = (*blockDAO)(nil)
```
zjshen14: @yutongp can you comment if this line is necessary

blockchain/blockdao.go
```diff
@@ -46,22 +53,37 @@ var (
...
+var (
+	cacheMtc = prometheus.NewCounterVec(
```
golangcibot: `cacheMtc` is a global variable (from `gochecknoglobals`)

blockchain/blockdao.go
```diff
@@ -35,6 +39,8 @@ const (
 	receiptsNS                       = "rpt"
 
 	hashOffset = 12
+
+	maxCacheSize = 10000
```
raullenchai: This can be a config field? There will be three cases:
- nodes that will not provide API service, cache is unneeded
- nodes with limited memory + provide API service, set maxCacheSize to something the memory can afford
- nodes need realtime data (e.g., for debugging), set maxCacheSize to 0 to disable it.

blockchain/blockdao_test.go
```diff
@@ -397,3 +407,68 @@ func TestBlockDao_putReceipts(t *testing.T) {
 		assert.Equal(t, receipt.ActHash, r.ActHash)
 	}
 }
+
+func BenchmarkBlockCache(b *testing.B) {
+	test := func(cacheSize int, b *testing.B) {
+		b.StopTimer()
+		path := filepath.Join(os.TempDir(), fmt.Sprintf("test-%d.db", rand.Int()))
```
golangcibot: G404: Use of weak random number generator (math/rand instead of crypto/rand) (from `gosec`)

zjshen14:
pkg: github.com/iotexproject/iotex-core/blockchain
BenchmarkBlockCache/cache-8         	   20000	     54903 ns/op
BenchmarkBlockCache/no-cache-8      	   10000	    125880 ns/op

Note that there's only 20000 ops. If more the results, will be much better

blockchain/blockdao.go
```diff
@@ -139,6 +155,16 @@ func (dao *blockDAO) getBlockHeight(hash hash.Hash256) (uint64, error) {
 
 // getBlock returns a block
 func (dao *blockDAO) getBlock(hash hash.Hash256) (*block.Block, error) {
+	if dao.cache != nil {
+		dao.cacheMutex.RLock()
```
CoderZhi: we may create three private functions to get, insert, and delete cacheBlock, with "defer dao.cacheMutex.RUnlock/Unlock()"

PR #751, opened 2019-03-15
IOTEX 336 Fix TestLocalTransfer flaky  panic on CircleCI build

zcchai-iotex:
Original TestLocalTransfer uses 1 second as the block produce interval, which is too aggressive for CircleCI server.
So this fix is to increase the block produce interval to  2 seconds.

### DIFF
--- a/e2etest/local_transfer_test.go
+++ b/e2etest/local_transfer_test.go
@@ -496,7 +496,7 @@ func newTransferConfig(
 	cfg.Chain.EnableAsyncIndexWrite = true
 	cfg.Consensus.Scheme = config.StandaloneScheme
 	cfg.API.Port = apiPort
-	cfg.Genesis.BlockInterval = 1 * time.Second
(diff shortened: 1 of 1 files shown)


e2etest/local_transfer_test.go
```diff
@@ -373,21 +374,22 @@ func TestLocalTransfer(t *testing.T) {
...
+			fmt.Println("PASS")
 		case TsfPending:
 			//Need to wait long enough to make sure the pending transfer is not minted, only stay in action pool
 			time.Sleep(cfg.Genesis.BlockInterval + time.Second)
 			_, err := ap.GetActionByHash(tsf.Hash())
 			require.NoError(err, tsfTest.message)
 			_, err = bc.GetActionByActionHash(tsf.Hash())
 			require.Error(err, tsfTest.message)
+			fmt.Println("PASS")
 		case TsfFinal:
```
dustinxie: print diff msg like "PASS TsfPending", "PASS TsfFinal", would that help debug?

Issue #716, opened 2019-03-11
CI for building on Ubuntu


raullenchai:
We already have circle ci build/test for macos and docker (debian). Need to build/test on ubuntu as well.

feigu1984:
circlie ci  latest  ubuntu 14.04, default golang 1.9  ;  ./go.test.sh
```
# github.com/iotexproject/iotex-core/vendor/github.com/multiformats/go-multiaddr
vendor/github.com/multiformats/go-multiaddr/codec.go:140:10: undefined: strings.Builder
vendor/github.com/multiformats/go-multiaddr/component.go:118:8: undefined: strings.Builder
vendor/github.com/multiformats/go-multiaddr/component.go:125:32: undefined: strings.Builder
FAIL	github.com/iotexproject/iotex-core/api [build failed]
Exited with code 2
```
1,  I'm not sure if it's a code problem?
2,  I'll try to update  golang > 1.11.5

feigu1984:
when golang update 1.11.6   ./go.test.sh  working
found out   last macos  
`go test github.com/iotexproject/iotex-core/action/protocol/vote/`

> protocol.go:81:15: Wrapf format %d has arg voteFrom.Balance of wrong type *math/big.Int

use go 1.12    change 1.11.5/6   ok

raullenchai:
@feigu1984 right, golang ≥ 1.11.5 is required.

raullenchai:
@lzxm160 pls dig out why unit test always fail under ubuntu -- low priority tho

lzxm160:
> @lzxm160 pls dig out why unit test always fail under ubuntu -- low priority tho

sure

lzxm160:
I tested with ubuntu 16.04 and golang 1.12.5,there's no error found.I also build a docker image ,and added it to config.yml.
https://github.com/iotexproject/iotex-core/pull/1170,this is the test demo.
I think my docker repo need auth,maybe someone can help me build it under docker account iotex

lzxm160:
The ubuntu ci test used default resouce_class medium,need add "resource_class: large" to config.yaml
Too few resources caused test results error

PR #756, opened 2019-03-16
CI for building on Ubuntu #716

feigu1984:
1, add building Ubuntu 
2, fix macos build "protocol.go:81:15: Wrapf format %d has arg voteFrom.Balance of wrong type *math/big.Int"
3,  Modify  describe


feigu1984:
only  build_test_ubuntu can be pass  
https://github.com/iotexproject/iotex-core/pull/756/commits/1982d7530b989da79f211f66564409a29394ebaf

PR #776, opened 2019-03-19
fix two bugs in consensus

CoderZhi:
1. When mint block, pass block creation timestamp as the raCtx blocktimestamp
2. On consensus reached, if commit fails, log error and early return


blockchain/blockchain.go
```diff
@@ -466,7 +466,7 @@ func (bc *blockchain) MintNewBlock(
 	ctx := protocol.WithRunActionsCtx(context.Background(),
 		protocol.RunActionsCtx{
 			BlockHeight:    newblockHeight,
-			BlockTimeStamp: bc.now(),
```
raullenchai: remove L1075.

PR #796, opened 2019-03-21
fix producing lock endorsement on nil

CoderZhi:
Lock endorsement shouldn't be generated if block hash is nil

### DIFF
--- a/consensus/scheme/rolldpos/rolldposctx.go
+++ b/consensus/scheme/rolldpos/rolldposctx.go
@@ -300,17 +300,19 @@ func (ctx *rollDPoSCtx) NewLockEndorsement(
 	case ErrInsufficientEndorsements:
 		return nil, nil
 	case nil:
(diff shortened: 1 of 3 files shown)


consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -300,17 +300,20 @@ func (ctx *rollDPoSCtx) NewLockEndorsement(
...
+		if len(blkHash) != 0 {
+			ctx.loggerWithStats().Info("Locked", log.Hex("block", blkHash))
+			return ctx.newEndorsement(
+				blkHash,
+				LOCK,
+				ctx.round.StartTime().Add(
+					ctx.cfg.FSM.AcceptBlockTTL+ctx.cfg.FSM.AcceptProposalEndorsementTTL,
+				),
+			)
+		} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

PR #799, opened 2019-03-21
Fix delegate cmd output for ioctl

Frankonly:


raullenchai:
> _No description provided._

pls provide samples

PR #808, opened 2019-03-22
IOTEX-355 Return action hash and block hash with the action core from API

zjshen14:


proto/api/api.proto
```diff
@@ -105,8 +105,14 @@ message GetActionsByBlockRequest {
   uint64 count = 3;
 }
 
+message Action {
```
puncsky: it feels strange to access iotextypes.Action  like getActionsRespons.actions[n].action

I suggest naming it `MintedAction` or `ActionResult` ... or other kinds of action you can name it

Frankonly:
ioctl is ok.

zjshen14:
Updated per comments

api/api.go
```diff
@@ -411,31 +411,35 @@ func (api *Server) getActions(start uint64, count uint64) (*iotexapi.GetActionsR
...
+			}
+			res = append(res, act)
 		}
 	}
 
-	return &iotexapi.GetActionsResponse{Actions: res}, nil
+	return &iotexapi.GetActionsResponse{ActionInfo: res}, nil
 }
 
 // getAction returns action by action hash
```
raullenchai: // getSingleAction

Issue #793, opened 2019-03-21
ioctl install should not assume user has golang environment


zjshen14:
https://github.com/iotexproject/iotex-core/blob/68d2e7ff8c9bdd8967d9b7ac731fc5a0b270fd12/install-cli.sh#L66

Issue #759, opened 2019-03-16
Add git hook to run circle ci locally before pushing a commit


zjshen14:
Before pushing the code, run https://circleci.com/docs/2.0/local-cli/ first, and fail the push if the build doesn't pass

PR #812, opened 2019-03-22
 fix pre-commit issue

feigu1984:
https://github.com/iotexproject/iotex-core/issues/793
https://github.com/iotexproject/iotex-core/issues/759


install-cli.sh
```diff
@@ -62,7 +64,28 @@ downloadFile() {
         exit 1
     fi
 }
-
+install_golang() {
```
raullenchai: No need to install golang...

Issue #811, opened 2019-03-22
Failed to get receipt of transfer


Frankonly:
### What version of iotex-core are you using (`ioctl version`)?

snapshot-v0.6.0

### What operating system and processor architecture are you using?

macOS

### What did you do? If possible, provide a recipe for reproducing the error.

ioctl action hash 7202b8fa3bc08ebbecb0b8ba32dd093b3d3d96aa1962e3774eac9f241967c0ab

### What did you expect to see?

receipt of transfer which has been run on testnet.

### What did you see instead?

rpc error: code = NotFound desc = failed to get receipt index for action 7202b8fa3bc08ebbecb0b8ba32dd093b3d3d96aa1962e3774eac9f241967c0ab: key = 32dd093b3d3d96aa1962e3774eac9f241967c0ab doesn't exist: not exist in DB

raullenchai:
cc @zjshen14

lizhefeng:
Action hash is not added to transfer receipt, which is used as the key to find the receipt in chain db. Will be fixed in next PR.

PR #813, opened 2019-03-22
Add action hash and status to transfer receipt and vote receipt

lizhefeng:
Add action hash to receipt so that receipt can be fetched by hash.


Frankonly:
still not work on my local single server

Frankonly:
It works on my single server.

Frankonly:
#811

PR #820, opened 2019-03-22
delete redundant and wrong proposer checking

CoderZhi:
If the proposal is from other users, we already check the proposer in handleconsensusmessage. If the proposal is from the itself, no checking is needed.


consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -328,7 +319,7 @@ func (ctx *rollDPoSCtx) NewPreCommitEndorsement(
 	case ErrInsufficientEndorsements:
 		return nil, nil
 	case nil:
-		ctx.loggerWithStats().Debug("Ready to pre-commit")
+		ctx.loggerWithStats().Info("Ready to pre-commit")
```
zjshen14: Necessarily info?

PR #829, opened 2019-03-23
Fix minicluster nightly build balance check error

zcchai-iotex:
Mini-cluster nightly build fails balance check. 
This fix is to check every action issued and update balance expectation only for successfully performed actions.


tools/util/injectorutil.go
```diff
@@ -393,6 +378,10 @@ func injectTransfer(
 		return err
 	}, bo); err != nil {
 		log.L().Error("Failed to inject transfer", zap.Error(err))
+	} else {
```
golangcibot: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (from `gocritic`)

tools/util/injectorutil.go
```diff
@@ -424,8 +414,11 @@ func injectVote(
 		return err
 	}, bo); err != nil {
 		log.L().Error("Failed to inject vote", zap.Error(err))
+	} else {
```
golangcibot: elseif: can replace 'else {if cond {}}' with 'else if cond {}' (from `gocritic`)

tools/util/injectorutil.go
```diff
@@ -658,92 +657,151 @@ func GetAllBalanceMap(
...
 
+// CheckPendingActionList will go through the pending action list, for an executed action:
+// 1) update the expectation balance map if the action has been run successfully
+// 2) remove the action from pending list
+func CheckPendingActionList(
+	cs *chainservice.ChainService,
+	pendingActionMap *map[hash.Hash256]*big.Int,
+	balancemap *map[string]*big.Int,
+) error {
+	for selphash, _ := range *pendingActionMap {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for selphash := range ...` (from `golint`)

tools/util/injectorutil.go
```diff
@@ -658,92 +654,166 @@ func GetAllBalanceMap(
...
 
+// CheckPendingActionList will go through the pending action list, for an executed action:
+// 1) update the expectation balance map if the action has been run successfully
+// 2) remove the action from pending list
+func CheckPendingActionList(
+	cs *chainservice.ChainService,
+	pendingActionMap *sync.Map,
+	balancemap *map[string]*big.Int,
+) (bool, error) {
+	var retErr error
```
golangcibot: retErr declared but not used (from `typecheck`)

PR #835, opened 2019-03-24
rebroadcast precommit endorsement on timeout

CoderZhi:
In a corner case that only 2/3 or less nodes enter the commit step, the chain will get stuck. To fix the issue, the nodes in commit step should rebroadcast their commit endorsements every few seconds. Meanwhile, for the other nodes, they should process those commit endorsements from the same height but previous round.

This PR changes three things:
1. Introduce a new field COMMIT_TTL to the config for FSM, and rebroadcast commit endorsement every COMMIT_TTL
2. Update IsStaleEvent logic to accept COMMIT endorsement from previous round
3. Fix a consensus event creation time of incoming events to clock.Now()

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -125,6 +125,7 @@ var (
 					AcceptBlockTTL:               4 * time.Second,
 					AcceptProposalEndorsementTTL: 2 * time.Second,
 					AcceptLockEndorsementTTL:     2 * time.Second,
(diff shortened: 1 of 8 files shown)


CoderZhi:
Test with enable gravity chain poll, reach height 2500+
Test with disable gravity chain poll, reach height 2200, and stuck on the height because of processing time issue.

PR #838, opened 2019-03-24
IOTEX-361 Fix error when granting epoch reward

zjshen14:
So the phenomenon is that in the last block of the epoch, if some one claim the reward and drain the balance to 0, the epoch reward grant will fail.

This is because the unclaimed balance account is deleted if it's drained up, and then the epoch reward should be set to the same account. Usually it's fine because if the state doesn't exist, we will create and insert one. However, if the state is deleted in the same transaction, querying it will return AlreadyDeleted instead of NotExist, which is unexpected to handle.

The quick walk around is not to delete the 0 balance account, which should affect storage much given only limited number of address to reward.

> Why not handling AlreadyDeleted?
This is a not expected semantics from underlying storage.

Will tackle the fundamental storage problem separately


action/protocol/rewarding/reward.go
```diff
@@ -268,11 +268,6 @@ func (p *Protocol) claimFromAccount(sm protocol.StateManager, addr address.Addre
 	balance := big.NewInt(0).Sub(acc.balance, amount)
 	if balance.Cmp(big.NewInt(0)) < 0 {
 		return errors.New("no enough available balance")
-	} else if balance.Cmp(big.NewInt(0)) == 0 {
```
raullenchai: I'd rather leave a TODO here at least.

PR #841, opened 2019-03-25
IOTEX-363 Add numActions to GetAccount API

lizhefeng:


api/api.go
```diff
@@ -139,11 +139,16 @@ func (api *Server) GetAccount(ctx context.Context, in *iotexapi.GetAccountReques
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	actions, err := api.getTotalActionsByAddress(in.Address)
```
zjshen14: This is too expensive to get the total action number

PR #842, opened 2019-03-25
Return not found even if it's deleted in the current transaction

zjshen14:


db/kvstore4trie.go
```diff
@@ -94,12 +94,15 @@ func (s *KVStoreForTrie) Put(key []byte, value []byte) error {
...
-	if err != nil {
-		if v, err = s.dao.Get(s.bucket, key); err != nil {
-			return nil, errors.Wrapf(err, "failed to get key %x", key)
+	if errors.Cause(err) == ErrNotExist {
+		if v, err = s.dao.Get(s.bucket, key); errors.Cause(err) == ErrNotExist {
+			return nil, errors.Wrapf(ErrNotExist, "failed to get key %x", key)
 		}
 		// TODO: put it back to cache
 	}
+	if errors.Cause(err) == ErrAlreadyDeleted {
```
dustinxie: move 103~105 to in front of 97?

db/kvstore4trie.go
```diff
@@ -94,12 +94,15 @@ func (s *KVStoreForTrie) Put(key []byte, value []byte) error {
 func (s *KVStoreForTrie) Get(key []byte) ([]byte, error) {
 	trieKeystoreMtc.WithLabelValues("get").Inc()
 	v, err := s.cb.Get(s.bucket, key)
-	if err != nil {
-		if v, err = s.dao.Get(s.bucket, key); err != nil {
-			return nil, errors.Wrapf(err, "failed to get key %x", key)
+	if errors.Cause(err) == ErrNotExist {
+		if v, err = s.dao.Get(s.bucket, key); errors.Cause(err) == ErrNotExist {
```
dustinxie: feel line 98~99 can simply reduce to:
v, err = s.dao.Get(s.bucket, key)
there's no need for another wrap

PR #856, opened 2019-03-27
install-cli add windows

feigu1984:
windows install to $HOME/ioctl.exe


install-cli.sh
```diff
@@ -77,11 +77,12 @@ findGoBinDirectory() {
             echo "Installation could not determine your \$GOPATH."
             exit 1
         fi
-        if [ -z "$GOBIN" ]; then
-            GOBIN=$(echo "${EFFECTIVE_GOPATH%%:*}/bin" | sed s#//*#/#g)
+        if [ -z "$GOROOT" ]; then
```
raullenchai: why check on GOROOT is better than check on GOBIN?

raullenchai: windows has no GOBIN?

raullenchai: I prefer to completely get rid of GO env. We should install in three fixed dirs for macos, linux and windows.

install-cli.sh
```diff
@@ -77,11 +77,12 @@ findGoBinDirectory() {
...
+            #GOBIN=$(echo "${EFFECTIVE_GOPATH%%:*}/bin" | sed s#//*#/#g)
+	    GOBIN=$GOROOT/bin
         fi
-        if [ ! -d "$GOBIN" ]; then
```
feigu1984: windows  but golang installed

install-cli.sh
```diff
@@ -77,11 +78,12 @@ findGoBinDirectory() {
             echo "Installation could not determine your \$GOPATH."
             exit 1
         fi
-        if [ -z "$GOBIN" ]; then
-            GOBIN=$(echo "${EFFECTIVE_GOPATH%%:*}/bin" | sed s#//*#/#g)
+        if [ -z "$GOROOT" ]; then
```
raullenchai: can remove findGoBinDirectory

install-cli.sh
```diff
@@ -128,10 +130,9 @@ initArch
 initOS
 
 # determine install directory if required
```
raullenchai: remove if not needed; update the comment in file header

PR #864, opened 2019-03-28
Fix P2P and blocksync efficiency (part 1)

zjshen14:
Three things:

1. Update the go-p2p lib. Change the way to get the neighbors. Previous way is too expensive to take almost 45s to finish probing nearest peers. It changes to use local cached peer info instead (excluding itself).

2. Do a little shuffling when choosing the peers to send blocksync request to avoid hitting bootnode always

3. Expose nat/active relay option for people to use


blocksync/worker.go
```diff
@@ -98,14 +97,15 @@ func (w *syncWorker) Sync() {
...
-		p := peers[w.rrIdx]
+		rrIdx %= len(peers)
+		p := peers[rrIdx]
 		if err := w.unicastHandler(ctx, p, &iotexrpc.BlockSync{
 			Start: interval.Start, End: interval.End,
 		}); err != nil {
 			log.L().Warn("Failed to sync block.", zap.Error(err))
 		}
-		w.rrIdx++
+		rrIdx++
```
raullenchai: reshuffle or ++?

zjshen14: updated

PR #870, opened 2019-03-28
IOTEX-297 Add e2e rewarding test

zcchai-iotex:


e2etest/rewarding_test.go
```diff
@@ -69,3 +87,460 @@ func TestBlockReward(t *testing.T) {
...
+			for i := 0; i < numNodes; i++ {
+
+				var amount *big.Int
+				rewardAddrStr := identityset.Address(i + numNodes).String()
+				rewardPriKey := identityset.PrivateKey(i + numNodes)
+				expectedSuccess := true
+
+				rand.Seed(time.Now().UnixNano())
+				switch r := rand.Intn(7); r {
+				case 0:
```
raullenchai: what does each int stand for?

zcchai-iotex: int r is a selector for claim action test cases. These test cases are commented under each switch-case branch.
For example, case 0 is to test claim 0 amount, case 1 is to test claim all balance, case 4 is to test claim negative amount , etc

raullenchai: can we at least comment in the code? or better have enum to make these explicit.

zcchai-iotex: I will update with explicit enum type.

e2etest/rewarding_test.go
```diff
@@ -69,3 +87,460 @@ func TestBlockReward(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, balance.Cmp(big.NewInt(0).Mul(blockReward, big.NewInt(5))) >= 0)
 }
+
+func TestBlockEpochReward(t *testing.T) {
```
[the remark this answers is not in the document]
zcchai-iotex: Now the test is set to run up to height 60 (runToHeight := uint64(60)), and block interval is 1 second, so plus some cooldown time to finish all the actions, the test will roughly run about over 1 minute.

PR #892, opened 2019-03-29
Blocksync efficiency (Part 2): improving block request scheduling

zjshen14:
Since the block is committed in order. If first few blocks are not downloaded they will bock the whole active sync window to move forward. I adjust the scheduling to send more requests for the first few intervals.

I use the following configs in one node in mainnet rehearsal:
```
  bufferSize: 600
  maxRepeat: 4
  repeatDecayStep: 4
```

It made my node download ~1500 blocks per min

It could be even fast, but I didn't try yet.

> Is network I/O a concern?

No, we use quite little network I/O so far, at the low tps case. If tps goes high in the future, we could adjust configs.


config/config.go
```diff
@@ -133,9 +133,11 @@ var (
 			},
 		},
 		BlockSync: BlockSync{
-			Interval:     10 * time.Second,
-			BufferSize:   50,
-			IntervalSize: 10,
+			Interval:        10 * time.Second,
+			BufferSize:      100,
+			IntervalSize:    10,
+			MaxRepeat:       3,
```
raullenchai: This config basically says for block 0 and 1 can try 3 and 2 times. Does this couple with number of allowed peers? If so, need to state clearly in the comment,

```
block #, repeated times
0                3
1                2
2                1
3                1
4                1
5                ...
```

zjshen14: 1 block -> block interval
2 Not couple with # of peers

PR #914, opened 2019-04-01
IOTEX-385 Expose reward details via log

zjshen14:


zjshen14:
cc @Frankonly 

We need to be able to visualize the reward action and the logs in ioctl

Frankonly:
> cc @Frankonly
> 
> We need to be able to visualize the reward action and the logs in ioctl

Please file an issue and provide more details.

Issue #910, opened 2019-04-01
Stripe ioctl binary to make it smaller


raullenchai:
**What would you like to be added**:
Reduce ioctl binary size (now 20+MB) by stripping the binary with `-ldflags="-s -w"`, e.g., https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/

Need to make sure the stripped binary works well on all platforms,

**Why is this needed**:
Faster download, faster install

PR #921, opened 2019-04-02
Shrink ioctl binary

Frankonly:
#910


zjshen14:
@Frankonly can you test what error stack will be printed if error happens after striping the symbols

Frankonly:
> @Frankonly can you test what error stack will be printed if error happens after striping the symbols

the same when I test

PR #924, opened 2019-04-02
IOTEX-386 Add block hash and height into receipt

zjshen14:
Also rename some fields to be consistent

PR #925, opened 2019-04-03
Fix rewarding test failure when block produce interval is short

zcchai-iotex:
The issue is caused by the state difference between two chain instances on different servers during a sync period. So the fix is to check action hash and reading account balance on the same chain instance.


zjshen14:
1. So it is the test only problem, right?

2. https://github.com/iotexproject/iotex-core/pull/924 is merged, where I changed ttl to be longer to mitigate the issue. Can you change the ttl config back

zcchai-iotex:
> 1. So it is the test only problem, right?
> 2. #924 is merged, where I changed ttl to be longer to mitigate the issue. Can you change the ttl config back

Yes, it is a test bug, not because of ttl, I will merge back original ttl

Issue #918, opened 2019-04-02
Add e2e tests for typical contracts 


raullenchai:
**What would you like to be added**:

This repo contains a great list of typical contracts 
https://github.com/raineorshine/solidity-by-example. We should integrate all of them into our e2etests. Please do this in a systematic way.

**Why is this needed**:
Guarantee correctness of execution layer of blockchain.

lzxm160:
all done

PR #939, opened 2019-04-04
Add e2e tests for typical contracts #918

lzxm160:
work on #918 
add array-passing.sol


tools/executiontester/assetcontract/assetcontract.go
```diff
@@ -24,6 +24,7 @@ type ReturnedContract struct {
 	Erc721Token blockchain.Erc721Token
 	ArrDelete   blockchain.ArrayDelete
 	ArrString   blockchain.ArrayString
+	ArrPassing  blockchain.ArrayPassing
```
dustinxie: can you combine this passing test, code, solidity binary into existing ArrDelete?

the function is much similar, pushing int value into an array, and delete one item in the array

lzxm160: ok

Issue #941, opened 2019-04-04
Better ioctl installation handling in linux env


zjshen14:
**What would you like to be added**:

I hotfix an installation issue on linux, but it only works for the latest release. We need to fix it for unstable. 

In addition, see how to make this backward compatible because our previous release candidate doesn't have libsect283k1_ubuntu.so.

If I missed other points of installing the shared library, feel free to raise them. Thanks

**Why is this needed**:

wqwqwq219:
mv ibsect283k1_ubuntu.so  to /usr/lib64 on linux will resolve it

zjshen14:
@feigu1984 can you fix @wqwqwq219 's issue too?

zjshen14:
We still need it until we push next release candidate

Issue #801, opened 2019-03-21
ioctl supports Windows


raullenchai:
1. need a way to build ioctl on windows
2. need the curl install script to support windows

raullenchai:
let's set up a windows machine with go environment, pull iotex-core every 24 hours, build a binary, upload to s3://ioctl.

feigu1984:
There are many problems
1, exec gcc executable file not found in path windows
2, C:\Go\pkg\tool\windows_amd64\link.exe: running gcc failed: exit status 1
3,  buildcli.sh   dos  fileformat -->   unix

PR #950, opened 2019-04-04
IOTEX-391 Solve the linux libsect283k1_ubuntu.so error and support windows

Frankonly:
ec283.go who needs libsect283k1_ubuntu.so is not used by our code but included in the crypto package, so I take away ec283.go and related test files from the crypto package and put them into a package that no files import. Therefore, we don't need libsect283k1_ubuntu and to export LD_LIBRARY_PATH now.

After moving these code, I try running the iotex-core in windows. reinstall gcc and restart the computer. New code can work on windows now. The way I tested on windows was `go build` in \server and then `./server.exe` to run. The server was running and producing blocks.


zjshen14:
Can you describe why it will fix the problem?

Frankonly:
> Can you describe why it will fix the problem?

if I use string “poll“ rather than poll.ProtocolID in nodedelegate.go, in which case “poll” should not be imported, ioctl can be executed without libsect283k1_ubuntu.so in linux, which means golang compiler packages all imported lib into binary. The ec283.go and related test files are not used by our code but packaged into binary due to importing the "crypto", so when I take away those idle code from crypt, the binary doesn't need the libsect283k1_ubuntu.so anymore.

CoderZhi:
I guess the lib directory didn't have a go file and was ignored. After moving one go file into it, the directory will be included automatically

zjshen14:
I still couldn't understand why we used not to need libsect283k1_ubuntu.so PREVIOUSLY. ioctl uses keystore, which uses crypto, which SHOULD use libsect283k1_ubuntu.so for signature eventually.

It found geth has signature_cgo.go and signature_nocgo.go. The latter uses a pure golang implementation. Did you somehow build ioctl and use signature_nocgo.go in the end, so that libsect283k1_ubuntu.so is not required before?

Frankonly:
> I still couldn't understand why we used not to need libsect283k1_ubuntu.so PREVIOUSLY. ioctl uses keystore, which uses crypto, which SHOULD use libsect283k1_ubuntu.so for signature eventually.
> 
> It found geth has signature_cgo.go and signature_nocgo.go. The latter uses a pure golang implementation. Did you somehow build ioctl and use signature_nocgo.go in the end, so that libsect283k1_ubuntu.so is not required before?

ioctl account use the keystore who imports crypto from go-ethereum rather than ours. As what you say, it is a pure golang implementation.

Frankonly:
#941 #801

PR #960, opened 2019-04-05
Update multisend bytecode generator

Frankonly:
Update multisend bytecode generator and add solidity source code.
The payload is to be consistent with transfer.


cli/tools/multisend/Multisend.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.4.24;
```
CoderZhi: change file name to lowercase

cli/tools/multisend/multisend.go
```diff
@@ -30,8 +30,9 @@ func main() {
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
-	Use:  "multisend [command]",
-	Args: cobra.ExactArgs(1),
+	Use:   "multisend 'JSON_DATA'",
```
CoderZhi: can we add some sample of JSON_DATA format in the usage?

PR #974, opened 2019-04-07
Add e2e tests for typical contracts #918

lzxm160:
work on #918 ,add no-variable-length-returns.sol f.value.sol no-variable-length-returns.sol public-mapping.sol remove-from-array.sol sha3.sol tail-recursion.sol tuple.sol


action/protocol/execution/testdata/f.value.json
```diff
@@ -29,7 +29,9 @@
         "rawAmount": "0",
         "rawGasLimit": 1200000,
         "rawGasPrice": "0",
-        "failed": true,
+        "failed": false,
```
CoderZhi: could you explain more? Previously, "failed" was true, did you observe that same result on Ethereum? Now the "failed" value is false, is it the same as Ethereum? If both are yes, which is the real expected value, what's the purpose of this test?

action/protocol/execution/testdata/f.value.sol
```diff
@@ -13,7 +13,7 @@ contract One{
 contract Two{
     One o;
    function Two(){
-       One o = One(0x675f1057f81e9e768e33faddbd5609c09f4c0a5c);
+       o = One(0x675f1057f81e9e768e33faddbd5609c09f4c0a5c);
```
CoderZhi: use the new framework to set the contract address

action/protocol/execution/testdata/f.value.json
```diff
@@ -29,7 +29,9 @@
...
+        "hasReturnValue": true,
```
CoderZhi: this field has been deprecated, rebase onto the latest version.

action/protocol/execution/testdata/no-variable-length-returns.json
```diff
@@ -0,0 +1,51 @@
...
+        "rawAmount": "0",
+        "rawGasLimit": 5000000,
+        "rawGasPrice": "0",
+        "expectedBalances": [],
+        "comment": "deploy B contract"
+    }],
+    "executions": [{
+        "contractIndex":1,
+        "rawPrivateKey": "cfa6ef757dee2e50351620dca002d32b9c090cfda55fb81f37f1d26b273743f1",
+        "rawByteCode": "85aba275000000000000000000000000675f1057f81e9e768e33faddbd5609c09f4c0a5c",
```
CoderZhi: use the new way of fetching deployed contract address, same for other tests

action/protocol/execution/testdata/send-eth.json
```diff
@@ -0,0 +1,34 @@
+{
+    "initBalances": [{
+        "account": "io1mflp9m6hcgm2qcghchsdqj3z3eccrnekx9p0ms",
+        "rawBalance": "1000000000000000000000000000"
+    }, {
+        "account": "io1rc2d2de7rtuucalsqv4d9ng0h297t63w7wvlph",
+        "rawBalance": "0"
+    }, {
+        "account": "io1zf69lmyzkkzlywwqzzggst45qupvx2cyye62fp",
```
CoderZhi: please remove the accounts which are not used in the test. Same for all other tests.

PR #975, opened 2019-04-08
Consensus node high availability (Part 1): allow consensus node stand-by

zjshen14:
Per feedback from the community, it seems to be useful to provide HA feature for delegates. In HA, 1 server will be active and the others will standby. Once the active node is down, another will be set to active by leader election (in Part 2). The goal is 99% availability. In terms of consensus node, the two modes behavior are as follows:

- Active: handle block sync + consensus messages
- Standby: handle block sync

> What we need to do?

Add the active flag. When the flag is set to false, wait consensus fsm go back to init state, and stand by there and process no events.

> How about state sharing among the servers?

Active server will have the consensus state for the next block that, but the stand-by servers won't. It's almost okay because the active server will decommission only after it finish the current round. There're  corner cases that the states will be lost, but it should be fine for 99% availability. Once we want to pursue 99.999% availability we could do state persistency

### DIFF
--- a/consensus/scheme/rolldpos/rolldpos.go
+++ b/consensus/scheme/rolldpos/rolldpos.go
@@ -236,6 +236,15 @@ func (r *RollDPoS) CurrentState() fsm.State {
 	return r.cfsm.CurrentState()
 }
 
+// Activate activates or pauses the roll-DPoS consensus. When it is deactivated, the node will finish the current
+// consensus round if it is doing the work and then return the the initial state
+func (r *RollDPoS) Activate(active bool) { r.ctx.Activate(active) }
+
+// Active is true if the roll-DPoS consensus is active, or false if it is stand-by
+func (r *RollDPoS) Active() bool {
+	return r.ctx.Active() || r.cfsm.CurrentState() != consensusfsm.InitState
+}
+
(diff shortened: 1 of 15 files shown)


zjshen14:
Verified in mainnet rehearsal 2

consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -236,6 +236,15 @@ func (r *RollDPoS) CurrentState() fsm.State {
 	return r.cfsm.CurrentState()
 }
 
+// Activate activates or pauses the roll-DPoS consensus. When it is deactivated, the node will finish the current
+// consensus round if it is doing the work and then return the the initial state
+func (r *RollDPoS) Activate(active bool) { r.ctx.Activate(active) }
+
+// Active is true if the roll-DPoS consensus is active, or false if it is stand-by
+func (r *RollDPoS) Active() bool {
+	return r.ctx.Active() || r.cfsm.CurrentState() != consensusfsm.InitState
```
CoderZhi: what's the second part for?

zjshen14: If the consensus fsm is not at init state, it's also considered as active

Issue #982, opened 2019-04-09
Make ioctl support --endpoint= when executing an command


zjshen14:
**What would you like to be added**:

--endpoint=blah blah when executing a command

**Why is this needed**:

Temporarily talk to another endpoint to get the data

PR #1000, opened 2019-04-11
Add an endpoint flag and fix a flag mistake

Frankonly:
#982


cli/ioctl/cmd/account/account.go
```diff
@@ -52,6 +52,8 @@ func init() {
 	AccountCmd.AddCommand(accountListCmd)
 	AccountCmd.AddCommand(accountNonceCmd)
 	AccountCmd.AddCommand(accountUpdateCmd)
+	AccountCmd.PersistentFlags().StringVar(&config.ReadConfig.Endpoint, "endpoint",
+		config.ReadConfig.Endpoint, "set endpoint for once")
```
raullenchai: set endpoint for the current execution

raullenchai: same below

PR #1029, opened 2019-04-15
Make foundation bonus exclude robot machines

zjshen14:
Like epoch reward, exclude the nodes whose operator address in the exempt list


action/protocol/rewarding/reward.go
```diff
@@ -150,13 +150,23 @@ func (p *Protocol) GrantEpochReward(
 		return nil, err
 	}
 
+	// Get the delegate list who exempts repoch reward
```
raullenchai: s/repoch/epoch

PR #1034, opened 2019-04-16
fail transfer is the recipient is a contract address

CoderZhi:
As discussed offline, we want to prevent transfer to a contract address with action.Transfer. We want users to use action.Execution if the recipient is a contract address.
Let me know whether the judgement make sense for you. And I will add some unit test later.


action/protocol/account/transfer.go
```diff
@@ -31,8 +32,13 @@ func (p *Protocol) handleTransfer(ctx context.Context, act action.Action, sm pro
...
 	}
-	if tsf.IsContract() {
-		return nil, nil
+	recipientAddr, err := address.FromString(tsf.Recipient())
+	if err != nil {
+		return nil, errors.Errorf("failed to decode recipient address %s", tsf.Recipient())
+	}
+	recipientAcct, err := accountutil.LoadAccount(sm, hash.BytesToHash160(recipientAddr.Bytes()))
+	if err == nil && len(recipientAcct.CodeHash) > 0 {
+		return nil, errors.New("failed to transfer to a contract address directly")
```
zjshen14: Since we don't validate it in the action pool, we'd better not throw error, but return failure receipt. Otherwise, people can submit bad transfer to block production.

PR #1054, opened 2019-04-18
Modify ReadContract API

lizhefeng:


zjshen14:
@ququzone how urgent/critical is this issue to unblock the antenna?

ququzone:
> @ququzone how urgent/critical is this issue to unblock the antenna?

not urgent, We can set gas manual.

zjshen14:
@ququzone can you check if the change here makes sense to antena integration?

ququzone:
> @ququzone can you check if the change here makes sense to antena integration?

how can I get executed normal method result by this way?

zjshen14:
Return data structure will be the same, but the request Paramus become execution plus caller addr @lizhefeng

PR #1059, opened 2019-04-19
Prevent actions with negative amount and gas price

zjshen14:


action/protocol/account/transfer.go
```diff
@@ -42,6 +42,10 @@ func (p *Protocol) handleTransfer(ctx context.Context, act action.Action, sm pro
 		return nil, action.ErrHitGasLimit
 	}
 
+	if tsf.Amount().Cmp(big.NewInt(0)) < 0 {
```
lizhefeng: tsf.Amount.Sign()<0 is simpler?

PR #1070, opened 2019-04-22
remove alias when delete account

ququzone:
#1069


cli/ioctl/cmd/account/accountdelete.go
```diff
@@ -70,6 +70,7 @@ func accountDelete(args []string) (string, error) {
 			if err := ks.Delete(v, password); err != nil {
 				return "", err
 			}
+			alias.Remove(args[0])
```
golangcibot: Error return value of `alias.Remove` is not checked (from `errcheck`)

PR #1077, opened 2019-04-23
Fix TestBlockEpochReward sporadic failure

zcchai-iotex:
The error occurs because the values from reading the real-time unclaimed balance and asynchronously updated expectation may be inconsistent if the chain updates fast.

So the fix is to introduce a tolerance value which allows a small number of consecutive unmatched balance between real-time unclaimed balance and expectation balance, which will eventually be checked by totally balance and claimed value.


zcchai-iotex:
> can we produce blocks slowly? and check them one by one?

Yes, if we produce blocks slowly, then most probably, reading the real-time unclaimed balance and reading the action-driven expectation will be consistent.  And this updated tolerance code will never be triggered.

PR #1098, opened 2019-04-27
Allow reading password from file

hao1022:
In some occasions, the user of ioctl wants full automation. For
example, auto-claim reward regularly. This change eliminates the
interaction between user and ioctl and fully automates ioctl's
action execution.

### DIFF
--- a/cli/ioctl/cmd/action/action.go
+++ b/cli/ioctl/cmd/action/action.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"math/big"
 	"syscall"
+	"io/ioutil"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/spf13/cobra"
@@ -36,6 +37,7 @@ var (
 	nonce    uint64
 	signer   string
 	bytecode []byte
+	pwdFile  string
 )
 
 // ActionCmd represents the account command
@@ -66,6 +68,7 @@ func setActionFlags(cmds ...*cobra.Command) {
 			"set gas price (unit: 10^(-6)Iotx)")
 		cmd.Flags().StringVarP(&signer, "signer", "s", "", "choose a signing account")
 		cmd.Flags().Uint64VarP(&nonce, "nonce", "n", 0, "set nonce")
+		cmd.Flags().StringVarP(&pwdFile, "pwdfile", "w", "", "file to read password from")
 		cmd.MarkFlagRequired("signer")
 		if cmd == actionDeployCmd || cmd == actionInvokeCmd {
 			cmd.Flags().BytesHexVarP(&bytecode, "bytecode", "b", nil, "set the byte code")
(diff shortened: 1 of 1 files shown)


cli/ioctl/cmd/action/action.go
```diff
@@ -36,6 +37,7 @@ var (
 	nonce    uint64
 	signer   string
 	bytecode []byte
+	pwdFile  string
```
golangcibot: `pwdFile` is a global variable (from `gochecknoglobals`)

cli/ioctl/cmd/action/action.go
```diff
@@ -93,11 +96,22 @@ func GetGasPrice() (*big.Int, error) {
...
 func sendAction(elp action.Envelope) (string, error) {
-	fmt.Printf("Enter password #%s:\n", signer)
-	bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
-	if err != nil {
-		log.L().Error("failed to get password", zap.Error(err))
-		return "", err
+	var bytePassword []byte
+	if pwdFile != "" {
+		var err error
+		bytePassword, err := ioutil.ReadFile(pwdFile)
```
golangcibot: ineffectual assignment to `bytePassword` (from `ineffassign`)

cli/ioctl/cmd/action/action.go
```diff
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"math/big"
 	"syscall"
+	"io/ioutil"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

cli/ioctl/cmd/action/action.go
```diff
@@ -93,11 +96,22 @@ func GetGasPrice() (*big.Int, error) {
...
+		var err error
+		bytePassword, err = ioutil.ReadFile(pwdFile)
+		if err != nil {
+			log.L().Error("failed to read password from file", zap.Error(err))
+			return "", err
+		}
+	} else {
+		var err error
+		fmt.Printf("Enter password #%s:\n", signer)
+		bytePassword, err = terminal.ReadPassword(int(syscall.Stdin))
```
golangcibot: unnecessary conversion (from `unconvert`)

raullenchai: +1

hao1022:
> This sounds like a security hole.
> 
> Does this approach help you out to the bash script for prompt interaction?
> https://stackoverflow.com/questions/3804577/have-bash-script-answer-interactive-prompts

True.

But I would argue this is no worse than storing private key in config files, which is already practiced in iotex. This is just an option. If the user wants to use it, then they are pretty much willing to take the risk or have a way to mitigate it.

We had attempted to enable automatic input, but it turned out to be difficult. Some relevant discussion here:

  https://github.com/golang/go/issues/19909

zjshen14:
> But I would argue this is no worse than storing private key in config files, which is already practiced in iotex. 

Afaik, we stored private key as an encrypted keystore file than a plain text in the config file.  However, the code change here wants to read password from a plain text file.

> We had attempted to enable automatic input, but it turned out to be difficult. Some relevant discussion here:

Thanks for pointing us the thread. According to it,  https://github.com/golang/go/issues/19909#issuecomment-462103555 seems to be a reasonable walk around?

hao1022:
> > But I would argue this is no worse than storing private key in config files, which is already practiced in iotex.
> 
> Afaik, we stored private key as an encrypted keystore file than a plain text in the config file. However, the code change here wants to read password from a plain text file.

I mean the private key stored in config.yaml when launch server.

> 
> > We had attempted to enable automatic input, but it turned out to be difficult. Some relevant discussion here:
> 
> Thanks for pointing us the thread. According to it, [golang/go#19909 (comment)](https://github.com/golang/go/issues/19909#issuecomment-462103555) seems to be a reasonable walk around?

Not really working.

zjshen14:
> I mean the private key stored in config.yaml when launch server.

This is the problem we need to fix.

> Not really working.

Maybe you want to try [js](v)/[go](https://github.com/iotexproject/iotex-antenna-go) sdk for more elegant implementation.

zjshen14:
@hao1022 filed an issue to track this security problem: https://github.com/iotexproject/iotex-core/issues/1130

pam4:
> > Thanks for pointing us the thread. According to it, https://github.com/golang/go/issues/19909#issuecomment-462103555 seems to be a reasonable walk around?
> 
> Not really working.

I have only a rough idea of the context here, but I'm curious about why.

Not sure what you mean with "we had attempted to enable automatic input", but if your aim is to read a password from a non-terminal `stdin` (and you are not using `stdin` for something else), then that kind of code should just work.

https://play.golang.org/p/UfFLRKSvyhr

Maybe I'm just misunderstanding your use case, but you said the discussion in golang/go#19909 is relevant.

zjshen14:
Close it.

@hao1022 , if you have different opinion, please feel free to reopen it.

Issue #1071, opened 2019-04-22
Cleanup native voting system code


zjshen14:
We eventually chose to vote via ethereum, so that the native voting system code is no longer useful and we could delete it. There are a bunch of things need to do:

1. Remove voting action's reference in the test code or replace it with transfer action if necessary to make the test pass.
2. Remove voting action protobuf and goland definition and reference in the prod code.
3. Remove voting protocol registration in the chain service, and the tests
4. Remove voting protocol, and the voting logic handling in account protocol (one important thing here is to evaluate if it's a breaking change)
5. Use lifeLongDelegatesProtocol by default to feed delegate info to make blockchain be able to run by default.

zjshen14:
@lzxm160 would you please help this task?

lzxm160:
no problem

PR #1127, opened 2019-05-03
Cleanup native voting system code #1071 Use lifeLongDelegatesProtocol…

lzxm160:
work on #1071 
5、Use lifeLongDelegatesProtocol by default to feed delegate info to make blockchain be able to run by default.


blockchain/blockchain.go
```diff
@@ -1351,6 +1351,15 @@ func (bc *blockchain) createPollGenesisStates(ctx context.Context, ws factory.Wo
 			ws,
 		)
 	}
+	pollProtocol := poll.NewLifeLongDelegatesProtocol(bc.config.Genesis.Delegates)
```
zjshen14: This code is not necessary

blockchain/blockchain_test.go
```diff
@@ -833,8 +833,9 @@ func TestBlockchainInitialCandidate(t *testing.T) {
 	defer func() {
 		require.NoError(bc.Stop(context.Background()))
 	}()
-	// TODO: we will fix this test case by testing using lifeLongDelegatesProtocol to initialize the candidates
-	// candidate, err := sf.CandidatesByHeight(0)
+	candidate, err := sf.CandidatesByHeight(1)
```
zjshen14: Check this: https://github.com/iotexproject/iotex-core/blob/master/server/itx/server.go#L283-L319

Like the one you fix in minicluster, you need to adjust the genesis config to enable lifeLongDelegatesProtocol

blockchain/blockchain_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/iotexproject/iotex-core/action/protocol/poll"
```
zjshen14: fix the group

blockchain/genesis/genesis.go
```diff
@@ -172,7 +172,7 @@ type (
 // New constructs a genesis config. It loads the default values, and could be overwritten by values defined in the yaml
 // config files
 func New() (Genesis, error) {
-	def := defaultConfig()
+	def := Default
```
zjshen14: No, we should still call defaultConfig()

blockchain/blockchain_test.go
lzxm160: sure

blockchain/genesis/genesis.go
lzxm160: defaultConfig() does not append delegates,this cause "make run" server error,but Default appended delegates in initTestDefaultConfig()

lzxm160:
move to #1139

PR #1138, opened 2019-05-06
delete an account should not require password #1114

svolence:


cli/ioctl/cmd/account/accountdelete.go
```diff
@@ -62,17 +60,6 @@ func accountDelete(args []string) (string, error) {
 				return "Quit", nil
 			}
 
-			fmt.Printf("Enter password #%s:\n", addr)
-			bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
-			if err != nil {
-				log.L().Error("failed to get password", zap.Error(err))
-				return "", err
-			}
-			password := string(bytePassword)
```
raullenchai: then it is not deleted...

svolence: > then it is not deleted...

Has been updated

PR #1141, opened 2019-05-07
Fix the bugs of double charging gas fee

zjshen14:
1. Simple way to handle the hard fork by a given height. Eth is doing in the similar way.
2. Make the fork height configurable by genesis, but not included in genesis proto now to be compatible on genesis hash for mainnet.
3. In test, all fork height is set 0 to test the new behavior. For test cluster, we could override the height in genesis.
4. Fix two errors:
 - In account protocol, avoid one account read/write shadow the other in the process of deposit gas.
 - In execution protocol, refund all deposit first to avoid double charge.


zjshen14:
> could we have a better test for execution part, e.g. gas consumption?

The gas consumption has been implicitly tested by asserting the balance. Previously it didn't fail because we don't include rewarding protocol to charge gas. The pr registers the rewarding protocol into the execution test

PR #1142, opened 2019-05-07
Cleanup native voting system code #1071 rm voting in account protocol

lzxm160:
work on #1071 
rm voting in account protocol


action/protocol/account/accountpb/account.proto
```diff
@@ -15,8 +15,5 @@ message Account {
     string balance  = 2;
     bytes root  = 3;
     bytes codeHash  = 4;
-    bool isCandidate = 5;
```
zjshen14: This maybe a breaking change. Can you do an experiment.
1 Before removing these fields, construct an Account proto message with only nonce/balance/root/codeHash are specified, and marshal it to bytes
2. After removing these fields, construct an Account proto message with the same parameters as above, and marshal it to bytes too
3. Compare if the two bytes are equal or not

lzxm160: sure

lzxm160: tested,the two bytes are equal

PR #1157, opened 2019-05-11
 Add GetActionsByAddress() endpoint

dustinxie:


api/api.go
```diff
@@ -544,34 +550,50 @@ func (api *Server) getActions(start uint64, count uint64) (*iotexapi.GetActionsR
...
-				return &iotexapi.GetActionsResponse{ActionInfo: res}, nil
+				return &iotexapi.GetActionsResponse{
+					Total:      totalActions,
+					ActionInfo: res,
+				}, nil
 			}
 			act, err := api.convertToAction(selps[i], true)
 			if err != nil {
-				return nil, err
+				continue
```
lizhefeng: Why continue?

api/api.go
```diff
@@ -544,34 +550,50 @@ func (api *Server) getActions(start uint64, count uint64) (*iotexapi.GetActionsR
...
+	for height := uint64(1); height <= api.bc.TipHeight(); height++ {
+		blk, err := api.bc.GetBlockByHeight(height)
 		if err != nil {
 			return nil, status.Error(codes.NotFound, err.Error())
 		}
 		selps := blk.Actions
+		if !hit && start >= uint64(len(selps)) {
+			start -= uint64(len(selps))
+			continue
+		}
```
lizhefeng: Shouldn't we set hit to true before the for loop?

api/api.go
dustinxie: as discussed, this is trade-off of robustness vs. accuracy

api/api.go
dustinxie: since start > len(selps), hit remains false before the for loop below

Issue #1155, opened 2019-05-10
Export Public Key and Sign Message


CoderZhi:
**What would you like to be added**:
1. Export Public Key of an account
2. Sign a message with private key

**Why is this needed**:
1. Share public key to other ppl
2. Sign a private message, e.g., identity for login

zjshen14:
@lzxm160 would you please help?

lzxm160:
> @lzxm160 would you please help?

sure

PR #1161, opened 2019-05-11
Export Public Key and Sign Message #1155

lzxm160:
work on #1155 
add two commands:
1、ioctl account exportpublic fist
Enter password #fist:
04d4f7be66bf121f0d1dfe0a4ebe30561d40d71932994d2fbc86fd4ef25e71b8923e5d39787f0a44a1c5574a7b8c5b8c4dcecea5c1be5a2b0c040c8d15a93d743b
2、ioctl account sign fist 0xdeadbeefdafa
Enter password #fist:
8aab5952493d0a5e7436115a226d6837a6f1b07f46cb49764194a7e12dd428107e8f877ec7c2f5bffa2f7c02d5a8c95e2ecb87be894314bc30df5d3cae6fba7500

msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(message), message)
signed msg is same as ethereum


cli/ioctl/cmd/account/account.go
```diff
@@ -9,24 +9,26 @@ package account
...
+	"encoding/hex"
 	"fmt"
 	"syscall"
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
-	"github.com/iotexproject/go-pkgs/crypto"
-	"github.com/iotexproject/iotex-address/address"
-	"github.com/iotexproject/iotex-proto/golang/iotexapi"
-	"github.com/iotexproject/iotex-proto/golang/iotextypes"
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
```
dustinxie: use go-pkgs/crypto for message signing
we don't want to introduce new dependency here

cli/ioctl/cmd/account/account.go
```diff
@@ -9,24 +9,26 @@ package account
...
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 	"golang.org/x/crypto/ssh/terminal"
 	"google.golang.org/grpc/status"
 
+	"github.com/iotexproject/go-pkgs/crypto"
```
dustinxie: change back to original import style, iotex-address, iotex-proto do not belong to iotex-core internal project

cli/ioctl/cmd/account/account.go
```diff
@@ -47,16 +49,34 @@ func init() {
...
 }
 
+// Sign sign message with signer
+func Sign(signer, password, message string) (signedMessage string, err error) {
+	pri, err := KsAccountToPrivateKey(signer, password)
+	if err != nil {
+		return
+	}
+	msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(message), message)
+	mes := ethcrypto.Keccak256([]byte(msg))
```
dustinxie: use go-pkgs/hash.Hash256()

lzxm160: sure

cli/ioctl/cmd/account/account.go
```diff
@@ -47,16 +49,34 @@ func init() {
...
 		"insecure connection for once")
 }
 
+// Sign sign message with signer
+func Sign(signer, password, message string) (signedMessage string, err error) {
+	pri, err := KsAccountToPrivateKey(signer, password)
+	if err != nil {
+		return
+	}
+	msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(message), message)
```
dustinxie: did some study, looks like we are trying to implement web3.eth_sign() here, so our user can use iotex address to sign eth message

https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign
it is taking message to sign in hexstring format, "0xdeadbeef"

so we need to interpret 'message' as a hex string, same as we discussed in antenna-go project, this also allows user to sign a message that is not necessarily a text string

b, err := hex.DecodeString(message)
if err != nil {
    return
}
prefix := fmt.Sprintf("\x19Ethereum Signed Message:\n%d", len(b))
msg := append([]byte(prefix), b...)

lzxm160: ok

PR #1172, opened 2019-05-16
Upgrade the ci resource for ubuntu job

zjshen14:


zjshen14:
cicrlci machine executor doesn't support resource class yet

PR #1184, opened 2019-05-21
Implement block streaming API

lizhefeng:


api/api.go
```diff
@@ -512,6 +512,37 @@ func (api *Server) SendSignedActionBytes(
...
+func (api *Server) StreamBlocks(in *iotexapi.StreamBlocksRequest, stream iotexapi.APIService_StreamBlocksServer) error {
+	subscriber := &blockListener{
+		pendingBlks: make(chan *block.Block, 64), // Actually 1 should be enough
+		cancelChan:  make(chan interface{}),
+	}
+	if err := api.bc.AddSubscriber(subscriber, true); err != nil {
+		return status.Error(codes.Internal, "failed to add the subscriber for new blocks")
+	}
+
+	for {
```
zjshen14: Can we encapsulate the logic into blockListener?

api/api.go
```diff
@@ -512,6 +512,37 @@ func (api *Server) SendSignedActionBytes(
...
+			for _, receipt := range blk.Receipts {
+				receiptsPb = append(receiptsPb, receipt.ConvertToReceiptPb())
+			}
+			if err := stream.Send(&iotexapi.StreamBlocksResponse{
+				Block: &iotexapi.BlockInfo{
+					Block:    blk.ConvertToBlockPb(),
+					Receipts: receiptsPb,
+				},
+			}); err != nil {
+				return status.Errorf(codes.Aborted, "failed to send block %d in the stream", blk.Height())
```
zjshen14: In this case we need to unsubscribe, right?

blockchain/blockchain.go
```diff
@@ -150,24 +150,28 @@ type Blockchain interface {
...
+	mu               sync.RWMutex // mutex to protect utk, tipHeight and tipHash
+	dao              *blockDAO
+	config           config.Config
+	tipHeight        uint64
+	tipHash          hash.Hash256
+	validator        Validator
+	lifecycle        lifecycle.Lifecycle
+	clk              clock.Clock
+	blocklistener    []BlockCreationSubscriber
+	apiBlockListener []BlockCreationSubscriber
```
zjshen14: We shouldn't differentiate blocklistener and  apiBlockListener

api/api.go
```diff
@@ -527,14 +541,22 @@ func (api *Server) Start() error {
 			log.L().Fatal("Node failed to serve.", zap.Error(err))
 		}
 	}()
+	if err := api.bc.AddSubscriber(api.blockListener); err != nil {
```
zjshen14: Can you please double check blockchain is started before api and close after it

api/blocklistener.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Stop stops the block listener
+func (bl *blockListener) Stop() error {
+	close(bl.cancelChan)
+	return nil
+}
+
+// HandleBlock handles the block
+func (bl *blockListener) HandleBlock(blk *block.Block) error {
+	if len(bl.streamList) > 0 {
```
zjshen14: yo may not necessary to do this check.

api/api.go
```diff
@@ -512,6 +520,12 @@ func (api *Server) SendSignedActionBytes(
 	})
 }
 
+// StreamBlocks streams blocks
+func (api *Server) StreamBlocks(in *iotexapi.StreamBlocksRequest, stream iotexapi.APIService_StreamBlocksServer) error {
+	api.blockListener.streamList = append(api.blockListener.streamList, stream)
```
zjshen14: Can you make it blockListener#AddStream() method

api/blocklistener.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/iotexproject/iotex-core/blockchain/block"
+	"github.com/iotexproject/iotex-core/pkg/log"
+)
+
+// blockListener defines the block listener in subscribed through API
+type blockListener struct {
+	pendingBlks chan *block.Block
+	cancelChan  chan interface{}
+	streamList  []iotexapi.APIService_StreamBlocksServer
```
zjshen14: Consider the problem of concurrent access to the streamList.

Can we use sync.Map instead? Not sure if there's sync.List

api/blocklistener.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Start starts the block listener
+func (bl *blockListener) Start() error {
+	go func() {
+		for {
+			select {
+			case <-bl.cancelChan:
+				return
```
zjshen14: on cancel, do we need to close all streams?

api/blocklistener.go
```diff
@@ -0,0 +1,64 @@
...
+							Block:    blk.ConvertToBlockPb(),
+							Receipts: receiptsPb,
+						},
+					}); err != nil {
+						log.L().Info(
+							"Error when streaming the block",
+							zap.Uint64("height", blk.Height()),
+							zap.Error(err),
+						)
+						bl.streamList = append(bl.streamList[:i], bl.streamList[i+1:]...)
```
zjshen14: Do we still need to explicitly close bl.streamList[i]?

api/api.go
lizhefeng: Yes. Confirmed.

PR #1187, opened 2019-05-21
Fix contract read return ni by setting correct nonce and gas limit in…

dustinxie:
… api.ReadContract()

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -86,7 +86,7 @@ type Server struct {
 
 // NewServer creates a new server
 func NewServer(
-	cfg config.API,
+	cfg config.Config,
 	chain blockchain.Blockchain,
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -266,6 +267,7 @@ type (
 		SuggestBlockWindow int    `yaml:"suggestBlockWindow"`
 		DefaultGas         uint64 `yaml:"defaultGas"`
 		Percentile         int    `yaml:"Percentile"`
+		ActionGasLimit     uint64 `yaml:"actionGasLimit"`
```
zjshen14: It's not necessary to define another config field. We just need to pass `Genesis.ActionGasLimit` into api

Issue #1185, opened 2019-05-21
getActions takes too long


puncsky:
### What version of iotex-core are you using (`ioctl version`)?

iotex-core v0.5.1

### What did you do? If possible, provide a recipe for reproducing the error.

https://testnet.iotexscan.io/api-gateway/
```gql
  query getActions($byIndex: GetActionsByIndexRequest) {
    getActions(byIndex: $byIndex) {
      actionInfo {
        actHash
        blkHash
      action {
        core {
          version
          nonce
          gasLimit
          gasPrice
          transfer {
            amount
            recipient
            payload
          }
          execution {
            amount
            contract
            data
          }
          depositToRewardingFund {
            amount
            data
          }
          claimFromRewardingFund {
            amount
            data
          }
          grantReward {
            type
          }
        }
        signature
        senderPubKey
      }
    }
    }
  }
```

variables

```json
{"byIndex": {"start":172066, "count": 20}}
```

### What did you expect to see?

https://testnet.iotexscan.io/action should not timeout

zjshen14:
@puncsky how does https://iotexscan.io/action query blockchain api?

puncsky:
> @puncsky how does https://iotexscan.io/action query blockchain api?

updated into the description

PR #1196, opened 2019-05-23
Reverse the order of db queries in GetActionsByIndex API

lizhefeng:
https://github.com/iotexproject/iotex-core/issues/1185

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -598,6 +598,7 @@ func (api *Server) readState(ctx context.Context, in *iotexapi.ReadStateRequest)
 }
 
 // GetActions returns actions within the range
+// This is a workaround for the slow access issue if the start index is very big
 func (api *Server) getActions(start uint64, count uint64) (*iotexapi.GetActionsResponse, error) {
 	if count == 0 || count > api.cfg.RangeQueryLimit {
 		return nil, status.Error(codes.InvalidArgument, "range exceeds the limit")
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -611,23 +611,29 @@ func (api *Server) getActions(start uint64, count uint64) (*iotexapi.GetActionsR
...
+		if !hit && reverseStart >= uint64(len(blk.Actions)) {
+			reverseStart -= uint64(len(blk.Actions))
 			continue
 		}
-		// now start < len(blk.Actions), we are going to fetch actions from this block
+		// now reverseStart < len(blk.Actions), we are going to fetch actions from this block
 		hit = true
-		act := api.actionsInBlock(blk, start, count)
+		act := api.reverseActionsInBlock(blk, reverseStart, count)
 		res = append(res, act...)
```
zjshen14: res = append(act,res...)

api/api.go
```diff
@@ -611,23 +611,29 @@ func (api *Server) getActions(start uint64, count uint64) (*iotexapi.GetActionsR
 		return nil, status.Error(codes.InvalidArgument, "start exceeds the limit")
 	}
 
+	reverseStart := totalActions - (start + count)
```
zjshen14: Pls leave a comment to tell why we do the reverse order

Issue #1194, opened 2019-05-23
Node crash after moving to iotex-election 0.1.10


zjshen14:
```
[...]

PR #1210, opened 2019-05-26

lzxm160:
work on #1194 

This fix crash on start and errors when syncing blocks


lzxm160:
open another branch

PR #1218, opened 2019-05-29

lzxm160:
work on #1194 

According to https://github.com/iotexproject/iotex-core/pull/1210#pullrequestreview-242851417

Tests in my local computer shows that wait for about 6 minutes util  initialize ok


action/protocol/poll/protocol.go
```diff
@@ -197,6 +198,24 @@ func (p *governanceChainCommitteeProtocol) Initialize(
 	log.L().Info("Initialize poll protocol", zap.Uint64("height", p.initGravityChainHeight))
 	var ds state.CandidateList
 	if ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight); err != nil {
+		errMsg := "bucket = electionNS doesn't exist: not exist in DB"
```
zjshen14: Don't check errorMsg string, but error type.

action/protocol/poll/protocol.go
```diff
@@ -197,6 +198,24 @@ func (p *governanceChainCommitteeProtocol) Initialize(
...
+		errChan := make(chan error, 1)
+		errChan <- err
+	loop:
```
zjshen14: No need to have loop label

action/protocol/poll/protocol.go
```diff
@@ -197,6 +198,24 @@ func (p *governanceChainCommitteeProtocol) Initialize(
 	log.L().Info("Initialize poll protocol", zap.Uint64("height", p.initGravityChainHeight))
 	var ds state.CandidateList
 	if ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight); err != nil {
+		errMsg := "bucket = electionNS doesn't exist: not exist in DB"
+		errChan := make(chan error, 1)
+		errChan <- err
```
zjshen14: You don't have to check error asynchronously. All could happen in a synchronous loop with sleep backoff.

action/protocol/poll/protocol.go
```diff
@@ -197,6 +198,24 @@ func (p *governanceChainCommitteeProtocol) Initialize(
...
+	loop:
+		for {
+			select {
+			case <-time.After(time.Second * 15):
```
zjshen14: backoff 15 sec is too conservative, and can we make it configurable

action/protocol/poll/protocol.go
lzxm160: sure

action/protocol/poll/protocol.go
lzxm160: ok

PR #1220, opened 2019-05-29
Add bloom filter in block header to capture all contract events

dustinxie:
our block lacks ability to quickly search/filter contract events, add a bloom filter into block header to address this. code used pacificHeight as knob to enable this change

Test plan: make test can pass => it is backward-compatible with current chain data


blockchain/block/bloomfilter.go
```diff
@@ -0,0 +1,47 @@
...
+	if len(b) != 256 {
+		return nil, errors.Errorf("wrong length %d, expecting 256", len(b))
+	}
+	f := bloomFilter{}
+	copy(f[:], b[:])
+	return &f, nil
+}
+
+// Add 32-byte key into bloom filter
+func (f *bloomFilter) Add(key hash.Hash256) {}
```
dustinxie: WIP to implement this and Exist()

blockchain/block/header.go
```diff
@@ -87,6 +89,10 @@ func (h *Header) BlockHeaderCoreProto() *iotextypes.BlockHeaderCore {
 		DeltaStateDigest: h.deltaStateDigest[:],
 		ReceiptRoot:      h.receiptRoot[:],
 	}
+	if h.logsBloom != nil && h.height >= genesis.PacificHeight() {
```
dustinxie: add in bloom filter for height >= pacificHeight

blockchain/blockchain.go
```diff
@@ -630,11 +631,16 @@ func (bc *blockchain) MintNewBlock(
 	if newblockHeight == 1 {
 		prevBlkHash = bc.config.Genesis.Hash()
 	}
+	var bloom block.BloomFilter
+	if newblockHeight >= genesis.PacificHeight() {
```
dustinxie: add in bloom filter for height >= pacificHeight

blockchain/block/bloomfilter.go
```diff
@@ -0,0 +1,75 @@
+package block
+
+import (
+	"github.com/iotexproject/go-pkgs/hash"
+	"github.com/pkg/errors"
+)
+
+type (
+	// BloomFilter interface
+	BloomFilter interface {
```
zjshen14: It seems not necessary to have an interface

blockchain/genesis/genesis.go
```diff
@@ -196,6 +196,11 @@ func New() (Genesis, error) {
 	return genesis, nil
 }
 
+// PacificHeight returns the pacific height
+func PacificHeight() uint64 {
```
zjshen14: We shouldn't always assume the default config. Test and testnet are using different ones.

blockchain/blockchain.go
zjshen14: get the height from config instead of reading the default value.

Let's define a new height as 0.6.x is released. For example, "CaliforniaHeight"?

Issue #1208, opened 2019-05-26
ioctl support 0x start bytecode


guo:

guo:
install `ioctl`
point to testnet.
try this command
```
ioctl action invoke vitatest1 -s buddytest -l 1000000 -b 70a082310000000000000000000000007fad0dcb9ea5fcabdc26c498d5f94161e54a8d44
```
add some code to make 
```
ioctl action invoke vitatest1 -s buddytest -l 1000000 -b 0x70a082310000000000000000000000007fad0dcb9ea5fcabdc26c498d5f94161e54a8d44
```
also works.

remove the prefix `0x` if it is detected.

PR #1223, opened 2019-05-30
ioctl support 0x start bytecode

JIUQIWANG:
Working on #1208 

Add support to 0x start bytecode by using string to be store the input parameter BYTE_CODE. Then modify and decode before Execution.

Test on my local computer that the 0x start bytecode could be accepted by ioctl now.


cli/ioctl/cmd/action/action.go
```diff
@@ -69,7 +69,7 @@ func setActionFlags(cmds ...*cobra.Command) {
 		cmd.Flags().Uint64VarP(&nonce, "nonce", "n", 0, "set nonce")
 		cmd.MarkFlagRequired("signer")
 		if cmd == actionDeployCmd || cmd == actionInvokeCmd || cmd == actionReadCmd {
-			cmd.Flags().BytesHexVarP(&bytecode, "bytecode", "b", nil, "set the byte code")
+			cmd.Flags().StringVarP(&bytecodeString, "bytecode", "b", "1", "set the byte code")
```
CoderZhi: default byte code string should be ""

cli/ioctl/cmd/action/actiondeploy.go
```diff
@@ -60,7 +62,9 @@ func deploy() (string, error) {
 			return "", err
 		}
 	}
-	tx, err := action.NewExecution("", nonce, big.NewInt(0), gasLimit, gasPriceRau, bytecode)
+	var bytecodeBytes []byte
+	bytecodeBytes, err =hex.DecodeString(strings.TrimLeft(bytecodeString, "0x"))
```
CoderZhi: ```bytecodeBytes, err := hex.DecodeString(strings.TrimLeft(bytecodeString, "0x"))```

CoderZhi: what if there is any error?

cli/ioctl/cmd/action/actioninvoke.go
```diff
@@ -7,11 +7,12 @@
 package action
 
 import (
+	"encoding/hex"
 	"fmt"
-	"math/big"
-
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
+	"math/big"
```
CoderZhi: wrong grouping. As you can tell from the other files, we put system packages in the first group, the 3rd party packages in the second group, and our own packages in the third group.

Issue #1140, opened 2019-05-06
Revive test coverage


zjshen14:
Test coverage reported by https://codecov.io/gh/iotexproject/iotex-core is growing lower, as we didn't write enough test cases for newly added code. Two packages are most worth adding more tests are: action/protocols and blockchain.

lzxm160:
still need improve test coverage

raullenchai:
Thanks for pushing this one consistently.

lzxm160:
need more tests to fix coverage

PR #1225, opened 2019-05-30
Revive test coverage #1140

lzxm160:
work on #1140 

Merging this into master will increase coverage by 2.56%


blockchain/actionpicker_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/iotexproject/iotex-core/action"
+	"github.com/iotexproject/iotex-core/actpool/actioniterator"
+	"github.com/iotexproject/iotex-core/test/identityset"
+)
+
+func TestPickAction(t *testing.T) {
```
zjshen14: The effectiveness of this test should be testing against the gas limit

server/itx/server_test.go
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) 2019 IoTeX
```
zjshen14: I feel these tests are not so useful, as it's a integration place. Instead, we need to write more e2e integration tests

server/itx/heartbeat_test.go
```diff
@@ -0,0 +1,44 @@
...
+	s, err := NewServer(cfg)
+	cfg.Consensus.Scheme = config.RollDPoSScheme
+	cfg.Genesis.EnableGravityChainVoting = true
+	require.NoError(err)
+	require.NotNil(s)
+	handler := NewHeartbeatHandler(s)
+	require.NotNil(handler)
+	require.Panics(func() { handler.Log() }, "P2pAgent is nil")
+
+	ctx, cancel := context.WithCancel(context.Background())
```
zjshen14: What does the following try to test?

server/itx/subchain_test.go
```diff
@@ -7,10 +7,20 @@
 package itx
```
zjshen14: You could ignore subchain related, which might be deprecated.

server/itx/heartbeat_test.go
lzxm160: test HeartbeatHandler.Log,but there's no returned value

server/itx/subchain_test.go
lzxm160: sure

Issue #1205, opened 2019-05-25
Skip system actions when estimating gas price


zjshen14:
We allow system actions (GrantReward and PutPollResult) to be included into a block with a 0 gas price, so that it will affect the estimation of gas price for the user actions if we include them to calculate the past average gas price in a block: https://github.com/iotexproject/iotex-core/blob/master/gasstation/gasstattion.go#L37. We need to exclude these system actions in SuggestGasPrice to make the suggested gas price more reasonable.

PR #1226, opened 2019-05-30
Skip system actions when estimating gas price

JIUQIWANG:
Working on #1205 

Write a function to determine whether one action is system action or user action.

Pass the unit test and get the correct result.


gasstation/gasstattion.go
```diff
@@ -33,6 +33,18 @@ func NewGasStation(bc blockchain.Blockchain, cfg config.API, limit uint64) *GasS
 	}
 }
 
+//WhetherSystemAction determine whether input action belongs to system action
```
CoderZhi: space after //

gasstation/gasstattion.go
```diff
@@ -33,6 +33,18 @@ func NewGasStation(bc blockchain.Blockchain, cfg config.API, limit uint64) *GasS
...
+func (gs *GasStation) WhetherSystemAction(act action.SealedEnvelope) bool {
```
CoderZhi: IsSystemAction

PR #1229, opened 2019-05-30
Skip system actions when estimating gas price

JIUQIWANG:
Working on #1205

Write a function to determine whether one action is system action or user action and check skip the block which only have system action.

I separate the TestSuggestGasPrice function in gasstation_test.go file to two function, one for normal block with one transfer user action and a system action, other function is testing for the blocks which only consist of system action. Then test their price result to see whether the logic is worked.

### DIFF
--- a/gasstation/gasstattion.go
+++ b/gasstation/gasstattion.go
@@ -33,8 +33,8 @@ func NewGasStation(bc blockchain.Blockchain, cfg config.API, limit uint64) *GasS
 	}
 }
 
-//WhetherSystemAction determine whether input action belongs to system action
-func (gs *GasStation) WhetherSystemAction(act action.SealedEnvelope) bool {
+//IsSystemAction determine whether input action belongs to system action
+func (gs *GasStation) IsSystemAction(act action.SealedEnvelope) bool {
 	switch act.Action().(type) {
 	case *action.GrantReward:
 		return true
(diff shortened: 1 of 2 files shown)


gasstation/gasstattion.go
```diff
@@ -54,11 +66,8 @@ func (gs *GasStation) SuggestGasPrice() (uint64, error) {
 
 		smallestPrice := blk.Actions[0].GasPrice()
```
zjshen14: This is problematic, because blk.Actions[0] could be a system action.

So it would happen that in a block there's just one GrantReward action, and no user action. In this case, you should have nil smallestPrice.

gasstation/gasstattion.go
```diff
@@ -51,14 +63,13 @@ func (gs *GasStation) SuggestGasPrice() (uint64, error) {
 		if len(blk.Actions) == 0 {
 			continue
 		}
-
+		if gs.IsSystemAction(blk.Actions[0]) && len(blk.Actions) == 1 {
```
zjshen14:
```suggestion
		if len(blk.Actions) == 1 && gs.IsSystemAction(blk.Actions[0]) {
```

Short circuit: https://kuree.gitbooks.io/the-go-programming-language-report/content/22/text.html

PR #1233, opened 2019-06-02

lzxm160:
work on #1194

According to #1210 (review)

Tests in my local computer shows that wait for about 6 minutes util initialize ok

### DIFF
--- a/action/protocol/poll/protocol.go
+++ b/action/protocol/poll/protocol.go
@@ -200,16 +200,16 @@ func (p *governanceChainCommitteeProtocol) Initialize(
 ) (err error) {
 	log.L().Info("Initialize poll protocol", zap.Uint64("height", p.initGravityChainHeight))
 	var ds state.CandidateList
-	if ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight); err != nil {
-		for {
-			log.L().Error("calling committee,waiting for a while", zap.Int64("duration", int64(p.initialCandidatesInterval.Seconds())), zap.String("unit", " seconds"))
-			time.Sleep(p.initialCandidatesInterval)
-			ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight)
-			if err == nil || errors.Cause(err) != db.ErrNotExist {
-				break
-			}
+
+	for {
+		ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight)
+		if err == nil || errors.Cause(err) != db.ErrNotExist {
+			break
 		}
+		log.L().Error("calling committee,waiting for a while", zap.Int64("duration", int64(p.initialCandidatesInterval.Seconds())), zap.String("unit", " seconds"))
+		time.Sleep(p.initialCandidatesInterval)
 	}
+
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -311,6 +312,9 @@ type (
 
 		// SQLite3 is the config for SQLITE3
 		SQLITE3 SQLITE3 `yaml:"SQLITE3"`
+
+		// InitTryInterval is the config for committee init db
+		InitTryInterval int `yaml:"initTryInterval"`
```
zjshen14: Move the config to poll, and try to default to 3sec

action/protocol/poll/protocol.go
```diff
@@ -194,9 +198,26 @@ func (p *governanceChainCommitteeProtocol) Initialize(
...
+	interval, ok := InitTryInterval.(int)
+	if !ok {
+		log.L().Error("Interval read from config error")
+		interval = 15
+	}
 	log.L().Info("Initialize poll protocol", zap.Uint64("height", p.initGravityChainHeight))
 	var ds state.CandidateList
 	if ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight); err != nil {
+		errMsg := "bucket = electionNS doesn't exist: not exist in DB"
+		for strings.Contains(err.Error(), errMsg) {
```
zjshen14: errors.Cause()

action/protocol/poll/protocol.go
```diff
@@ -45,6 +46,9 @@ type GetEpochHeight func(uint64) uint64
 // GetEpochNum defines a function to get epoch number given a block height
 type GetEpochNum func(uint64) uint64
 
+// InitTryIntervalCtxKey is for InitTryInterval context
+type InitTryIntervalCtxKey struct{}
```
zjshen14: You can set the interval when constructing the protocol

No need to pass it over context

config/config.go
lzxm160: got it

action/protocol/poll/protocol.go
lzxm160: sure

lzxm160:
> Would you please test it against mainnet (sync from a clean state) after applying this change?

already tested

action/protocol/poll/protocol.go
```diff
@@ -197,6 +201,13 @@ func (p *governanceChainCommitteeProtocol) Initialize(
 	log.L().Info("Initialize poll protocol", zap.Uint64("height", p.initGravityChainHeight))
 	var ds state.CandidateList
 	if ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight); err != nil {
+		for err != nil && errors.Cause(err) == db.ErrNotExist {
```
zjshen14: L203-L209, how about the following?

```suggestion
    var err error
		for {
			log.L().Error("calling committee,waiting for a while", zap.Int64("duration", int64(p.initialCandidatesInterval.Seconds())), zap.String("unit", " seconds"))
			time.Sleep(p.initialCandidatesInterval)
			ds, err = p.delegatesByGravityChainHeight(p.initGravityChainHeight)
			if err == nil || errors.Cause(err) != db.ErrNotExist {
			    break
			}
		}
```

Issue #1234, opened 2019-06-03
Move total actions count from index service to basic block store


lizhefeng:
**What would you like to be added**:
Move total actions count from index builder to block store
**Why is this needed**:
Total action count should not be considered as an index. It is a block-level datum.

PR #1235, opened 2019-06-04
Move total actions count from index service to basic block store #1234

lzxm160:
work on #1234


lzxm160:
need check if user change server start with plugin and without plugin,actions number is equal or not

PR #1238, opened 2019-06-04

lzxm160:
work on #1194

According to #1210 (review)

Tests in my local computer shows that wait for about 6 minutes util initialize ok

This is step 2

### DIFF
--- a/blocksync/buffer.go
+++ b/blocksync/buffer.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
+	"github.com/iotexproject/iotex-core/action/protocol/poll"
 	"github.com/iotexproject/iotex-core/actpool"
 	"github.com/iotexproject/iotex-core/blockchain"
 	"github.com/iotexproject/iotex-core/blockchain/block"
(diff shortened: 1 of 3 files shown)


blocksync/buffer.go
```diff
@@ -78,7 +79,11 @@ func (b *blockBuffer) Flush(blk *block.Block) (bool, bCheckinResult) {
 		}
 		delete(b.blocks, heightToSync)
 		if err := commitBlock(b.bc, b.ap, b.cs, blk); err != nil && errors.Cause(err) != blockchain.ErrInvalidTipHeight {
-			l.Error("Failed to commit the block.", zap.Error(err), zap.Uint64("syncHeight", heightToSync))
+			if errors.Cause(err) == poll.ErrProposedDelegatesLength || errors.Cause(err) == poll.ErrDelegatesNotAsExpected {
```
zjshen14: Can you check if the error you want to ignore actually comes from here instead: https://github.com/iotexproject/iotex-core/blob/8f7700a735449676a2fa8fe87312f02be8138940/consensus/scheme/rolldpos/roundcalculator.go#L215

lzxm160: just msg error use errors.Wrapf ,not include a public define

zjshen14:
Approved, but can you help me to investigate why https://github.com/iotexproject/iotex-core/blob/4db795f9d3c919bb7d39e185a3091004ad742d10/blocksync/util.go#L17 doesn't throw error when the eth data is not available yet?

PR #1239, opened 2019-06-05
Update go-ethereum to fix arm64 build failure

dustinxie:


pkg/log/redirectStderr_armOS.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"log"
+	"os"
+	"syscall"
+)
+
+// redirectStderr to the file passed in
+func redirectStderr(f *os.File) error {
+	err := syscall.Dup3(int(f.Fd()), 2, 0)
```
dustinxie: arm's name is Dup3, not Dup2 as others

Issue #1242, opened 2019-06-07
change GetChainMeta api Response to include numActions when gateway plugin not enabled


lzxm160:
when gateway plugin is not enabled,response like this:
Response contents:
{
  "chainMeta": {
    "height": 381780,
    "epoch": {
      "num": 1061,
      "height": 381601,
      "gravityChainStartHeight": 7898600
    }
  }
}
change to:
**{
  "chainMeta": {
    "height": 381780,
    "numActions": 388097,
    "epoch": {
      "num": 1061,
      "height": 381601,
      "gravityChainStartHeight": 7898600
    }
  }
}**

PR #1260, opened 2019-06-11
change GetChainMeta api Response

lzxm160:
work on #1242


zjshen14:
Btw, GetByIndex should work as well. Could you take a look?

PR #1273, opened 2019-06-13
change transfer amount type to big int

JIUQIWANG:
I just change the type of transferAmount variable from uint64 to big.int to have a better ability to store large Integer value.

### DIFF
--- a/cli/ioctl/cmd/action/xrc20approve.go
+++ b/cli/ioctl/cmd/action/xrc20approve.go
@@ -8,8 +8,6 @@ package action
 
 import (
 	"fmt"
-	"math/big"
-	"strconv"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


cli/ioctl/cmd/action/xrc20approve.go
```diff
@@ -33,11 +31,7 @@ var Xrc20ApproveCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		transfer, err := strconv.ParseInt(args[1], 10, 64)
-		if err != nil {
-			return err
-		}
-		xrc20TransferAmount = uint64(transfer)
+		xrc20TransferAmount.SetString(args[1], 10)
```
zjshen14: So do you expect users to input 10^18 to represent 1 erc20 token?

zjshen14:
I think we need to ether fetch or let user provide the decimal of erc20, and you generate the big.Int for them.

PR #1275, opened 2019-06-13
refactor ioctl action and xrc20 commands

CoderZhi:
1. Introduce Flag to simplify the "set" and "read"
2. Refactor action commands, remove unnecessary returns
3. Add AMOUNT_IN_IOTEX to deploy command for the case that the constructor is payable
4. Add default gas price, gas limit, and nonce
5. Support reading contract as alias

### DIFF
--- a/cli/ioctl/cmd/action/actiondeploy.go
+++ b/cli/ioctl/cmd/action/actiondeploy.go
@@ -7,55 +7,37 @@
 package action
 
 import (
-	"fmt"
 	"math/big"
 
-	"github.com/spf13/cobra"
-	"go.uber.org/zap"
-
-	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/cli/ioctl/util"
-	"github.com/iotexproject/iotex-core/pkg/log"
+	"github.com/pkg/errors"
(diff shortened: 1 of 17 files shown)


cli/ioctl/cmd/action/actiondeploy.go
```diff
@@ -7,55 +7,37 @@
...
 	"github.com/iotexproject/iotex-core/cli/ioctl/util"
-	"github.com/iotexproject/iotex-core/pkg/log"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
 )
 
 // actionDeployCmd represents the action deploy command
 var actionDeployCmd = &cobra.Command{
-	Use:   "deploy -s SIGNER -b BYTE_CODE -l GAS_LIMIT [-p GAS_PRICE]",
+	Use:   "deploy [AMOUNT_IOTX] -s SIGNER -b BYTE_CODE [-l GAS_LIMIT] [-p GAS_PRICE]",
```
zjshen14: `[AMOUNT_IOTX]` becomes a must value to fill?

CoderZhi: default value is 0, we may leave it as optional.

PR #1281, opened 2019-06-14
Fix one bug after system interrupt during typing password

JIUQIWANG:
During user should typing the password, if there is a system interrupt like CTRL+C will cause terminal remain in hidden input status. So I create a listener to reset the terminal status when system interrupt approach.

### DIFF
--- a/cli/ioctl/cmd/account/accountimport.go
+++ b/cli/ioctl/cmd/account/accountimport.go
@@ -10,14 +10,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"strings"
-	"syscall"
 
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
-	"golang.org/x/crypto/ssh/terminal"
 	"gopkg.in/yaml.v2"
 
 	"github.com/iotexproject/iotex-core/cli/ioctl/cmd/config"
+	"github.com/iotexproject/iotex-core/cli/ioctl/util"
 	"github.com/iotexproject/iotex-core/cli/ioctl/validator"
 	"github.com/iotexproject/iotex-core/pkg/log"
 )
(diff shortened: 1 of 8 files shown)


cli/ioctl/cmd/account/accountimport.go
```diff
@@ -85,15 +83,11 @@ func writeToFile(alias, addr string) (string, error) {
 		alias), nil
 }
 func readPasswordFromStdin() (string, error) {
```
CoderZhi: this is a misleading function name. What it is intended to do is to ready private key, but not password. This is the reason spaces are trimmed. Moreover, this is a wrong logic, if the private key length is not valid, we are supposed to reject the private key. Let's keep it as it is for now, and fix the problem in another PR.

CoderZhi: For readPasswordFromStdin in line 119, you can replace it with the function you write "util.TypePassword()"

cli/ioctl/cmd/action/action.go
```diff
@@ -145,12 +143,12 @@ func execute(contract string, amount *big.Int, bytecode []byte) (err error) {
 
 func sendAction(elp action.Envelope, signer string) error {
 	fmt.Printf("Enter password #%s:\n", signer)
-	bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
+	bytePassword, err := util.TypePassword()
```
CoderZhi: misleading naming, this is a string, not byte

cli/ioctl/util/util.go
```diff
@@ -109,3 +115,37 @@ func StringToIOTX(amount string) (iotx string, err error) {
 	iotx = RauToString(amountInt, 18)
 	return
 }
+
+//TypePassword used to safely get password input
```
CoderZhi: space after "//"

cli/ioctl/util/util.go
```diff
@@ -109,3 +115,37 @@ func StringToIOTX(amount string) (iotx string, err error) {
...
+func TypePassword() (string, error) {
```
CoderZhi: let's rename it to ReadSecretFromStdin

cli/ioctl/cmd/account/accountimport.go
```diff
@@ -85,15 +83,11 @@ func writeToFile(alias, addr string) (string, error) {
...
-	passwordBytes, err := terminal.ReadPassword(int(syscall.Stdin))
+	password, err := util.ReadSecretFromStdin()
 	if err != nil {
 		log.L().Error("failed to get password", zap.Error(err))
 		return "", err
 	}
-	password := strings.TrimSpace(string(passwordBytes))
```
CoderZhi: as we discussed, we need to verify the length of the private key in a follow up PR.

PR #1291, opened 2019-06-19
Change epoch reward from 12.5K to 15K

zjshen14:
Implementation of https://github.com/iotexproject/iips/blob/master/iip-2.md

Also, cleanup some default rewarding settings in genesis.go, to make it consistent with the number we used in mainnet.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -1388,7 +1388,7 @@ func (bc *blockchain) createPollGenesisStates(ctx context.Context, ws factory.Wo
 func (bc *blockchain) updateAleutianEpochRewardAmount(ctx context.Context, ws factory.WorkingSet) error {
 	p, ok := bc.registry.Find(rewarding.ProtocolID)
 	if !ok {
-		return errors.New("rewarding protocol isn't found")
+		return nil
 	}
 	rp, ok := p.(*rewarding.Protocol)
 	if !ok {
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -28,7 +28,7 @@ import (
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/action/protocol"
 	"github.com/iotexproject/iotex-core/action/protocol/account"
-	"github.com/iotexproject/iotex-core/action/protocol/account/util"
+	accountutil "github.com/iotexproject/iotex-core/action/protocol/account/util"
 	"github.com/iotexproject/iotex-core/action/protocol/execution/evm"
 	"github.com/iotexproject/iotex-core/action/protocol/poll"
 	"github.com/iotexproject/iotex-core/action/protocol/rewarding"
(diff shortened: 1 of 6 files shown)


blockchain/genesis/genesis.go
```diff
@@ -56,14 +56,15 @@ func defaultConfig() Genesis {
...
-			EpochRewardStr:                 unit.ConvertIotxToRau(300000).String(),
+			EpochRewardStr:                 unit.ConvertIotxToRau(12500).String(),
+			AleutianEpochRewardStr:         unit.ConvertIotxToRau(15000).String(),
 			NumDelegatesForEpochReward:     100,
 			ExemptAddrStrsFromEpochReward:  []string{},
-			FoundationBonusStr:             unit.ConvertIotxToRau(2880).String(),
+			FoundationBonusStr:             unit.ConvertIotxToRau(80).String(),
 			NumDelegatesForFoundationBonus: 36,
-			FoundationBonusLastEpoch:       365,
+			FoundationBonusLastEpoch:       8760,
```
dustinxie: these change means we are changing epoch to 1 hour?

blockchain/blockchain.go
```diff
@@ -611,6 +611,12 @@ func (bc *blockchain) MintNewBlock(
 			ActionGasLimit: bc.config.Genesis.ActionGasLimit,
 			Registry:       bc.registry,
 		})
+
+	if newblockHeight == bc.config.Genesis.AleutianBlockHeight {
```
dustinxie: not sure if I understand correctly, only the delegate who mint the aleutian block will perform the below update, others won't update new value 15K into admin state?

i think we need a special 'governance' type action to update 15K, and in commitBlock() everyone take that action to update to 15K? that seems a lot of change, but consider in the future we may run into similar change

or we should pass AleutianHeight to rewarding protocol, same as what we did for evm protocol?

blockchain/genesis/genesis.go
zjshen14: No, just make it equivalent to 1yr if epoch length is 1hr, make the default consistent with the mainnet setting

blockchain/blockchain.go
zjshen14: > others won't update new value 15K into admin state?

Validator will update too

> that seems a lot of change, but consider in the future we may run into similar change

That's true, and it's not going to be a lot adjustment. Hence, it's not worthwhile yet of making an rewarding governance protocol

blockchain/blockchain.go
```diff
@@ -1051,6 +1057,11 @@ func (bc *blockchain) runActions(
 			Registry:       bc.registry,
 		})
 
+	if acts.BlockHeight() == bc.config.Genesis.AleutianBlockHeight {
+		if err := bc.updateAleutianEpochRewardAmount(ctx, ws); err != nil {
```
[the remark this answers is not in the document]
zjshen14: Good call. Let's skip

blockchain/blockchain.go
dustinxie: agreed, not worthwhile for now

PR #1297, opened 2019-06-21
Fix calculated tps = 0 issue

dustinxie:


api/api.go
```diff
@@ -252,13 +253,11 @@ func (api *Server) GetChainMeta(ctx context.Context, in *iotexapi.GetChainMetaRe
...
-	// if time duration is less than 1 second, we set it to be 1 second
-	if timeDuration < 1 {
-		timeDuration = 1
-	}
-
-	tps := numActions / timeDuration
+	t1 := time.Unix(blks[0].Timestamp.GetSeconds(), int64(blks[0].Timestamp.GetNanos()))
+	t2 := time.Unix(blks[len(blks)-1].Timestamp.GetSeconds(), int64(blks[len(blks)-1].Timestamp.GetNanos()))
+	// duration of tps window in milli-seconds
+	timeDuration := t2.Sub(t1) / time.Millisecond
```
CoderZhi: time duration will be 0 if len(blks) is 1. It is a corner case, which rarely happens.

api/api_test.go
```diff
@@ -718,9 +718,6 @@ func TestServer_GetActionsByAddress(t *testing.T) {
 		res, err := svr.GetActions(context.Background(), request)
 		require.NoError(err)
 		require.Equal(test.numActions, len(res.ActionInfo))
-		if test.numActions > 0 {
```
dustinxie: changed timestamp in addTestingBlocks(), which would caused returned ActionInfo[] order change since it is sorted by timestamp, so the test breaks

remove it since this is not a very important test

api/api.go
dustinxie: fixed in latest push

api/api.go
```diff
@@ -252,13 +253,17 @@ func (api *Server) GetChainMeta(ctx context.Context, in *iotexapi.GetChainMetaRe
...
-	timeDuration := blks[len(blks)-1].Timestamp.GetSeconds() - blks[0].Timestamp.GetSeconds()
-	// if time duration is less than 1 second, we set it to be 1 second
-	if timeDuration < 1 {
-		timeDuration = 1
+	t1 := time.Unix(blks[0].Timestamp.GetSeconds(), int64(blks[0].Timestamp.GetNanos()))
+	t2 := time.Unix(blks[len(blks)-1].Timestamp.GetSeconds(), int64(blks[len(blks)-1].Timestamp.GetNanos()))
+	// duration of time difference in milli-seconds
+	var timeDiff time.Duration
+	if t1 != t2 {
+		timeDiff = t2.Sub(t1) / time.Millisecond
```
CoderZhi: add 10 here

PR #1301, opened 2019-06-22
Add StreamFilterLogs() to enable querying logs by topics

dustinxie:
add a Responder interface to define the actual behavior when a new block arrives, this turns the Listener.Start() into common code. 

the Listener on chain just keep adding Responder -- each implements functionality for different streaming APIs

the original blockListener becomes a Responder, so is log filter added in this PR

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -524,6 +524,56 @@ func (api *Server) StreamBlocks(in *iotexapi.StreamBlocksRequest, stream iotexap
 	}
 }
 
+// StreamFilterLogs streams logs that match the filter condition
+func (api *Server) StreamFilterLogs(in *iotexapi.FilterLogsRequest, stream iotexapi.APIService_StreamFilterLogsServer) error {
+	if len(in.BlockHash) > 0 {
+		// query logs in a specific block
+		blk, err := api.bc.GetBlockByHash(hash.BytesToHash256(in.BlockHash))
+		if err != nil {
+			return status.Error(codes.InvalidArgument, err.Error())
+		}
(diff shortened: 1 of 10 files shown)


api/logfilter.go
```diff
@@ -0,0 +1,109 @@
...
+				logs = append(logs, log)
+			}
+		}
+	}
+	return logs
+}
+
+// match checks if a given log matches the filter
+func (l *LogFilter) match(log *iotextypes.Log) bool {
+	if l.Address != log.ContractAddress {
```
zjshen14: This match should be optional if Address is given

api/logfilter.go
```diff
@@ -0,0 +1,109 @@
...
+		return false
+	}
+	if len(l.Topics) > len(log.Topics) {
```
zjshen14: L85-L107 could be simplified into the logic that if there exists at least one topic in log.Topics that doesn't match a one in l.Topics (note that missing isn't considered as mismatch)

api/api.go
```diff
@@ -524,6 +524,56 @@ func (api *Server) StreamBlocks(in *iotexapi.StreamBlocksRequest, stream iotexap
 	}
 }
 
+// StreamFilterLogs streams logs that match the filter condition
+func (api *Server) StreamFilterLogs(in *iotexapi.FilterLogsRequest, stream iotexapi.APIService_StreamFilterLogsServer) error {
```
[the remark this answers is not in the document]
dustinxie: fixed in new push

api/logfilter.go
dustinxie: fixed in new push

Issue #1299, opened 2019-06-22
ioctl: add balance check when sending out any iotx (possibly check gas fee too)


guo:
**What would you like to be added**:

**Why is this needed**:

atorrente75:
also, a dry-run option would be great, in order to check if everything is ok (balance, gas limit, etc) ;-)

PR #1306, opened 2019-06-24
ioctl: add balance check when sending out any iotx (possibly check ga…

lzxm160:
…s fee too)

work  on #1299 

https://github.com/iotexproject/iotex-core/pull/1302#pullrequestreview-253226543

### DIFF
--- a/cli/ioctl/cmd/action/action.go
+++ b/cli/ioctl/cmd/action/action.go
@@ -159,7 +159,7 @@ func sendAction(elp action.Envelope, signer string) error {
 		log.L().Error("failed to sign action", zap.Error(err))
 		return err
 	}
-	err = isBalanceEnough(signer, sealed)
+	if err := isBalanceEnough(signer, sealed); err != nil {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cli/ioctl/cmd/action/action.go
```diff
@@ -133,6 +133,21 @@ func execute(contract string, amount *big.Int, bytecode []byte) (err error) {
 		log.L().Error("error when invoke an execution", zap.Error(err))
 		return
 	}
+	accountMeta, err := account.GetAccountMeta(signer)
```
zjshen14: Can you avoid dup code, and make this check as a method?

lzxm160: sure

cli/ioctl/cmd/action/action.go
```diff
@@ -159,6 +159,10 @@ func sendAction(elp action.Envelope, signer string) error {
 		log.L().Error("failed to sign action", zap.Error(err))
 		return err
 	}
+	err = isBalanceEnough(signer, sealed)
```
zjshen14: The code style convention is: 

```suggestion
	if err := isBalanceEnough(signer, sealed); err != nil {
```

Issue #1288, opened 2019-06-19
Flaky rolldpos test


zjshen14:
**Which jobs are failing**:
On laptop
**Which test(s) are failing**:
Only know inside rolldpos package
**Since when has it been failing**:
For quite a long time
**Testgrid link**:
N/A
**Reason for failure**:
N/A
**Anything else we need to know**:
Log: https://gist.github.com/zjshen14/e0224767d3bd5a38f2884e62b91680ea

zjshen14:
Encounter another related failure

```
2019-06-21T12:14:50.551-0700	ERROR	rolldpos/rolldposctx.go:475	invalid message type	{"height": 33, "epoch": 3, "round": 1, "proposer": "io1juvx5g063eu4ts832nukp4vgcwk2gnc5cu9ayd", "numProposals:": 0, "numLocks:": 0, "numCommits:": 0, "message": null}
github.com/iotexproject/iotex-core/consensus/scheme/rolldpos.(*rollDPoSCtx).Broadcast
	/Users/zjshen/Projects/go/src/github.com/iotexproject/iotex-core/consensus/scheme/rolldpos/rolldposctx.go:475
github.com/iotexproject/iotex-core/consensus/consensusfsm.(*ConsensusFSM).onBroadcastPreCommitEndorsement
	/Users/zjshen/Projects/go/src/github.com/iotexproject/iotex-core/consensus/consensusfsm/fsm.go:523
github.com/iotexproject/go-fsm.(*fsm).Handle
	/Users/zjshen/Projects/go/pkg/mod/github.com/iotexproject/go-fsm@v1.0.0/fsm.go:178
github.com/iotexproject/iotex-core/consensus/consensusfsm.(*ConsensusFSM).handle
	/Users/zjshen/Projects/go/src/github.com/iotexproject/iotex-core/consensus/consensusfsm/fsm.go:332
github.com/iotexproject/iotex-core/consensus/consensusfsm.(*ConsensusFSM).Start.func1
	/Users/zjshen/Projects/go/src/github.com/iotexproject/iotex-core/consensus/consensusfsm/fsm.go:231
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4e57c98]

goroutine 5157 [running]:
[...]

PR #1311, opened 2019-06-25
Flaky rolldpos test

lzxm160:
work on #1288

fix all test fail when run make test

### DIFF
--- a/consensus/scheme/rolldpos/rolldpos_test.go
+++ b/consensus/scheme/rolldpos/rolldpos_test.go
@@ -539,7 +539,7 @@ func TestRollDPoSConsensus(t *testing.T) {
 		}()
 		assert.NoError(t, testutil.WaitUntil(200*time.Millisecond, 60*time.Second, func() (bool, error) {
 			for _, chain := range chains {
-				if chain.TipHeight() < 5 {
+				if chain.TipHeight() < 48 {
 					return false, nil
 				}
 			}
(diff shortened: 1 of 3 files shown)


consensus/scheme/rolldpos/rolldpos_test.go
```diff
@@ -539,7 +539,7 @@ func TestRollDPoSConsensus(t *testing.T) {
 		}()
 		assert.NoError(t, testutil.WaitUntil(200*time.Millisecond, 60*time.Second, func() (bool, error) {
 			for _, chain := range chains {
-				if chain.TipHeight() < 48 {
+				if chain.TipHeight() < 5 {
```
zjshen14: This is one epoch test, so that 48 blocks are desired.

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -473,6 +473,7 @@ func (ctx *rollDPoSCtx) Broadcast(endorsedMsg interface{}) {
 	ecm, ok := endorsedMsg.(*EndorsedConsensusMessage)
 	if !ok {
 		ctx.loggerWithStats().Error("invalid message type", zap.Any("message", ecm))
+		return
```
zjshen14: good catch!

consensus/scheme/rolldpos/rolldpos_test.go
lzxm160: see

Issue #1024, opened 2019-04-15
estimate gas for deploy contract is zero


ququzone:
https://github.com/iotexproject/iotex-antenna/blob/3d9eabb6c2da4fc322bd0491c0d20bf64c0929ca/src/action/__test__/method.test.ts#L45

delete L50 for above file, then estimate gas is zero.

zjshen14:
Is it still an issue? @ququzone

ququzone:
estimate action gas error message

```
Error {
    code: 13,
    details: 'Out of gas',
    metadata: Metadata {
      _internal_repr: {
        date: Array [ … ],
        server: Array [ … ],
        'x-envoy-upstream-service-time': Array [ … ],
      },
    },
    message: '13 INTERNAL: Out of gas',
  }
```

PR #1327, opened 2019-06-28
estimate gas for deploy contract is zero 1024

lzxm160:
work on #1024 

**example for this change:**

grpcurl -v -plaintext -d '{"execution": {"amount": "0", "contract": "","data":"YIBgQFI0gBVhABBXYACA/VtQYQEjYAECYAKBYAAZFpBVUDNz//////////////////////////8WYAECYAOBYAAZFpBVUGADVGACVBdgBIFgABkWkFVQYQKugGEAZWAAOWAA8wBggGBAUmAENhBhAHhXYAA1fAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAkARj/////xaAYwzA4fsUYQB9V4BjKPNxqhRhALBXgGNrHXUrFGEA31eAY9S4OZIUYQESV4Bj2uqFxRRhAUVXgGPrb9lqFGEBiFdbYACA/Vs0gBVhAIlXYACA/VtQYQCSYQG7VltgQFGAgmAAGRZgABkWgVJgIAGRUFBgQFGAkQOQ81s0gBVhALxXYACA/VtQYQDFYQHBVltgQFGAghUVFRWBUmAgAZFQUGBAUYCRA5DzWzSAFWEA61dgAID9W1BhAPRhAddWW2BAUYCCYAAZFmAAGRaBUmAgAZFQUGBAUYCRA5DzWzSAFWEBHldgAID9W1BhASdhAd1WW2BAUYCCYAAZFmAAGRaBUmAgAZFQUGBAUYCRA5DzWzSAFWEBUVdgAID9W1BhAYZgBIA2A4EBkICANXP//////////////////////////xaQYCABkJKRkFBQUGEB41ZbAFs0gBVhAZRXYACA/VtQYQGdYQJ8VltgQFGAgmAAGRZgABkWgVJgIAGRUFBgQFGAkQOQ81tgA1SBVltgAGAEVGAAGRZgAVRgABkWFJBQkFZbYAJUgVZbYARUgVZbM3P//////////////////////////xZgAQKBc///////////////////////////FmABAhdgAYFgABkWkFVQYAFgAICDc///////////////////////////FnP//////////////////////////xaBUmAgAZCBUmAgAWAAIGAAYQEACoFUgWD/AhkWkIMVFQIXkFVQUFZbYAFUgVYAoWVienpyMFggibX5lHbWQrZqITwSzRmCB7LoE7scrzvXXiK+U16/XRMAKQ=="},"callerAddress":"io1juvx5g063eu4ts832nukp4vgcwk2gnc5cu9ayd"}' 127.0.0.1:14014 iotexapi.APIService.EstimateGasForAction


gasstation/gasstattion.go
```diff
@@ -100,10 +101,13 @@ func (gs *GasStation) EstimateGasForAction(actPb *iotextypes.Action) (uint64, er
 		if err != nil {
 			return 0, err
 		}
-		_, receipt, err := gs.bc.ExecuteContractRead(callerAddr, sc)
+		_, receipt, err := gs.bc.ExecuteContractRead(callerAddr, sc, true)
```
zjshen14: Instead of adding the flag, can we reset the gas limit here for the execution?

zjshen14:
In addition, estimate gas api could be changed to accept execution and caller address, instead of the sealed envelope, so that users don't need to sign an action to get gas estimated. Check how ReadContract api is defined. BTW, this is a breaking api change.

gasstation/gasstattion.go
lzxm160: This is a sealed action,cannot change any value inside

lzxm160:
> In addition, estimate gas api could be changed to accept execution and caller address, instead of the sealed envelope, so that users don't need to sign an action to get gas estimated. Check how ReadContract api is defined. BTW, this is a breaking api change.

First step,I need to change api define in iotex-proto

Issue #1303, opened 2019-06-23
iotex-core/cli/ioctl/ => iotex-core/ioctl/


raullenchai:
cli is the dir that should be removed.

PR #1328, opened 2019-06-28
Change cli directory

JIUQIWANG:
Working on #1303 
remove cli directory and also change all cli/ioctl/ references to ioctl/
use make build to test.


raullenchai:
pls make sure this won't break the nightly build

PR #1332, opened 2019-06-30
improve evmadaptor and contract test coverage

dustinxie:
This would enable solidity 0.5.1

Please focus on 2nd commit, which adds the 2 new interfaces. It also refactors the implementation of SetCode/State, GetState

The 1st commit is just to make the test code more readable by adding test vector, and also improve test coverage by testing both Trie and Trieless DB

### DIFF
--- a/action/protocol/execution/evm/contract_test.go
+++ b/action/protocol/execution/evm/contract_test.go
@@ -14,15 +14,14 @@ import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 
-	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-core/action/protocol"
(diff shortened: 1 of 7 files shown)


action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -526,64 +535,66 @@ func (stateDB *StateDBAdapter) GetCodeSize(evmAddr common.Address) int {
 // SetCode sets contract's code
 func (stateDB *StateDBAdapter) SetCode(evmAddr common.Address, code []byte) {
 	addr := hash.BytesToHash160(evmAddr[:])
-	if contract, ok := stateDB.cachedContract[addr]; ok {
-		contract.SetCode(hash.Hash256b(code), code)
-		return
-	}
 	contract, err := stateDB.getContract(addr)
```
dustinxie: if contract, ok := stateDB.cachedContract[addr]; ok{} is merged into stateDB.getContract(), to return the Contract object, then call SetCode()

the same is done for SetState(), GetState(), this simplifies the code, and as a result, removes getContractState() and setContractState()

zjshen14:
Can we split the PR into two. The first commit sound good to go while we wanna carefully examine the second one.

PR #1336, opened 2019-07-02
ioctl -s [address] should ask for private key instead of password (if…

lzxm160:
… wallet is not imported) 1334
work on #1334 

example for this change:

**ioctl action transfer io1mflp9m6hcgm2qcghchsdqj3z3eccrnekx9p0ms 1000 "" -s io1p0q5l362x9rulctvajfrhudqs958ndhftfj8kw
Enter private key #io1p0q5l362x9rulctvajfrhudqs958ndhftfj8kw:**

### DIFF
--- a/ioctl/cmd/action/action.go
+++ b/ioctl/cmd/action/action.go
@@ -16,7 +16,6 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/golang/protobuf/proto"
-	"github.com/iotexproject/go-pkgs/crypto"
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
 	"github.com/iotexproject/iotex-proto/golang/iotexapi"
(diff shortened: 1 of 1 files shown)


ioctl/cmd/action/action.go
```diff
@@ -214,3 +233,18 @@ func isBalanceEnough(address string, act action.SealedEnvelope) (err error) {
...
+func signerIsExist(signer string) bool {
+	address, err := address.FromString(signer)
+	if err != nil {
+		return false
+	}
+	// find the account in keystore
+	ks := keystore.NewKeyStore(config.ReadConfig.Wallet,
+		keystore.StandardScryptN, keystore.StandardScryptP)
+	for _, account := range ks.Accounts() {
+		if bytes.Equal(address.Bytes(), account.Address.Bytes()) {
```
zjshen14: Replace address with addr
```suggestion
		if address.Equal(addr, account.Address) {
```

Issue #1320, opened 2019-06-27
Make crypto computation tolerate client heterogeneity


zjshen14:
**What would you like to be added**:

1. Public key may start or not start with 04.
2. V of signature (the last byte) may be 0, 1 or 27, 28.

**Why is this needed**:

We're now increasing the support from the clients in different programming languages and libraries.

zjshen14:
1. createSecp256k1KeyPair in web3j will create a keypair, which could export the public key (64 bytes) without 04 prefix, but geth will.

2. web3j signs the message, and v of the signature is 27 or 28, but geth is 0 or 1 (subtracted by 27)

We use geth at our server side, but we have more clients written with different programming language and libraries.

It's better to provide tolerance to different version of pk, sig (or even something else), it is worthwhile to study how geth handle this first

lzxm160:
ok,no problem

PR #1342, opened 2019-07-03
Make crypto computation tolerate client heterogeneity 1320

lzxm160:
work on #1320

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -54,6 +54,10 @@ var (
 	ErrAction = errors.New("invalid action")
 )
 
+const (
+	secp256pubKeyLength = 65
+)
+
 // BroadcastOutbound sends a broadcast message to the whole network
 type BroadcastOutbound func(ctx context.Context, chainID uint32, msg proto.Message) error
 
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -289,6 +293,12 @@ func (api *Server) GetServerMeta(ctx context.Context,
 // SendAction is the API to send an action to blockchain.
 func (api *Server) SendAction(ctx context.Context, in *iotexapi.SendActionRequest) (res *iotexapi.SendActionResponse, err error) {
 	log.L().Debug("receive send action request")
+	if len(in.Action.SenderPubKey) == secp256pubKeyLength-1 {
```
zjshen14: Let's try not to tamper user data, but when verifying the signature (action.Verify), we do this massage on pk and sig before calling the verification cgo function. How do you think?

lzxm160: https://github.com/iotexproject/go-pkgs/pull/9 already did this,but test failed,I think the action's hash is not same as original action,have to hard fork to make this work

zjshen14:
Let's take the other approach in: https://github.com/iotexproject/go-pkgs/pull/9#pullrequestreview-259097124

PR #1344, opened 2019-07-04
Fix TestBatchRollback for windows

Frankonly:
The testutil.CleanupPath(t, testPath) meets error in TestBatchRollback on windows
`The process cannot access the file because it is being used by another process.`

This test case uses os.TempDir, so we can remove those CleanupPath()

Tested on windows and centos


Frankonly:
test failed on CircleCI

PR #1355, opened 2019-07-09
 Format output for alias Cmds

Frankonly:
Tested on windows and centos.

### DIFF
--- a/ioctl/cmd/alias/aliasimport.go
+++ b/ioctl/cmd/alias/aliasimport.go
@@ -96,6 +96,7 @@ func printImport(message importMessage) {
 		for _, alias := range message.Unimported {
 			fmt.Print(" " + alias.Name)
 		}
+		fmt.Println()
 	case "json":
 		out := output.Output{MessageType: output.Result, Message: message}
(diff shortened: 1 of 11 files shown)


ioctl/cmd/alias/aliasimport.go
```diff
@@ -24,58 +26,83 @@ var aliasImportCmd = &cobra.Command{
...
 	if err := ioutil.WriteFile(config.DefaultConfigFile, out, 0600); err != nil {
-		return "", fmt.Errorf("failed to write to config file %s", config.DefaultConfigFile)
+		return output.PrintError(output.WriteFileError,
+			fmt.Sprintf("failed to write to config file %s", config.DefaultConfigFile))
+	}
+	printImport(importMessage)
+	return nil
+}
+
+func printImport(message importMessage) {
```
zjshen14: I recommend you define a `String(format string)` interface for Message as well as a method for Output. Then . you implement the interface for all kinds of messages, which will make your code look cleaner

Frankonly: Cool, I will do that.

Issue #1326, opened 2019-06-28
Support estimate gas consumption when executing the action from ioctl


zjshen14:
**What would you like to be added**:

If user doesn't specify, call the estimate gas api and get the gas consumption estimation to fill the action to send.

**Why is this needed**:

For many smart contract calls, users don't know how much gas limit they should set.

lzxm160:
after https://github.com/iotexproject/iotex-core/issues/1024 this issue, ioctl can call EstimateActionGasConsumption api

PR #1361, opened 2019-07-10
Support estimate gas consumption when executing the action from ioctl…

lzxm160:
… 1326
work on #1326

### DIFF
--- a/ioctl/cmd/action/action.go
+++ b/ioctl/cmd/action/action.go
@@ -36,7 +36,7 @@ import (
 
 // Flags
 var (
-	gasLimitFlag = flag.NewUint64VarP("gas-limit", "l", 300000, "set gas limit")
+	gasLimitFlag = flag.NewUint64VarP("gas-limit", "l", 0, "set gas limit")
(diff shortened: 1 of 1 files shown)


ioctl/cmd/action/action.go
```diff
@@ -132,6 +156,15 @@ func execute(contract string, amount *big.Int, bytecode []byte) (err error) {
 		log.L().Error("error when invoke an execution", zap.Error(err))
 		return
 	}
+	if gasLimit == 300000 {
```
zjshen14: why 300000?

zjshen14:
else looks good

ioctl/cmd/action/action.go
lzxm160: Is there a better way to check if user sets gasLimitFlag?need to find out

zjshen14: Default to 0, as long as 0, we will estimate?

lzxm160: ok

PR #1373, opened 2019-07-15
refactor HeightChange, make it easy for adding new heights in future

dustinxie:
HeightChange is defined by heights in config file, so move its definition and code from EVM params to config.

To add a new height in future, we only need to add in config yaml, and modify a single file, no need to change EVM params function interface


config/heightchange.go
```diff
@@ -0,0 +1,50 @@
...
+
+// IsPost return true if height is after the height change
+func (hc *HeightChange) IsPost(height uint64, name HeightName) bool {
+	var h uint64
+	if name == Pacific {
+		h = hc.pacificHeight
+	} else if name == Aleutian {
+		h = hc.aleutianHeight
+	} else {
+		panic("invalid height name!")
```
zjshen14: log.Panic

config/heightchange.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return height >= h
+}
+
+// IsPre return true if height is before the height change
+func (hc *HeightChange) IsPre(height uint64, name HeightName) bool {
```
zjshen14: This method is not necessary

config/heightchange.go
```diff
@@ -0,0 +1,50 @@
...
+	Pacific = iota
+	Aleutian
+)
+
+type (
+	// HeightName is codename for height changes
+	HeightName int
+
+	// HeightChange lists heights at which certain fixes take effect
+	HeightChange struct {
```
zjshen14: rename it to UpgradeHeight?

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -55,15 +56,15 @@ type (
 		preimageSnapshot map[int]preimageMap
 		dao              db.KVStore
 		cb               db.CachedBatch
-		hc               *HeightChange
+		hc               *config.HeightChange
```
zjshen14: Let's avoid using ptr

config/heightchange.go
dustinxie: in the first code change, i don't have it, but then i found
!hc.IsPost(), you might miss the ! at front
so add this IsPre() for better readability

dustinxie:
> Good refactor.
> 
> Please search for PacificHeight and refactor the relevant code as well

yes, that's done

Issue #1360, opened 2019-07-10
Support check a single address' votes


zjshen14:
1. Needs to add an endpoint intake the query, the data should be pull-able from some api of the embedded election service.

2. Add the cli to integrate with the api.

PR #1380, opened 2019-07-17
Support check a single address votes 1360

lzxm160:
work on #1360  first step
https://github.com/iotexproject/iotex-core/issues/1360#issue-466431194

example：

grpcurl -v -plaintext -d '{"votee": "726f626f7462703030303136", "height": "8164200", "offset": 0,"limit":10}' 127.0.0.1:14014 iotexapi.APIService.GetVotes

Response contents:
{
  "buckets": [
    {
      "voter": "13f8ebd57488019a9652120a5b2be7554aa31fba",
      "votes": "136000000000000000000",
      "weightedVotes": "3010304432372308021120000",
      "remainingDuration": "168h0m0s"
    }
  ]
}

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -21,6 +21,7 @@ import (
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
+	"github.com/iotexproject/iotex-election/committee"
 	"github.com/iotexproject/iotex-proto/golang/iotexapi"
 	"github.com/iotexproject/iotex-proto/golang/iotextypes"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


api/api.go
```diff
@@ -592,6 +596,60 @@ func (api *Server) StreamLogs(in *iotexapi.StreamLogsRequest, stream iotexapi.AP
...
+	}
+	result, err := api.electionCommittee.ResultByHeight(height)
+	if err != nil {
+		return nil, err
+	}
+	name, err := hex.DecodeString(in.Votee)
+	if err != nil {
+		return nil, err
+	}
+	if len(name) != 12 {
```
zjshen14: Avoid use hard code number inline

Define 12 in advance

Btw, did you confirm 12 is the rule?

api/api.go
```diff
@@ -592,6 +596,60 @@ func (api *Server) StreamLogs(in *iotexapi.StreamLogsRequest, stream iotexapi.AP
...
+		return nil, errors.New("invalid candidate name")
+	}
+	votes := result.VotesByDelegate(name)
```
zjshen14: Hm..., given this, the start and count doesn’t make too much sense, because the we still fetch all votes of a delegate

But this could be a followup

api/api.go
lzxm160: Its the same code as https://github.com/iotexproject/iotex-election/blob/master/server/server.go#L266-L297,I'll try add some test

PR #1387, opened 2019-07-18
 Refactor error output for ioctl

Frankonly:
Using `output.NewError()` instead of `output.PrintError()` to deliver error message among functions.
```
func NewError(code ErrorCode, newInfo string, pre error)
```
`pre` error will be parsed and recomposed with `newInfo`. In this way, we can retain low layer function error message and deliver upper layer function the same time.

At the top layer of a command, new `output.PrintError()` is used to print error message in format.

This PR also fixes some minor problems that has not been merged into master branch (#1381) and refactor some ioctl command to adapt new error output mode.

After merging this PR, most clusters of ioctl command except `node` and `config` commands support format output (json).

Part of commands has been tested on CentOS and worked well.

#935

### DIFF
--- a/ioctl/cmd/action/action.go
+++ b/ioctl/cmd/action/action.go
@@ -92,7 +92,7 @@ func nonce(executor string) (uint64, error) {
 	}
 	accountMeta, err := account.GetAccountMeta(executor)
 	if err != nil {
-		return 0, err
+		return 0, output.NewError(0, "failed to get account meta", err)
 	}
 	return accountMeta.PendingNonce, nil
 }
@@ -114,7 +114,7 @@ func gasPriceInRau() (*big.Int, error) {
 	}
 	conn, err := util.ConnectToEndpoint(config.ReadConfig.SecureConnect && !config.Insecure)
 	if err != nil {
-		return nil, err
+		return nil, output.NewError(output.NetworkError, "failed to connect to endpoint", err)
 	}
 	defer conn.Close()
 	cli := iotexapi.NewAPIServiceClient(conn)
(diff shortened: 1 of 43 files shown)


ioctl/output/format.go
```diff
@@ -21,6 +21,10 @@ type ErrorCode int
 const (
 	// UndefinedError used when an error cat't be classified
 	UndefinedError ErrorCode = iota
+	// UpdateError used when an error occurs when running update command
+	UpdateError
```
zjshen14: Prefer to add errors to the end of the list for compatibility

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/iotexproject/go-pkgs v0.1.1-0.20190708233003-85a24189bbd4
 	github.com/iotexproject/iotex-address v0.2.0
 	github.com/iotexproject/iotex-election v0.1.10
-	github.com/iotexproject/iotex-proto v0.2.1-0.20190717000031-25b6ccc65ebd // indirect
+	github.com/iotexproject/iotex-proto v0.2.1-0.20190717000031-25b6ccc65ebd
```
zjshen14: Why proto change?

ioctl/cmd/action/action.go
```diff
@@ -92,7 +92,7 @@ func nonce(executor string) (uint64, error) {
 	}
 	accountMeta, err := account.GetAccountMeta(executor)
 	if err != nil {
-		return 0, err
+		return 0, output.NewError(0, "failed to get account meta", err)
```
zjshen14: What's error code 0? Unknown?

ioctl/cmd/action/action.go
```diff
@@ -272,25 +241,93 @@ func sendAction(elp action.Envelope, signer string) error {
...
-	return sendRaw(selp)
+	return SendRaw(selp)
 }
 
-func isBalanceEnough(address string, act action.SealedEnvelope) (err error) {
+// Execute sends signed execution transaction to blockchain
+func Execute(contract string, amount *big.Int, bytecode []byte) error {
+	gasPriceRau, err := gasPriceInRau()
+	if err != nil {
+		return output.NewError(0, "failed to get gas price", err)
```
zjshen14: ditto

go.mod
Frankonly: I have no idea. Go mod did that automatically.

ioctl/output/format.go
Frankonly: I have concerned about that, but UpdateError is special, so I put it at second place.

go.mod
Frankonly: When I run `make ioctl`

ioctl/cmd/action/action.go
Frankonly: The first kind of Error Message.

Frankonly: `UndefinedError`

ioctl/cmd/action/action.go
Frankonly: Using `0` when we don't hope previous error code to be overrode.
Here, the new error's code is got from `gasPrinceInRau`.

wowa1234567:
Http setting man müßte ihn brauser eingeben oder nur kurze frage?

Frank <notifications@github.com> schrieb am Mo., 22. Juli 2019, 17:33:

Issue #1386, opened 2019-07-18
Reduplicative ations with same hash shows in action list


anshuomomo:
https://iotexscan.io/address/io15ed8a4lyqgglm7z9wdq40kjksc0j5z0erw5v3x

![image](https://user-images.githubusercontent.com/16026265/61455915-6881be80-a997-11e9-977f-e663568ee57d.png)

![image](https://user-images.githubusercontent.com/16026265/61455929-72a3bd00-a997-11e9-9ac0-89bc56b24d1b.png)

zjshen14:
@lzxm160 Is this the bug you're talking about in https://github.com/iotexproject/iotex-core/pull/1384

lzxm160:
> @lzxm160 Is this the bug you're talking about in #1384

yes

wowa1234567:
Würde das gehen auf N26 app zu übertragen? 😅

PR #1390, opened 2019-07-19
Reduplicative ations with same hash shows in action list 1386

lzxm160:
work on #1386 
example:
server -config-path=/etc/iotex/config_testnet.yaml \
        -genesis-path=/etc/iotex/genesis_testnet.yaml \
        -plugin=gateway -reindex=true
or
server -config-path=/etc/iotex/config_testnet.yaml \
        -genesis-path=/etc/iotex/genesis_testnet.yaml \
        -plugin=gateway -reindex

The index of actions will reindex from height 1 and index 0 with '-reindex'

### DIFF
--- a/blockchain/indexbuilder.go
+++ b/blockchain/indexbuilder.go
@@ -44,10 +44,16 @@ type IndexBuilder struct {
 	cancelChan   chan interface{}
 	timerFactory *prometheustimer.TimerFactory
 	dao          *blockDAO
+	reindex      bool
+}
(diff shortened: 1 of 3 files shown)


lzxm160:
> lgtm,
> 
> but could you do some experiment with our mainnet data, to see if the problem is go fixed?
> 
> You could download the snapshot here: https://t.iotex.me/mainnet-data-with-idx-latest

sure

Issue #1394, opened 2019-07-22
Remove badger implementation of db


zjshen14:
Let's cleanup unused implementation to avoid unnecessary maintenance overhead

PR #1401, opened 2019-07-23
removed badger implementation of db #1394

koseoyoung:
removed badger implementation of db 
works on #1394


db/db.go
```diff
@@ -126,8 +126,5 @@ func (m *memKVStore) Commit(b KVStoreBatch) (e error) {
 
 // NewOnDiskDB instantiates an on-disk KV store
 func NewOnDiskDB(cfg config.DB) KVStore {
```
zjshen14: There's no need to have `NewOnDiskDB`. People should instantiate BoltDB explicitly.

db/db_test.go
```diff
@@ -104,11 +97,7 @@ func TestBatchRollback(t *testing.T) {
 		if kvboltDB, ok := kvStore.(*boltDB); ok {
 			err = kvboltDB.batchPutForceFail(bucket1, testK1[:], testV[:])
 			assert.NotNil(err)
-		} else if kvbadgerDB, ok := kvStore.(*boltDB); ok {
```
zjshen14: No need to have this if/else logic, as we only test against boltdb impl

zjshen14: Also you could give up the internal func `NewOnDiskDB`

db/db.go
koseoyoung: NewOnDiskDB function is now used in other files, so I think it is little bit tricky to change it whole stuff

db/db_test.go
koseoyoung: I'll get rid of if logic

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -38,6 +38,14 @@ var (
 		[]string{},
 	)
 
+	consensusHeightMtc = prometheus.NewGaugeVec(
```
zjshen14: Irrelevant?

koseoyoung: sorry I fixed it!

PR #1402, opened 2019-07-24
Using zero address for ioctl when failed to get caller to read contract

Frankonly:


ioctl/cmd/action/action.go
```diff
@@ -284,7 +284,7 @@ func Execute(contract string, amount *big.Int, bytecode []byte) error {
 func Read(contract address.Address, bytecode []byte) (string, error) {
 	caller, err := signer()
 	if err != nil {
-		return "", output.NewError(0, "failed to get signer address", err)
+		caller = "io1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqd39ym7" // zero address after encode in IoTeX blockchain
```
zjshen14: define it somewhere as a const?

Frankonly: ok

PR #1404, opened 2019-07-25
Reduplicative ations with same hash shows in action list 1386

lzxm160:
work on #1386
patch for #1390

### DIFF
--- a/blockchain/indexbuilder.go
+++ b/blockchain/indexbuilder.go
@@ -182,10 +182,10 @@ func (ib *IndexBuilder) initAndLoadActions() error {
 			return err
 		}
 	} else {
-		if err = ib.dao.deleteBucket(blockAddressActionMappingNS); err != nil {
+		if err = ib.dao.kvstore.Delete(blockAddressActionMappingNS, nil); err != nil {
 			return err
 		}
-		if err = ib.dao.deleteBucket(blockAddressActionCountMappingNS); err != nil {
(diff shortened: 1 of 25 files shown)


blockchain/indexbuilder.go
```diff
@@ -181,6 +181,13 @@ func (ib *IndexBuilder) initAndLoadActions() error {
 		if err != nil {
 			return err
 		}
+	} else {
+		if err = ib.dao.deleteBucket(blockAddressActionMappingNS); err != nil {
```
zjshen14: So if we want to reindex, we need to cleanup existing addr <-> act mapping?

Why don't we need to cleanup other index data?

db/db.go
```diff
@@ -38,6 +38,8 @@ type KVStore interface {
 	Delete(string, []byte) error
 	// Commit commits a batch
 	Commit(KVStoreBatch) error
+	// DeleteBucket delete a Bucket
+	DeleteBucket(key []byte) error
```
zjshen14: Can we avoid adding a new api. Instead, `Delete(string, []byte) error` means delete a whole bucket if the second param is not given

lzxm160: sure

blockchain/indexbuilder.go
lzxm160: The index of all ations will be covered when reindex

PR #1411, opened 2019-07-27
fix ci nightly build bug

JIUQIWANG:
I have done research on every PR on Jun 24th which is the date we failed the ci test

That basically because previously our general ActionGasLimit is 10000000 and after we remove the per action gas limit cap it become to 5000000.

### DIFF
--- a/tools/minicluster/minicluster.go
+++ b/tools/minicluster/minicluster.go
@@ -93,6 +93,7 @@ func main() {
 			config.Network.BootstrapNodes = []string{}
 			config.Network.MasterKey = "bootnode"
 		}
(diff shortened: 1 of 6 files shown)


tools/minicluster/minicluster.go
```diff
@@ -93,6 +93,7 @@ func main() {
 			config.Network.BootstrapNodes = []string{}
 			config.Network.MasterKey = "bootnode"
 		}
+		config.Genesis.AleutianBlockHeight = 1
```
zjshen14: also set pacific height to 1 as well

JIUQIWANG: ok

Issue #1395, opened 2019-07-22
Support partitioning db by block height


zjshen14:
**What would you like to be added**:

Nowadays all block data are written into the same db file. It is not friendly to daily backup, which theoretically only needs to backup incremental block data.

**Why is this needed**:

Let's make the block data storage to partitioned by block height physically. For example, 1-8640 blocks are stored in chain-1.db, and 8641-19280 blocks are stored in chain-2.db

lzxm160:
1421,split by db size

PR #1412, opened 2019-07-27
Support partitioning db by block height 1395

lzxm160:
work on #1395

Block header,body ,footer and receipt map to different dbs according to block height,
other datas still store in default db,chain.db.
The default split length is 0，it means don't need to split db.
The split height config means it will split to diff db since that height.

If split length is 100000,every db's size is about 400M.
The new db's name like this:
chain-1.db
chain-2.db

### DIFF
--- a/blockchain/blockdao.go
+++ b/blockchain/blockdao.go
@@ -8,6 +8,10 @@ package blockchain
 
 import (
 	"context"
+	"fmt"
+	"path"
+	"strings"
+	"sync"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/iotexproject/go-pkgs/hash"
@@ -19,6 +23,7 @@ import (
 
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/blockchain/block"
+	"github.com/iotexproject/iotex-core/config"
 	"github.com/iotexproject/iotex-core/db"
 	"github.com/iotexproject/iotex-core/pkg/cache"
 	"github.com/iotexproject/iotex-core/pkg/compress"
@@ -52,6 +57,7 @@ var (
 	heightPrefix             = []byte("he.")
 	actionFromPrefix         = []byte("fr.")
 	actionToPrefix           = []byte("to.")
+	blockHashDBPrefix        = []byte("bhdb.")
 )
 
 var (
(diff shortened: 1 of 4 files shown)


blockchain/blockdao.go
```diff
@@ -700,3 +822,11 @@ func deleteActions(dao *blockDAO, blk *block.Block, batch db.KVStoreBatch) error
 
 	return nil
 }
+
+// getDBIndex get db index from block height
+func getDBIndex(hei uint64, split uint64) int {
+	if split == 0 {
```
zjshen14: Is it to support when SplitDBLength is not set (default 0), all data will be written into the same db?

blockchain/blockdao.go
```diff
@@ -434,9 +447,20 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
...
+	whichDB := getDBIndex(blk.Height(), dao.cfg.SplitDBLength)
+	kv, err := dao.getDBFromIndex(whichDB)
+	if err != nil {
+		return err
+	}
+	err = kv.Commit(batchForBlock)
+	if err != nil {
+		return err
+	}
+	blockHashDBKey := append(blockHashDBPrefix, hash[:]...)
```
zjshen14: When lookup by blockHash, can we just lookup for height first? We will know which db to reach as long as the height is known.

blockchain/blockdao.go
```diff
@@ -606,6 +652,82 @@ func (dao *blockDAO) deleteTipBlock() error {
...
+	cfg := dao.cfg
+	withSuffix = path.Base(cfg.DbPath)
+	suffix = path.Ext(withSuffix)
+	name = strings.TrimSuffix(withSuffix, suffix)
+	name += fmt.Sprintf("-%d", index) + ".db"
+	cfg.DbPath = path.Dir(cfg.DbPath) + "/" + name
+
+	kvstore = db.NewBoltDB(cfg)
+	dao.kvstores.Store(index, kvstore)
+	err = kvstore.Start(context.Background())
```
zjshen14: A follow up may be closing the keystore when it's not been used for a while

blockchain/blockdao.go
lzxm160: sure,now default is 100000,I'll change to 0

blockchain/blockdao.go
lzxm160: see,I'll try delete this key

blockchain/blockdao.go
lzxm160: I  think it will close when call dao.lifecycle.Stop()

lzxm160:
> And think of the compatibility or migration plan. Currently all the blocks are stored in the one db, which couldn't be found with you new height->db mapping

I'll think about it

Issue #1388, opened 2019-07-18
Chain has moved to tip height but consensus is stuck S_ACCEPT_PRECOMMIT_ENDORSEMENT


zjshen14:
It should be S_ACCEPT_PRECOMMIT_ENDORSEMENT of a previous block height. Suspect that the clibration doesn't happen on block sync.

raullenchai:
Any update?

dustinxie:
Zhi fixed in https://github.com/iotexproject/iotex-core/commit/c06239d6a6b866670769ee8e31c33f3296d63fd2

PR #1415, opened 2019-07-29
Chain has moved to tip height but consensus is stuck S_ACCEPT_PRECOMM…

lzxm160:
…IT_ENDORSEMENT 1388

work on #1388

### DIFF
--- a/consensus/consensusfsm/fsm.go
+++ b/consensus/consensusfsm/fsm.go
@@ -202,6 +202,13 @@ func NewConsensusFSM(cfg Config, ctx Context, clock clock.Clock) (*ConsensusFSM,
 			[]fsm.State{
 				sAcceptPreCommitEndorsement,
 				sPrepare, // reach consensus, start next epoch
+			}).
+		AddTransition(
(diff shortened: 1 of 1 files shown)


zjshen14:
Why will this fix the bug?

lzxm160:
> Why will this fix the bug?

I think it's this error caused stuck:
{"level":"debug","ts":1564397364.4860137,"caller":"consensusfsm/fsm.go:348","msg":"consensus state transition could find the match","networkAddr":"0.0.0.0:4692","height":3,"epoch":1,"round":0,"proposer":"io19kshh892255x4h5ularvr3q3al2v8cgl80fqrt","src":"S_ACCEPT_PRECOMMIT_ENDORSEMENT","evt":"E_RECEIVE_LOCK_ENDORSEMENT","error":"transition and destinations are found for state S_ACCEPT_PRECOMMIT_ENDORSEMENT, event E_RECEIVE_LOCK_ENDORSEMENT: unable to find a valid transition","errorVerbose":"unable to find a valid transition

Issue #1346, opened 2019-07-04
Keep track of account state changes


lizhefeng:
**What would you like to be added**:
A journal-like structure to keep the record of state changes.
**Why is this needed**:
We want to have an overview of all the accounts (including account balance) currently existing on the chain. But we don't have the record of state changes (mostly the state changes resulted from executions). Need to figure out a way to keep track of the account state changes in all kinds of actions. Please take a look at https://github.com/ethereum/go-ethereum/blob/36f81118f6d00713dbf8d1caf168ad937e0f1a90/core/state/statedb.go#L90 for your reference.

raullenchai:
@Frankonly what is the progress on this one?

zjshen14:
@lizhefeng if you want to iterate through all the accounts you can use the cursor from boltdb: https://github.com/etcd-io/bbolt. Check the readme

PR #1417, opened 2019-07-30
Add transfer log in stateDB for evm transfer tracking

Frankonly:
One solution of evm transfer tracking. #1346 
I define a new stateDB namespace "transferLog" to storage transfer logs. The key is execution hash and the value is a set of transfer logs that records all transfers happened in this execution.

### DIFF
--- a/state/transferlog.go
+++ b/state/transferlog.go
@@ -15,7 +15,7 @@ type TransferLogs struct {
 	Transfers []TransferLog `json:"transfers"`
 }
 
-// TransferLogs represents one evm transfer trace
+// TransferLog represents one evm transfer trace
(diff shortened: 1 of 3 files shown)


state/transferlog.go
```diff
@@ -0,0 +1,54 @@
...
+// License 2.0 that can be found in the LICENSE file.
+
+package state
+
+import (
+	"encoding/json"
+)
+
+// TransferLogs represents a group of evm transfer traces
+type TransferLogs struct {
```
zjshen14: We use protobuf internally for serde

zjshen14:
We decided to go with another approach

PR #1421, opened 2019-08-01
Support partitioning db by block height 1395

lzxm160:
work on #1395

Block header,body ,footer and receipt map to different dbs according to block height,
other datas still store in default db：chain.db.
The default split size is 640M，it don't need to split db when it's 0.
The split height config means it will split to diff db since that height.

The new db's name like this:
chain-00000001.db
chain-00000002.db

### DIFF
--- a/blockchain/blockdao.go
+++ b/blockchain/blockdao.go
@@ -62,7 +62,6 @@ var (
 	actionFromPrefix         = []byte("fr.")
 	actionToPrefix           = []byte("to.")
 	blockHeightToFileKey     = []byte("bhf.")
-	receiptHeightToFileKey   = []byte("rhf.")
 )
 
 var (
@@ -433,7 +432,7 @@ func (dao *blockDAO) getReceiptByActionHash(h hash.Hash256) (*action.Receipt, er
 		return nil, errors.Wrapf(err, "failed to get receipt index for action %x", h)
 	}
 	height := enc.MachineEndian.Uint64(heightBytes)
-	kvstore, _, err := dao.getDBFromHeight(height, receiptHeightToFileKey)
+	kvstore, _, err := dao.getDBFromHeight(height)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


blockchain/blockdao.go
```diff
@@ -52,6 +61,8 @@ var (
 	heightPrefix             = []byte("he.")
 	actionFromPrefix         = []byte("fr.")
 	actionToPrefix           = []byte("to.")
+	blockHeightToFileKey     = []byte("bhf.")
```
zjshen14: I wonder if it's necessary two mappings? block and receipt of the same height should be in the same file.

blockchain/blockdao.go
```diff
@@ -123,13 +138,50 @@ func (dao *blockDAO) Start(ctx context.Context) error {
...
 }
+func (dao *blockDAO) initStores() error {
+	cfg := dao.cfg
+	model, dir := getFileNameAndDir(cfg.DbPath)
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+	var maxN uint64
+	for _, file := range files {
```
zjshen14: what if the directory mix some other files in? for example, we currently put chain.db and trie.db in the same dir

blockchain/blockdao.go
```diff
@@ -606,6 +701,129 @@ func (dao *blockDAO) deleteTipBlock() error {
 	return dao.kvstore.Commit(batch)
 }
 
+// getDBFromHash returns db of this block stored
+func (dao *blockDAO) getDBFromHash(h hash.Hash256) (db.KVStore, uint64, error) {
+	hei, err := dao.getBlockHeight(h)
```
zjshen14: hei->height

blockchain/blockdao.go
```diff
@@ -606,6 +701,129 @@ func (dao *blockDAO) deleteTipBlock() error {
...
+	if blkHeight <= dao.cfg.SplitDBHeight {
+		return dao.kvstore, 0, nil
+	}
+	topIndex := dao.topIndex.Load().(uint64)
+	file, dir := getFileNameAndDir(dao.cfg.DbPath)
+	if err != nil {
+		return
+	}
+	longFileName := dir + "/" + file + fmt.Sprintf("-%08d", topIndex) + ".db"
+	dat, err := os.Stat(longFileName)
```
zjshen14: Can you measure the overhead of this call, which seems to have to execute on every put block now

blockchain/blockdao.go
```diff
@@ -473,11 +542,16 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
 
 // putReceipts store receipt into db
 func (dao *blockDAO) putReceipts(blkHeight uint64, blkReceipts []*action.Receipt) error {
+	kvstore, fileindex, err := dao.getTopDB(blkHeight)
```
zjshen14: oh, now I can see it could have the corner case that a block and its receipt are in different files.

hm..., is this going to over complex?

blockchain/blockdao.go
```diff
@@ -606,6 +701,129 @@ func (dao *blockDAO) deleteTipBlock() error {
...
+		return dao.kvstore, 0, nil
+	}
+	topIndex := dao.topIndex.Load().(uint64)
+	file, dir := getFileNameAndDir(dao.cfg.DbPath)
+	if err != nil {
+		return
+	}
+	longFileName := dir + "/" + file + fmt.Sprintf("-%08d", topIndex) + ".db"
+	dat, err := os.Stat(longFileName)
+	if err != nil {
```
zjshen14: Is it possible that the error is not because of non-existing file?

blockchain/blockdao.go
```diff
@@ -606,6 +701,129 @@ func (dao *blockDAO) deleteTipBlock() error {
...
+		// db file is not exist,this will create
+		return dao.openDB(topIndex)
+	}
+	// dao.cfg.SplitDBSize need convert to M
+	if uint64(dat.Size()) > dao.cfg.SplitDBSize*1024*1024 {
```
zjshen14: Better to define a SplitDBSize() = SplitDBSizeMB * 1024 * 1024 in config

blockchain/blockdao.go
```diff
@@ -606,6 +701,129 @@ func (dao *blockDAO) deleteTipBlock() error {
...
+	topIndex := dao.topIndex.Load().(uint64)
+	file, dir := getFileNameAndDir(dao.cfg.DbPath)
+	if err != nil {
+		return
+	}
+	longFileName := dir + "/" + file + fmt.Sprintf("-%08d", topIndex) + ".db"
+	dat, err := os.Stat(longFileName)
+	if err != nil {
+		// db file is not exist,this will create
+		return dao.openDB(topIndex)
```
zjshen14: It could potentially race on opening the new db file by two go routine

blockchain/blockdao.go
lzxm160: putblock and putreceipt in different method,I think sometimes they will be put in different db file

blockchain/blockdao.go
lzxm160: only find files like this:chain-xxxxxxxx.db

blockchain/blockdao.go
lzxm160: ok

blockchain/blockdao.go
zjshen14: blockHeightToFileKey -> heightToFilePrefix

bhf. -> hf.

blockchain/blockdao.go
```diff
@@ -62,25 +72,32 @@ var (
 		},
 		[]string{"result"},
 	)
+	modelLen  = len("00000000.db")
+	suffexLen = len(".db")
```
zjshen14: suffexLen -> suffixLen

blockchain/blockdao.go
```diff
@@ -62,25 +72,32 @@ var (
 		},
 		[]string{"result"},
 	)
+	modelLen  = len("00000000.db")
```
zjshen14: modelLen -> patternLen?

blockchain/blockdao.go
```diff
@@ -455,29 +533,33 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
...
+		return
 	}
-	return dao.kvstore.Commit(batch)
+	err = dao.kvstore.Commit(batch)
+	return
 }
 
 // putReceipts store receipt into db
-func (dao *blockDAO) putReceipts(blkHeight uint64, blkReceipts []*action.Receipt) error {
+func (dao *blockDAO) putReceipts(blkHeight uint64, blkReceipts []*action.Receipt, kvstore db.KVStore) error {
```
zjshen14: Isn't an easier fix that preventing opening a new db file when putting receipt?

lzxm160: need to think about it

blockchain/blockdao.go
```diff
@@ -606,6 +705,135 @@ func (dao *blockDAO) deleteTipBlock() error {
...
+	if err != nil && os.IsNotExist(err) {
+		// db file is not exist,this will create
+		return dao.openDB(topIndex)
+	}
+	// other errors except file is not exist
+	if err != nil {
+		return
+	}
+	// dao.cfg.SplitDBSize need convert to M
+	if uint64(dat.Size()) > dao.cfg.SplitDBSize() {
```
zjshen14: You can choose to only create a new db file when you want to insert block, but not receipt to make sure block and receipt are always of the same height are the same db file. This is going to prevent your passing kvstore reference around

zjshen14: BTW, should this check be done earlier? If we want to create a new db file, we wouldn't want to open the previous one (L733)?

Issue #1298, opened 2019-06-21
Sporadic TestBroadcast failure


zjshen14:
**Which jobs are failing**:

Local `make`

**Which test(s) are failing**:

TestBroadcast

**Since when has it been failing**:

Not sure

**Testgrid link**:

N/A

**Reason for failure**:

Not sure

**Anything else we need to know**:

Error log:

```
--- FAIL: TestBroadcast (21.99s)
    require.go:794:
        	Error Trace:	agent_test.go:73
        	Error:      	Received unexpected error:
        	            	timed out
        	            	github.com/iotexproject/iotex-core/testutil.init.ializers
        	            		/Users/zjshen/Projects/go/src/github.com/iotexproject/iotex-core/testutil/wait.go:16
        	            	runtime.main
        	            		/usr/local/Cellar/go/1.12.5/libexec/src/runtime/proc.go:188
        	            	runtime.goexit
        	            		/usr/local/Cellar/go/1.12.5/libexec/src/runtime/asm_amd64.s:1337
        	Test:       	TestBroadcast
2019-06-21T12:04:07.659-0700	INFO	go-p2p@v0.2.10/p2p.go:334	P2p host started.	{"address": ["/ip4/127.0.0.1/tcp/37186/p2p/12D3KooWQTP7bbdbSUQTRLJ7ZB7Rzi416cT3BGdMd9C4XuepSg83"], "secureIO": true, "gossip": true}
2019-06-21T12:04:07.664-0700	INFO	go-p2p@v0.2.10/p2p.go:334	P2p host started.	{"address": ["/ip4/127.0.0.1/tcp/31993/p2p/12D3KooWDDRnfR7mpMZ2eKk7QD3fqKSoP3o9ZhFvhbtgXLvpbGtt"], "secureIO": true, "gossip": true}
2019-06-21T12:04:07.690-0700	INFO	p2p/agent.go:250	Connected bootstrap node.	{"address": "/ip4/127.0.0.1/tcp/37186/p2p/12D3KooWQTP7bbdbSUQTRLJ7ZB7Rzi416cT3BGdMd9C4XuepSg83"}
[...]

PR #1437, opened 2019-08-07
Sporadic TestBroadcast failure 1298

lzxm160:
work on #1298


p2p/agent_test.go
```diff
@@ -48,7 +48,7 @@ func TestBroadcast(t *testing.T) {
 		}
 	}
 	u := func(_ context.Context, _ uint32, _ peerstore.PeerInfo, _ proto.Message) {}
-	bootnodePort := testutil.RandomPort()
+	bootnodePort := 14689
```
zjshen14: Can we still random, but if we randomly get a port is used before, we randomly pick again?

lzxm160: I think it's too much complexity,we have to save all the port opened

Issue #1370, opened 2019-07-12
Full state history node


zjshen14:
Currently a consensus node only keeps the latest version of states, which is enough for consensus, and serving most use cases.

However, it might grow more use cases that need to access the account/contract's state at a certain block height. To support this efficiently, a node has to keep all versions of states. It may not necessarily have every node to remember full state history, but add it as an option for some node.

Or it's also possible not full state history, but last 100 block, 1k, 10k block height for example, depending on the concrete use cases.

lzxm160:
Need to research trie's way about state history

lzxm160:
changed code in branch 13704,tests show that still cannot get history of evm,still needs to work on it

lzxm160:
A lot of changes happend to iotex-core,need to retest state history

lzxm160:
According to https://github.com/lzxm160/iotex-core/tree/13709's test:
1. account's history state is good
2. evm can get storage history through readstate api adding a new command
3. readcontract api is not good,need to change StateDBAdapter later

lzxm160:
This will be retest today based on latest code

lzxm160:
Based on the disscussion with dustinxie,I'll change this implementation and retest

lzxm160:
The code has been modified,need test a lot and optimization

lzxm160:
https://github.com/iotexproject/iotex-core/tree/historystate for dustin check and discuss later

lzxm160:
https://github.com/iotexproject/iotex-core/tree/historystate modify according to comments,
https://github.com/lzxm160/iotex-core branch 137023 with api test

lzxm160:
https://github.com/lzxm160/iotex-core branch 137023 with api test
according to #1569's rangeindex modified codes,test failed with disgest match problem,need to find out

lzxm160:
https://github.com/lzxm160/iotex-core branch 137023 with api test
according to #1569's new commit https://github.com/iotexproject/iotex-core/commit/b58f62cabac6d92e15f6b6a9ca022c308884b588,add rangeindex and refactor again

lzxm160:
Based on the disscussion with dustinxie today,account's history almost ok,need to adjust a little,trie's history still pending

lzxm160:
https://github.com/iotexproject/iotex-core/tree/trienode add method to return all trie's nodes for the next step to record reference to trie's node

lzxm160:
refer to #1604 and #1619

[...]

PR #1439, opened 2019-08-07
Full state history node 1370

lzxm160:
work on #1370


zjshen14:
Per discussion offline, this won't be a breaking change

lzxm160:
need to add storage trie's history

Issue #1441, opened 2019-08-08
Make GetBlockMetas API only read block header


zjshen14:
Currently it uses the getBlock method form blockchain interface, which is more expensive than reading a header, especially when there are a lot of transactions. Blockchain interface has the getBlockHeader method, which only reads the the header data. We should migrate to use this method instead. One challenge is that the number of transactions is not in the header. We need to add a field. Also, this will change the data structure of header, which is a breaking change. It needs to be gated by the Bering height

raullenchai:
@5sWind

PR #1445, opened 2019-08-10
Make GetBlockMetas API only read block header. #1441

5sWind:
Related to issue #1441.
From my understanding, the number of all transactions are even with the block height.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -39,6 +39,7 @@ import (
 	"github.com/iotexproject/iotex-core/blockchain"
 	"github.com/iotexproject/iotex-core/blockchain/block"
 	"github.com/iotexproject/iotex-core/config"
+	"github.com/iotexproject/iotex-core/db"
 	"github.com/iotexproject/iotex-core/dispatcher"
 	"github.com/iotexproject/iotex-core/gasstation"
 	"github.com/iotexproject/iotex-core/pkg/log"
@@ -909,7 +910,7 @@ func (api *Server) getActionsByBlock(blkHash string, start uint64, count uint64)
 	}, nil
 }
 
-// getBlockMetas gets block within the height range
+// getBlockMetas returns blockmetas response within the height range
 func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlockMetasResponse, error) {
 	if count > api.cfg.API.RangeQueryLimit {
 		return nil, status.Error(codes.InvalidArgument, "range exceeds the limit")
@@ -921,32 +922,14 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
 	}
 	var res []*iotextypes.BlockMeta
 	for height := start; height <= tipHeight && count > 0; height++ {
-		blk, err := api.bc.GetBlockByHeight(height)
-		if err != nil {
-			return nil, status.Error(codes.NotFound, err.Error())
-		}
-
-		hash := blk.HashBlock()
-		ts, _ := ptypes.TimestampProto(blk.Header.Timestamp())
-		txRoot := blk.TxRoot()
-		receiptRoot := blk.ReceiptRoot()
-		deltaStateDigest := blk.DeltaStateDigest()
-		transferAmount := getTranferAmountInBlock(blk)
-		logsBloom := blk.Header.LogsBloomfilter()
-
-		blockMeta := &iotextypes.BlockMeta{
-			Hash:             hex.EncodeToString(hash[:]),
-			Height:           blk.Height(),
-			Timestamp:        ts,
-			NumActions:       int64(len(blk.Actions)),
-			ProducerAddress:  blk.ProducerAddress(),
-			TransferAmount:   transferAmount.String(),
-			TxRoot:           hex.EncodeToString(txRoot[:]),
-			ReceiptRoot:      hex.EncodeToString(receiptRoot[:]),
-			DeltaStateDigest: hex.EncodeToString(deltaStateDigest[:]),
-		}
-		if logsBloom != nil {
-			blockMeta.LogsBloom = hex.EncodeToString(logsBloom.Bytes())
+		blockMeta, err := api.getBlockMetasByHeader(height)
+		if errors.Cause(err) == action.ErrNotFound {
+			blockMeta, err = api.getBlockMetasByBlock(height)
+			if err != nil {
+				return nil, err
+			}
+		} else if err != nil {
+			return nil, err
 		}
 		res = append(res, blockMeta)
 		count--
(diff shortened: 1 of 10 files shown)


blockchain/blockchain.go
```diff
@@ -1424,3 +1425,11 @@ func calculateLogsBloom(cfg config.Config, height uint64, receipts []*action.Rec
 	}
 	return bloom
 }
+
+func calculateTxNumber(cfg config.Config, height uint64) uint64 {
+	if height < cfg.Genesis.BeringBlockHeight {
+		return 0
+	}
+	tn := height
```
zjshen14: You want to provide num of actions instead?

5sWind: Committed as new version.

api/api.go
```diff
@@ -919,30 +919,32 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
 	if start > tipHeight {
 		return nil, status.Error(codes.InvalidArgument, "start height should not exceed tip height")
 	}
-	var res []*iotextypes.BlockMeta
+	var res []*iotextypes.BlockHeaderCore
```
zjshen14: Like mentioned in another pr, let’s continue to use blockemeta as the output data structure

api/api.go
```diff
@@ -919,30 +919,32 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
...
 	for height := start; height <= tipHeight && count > 0; height++ {
-		blk, err := api.bc.GetBlockByHeight(height)
+		header, err := api.bc.BlockHeaderByHeight(height)
```
zjshen14: Since there’s no numActions bellow the Bering height, you need to continue get the full block data instead

zjshen14:
Also, please add test cases to cover the change

blockchain/block/header.go
```diff
@@ -95,6 +99,7 @@ func (h *Header) BlockHeaderCoreProto() *iotextypes.BlockHeaderCore {
 	if h.logsBloom != nil {
 		header.LogsBloom = h.logsBloom.Bytes()
 	}
+	header.NumActions = h.numActions;
```
dustinxie: you'll need to update go.mod file to point to the new iotex-proto version with NumActions added in header definition, w/o it the code won't compile

blockchain/block/header.go
```diff
@@ -129,6 +134,7 @@ func (h *Header) loadFromBlockHeaderCoreProto(pb *iotextypes.BlockHeaderCore) er
 	if pb.GetLogsBloom() != nil {
 		h.logsBloom, err = bloom.BloomFilterFromBytes(pb.GetLogsBloom(), 2048, 3)
 	}
+	h.numActions = pb.GetNumActions()
```
dustinxie: same as above

blockchain/block/header.go
5sWind: Will update go.mod file once this is merged: https://github.com/iotexproject/iotex-proto/pull/37.

api/api.go
dustinxie: +1

5sWind:
I’m uploading a new commit which changed to put the data into block, instead of break changing the header structure.

blockchain/blockchain.go
```diff
@@ -1006,6 +1020,20 @@ func (bc *blockchain) commitBlock(blk *block.Block) error {
 	if err != nil {
 		return err
 	}
+	// write numActions into DB
+	numActionsTimer := bc.timerFactory.NewTimer("putNumActions")
```
zjshen14: Can we encapsulate putNumActions and putTransferAmount into putBlock, and execute in one db batch

blockchain/blockdao.go
```diff
@@ -544,6 +547,87 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
...
+		return 0, errors.Wrap(err, "failed to get num actions")
+	}
+	if len(value) == 0 {
+		return 0, errors.Wrap(db.ErrNotExist, "num actions missing")
+	}
+	return enc.MachineEndian.Uint64(value), nil
+}
+
+// putTransferAmount store transferAmount into db
+func (dao *blockDAO) putTransferAmount(blkHeight uint64, blk *block.Block) error {
```
zjshen14: Please dedup putNumActions and putTransferAmount duplicate code.

5sWind:
> The new solution looks much better. I think you may consider two more things:
> 
> 1. You may need to implement the backfill logic to fill the numActions and transferAmount for the old blocks on mainnet.
>    2 Or in api.go, you need to keep the old logic of pulling the whole block as the fallback solution in the case you can't find numActions/transferAmount.
> 
> I think latter one will be easier to do.

Then we should make it gated by bering height if I choose the later one?

zjshen14:
> Then we should make it gated by bering height if I choose the later one?

Not necessary. You could do following for example:

```golang
numActions, err := GetNumActions(...)
if err == db.NotExist {
    // fallback to old logic
}
```

blockchain/blockchain.go
```diff
@@ -490,6 +494,27 @@ func (bc *blockchain) GetTotalActions() (uint64, error) {
...
+// GetNumActions returns the number of actions
+func (bc *blockchain) GetNumActions(height uint64) (uint64, error) {
+	return bc.dao.getNumActions(height)
+}
+
+// GetTranferAmount returns the transfer amount
+func (bc *blockchain) GetTranferAmount(height uint64) (*big.Int, error) {
+	transferAmount, err := bc.dao.getTranferAmount(height)
+	if err != nil {
+		if errors.Cause(err) == db.ErrNotExist {
```
5sWind: Only the transferAmount I can not find in db when I run the tests, so do this as my sense.

zjshen14: We may not want to handle the error here. Instead, when we get `db.ErrNotExist` while ether calling  `GetNumActions` or `GetTranferAmount` in api.go, we could suspect that we're trying to get the block meta data for an old block when we still didn't apply this code change and didn't index numActions and transferAmount. Therefore, we just need to fallback to the original way of getting the whole block. Though it might be time consuming, but it would guarantee the correctness.

5sWind: I didn't handle the error in api.go is because the `db.` reference does not exist and not to be imported into api package. From my own view, the stack trace would be `api` -> `blockchain`, and the `blockchain` would either to be calling `blockdao` or `db`. Considering the original design, I may ask this would be feasible to implement db in api.go directly.

api/api.go
```diff
@@ -921,32 +924,20 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
...
-			NumActions:       int64(len(blk.Actions)),
-			ProducerAddress:  blk.ProducerAddress(),
-			TransferAmount:   transferAmount.String(),
-			TxRoot:           hex.EncodeToString(txRoot[:]),
-			ReceiptRoot:      hex.EncodeToString(receiptRoot[:]),
-			DeltaStateDigest: hex.EncodeToString(deltaStateDigest[:]),
-		}
-		if logsBloom != nil {
-			blockMeta.LogsBloom = hex.EncodeToString(logsBloom.Bytes())
+		var blockMeta *iotextypes.BlockMeta
```
zjshen14: Almost good. Please see if it's better to refactor by following this pseudo code:

```golang
blkMeta, err := getBlockMetaByHeader(height)
if errors.Cause(err) == db.NotExist && (db.Error() == "missing numActions" || db.Error == ""missing transferAmount") {
    blkMeta, err := getBlockMetaByBlock(height)
}

...


func getBlockMetaByBlock(height) (BlockMeta, error) {
    // copy of existing code
    ...
}

func getBlockMetaByHeader(height) (BlockMeta, error) {
    // new logic of getting block meta by header
    ...
    numActions, err := ...
    if err != null {
        return nil, errors.Wrap(err, "missing numActions")
    }
    ...
    transferAmount, err := ...
    if err != null {
        return nil, errors.Wrap(err, "missing transferAmount")
    }
    ...
}
```

blockchain/blockchain.go
```diff
@@ -96,6 +96,12 @@ type Blockchain interface {
 	BlockFooterByHash(h hash.Hash256) (*block.Footer, error)
 	// GetTotalActions returns the total number of actions
 	GetTotalActions() (uint64, error)
+	// GetNumActions returns the number of actions
+	GetNumActions(height uint64) (uint64, error)
+	// GetTranferAmount returns the transfer amount
+	GetTranferAmount(height uint64) (*big.Int, error)
+	// CalculateTransferAmount returns the calculated transfer amount
+	CalculateTransferAmount(blkActions []action.SealedEnvelope) *big.Int
```
zjshen14: This doesn't need to be part of blockchain API. It seems to be an utility method, and doesn't access any blockchain resource. Consider move it to an util.go if we want to avoid duplicate CalculateTransferAmount usage in api.go and blockdao.go

api/api.go
5sWind: Could you specify the meaning of db.Error()? Is err.Error()? The error I can't find in db package.

api/api.go
```diff
@@ -957,43 +940,133 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
...
+	return blockMeta, nil
+}
+
+// getCommonBlockMeta gets blockmeta by empty interface
+func (api *Server) getCommonBlockMeta(common interface{}) *iotextypes.BlockMeta {
+	header, ok := common.(*block.Header)
+	if !ok {
+		blk := common.(*block.Block)
+		header = &blk.Header
+	}
```
5sWind: This commit compresses code and makes to always get header param.

blockchain/block/utils.go
```diff
@@ -22,3 +24,16 @@ func calculateTxRoot(acts []action.SealedEnvelope) hash.Hash256 {
 	}
 	return crypto.NewMerkleTree(h).HashTree()
 }
+
+// calculateTransferAmount returns the calculated transfer amount
+func calculateTransferAmount(acts []action.SealedEnvelope) *big.Int {
```
5sWind: calculateTransferAmount changed here.

api/api.go
```diff
@@ -921,32 +922,14 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
...
-			ProducerAddress:  blk.ProducerAddress(),
-			TransferAmount:   transferAmount.String(),
-			TxRoot:           hex.EncodeToString(txRoot[:]),
-			ReceiptRoot:      hex.EncodeToString(receiptRoot[:]),
-			DeltaStateDigest: hex.EncodeToString(deltaStateDigest[:]),
-		}
-		if logsBloom != nil {
-			blockMeta.LogsBloom = hex.EncodeToString(logsBloom.Bytes())
+		blockMeta, err := api.getBlockMetasByHeader(height)
+		if errors.Cause(err) == db.ErrNotExist {
```
5sWind: Still working on it.

api/api.go
zjshen14: Sorry my bad.

db.Error() -> err.Error(). It's the interface of error

And check https://godoc.org/github.com/pkg/errors#Cause for how to use errors.Cause. Also, search for the examples in our codebase.

5sWind: Should we only checks `errors.Cause(err) == db.NotExist`.
I tried to implement above code and found that both err.Error() and errors.Cause(err).Error() aren't worked as well for if.
Our codebase example also only has the `errors.Cause(err) == db.NotExist` check.

api/api.go
```diff
@@ -957,43 +938,131 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
...
+
+// getBlockMetasUpgrade checks the upgrade of numActions and transferAmount from DB
+func (api *Server) getBlockMetaUpgrade(height uint64) error {
+	if _, err := api.bc.GetTranferAmount(height); err != nil && errors.Cause(err) == db.ErrNotExist {
+		return errors.Wrap(action.ErrNotFound, "missing transfer amount")
+	}
+	if _, err := api.bc.GetNumActions(height); err != nil && errors.Cause(err) == db.ErrNotExist {
+		return errors.Wrap(action.ErrNotFound, "missing num actions")
+	}
+	return nil
```
5sWind: Is this a feasible way to handle the error.

api/api.go
```diff
@@ -957,43 +940,137 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
...
+// getBlockMetasByHeader gets block header by height
+func (api *Server) getBlockMetasByHeader(height uint64) (*iotextypes.BlockMeta, error) {
+	header, err := api.bc.BlockHeaderByHeight(height)
+	if err != nil {
+		return nil, status.Error(codes.NotFound, err.Error())
+	}
+	if err := api.getBlockMetaUpgrade(header.Height()); err != nil {
+		return nil, err
+	}
+	blockMeta := api.getCommonBlockMeta(header)
```
5sWind: This commit lets check error firstly, then gets blockmeta.

zjshen14:
Congrats! First blood!

Issue #1447, opened 2019-08-13
GetLogs APIs unexpectedly returns empty list


zjshen14:
Here's the query:

```
const getLogs = await client.getLogs({
    filter: {
      address: [],
      topics: [],
    },
    byRange: {
      fromBlock: 1,
      count: 100,
    }
  });
```

It shouldn't be empty. We might have bugs in the filtering logic.

zjshen14:
cc @ququzone

zjshen14:
cc @dustinxie

PR #1448, opened 2019-08-13
changed getLogsInBlock

koseoyoung:
works on #1447 
The problem of this issue is not because of filtering logic, but that the receipt field in block struct was empty. 
So I changed getLogsInBlock() to get a receipt by height and put it into the block struct.
But in the test file, the length of logs is still 0 because the block's tipheight is just around 4.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1145,7 +1145,7 @@ func (api *Server) getLogsInBlock(filter *LogFilter, start, count uint64) ([]*io
 		if err != nil {
 			return logs, status.Error(codes.InvalidArgument, err.Error())
 		}
-		logs = append(logs, filter.MatchBlock(receipts)...)
+		logs = append(logs, filter.MatchLogs(receipts)...)
 	}
 	return logs, nil
 }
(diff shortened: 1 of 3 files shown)


api/api.go
```diff
@@ -1145,6 +1145,11 @@ func (api *Server) getLogsInBlock(filter *LogFilter, start, count uint64) ([]*io
 		if err != nil {
```
dustinxie: remove this call
use an empty block.Block{}

dustinxie: directly pass Receipts to MatchBlock()

api/logfilter.go
```diff
@@ -63,9 +64,9 @@ func (l *LogFilter) Exit() {
 }
 
 // MatchBlock returns matching logs in a given block
-func (l *LogFilter) MatchBlock(blk *block.Block) []*iotextypes.Log {
+func (l *LogFilter) MatchBlock(receipts []*action.Receipt) []*iotextypes.Log {
```
dustinxie: rename to MatchLogs()

Issue #1444, opened 2019-08-09
LogsBloom hasn't been added to BlockMeta for API


zjshen14:
logsBloom was added to block header data structure since 0.7.x. However, it's not added to BlockMeta, so that when we query the block meta info, we wouldn't get this field.

PR #1449, opened 2019-08-14
Logsbloom in blockmeta

koseoyoung:
works on #1444 
added logsbloom field in blockmeta


api/api_test.go
```diff
@@ -282,16 +282,19 @@ var (
...
 			2,
 			7,
 			"6",
+			"",
 		},
 		{
 			4,
 			5,
 			"2",
+			"",
```
dustinxie: is it possible to test for a non-empty logsBloom?

Issue #1413, opened 2019-07-27
Error classification in smart contract execution


xiaxiangzhou:
**What would you like to be added**:
Error classification

**Why is this needed**:
When a client deploy smart contract, there is no error massage when the deploy failed. This make smart contract development very struggle.

For instance, https://testnet.iotexscan.io/action/985de7e30ab84e9c7b24cd3734a7f0fcecff6c24fc2f1df81d3aa20b31daa0b6 . It just say the execution is failed

PR #1451, opened 2019-08-15
Add error status description in ioctl

koseoyoung:
works on #1413 
I tested the error code on the TestNet, it works successfully so far. 
And we need to modifiy the front-end (ioctl, iotexscan) too.

So I changed ioctl by adding error status description when we print out the receipt status.

### DIFF
--- a/ioctl/cmd/action/actionhash.go
+++ b/ioctl/cmd/action/actionhash.go
@@ -226,30 +226,30 @@ func Match(in string, matchType string) string {
 		}
 		return "(" + alias + ")"
 	case "status":
-		if in == "0" {
+		switch in {
+		case "0":
 			return "(Failure)"
-		} else if in == "1" {
+		case "1":
 			return "(Success)"
-		} else if in == "100" {
+		case "100":
 			return "(Failure : Unknown)"
-		} else if in == "101" {
+		case "101":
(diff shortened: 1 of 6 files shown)


ioctl/cmd/action/actionhash.go
```diff
@@ -230,7 +230,26 @@ func Match(in string, matchType string) string {
...
+		} else if in == "104" {
+			return "(Failure : Contract address collision)"
+		} else if in == "105" {
+			return "(Failure : No compatible interpreter)"
+		} else if in == "106" {
+			return "(Failure : Execution reverted)"
+		} else if in == "107" {
+			return "(Failure : Max code size exceeded)"
+		} else if in == "108" {
+			return "(Failure : Write protection)"
```
dustinxie: is it possible to have an error code not equal to all these here?
in that case, add an else {} clause

koseoyoung: Actually, if the error type is not equal to known error, default code is 100(unknown). So, I think it isn't possible to have other than these code (and at the end if it's returning ""empty just in case)

zjshen14: BTW, how about using switch?

koseoyoung: Oh ok !!!

Issue #1042, opened 2019-04-17
Write status into a file and reload it on start


CoderZhi:
**What would you like to be added**:
Add a status file, which contains the status of consensus. On restart, the consensus component will resume the status from the status file.

**Why is this needed**:
Cover a corner case of consensus.

PR #1457, opened 2019-08-22
Move consensus status into db

koseoyoung:
works on #1042 
To store the status of consensus into DB, I modified 

1. Add endorsementpb.proto
2. In endorsementmanager, by using new pb, it's reading from db when it is instantiated, and putting into db when it register the block, addVote and clean.
In the addVote, it puts into db only it changes the status of consensus(over the majority) 
3. To distinguish the situation that db connection is needed from other situation(not needed), I use dbflag. And it is true only when it is about consensus status changing (Otherwise, it is false). Now I'm reusing withTolerance flag because it has similar function(it's only true when it is about consensus working - NewRoundWithTolerance()), but I'm not sure it is ok to reuse this flag.


consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -7,21 +7,45 @@
...
 
+//EndorsedByMajorityFunc defines a function to give an information of consensus status
+type EndorsedByMajorityFunc func(blockHash []byte, topics []ConsensusVoteTopic) bool
+
+func init() {
+	path := "consensus-status-db.bolt"
+	cfg.DbPath = path
+	collectionDB = db.NewBoltDB(cfg)
+	ctx := context.Background()
+	err := collectionDB.Start(ctx)
```
CoderZhi: if this is a db with "Start", putting it in "init" may not be a good way.

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -7,21 +7,45 @@
...
+	if err != nil {
+		return
```
CoderZhi: if there is an error in init function, usually panic

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -143,22 +233,85 @@ func (bc *blockEndorsementCollection) Endorsements(
...
+			dbflag:      true,
+			collections: map[string]*blockEndorsementCollection{},
+		}
+	}
+	//Get from DB
+	manager := &endorsementManager{}
+	managerProto := &endorsementpb.EndorsementManager{}
+	err = proto.Unmarshal(bytes, managerProto)
+	if err != nil {
+		fmt.Println(err)
```
CoderZhi: error handling

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -143,22 +233,85 @@ func (bc *blockEndorsementCollection) Endorsements(
...
+
+func (m *endorsementManager) SetIsMarjorityFunc(isMajorityFunc EndorsedByMajorityFunc) {
+	m.isMajorityFunc = isMajorityFunc
+}
+
+func (m *endorsementManager) fromProto(managerPro *endorsementpb.EndorsementManager) error {
+	m.collections = make(map[string]*blockEndorsementCollection)
+	for _, block := range managerPro.BlockEndorsements {
+		bc := &blockEndorsementCollection{}
+		if block.Blk == nil {
```
CoderZhi: block.Blk == nil is a valid case.

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -183,13 +336,32 @@ func (m *endorsementManager) RegisterBlock(blk *block.Block) error {
 	}
 	m.collections[encodedBlockHash] = newBlockEndorsementCollection(blk)
 
+	if m.dbflag {
```
CoderZhi: we can move collectionDB into endorsementManager, then dbflag could be deleted

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -200,16 +372,50 @@ func (m *endorsementManager) AddVoteEndorsement(
 	}
 	m.collections[encoded] = c
 
+	if m.dbflag && m.isMajorityFunc != nil {
+		afterVote = m.isMajorityFunc(vote.BlockHash(), []ConsensusVoteTopic{vote.Topic()})
+		if !beforeVote && afterVote {
+			//put into DB only it changes the status of consensus
```
CoderZhi: duplicate code

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -200,16 +372,50 @@ func (m *endorsementManager) AddVoteEndorsement(
...
 	}
+	managerProto, err := cleaned.toProto()
+	if err != nil {
+		return err
+	}
+	m = cleaned
 
-	return cleaned
+	if m.dbflag {
+		valBytes, err := proto.Marshal(managerProto)
```
CoderZhi: same here

consensus/scheme/rolldpos/endorsementpb/endorsementmanager.proto
```diff
@@ -0,0 +1,23 @@
...
+
+
+message endorserEndorsementCollection{
+    string endorser = 1;
+    repeated uint32 topics = 2;
+    repeated iotextypes.Endorsement endorsements = 3;
+}
+
+message blockEndorsementCollection{
+    iotextypes.Block blk=1;
```
CoderZhi: spaces before and after =

consensus/scheme/rolldpos/roundcalculator.go
```diff
@@ -56,22 +56,25 @@ func (c *roundCalculator) UpdateRound(round *roundCtx, height uint64, now time.T
...
 		blockInLock = round.blockInLock
 		proofOfLock = round.proofOfLock
-	} else {
-		eManager = newEndorsementManager()
-	}
-	proposer, err := c.calculateProposer(height, roundNum, delegates)
-	if err != nil {
-		return nil, err
 	}
+	eManager = round.eManager
```
CoderZhi: we need an empty manager in the else case

consensus/scheme/rolldpos/roundcalculator_test.go
```diff
@@ -34,7 +34,7 @@ func TestUpdateRound(t *testing.T) {
 	bc, roll := makeChain(t)
 	rc := &roundCalculator{bc, time.Second, time.Second, true, roll, bc.CandidatesByHeight}
 	ra, err := rc.NewRound(1, time.Unix(1562382392, 0))
-	require.NoError(err)
```
CoderZhi: this cannot be deleted

PR #1461, opened 2019-08-28
 improve address action index

dustinxie:
1. improve address <--> action indexing using counting index
2. this change allows to get rid of the time-consuming sort() when performing getActionsByIndex() and getActionsByAddress()
3. removed redundant indexing namespace "a2r"
4. improved test coverage for blockdao_test.go, and added indexbuilder_test.go

### DIFF
--- a/blockchain/blockdao.go
+++ b/blockchain/blockdao.go
@@ -21,10 +21,8 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
-	"github.com/iotexproject/iotex-proto/golang/iotextypes"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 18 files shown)


blockchain/blockdao.go
```diff
@@ -722,39 +664,71 @@ func (dao *blockDAO) deleteTipBlock() error {
 	batch.Delete(blockHashHeightMappingNS, heightKey, "failed to delete height -> hash mapping")
```
lzxm160: need delete receipts here?

blockchain/blockdao.go
```diff
@@ -190,34 +164,6 @@ func (dao *blockDAO) initStores() error {
 	dao.topIndex.Store(maxN)
 	return nil
 }
-func (dao *blockDAO) countActions() error {
```
lzxm160: if this function deleted,GetChainMeta api cannot get total actions when -plugin=gateway is not enabled or cfg.Chain.EnableAsyncIndexWrite is false

Issue #1462, opened 2019-08-29
After 1421,split db default config should be "not split"


lzxm160:
Now the default split size is 640M，it don't need to split db when it's 0.
I think SplitDBSizeMB should set 0 in case of SplitDBHeight is lower than current height

PR #1463, opened 2019-08-29
set SplitDBSizeMB default to 0

lzxm160:
work on #1462


dustinxie:
what is the purpose of change? we want to disable separation DB by size?

lzxm160:
> what is the purpose of change? we want to disable separation DB by size?

Default SplitDBHeight is lower than current height,if user dont set a new height ,split db will cause error. If we set default disable split db,so user will set config correct

Issue #1464, opened 2019-08-29
Check blacklist when processing a block


zjshen14:
We have the blacklist feature to stop processing the tx sent from some particular addresses. It's handled at action pool: https://github.com/iotexproject/iotex-core/blob/8993fd2e55eef0ed87bc1495dcf3cd9034cc3d8e/actpool/actpool.go#L167. That said, delegate node will reject tx getting into the action pool if the sender is blacklisted.

However, there's security hole. The delegate node could become malicious, or just doesn't set the blacklist accordingly, so that it may propose the block that contains the tx that should be blacklisted. Therefore, when validating a block: https://github.com/iotexproject/iotex-core/blob/master/blockchain/blockvalidator.go. We should check agains the same blacklist. If the block contains the blacklisted tx, we should raise error to reject the block.

PR #1465, opened 2019-08-30
Check blacklist when processing a block #1464

xuecanlong:
done


blockchain/blockvalidator_test.go
```diff
@@ -8,6 +8,7 @@ package blockchain
 
 import (
 	"context"
+	"github.com/iotexproject/iotex-address/address"
```
zjshen14: let's group imports in the order:

std

third party

first party

xuecanlong: ok, fixed

zjshen14:
@xuecanlong can you check why the ci build failed?

xuecanlong:
ok, already fixed.

PR #1466, opened 2019-08-31
State-tracker

Frankonly:
This is the state-tracker version of iotex-core who can track state changes like balance changes. The Tracker is written in stateDB, who records particular state changes when server is handling a new block and stores them into datebase as working set is committed into stateDB.

To run the state-tracker, we need to specify MySQL connection string and datababse name:
```
export CONNECTION_STRING=username:password@protocol(address)/
export DB_NAME=dbname
```


state/tracker/tracker.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+// Append appends new state change
+func (t *stateTracker) Append(c StateChange) {
+	t.changes = append(t.changes, c)
+}
+
+// Snapshot records current status of changes
+func (t *stateTracker) Snapshot() {
```
zjshen14: What the meaning of the snapshot?

db/sql/analyticssql/mysql_test.go
```diff
@@ -0,0 +1,51 @@
...
+package sql
+
+import (
+	"database/sql"
+	"fmt"
+	"testing"
+)
+
+const (
+	connectStr = "ba8df54bd3754e:9cd1f263@tcp(us-cdbr-iron-east-02.cleardb.net:3306)/"
```
zjshen14: Don’t leave this in the codebase

db/sql/analyticssql/storebase_tests.go
```diff
@@ -0,0 +1,248 @@
...
+	"database/sql"
+	"errors"
+	"testing"
+
+	"github.com/iotexproject/go-pkgs/hash"
+	"github.com/stretchr/testify/require"
+)
+
+// ActionHistory define the schema for action history
+type ActionHistory struct {
```
zjshen14: Is this copied from previous code of action index?

state/tracker/tracker.go
Frankonly: snapshot the cached state changes when the workingset calls Snapshot.

db/sql/analyticssql/storebase_tests.go
Frankonly: Yes

Frankonly: To avoid cycle import

PR #1468, opened 2019-09-03
store tip block hash

dustinxie:
1. store tip block hash into a key "ts"
2. this PR is to prep for the upcoming change of separating index db. With this change, we can remove the namesapce "h2h" in current chain.db

### DIFF
--- a/blockchain/blockdao.go
+++ b/blockchain/blockdao.go
@@ -57,6 +57,7 @@ const (
 
 var (
 	topHeightKey             = []byte("th")
+	topHashKey               = []byte("ts")
 	totalActionsKey          = []byte("ta")
 	indexActionsTipIndexKey  = []byte("iati")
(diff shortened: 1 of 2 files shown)


blockchain/blockdao.go
```diff
@@ -528,6 +538,7 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
 	if blk.Height() > topHeight {
 		batch.Put(blockNS, topHeightKey, height, "failed to put top height")
 	}
+	batch.Put(blockNS, topHashKey, hash[:], "failed to put top hash")
```
zjshen14: Shall we still put it inside `if blk.Height() > topHeight {` condition block, even though we only append block in sequence currently?

dustinxie: ok

PR #1477, opened 2019-09-06
change last block time and footer validation rule

CoderZhi:
The time of last block was calculated based on the commit time saved in block footer. As the commit time may not be consistent from delegate to delegate, it causes the problem that some nodes with a different commit time reject blocks from block sync.
Fixes:
1. Instead of checking the block producer is the right proposer according to the block time, check whether the producer is a delegate instead. With this fix, even the block commit time is different, it will still be accepted as long as the producer is a consensus delegate and there are enough endorsements.
2. Use the block timestamp as the last block time instead of commit time, to make sure that the consensus delegates get the same round number. This will cause problem that if it took more than one round on a height, the delegate who suppose to be the first proposer for the next height will get passed as the next height won't start from round 0.

### DIFF
--- a/consensus/scheme/rolldpos/roundcalculator.go
+++ b/consensus/scheme/rolldpos/roundcalculator.go
@@ -133,12 +133,11 @@ func (c *roundCalculator) roundInfo(
 ) (roundNum uint32, roundStartTime time.Time, err error) {
 	lastBlockTime := time.Unix(c.chain.GenesisTimestamp(), 0)
 	if height > 1 {
-		var lastBlock *block.Footer
-		if lastBlock, err = c.chain.BlockFooterByHeight(height - 1); err != nil {
+		var lastBlock *block.Header
+		if lastBlock, err = c.chain.BlockHeaderByHeight(height - 1); err != nil {
 			return
 		}
-		lastBlockCommitTime := lastBlock.CommitTime()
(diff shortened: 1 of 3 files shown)


consensus/scheme/rolldpos/roundcalculator.go
```diff
@@ -133,12 +133,11 @@ func (c *roundCalculator) roundInfo(
...
 	if height > 1 {
-		var lastBlock *block.Footer
-		if lastBlock, err = c.chain.BlockFooterByHeight(height - 1); err != nil {
+		var lastBlock *block.Header
+		if lastBlock, err = c.chain.BlockHeaderByHeight(height - 1); err != nil {
 			return
 		}
-		lastBlockCommitTime := lastBlock.CommitTime()
-		lastBlockTime = lastBlockTime.Add(lastBlockCommitTime.Sub(lastBlockTime) / c.blockInterval * c.blockInterval)
+		lastBlockTime = lastBlockTime.Add(lastBlock.Timestamp().Sub(lastBlockTime) / c.blockInterval * c.blockInterval)
```
CoderZhi: not sure whether we should do this. It may cause problem that delegates upgrade their nodes at different time, that they will have different "last block time", some use "commit time", some use "block time".
Potential solution: Only use block time after berling, and then delete it in the next version after berling.
Open to discussion.

dustinxie: works for me, will have a PR to follow this check-in

PR #1479, opened 2019-09-08
fixed #1476 Refactor the code of getBlockMeta in api.go

### LINKED ISSUE #1476: Refactor the code of getBlockMeta in api.go
https://github.com/iotexproject/iotex-core/commit/c2d2f093f84d76bbfb8e6d6025568fe309c7f93f, introduces the fallback logic of requesting full block when other block meta data isn't found. However, the code could be written in a cleaner way:

1. `getBlockMetaByHeader` and `getBlockMetasByHeader`, and `getBlockMetaByBlock` and `getBlockMetasByBlock` are almost identical code.

2. It's not accurate to use action.ErrNotFound. The error has nothing to do with action: https://github.com/iotexproject/iotex-core/blob/c53ab68a1e0f2801566535339f1275df97eba53d/api/api.go#L950

3. There's no need to call getBlockMetaUpgrade, which causes access the data twice from the db.

xuecanlong:
1. almost identical code not need to refactor
2. fixed error code
3. no call getBlockMetaUpgrade

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -923,7 +923,7 @@ func (api *Server) getBlockMetas(start uint64, count uint64) (*iotexapi.GetBlock
 	var res []*iotextypes.BlockMeta
 	for height := start; height <= tipHeight && count > 0; height++ {
 		blockMeta, err := api.getBlockMetasByHeader(height)
-		if errors.Cause(err) == action.ErrNotFound {
+		if errors.Cause(err) == db.ErrNotExist {
 			blockMeta, err = api.getBlockMetasByBlock(height)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -998,9 +998,6 @@ func (api *Server) getBlockMetasByHeader(height uint64) (*iotextypes.BlockMeta,
 	if err != nil {
 		return nil, status.Error(codes.NotFound, err.Error())
 	}
-	if err := api.getBlockMetaUpgrade(header.Height()); err != nil {
```
zjshen14: After removing this, shall we handle the error here: https://github.com/iotexproject/iotex-core/blob/89483c317149092ea2adcd7335c45344c423539f/api/api.go#L988-L989

PR #1488, opened 2019-09-11
native staking

dustinxie:
tested native staking works on nightly-2 -- deployed the staking contract and added 3 candidates, below shows the votes change

| Candidate address        | Votes on Ethereum           | Votes after merging native staking  |
| ------------- | ------------- | ----- |
| io1a64fdsxjdn62rgzjy9rv6pawnlyc887zhnlmk9 | 17375654260384277916000000 | 17375764933341350427324306 |
| io18f909pzsqrpdxhpeqegmcgc7czphk3vxjf98yy | 17264981303311766591700000 | 17265091976268839103024306 |
| io1yvfvf4tw7dau0u55zu26vdz3a2h6tvxsdqeagc | 17154308346239255267400000 | 17154419019196327778724306 |


action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,95 @@
...
+	"math/big"
+	"strings"
+	"time"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/iotexproject/iotex-election/types"
+)
+
+var dummyCaller, _ = address.FromString("io1emxf8zzqckhgjde6dqd97ts0y3q496gm3fdrl6")
```
zjshen14: Use all zero caller instead?

action/protocol/poll/protocol.go
```diff
@@ -187,10 +194,14 @@ func NewGovernanceChainCommitteeProtocol(
 	if err != nil {
 		log.L().Panic("Error when constructing the address of poll protocol", zap.Error(err))
 	}
-
+	ns, err := NewNativeStaking(cm, nativeStakingContract)
```
zjshen14: This needs to be gated by another fork height?

action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,95 @@
...
+func NewNativeStaking(cm protocol.ChainManager, staking string) (*NativeStaking, error) {
+	abi, err := abi.JSON(strings.NewReader(NsAbi))
+	if err != nil {
+		return nil, err
+	}
+	return &NativeStaking{cm, staking, abi}, nil
+}
+
+// Votes returns the votes on height
+func (ns *NativeStaking) Votes(height uint64) ([]*types.Vote, error) {
```
zjshen14: I recommend add height later if we really need it

action/protocol/poll/protocol.go
```diff
@@ -269,6 +281,20 @@ func (p *governanceChainCommitteeProtocol) delegatesByGravityChainHeight(height
 			Votes:         c.Score(),
 			RewardAddress: rewardAddress,
 		})
+		n = append(n, c.Name())
```
zjshen14: You may consider add Name to Candidate, but do not add it to proto definition if we don't need to materialize it

action/protocol/poll/protocol.go
```diff
@@ -396,6 +422,23 @@ func (p *governanceChainCommitteeProtocol) getGravityHeight(height uint64) (uint
...
 
+func (p *governanceChainCommitteeProtocol) mergeDelegates(cand state.CandidateList, name [][]byte, votes []*types.Vote) error {
+	// as of now, native staking does not have register contract, only voting/staking contract
+	// it is assumed that all votes done on native staking target for delegates registered on Ethereum
+	// votes cast to all outside address will not be counted and simply ignored
+	for i, c := range cand {
+		for _, v := range votes {
+			if bytes.Compare(v.Candidate(), name[i]) == 0 {
+				score := big.NewInt(0)
+				score.Add(c.Votes, v.Amount())
```
zjshen14: `c.Votes.add(v.Amount())`?

action/protocol/poll/protocol.go
```diff
@@ -396,6 +422,23 @@ func (p *governanceChainCommitteeProtocol) getGravityHeight(height uint64) (uint
 	return p.electionCommittee.HeightByTime(blkTime)
 }
 
+func (p *governanceChainCommitteeProtocol) mergeDelegates(cand state.CandidateList, name [][]byte, votes []*types.Vote) error {
```
zjshen14: You may want to make votes to be a map instead? Then the following merging logic could be simpler

action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,95 @@
...
+
+type (
+	// NativeStaking represents native staking struct
+	NativeStaking struct {
+		cm       protocol.ChainManager
+		contract string
+		abi      abi.ABI
+	}
+
+	pygg struct {
```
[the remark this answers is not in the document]
dustinxie: right, tested need to be Capital

action/protocol/poll/protocol.go
dustinxie: yeah, i added to Candidate earlier but found need to modify protobuf. we can do this if omitting this field from proto

action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,116 @@
...
+// Votes returns the votes on height
+func (ns *NativeStaking) Votes(height uint64) ([]*types.Vote, error) {
+	if ns.contract == "" {
+		return nil, ErrNoData
+	}
+
+	// read voter list from staking contract
+	votes := []*types.Vote{}
+	prevIndex := big.NewInt(0)
+	limit := big.NewInt(256)
```
zjshen14: check if it's uint8, if it is, 256 will overflow

action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,116 @@
...
+func (ns *NativeStaking) readVotes(prevIndx, limit *big.Int) ([]*types.Vote, error) {
+	data, err := ns.abi.Pack("getActivePyggs", prevIndx, limit)
+	if err != nil {
+		return nil, err
+	}
+
+	// read the staking contract
+	ex, _ := action.NewExecution(ns.contract, 1, big.NewInt(0), 1000000, big.NewInt(0), data)
+	data, _, err = ns.cm.ExecuteContractRead(dummyCaller, ex)
+	if err != nil {
```
zjshen14: You don't catch the errors two lines above. This error check is not effective?

action/protocol/poll/protocol.go
```diff
@@ -396,6 +421,22 @@ func (p *governanceChainCommitteeProtocol) getGravityHeight(height uint64) (uint
...
 }
 
+func (p *governanceChainCommitteeProtocol) mergeDelegates(cand state.CandidateList, votes []*types.Vote) error {
+	// as of now, native staking does not have register contract, only voting/staking contract
+	// it is assumed that all votes done on native staking target for delegates registered on Ethereum
+	// votes cast to all outside address will not be counted and simply ignored
+	for i, c := range cand {
+		for _, v := range votes {
+			if bytes.Compare(v.Candidate(), c.CanName) == 0 {
+				c.Votes.Add(c.Votes, v.Amount())
```
zjshen14:
```suggestion
				vs = big.NewInt(0).add(c.Votes, v.Amount())
				cand[i] = &state.Candidate{c.Address, vs, c.RewardAddress, c.CanName}
```

Avoid mutating the existing var?

CoderZhi: Agree, let's return a new list

action/protocol/protocol.go
```diff
@@ -75,3 +79,45 @@ type StateManager interface {
 	GetDB() db.KVStore
 	GetCachedBatch() db.CachedBatch
 }
+
```
dustinxie: move to here, so nativestaking_test.go can use it to mock ChainManager

action/protocol/poll/nativestaking.go
dustinxie: it is uint256

action/protocol/poll/nativestaking.go
dustinxie: action.NewExecution() will always return nil, that is why use _

action/protocol/poll/protocol.go
dustinxie: @zjshen14   in the first version review, you suggested c.Votes.add(v.Amount())? that is why change to c.Votes.Add(c.Votes, v.Amount()) 

c and c.Votes are both pointer, and i added test to verify this is correct, so should be fine i think

action/protocol/poll/protocol.go
dustinxie: this is addressed in latest change

action/protocol/poll/nativestaking.go
dustinxie: yes, we'll need to pass in height now

action/protocol/poll/protocol.go
dustinxie: this is fixed in latest change

action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Votes returns the votes on height
+func (ns *NativeStaking) Votes(height uint64) (VoteTally, error) {
+	if ns.contract == "" {
+		return nil, ErrNoData
+	}
+
+	now, err := ns.getBlockTime(height)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get current block time")
```
CoderZhi: failed to get time of block %d", height

action/protocol/poll/nativestaking.go
CoderZhi: just in case of any future change to NewExecution, we can catch the error

action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// read the staking contract
+	ex, _ := action.NewExecution(ns.contract, 1, big.NewInt(0), 1000000, big.NewInt(0), data)
+	data, _, err = ns.cm.ExecuteContractRead(dummyCaller, ex)
+	if err != nil {
+		return nil, err
+	}
+
+	// decode the contract read result
+	pygg := &pygg{}
```
CoderZhi: shall we change the struct to Pygg? Actually, it is Pyggs

action/protocol/poll/nativestaking.go
```diff
@@ -0,0 +1,166 @@
...
+	return &NativeStaking{cm, getBlockTime, staking, abi}, nil
+}
+
+// Votes returns the votes on height
+func (ns *NativeStaking) Votes(height uint64) (VoteTally, error) {
+	if ns.contract == "" {
+		return nil, ErrNoData
+	}
+
+	now, err := ns.getBlockTime(height)
```
CoderZhi: concern: the height is actually epoch height, and we use epoch height's time to calculate votes based on the buckets of current state. This may cause inconsistency in the future once we support archive state db.

action/protocol/poll/nativestaking.go
```diff
@@ -88,6 +88,10 @@ func (ns *NativeStaking) Votes(height uint64) (VoteTally, error) {
 
 	for {
 		vote, err := ns.readBuckets(prevIndex, limit)
+		if err == ErrNoData {
+			// all data been read
+			return votes, nil
```
CoderZhi: break

CoderZhi:
We'd better test it before merging to master

action/protocol/poll/staking_committee.go
```diff
@@ -0,0 +1,116 @@
...
+	initGravityChainHeight uint64,
+	getBlockTime GetBlockTime,
+	getEpochHeight GetEpochHeight,
+	getEpochNum GetEpochNum,
+	numCandidateDelegates uint64,
+	numDelegates uint64,
+	initialCandidatesInterval time.Duration,
+	staking string,
+) (Protocol, error) {
+	gs, err := NewGovernanceChainCommitteeProtocol(
```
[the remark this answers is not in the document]
dustinxie: fixed in latest push

action/protocol/poll/nativestaking.go
dustinxie: fixed in latest push

Issue #1490, opened 2019-09-11
Using interface to define chain used in consensus


CoderZhi:
Make components more independent, we need to define an interface of blockchain used in consensus. Similar to https://github.com/iotexproject/iotex-core/blob/ca3bad55a92109e6a1a49ef88520c024a2ba3ea8/action/protocol/protocol.go#L50, the interface should only contain the functions needed in the consensus component.

PR #1493, opened 2019-09-11
make chainmanager interface for consensus

koseoyoung:
works on #1490

### DIFF
--- a/consensus/scheme/rolldpos/rolldpos.go
+++ b/consensus/scheme/rolldpos/rolldpos.go
@@ -17,15 +17,16 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 
+	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/action/protocol/rolldpos"
 	"github.com/iotexproject/iotex-core/actpool"
-	"github.com/iotexproject/iotex-core/blockchain"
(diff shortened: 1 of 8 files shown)


consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -37,6 +38,32 @@ var (
 	ErrNotEnoughCandidates = errors.New("Candidate pool does not have enough candidates")
 )
 
+// ChainManager defines the blockchain interface
+type ChainManager interface {
```
zjshen14: What's the reason of not reusing the existing chain manager interface?

koseoyoung: Because it has different purpose from existing interface, so the methods in this interface are different.

consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -246,7 +273,7 @@ func (b *Builder) SetPriKey(priKey crypto.PrivateKey) *Builder {
 }
 
 // SetBlockchain sets the blockchain APIs
-func (b *Builder) SetBlockchain(chain blockchain.Blockchain) *Builder {
+func (b *Builder) SetBlockchain(chain ChainManager) *Builder {
```
dustinxie: SetChainManager?

PR #1499, opened 2019-09-18
Full state history node 1370

lzxm160:
work on #1370
add config for save how many histories (e.g., past 10,100,1000,... block heights') ,default is 2000
every 30 block height will check if it's time to delete history asynchronous.

After this we will add api to support history state


lzxm160:
need merge conflicts and test again

PR #1507, opened 2019-09-20
Fix new poll protocol

CoderZhi:
Fix three issues related to new poll protocol:
1. Validate function should use the new protocol
2. After bering height, the native staking shouldn't be nil
3. Score threshold filtering happens in iotex-core

### DIFF
--- a/action/protocol/poll/staking_committee.go
+++ b/action/protocol/poll/staking_committee.go
@@ -8,6 +8,8 @@ package poll
 
 import (
 	"context"
+	"math/big"
+
 	"github.com/pkg/errors"
 
 	"github.com/iotexproject/iotex-core/action"
@@ -17,9 +19,11 @@ import (
 )
 
 type stakingCommittee struct {
-	hu                config.HeightUpgrade
-	governanceStaking Protocol
-	nativeStaking     *NativeStaking
+	hu                   config.HeightUpgrade
+	governanceStaking    Protocol
+	nativeStaking        *NativeStaking
+	scoreThreshold       *big.Int
+	selfStakingThreshold *big.Int
 }
 
 // NewStakingCommittee creates a staking committee which fetch result from governance chain and native staking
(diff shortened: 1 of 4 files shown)


action/protocol/poll/staking_committee.go
```diff
@@ -74,26 +83,27 @@ func (sc *stakingCommittee) DelegatesByHeight(height uint64) (state.CandidateLis
...
 }
 
-func (sc *stakingCommittee) mergeDelegates(cand state.CandidateList, votes VoteTally) error {
+func (sc *stakingCommittee) mergeDelegates(list state.CandidateList, votes VoteTally) state.CandidateList {
 	// as of now, native staking does not have register contract, only voting/staking contract
 	// it is assumed that all votes done on native staking target for delegates registered on Ethereum
 	// votes cast to all outside address will not be counted and simply ignored
-	for i, c := range cand {
-		name := to12Bytes(c.CanName)
+	var merged state.CandidateList
```
dustinxie: i think it's not necessary to clone the list, we just modify the Votes
but it's OK if you feel strongly to do so

action/protocol/poll/staking_committee.go
```diff
@@ -17,9 +19,11 @@ import (
...
 
 type stakingCommittee struct {
-	hu                config.HeightUpgrade
-	governanceStaking Protocol
-	nativeStaking     *NativeStaking
+	hu                   config.HeightUpgrade
+	governanceStaking    Protocol
+	nativeStaking        *NativeStaking
+	scoreThreshold       *big.Int
+	selfStakingThreshold *big.Int
```
dustinxie: remove this

action/protocol/poll/staking_committee.go
CoderZhi: This is a coding practice problem. I believe it is a good way, and I would suggest doing similar thing for future code.

PR #1508, opened 2019-09-20
fix bug when start from genesis

koseoyoung:
according to change of iotex-election, fixed a bug when starting from genesis


koseoyoung:
if it has to make empty blockchain, we get result from iotex-election at the initialize part. previously, iotex-election didn't return err in resultbyheight function (it was a bug) but we changed it to return the err(db.notexist), so we should change the import path of db to call election-db to handle that error

PR #1510, opened 2019-09-24
Delete get votes api

CoderZhi:
As stated below, this is a confusing API. Two ways to query votes by height:
1. Query Ethereum contract
2. Query election committee

Providing an API for users to query iotex-core by ethereum height is not the right way.


lzxm160:
It's related to #1360 ,if you think it's deprecated,cli and iotex-proto should be deleted too.
pls make sure this @zjshen14

CoderZhi:
@lzxm160 Created issue https://github.com/iotexproject/iotex-core/issues/1514 as a follow up.

Issue #1501, opened 2019-09-19
Consensus metrics error when the blockchain is empty


zjshen14:
```
[...]

PR #1511, opened 2019-09-24
Consensus metrics error

koseoyoung:
works on #1501 
not return the error, but print out log when it is an empty blockchain


zjshen14:
To fix the problem, is it as simple as that when height = 0, we default the candidate list to empty?

server/itx/heartbeat.go
```diff
@@ -111,6 +114,10 @@ func (h *HeartbeatHandler) Log() {
 			// RollDpos Concensus Metrics
 			consensusMetrics, err = rolldpos.Metrics()
 			if err != nil {
+				if height == 0 && errors.Cause(err) == statedb.ErrStateNotExist {
+					log.L().Warn("consensus metrics are not ready yet, candidates are empty")
+					return
```
CoderZhi: is it intended to return error in this case?

koseoyoung: Oh no, I think we don't have to return in here

server/itx/heartbeat.go
```diff
@@ -111,8 +114,12 @@ func (h *HeartbeatHandler) Log() {
 			// RollDpos Concensus Metrics
 			consensusMetrics, err = rolldpos.Metrics()
 			if err != nil {
-				log.L().Error("failed to read consensus metrics", zap.Error(err))
-				return
+				if height == 0 && errors.Cause(err) == statedb.ErrStateNotExist {
```
zjshen14:
```
if height > 0 || errors.Cause(err) != statedb.ErrStateNotExist {
	log.L().Error("failed to read consensus metrics", zap.Error(err))
	return
}
```

PR #1512, opened 2019-09-25
use epoch start height in stakingCommittee.DelegatesByHeight()

dustinxie:


action/protocol/poll/staking_committee.go
```diff
@@ -31,13 +31,15 @@ func NewStakingCommittee(
...
+	getEpochHeight GetEpochHeight,
+	getEpochNum GetEpochNum,
 	staking string,
 	scoreThreshold *big.Int,
 ) (Protocol, error) {
 	var ns *NativeStaking
 	if staking != "" {
 		var err error
-		if ns, err = NewNativeStaking(cm, getTipBlockTime, staking); err != nil {
+		if ns, err = NewNativeStaking(cm, getTipBlockTime, getEpochHeight, getEpochNum, staking); err != nil {
```
CoderZhi: should we keep getEpochHeight and getEpochNum in staking_committee? They are not used in native staking.

dustinxie: fixed in latest push

Issue #1525, opened 2019-10-01
Prevent exposing internal error message to users


zjshen14:
For example, when using cli to query an unconfirmed action, I got the bellow:

```
root@iotex-delegate-0:/go/apps/iotex-core# ioctl action hash e7fc240f3eb14b8745e5703594ecf2fbe88b3dbe0e2f49cd6f82cac2375067cf
Error: failed to get action e7fc240f3eb14b8745e5703594ecf2fbe88b3dbe0e2f49cd6f82cac2375067cf: key = 94ecf2fbe88b3dbe0e2f49cd6f82cac2375067cf doesn't exist: not exist in DB
```

Rather than return the internal error such as key not exist in DB, we should only raise user friendly error, e.g., action xxxx isn't found.

BTW, more than this particular APIs have the same problem.

lzxm160:
This will be fixed when version is stable

lzxm160:
not necessary

PR #1528, opened 2019-10-02
Prevent exposing internal error message to users 1525

lzxm160:
work on #1525 
need to modify one by one


lzxm160:
> It's better to be a server side fix, as we will have more than just ioctl talk to our apis

ok

lzxm160:
keep error for debug request,change it when user feel not good

PR #1546, opened 2019-10-11
Update default cook height to 1641601

zjshen14:


blockchain/genesis/genesis.go
```diff
@@ -48,8 +49,8 @@ func defaultConfig() Genesis {
 			TimeBasedRotation:     false,
 			PacificBlockHeight:    432001,
 			AleutianBlockHeight:   864001,
-			BeringBlockHeight:     1382401,
-			CookBlockHeight:       1555201,
+			BeringBlockHeight:     1512001,
```
zjshen14: Previously we forgot to update this height on master

PR #1589, opened 2019-10-31
rewrite error message of execution simulation

Frankonly:
When I invoke an execution from ioctl without setting gas limit, ioctl will call ExecuteContractRead to estimate gas consume. If the execution gets failure status, the error message user gets will be `Error: failed to fix Execution gaslimit: receipt status failed` that may be confusing for a non-developer.

Therefore, I'd like to change the error message that node returns to ioctl into `execution simulation gets failure status`. And user may get `Error: failed to fix Execution gaslimit: execution simulation gets failure status`


Frankonly:
Maybe `execution simulation gets failure receipt`  is better.

PR #1606, opened 2019-11-04
set up dardanelles height

yutongp:
enable around sometime in 2019 nov 19 afternoon


blockchain/genesis/genesis.go
```diff
@@ -39,19 +39,20 @@ func init() {
...
+			NumSubEpochs:            2,
+			DardanellesNumSubEpochs: 30,
+			NumDelegates:            24,
+			NumCandidateDelegates:   36,
+			TimeBasedRotation:       false,
+			PacificBlockHeight:      432001,
+			AleutianBlockHeight:     864001,
+			BeringBlockHeight:       1512001,
+			CookBlockHeight:         1641601,
+			DardanellesBlockHeight:  1816201,
```
dustinxie: this block translates to Tues 11/19 3:49pm

can we change it to around 9~10am? in case sth happens, we have the remaining day to check

dustinxie: after discuss, 3:49pm caters to delegates in Asian time zone

PR #1622, opened 2019-11-07
define blockindexer interface in blockdao

CoderZhi:
Blockdao is using a complicated interface defined in blockindex now. The PR tries to simplify the interface definition, and define it in blockdao.

Test plan: make test && make minicluster

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -40,7 +40,6 @@ import (
 	"github.com/iotexproject/iotex-core/actpool/actioniterator"
 	"github.com/iotexproject/iotex-core/blockchain/block"
 	"github.com/iotexproject/iotex-core/blockchain/blockdao"
-	"github.com/iotexproject/iotex-core/blockindex"
 	"github.com/iotexproject/iotex-core/config"
 	"github.com/iotexproject/iotex-core/crypto"
 	"github.com/iotexproject/iotex-core/db"
@@ -151,7 +150,6 @@ type Blockchain interface {
 type blockchain struct {
 	mu            sync.RWMutex // mutex to protect utk, tipHeight and tipHash
 	dao           blockdao.BlockDAO
-	indexer       blockindex.Indexer
 	config        config.Config
 	tipHeight     uint64
 	tipHash       hash.Hash256
@@ -222,7 +220,7 @@ func BoltDBDaoOption() Option {
 		cfg.DB.DbPath = cfg.Chain.ChainDBPath // TODO: remove this after moving TrieDBPath from cfg.Chain to cfg.DB
 		bc.dao = blockdao.NewBlockDAO(
 			db.NewBoltDB(cfg.DB),
-			bc.indexer,
+			nil,
 			cfg.Chain.CompressBlock,
 			cfg.DB,
 		)
@@ -238,7 +236,7 @@ func InMemDaoOption() Option {
 		}
 		bc.dao = blockdao.NewBlockDAO(
 			db.NewMemKVStore(),
-			bc.indexer,
+			nil,
 			cfg.Chain.CompressBlock,
 			cfg.DB,
 		)
@@ -901,7 +899,7 @@ func (bc *blockchain) commitBlock(blk *block.Block) error {
 	}
 	// write block into DB
 	putTimer := bc.timerFactory.NewTimer("putBlock")
-	if err = bc.dao.PutBlock(blk); err == nil {
+	if err = bc.dao.PutBlock(blk); err != nil {
 		err = bc.dao.Commit()
 	}
 	putTimer.End()
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -899,7 +899,7 @@ func (bc *blockchain) commitBlock(blk *block.Block) error {
 	}
 	// write block into DB
 	putTimer := bc.timerFactory.NewTimer("putBlock")
-	if err = bc.dao.PutBlock(blk); err != nil {
+	if err = bc.dao.PutBlock(blk); err == nil {
 		err = bc.dao.Commit()
 	}
 	putTimer.End()
--- a/blockchain/blockdao/blockdao.go
+++ b/blockchain/blockdao/blockdao.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/blockchain/block"
-	"github.com/iotexproject/iotex-core/blockindex"
 	"github.com/iotexproject/iotex-core/config"
 	"github.com/iotexproject/iotex-core/db"
 	"github.com/iotexproject/iotex-core/pkg/cache"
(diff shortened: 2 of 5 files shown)


blockindex/indexer_test.go
```diff
@@ -277,7 +277,7 @@ func TestIndexer(t *testing.T) {
 				// tests[0] is the whole address/action data at block height 3
 				continue
 			}
-			require.NoError(indexer.DeleteBlock(blks[3-i]))
+			require.NoError(indexer.DeleteTipBlock(blks[3-i]))
```
golangcibot: expressions should not be cuddled with blocks (from `wsl`)

blockchain/blockdao/blockdao.go
```diff
@@ -293,15 +303,25 @@ func (dao *blockDAO) PutBlock(blk *block.Block) error {
...
-	blk, err := dao.deleteTipBlock()
+	dao.mutex.Lock()
+	defer dao.mutex.Unlock()
+	// Obtain tip block hash
+	hash, err := dao.getTipHash()
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed to get tip block hash")
+	}
+	blk, err := dao.getBlock(hash)
```
golangcibot: assignments should only be cuddled with other assignments (from `wsl`)

blockchain/blockdao/blockdao.go
```diff
@@ -293,15 +303,25 @@ func (dao *blockDAO) PutBlock(blk *block.Block) error {
...
+	if err != nil {
```
golangcibot: only one cuddle assignment allowed before if statement (from `wsl`)

blockchain/blockdao/blockdao.go
```diff
@@ -660,25 +675,25 @@ func (dao *blockDAO) deleteTipBlock() (*block.Block, error) {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to get tip block hash")
+		return errors.Wrap(err, "failed to get tip block hash")
 	}
 	if err := dao.kvstore.Put(blockNS, topHashKey, hash[:]); err != nil {
-		return nil, errors.Wrap(err, "failed to get tip block hash")
+		return errors.Wrap(err, "failed to get tip block hash")
 	}
-	return blk, nil
+	return nil
```
golangcibot: return statements should not be cuddled if block has more than two lines (from `wsl`)

blockchain/blockchain.go
```diff
@@ -901,7 +899,7 @@ func (bc *blockchain) commitBlock(blk *block.Block) error {
 	}
 	// write block into DB
 	putTimer := bc.timerFactory.NewTimer("putBlock")
-	if err = bc.dao.PutBlock(blk); err == nil {
+	if err = bc.dao.PutBlock(blk); err != nil {
```
golangcibot: if statements should only be cuddled with assignments used in the if statement itself (from `wsl`)

blockchain/blockdao/blockdao.go
```diff
@@ -81,48 +80,59 @@ var (
...
+		PutBlock(*block.Block) error
+		Commit() error
+		DeleteTipBlock() error
+		IndexFile(uint64, []byte) error
+		GetFileIndex(uint64) ([]byte, error)
+		KVStore() db.KVStore
+	}
+
+	// BlockIndexer defines an interface to accept block to build index
+	BlockIndexer interface {
```
dustinxie: if i understand correctly, the intention is that the indexer in BlockDAO only needs to have these 5 API.

would suggest name this interface to like "IndexPutter" and still define it under /blockindex, define another BlockIndexer interface under BlockDAO is kind of confusing

blockchain/blockdao/blockdao.go
```diff
@@ -81,48 +80,59 @@ var (
...
+		Stop(ctx context.Context) error
+		PutBlock(blk *block.Block) error
+		DeleteTipBlock(blk *block.Block) error
+		Commit() error
+	}
+
+	blockDAO struct {
+		compressBlock bool
+		kvstore       db.KVStore
+		indexer       BlockIndexer
```
dustinxie: and here blockindex.IndexPutter

blockchain/blockdao/blockdao.go
```diff
@@ -620,25 +640,20 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
 }
 
 // deleteTipBlock deletes the tip block
-func (dao *blockDAO) deleteTipBlock() (*block.Block, error) {
+func (dao *blockDAO) deleteTipBlock() error {
```
dustinxie: this is good catch, since we move indexer back into DAO, no need to return the deleted block

blockindex/indexer.go
```diff
@@ -47,7 +47,7 @@ type (
 		Stop(context.Context) error
 		Commit() error
 		PutBlock(*block.Block) error
-		DeleteBlock(*block.Block) error
+		DeleteTipBlock(*block.Block) error
```
dustinxie: and Indexer becomes:
Indexer interface {
    IndexPutter;
    all remaining API
}

blockchain/blockchain.go
dustinxie: in case PutBlock() does return an err != nil, err will be overwritten since dao.Commit() always return nil
so should use == here

actually would suggest we remove the Commit() interface from DAO for 2 reasons:

1. I know the intention is that DAO can be also considered an Indexer so it also have Commit() to unify interface, but DAO is a very basic module to put raw blocks, it won't be actually used as Indexer

2. Indexer's DeleteTipBlock(*Block) needs an input block, while DAO's DeleteTipBlock() does not, so anyway DAO has a different interface from Indexer now

blockchain/blockdao/blockdao.go
CoderZhi: Naming doesn't matter. For example, each package could define their own blockchain interfaces. Moreover, defining this in blockdao is on purpose. We discussed this problem before.

blockchain/blockdao/blockdao.go
```diff
@@ -81,48 +80,59 @@ var (
...
+		DeleteTipBlock(blk *block.Block) error
+		Commit() error
+	}
+
+	blockDAO struct {
+		compressBlock bool
+		kvstore       db.KVStore
+		indexer       BlockIndexer
+		htf           db.RangeIndex
+		kvstores      sync.Map //store like map[index]db.KVStore,index from 1...N
```
golangcibot: commentFormatting: put a space between `//` and comment text (from `gocritic`)

blockchain/blockdao/blockdao.go
```diff
@@ -81,48 +80,59 @@ var (
...
+		bodyCache     *cache.ThreadSafeLruCache
+		footerCache   *cache.ThreadSafeLruCache
+		cfg           config.DB
+		mutex         sync.RWMutex // for create new db file
+	}
+)
 
 // NewBlockDAO instantiates a block DAO
-func NewBlockDAO(kvstore db.KVStore, indexer blockindex.Indexer, compressBlock bool, cfg config.DB) BlockDAO {
+func NewBlockDAO(kvstore db.KVStore, indexer BlockIndexer, compressBlock bool, cfg config.DB) BlockDAO {
```
golangcibot: hugeParam: cfg is heavy (136 bytes); consider passing it by pointer (from `gocritic`)

blockchain/blockdao/blockdao.go
```diff
@@ -293,15 +303,25 @@ func (dao *blockDAO) PutBlock(blk *block.Block) error {
 }
 
 func (dao *blockDAO) DeleteTipBlock() error {
-	blk, err := dao.deleteTipBlock()
+	dao.mutex.Lock()
+	defer dao.mutex.Unlock()
+	// Obtain tip block hash
+	hash, err := dao.getTipHash()
```
golangcibot: importShadow: shadow of imported from 'github.com/iotexproject/go-pkgs/hash' package 'hash' (from `gocritic`)

PR #1635, opened 2019-11-11
Update heartbeat.go

EvenBoom:
Fix a misspelling at line 114 which was "Concensus"


raullenchai:
pls sign CLA @EvenBoom

PR #1638, opened 2019-11-11
Update apitestclient.go

EvenBoom:
Fix a misspelling at line 57 which was "elction"


raullenchai:
@pls sign CLA @EvenBoom

Issue #1617, opened 2019-11-06
TipHeight and TipHash should not be stored in a separate db


CoderZhi:
If we want to split chain db into multiple files ($start_height.db), the tip height and hash are now stored in a separate db. As a result, there is a potential data conflict in the db files because of multiple db writes. Ideally, the tip height and hash could be stored in the same db file as the block data.

Solution:
Each $start_height.db file will contain a pair of tip height and tip hash. In blockDao, we read the latest tip height and hash from the largest $start_height.db and store them in memory. In this way, we don't need to have multiple commits in one putBlock call.

lzxm160:
According to #1421 ,only header,body ,footer and receipt maped to split db,tip height and tip hash both in the default db

lzxm160:
As discussed offline,I'll add checking func to make sure chain.db and chain_xxxx.db have the same block height

lzxm160:
As discussed offline,today I'll add height->hash,start and end height for every chainxx.db

lzxm160:
As discussed offline,this needs to move block datas to chainxx.db totally,it's a big change and need a lot of tests

lzxm160:
try to resume this

lzxm160:
already done with latest code

PR #1640, opened 2019-11-12
add check func to sync chain and chainxxdb

lzxm160:
work on #1617

### DIFF
--- a/blockchain/blockdao/blockdao.go
+++ b/blockchain/blockdao/blockdao.go
@@ -216,8 +216,8 @@ func (dao *blockDAO) correct() (err error) {
 	if err != nil {
 		return dao.rewrite(tipHeight)
 	}
-	hash := blk.HashBlock()
-	if !bytes.Equal(tipHash[:], hash[:]) {
+	blkHash := blk.HashBlock()
+	if !bytes.Equal(tipHash[:], blkHash[:]) {
 		return dao.rewrite(tipHeight)
 	}
 	return nil
--- a/blockchain/blockdao/blockdao.go
+++ b/blockchain/blockdao/blockdao.go
@@ -181,7 +181,7 @@ func (dao *blockDAO) Start(ctx context.Context) error {
 	}
 	return dao.correct()
 }
-func (dao *blockDAO) correct() (err error) {
+func (dao *blockDAO) correct() error {
 	// have any error will rewrite block
 	tipHeight, err := dao.getTipHeight()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


blockchain/blockdao/blockdao.go
```diff
@@ -174,9 +175,70 @@ func (dao *blockDAO) Start(ctx context.Context) error {
...
+	// check block hash->height
+	if height != tipHeight {
+		return dao.rewrite(tipHeight)
+	}
+	// check block in chainxxx.db
+	blk, err := dao.getBlock(tipHash)
+	if err != nil {
+		return dao.rewrite(tipHeight)
+	}
+	hash := blk.HashBlock()
```
golangcibot: importShadow: shadow of imported from 'github.com/iotexproject/go-pkgs/hash' package 'hash' (from `gocritic`)

blockchain/blockdao/blockdao.go
```diff
@@ -174,9 +175,70 @@ func (dao *blockDAO) Start(ctx context.Context) error {
...
 		}
 	}
-	return dao.initStores()
+	err = dao.initStores()
+	if err != nil {
+		return err
+	}
+	return dao.correct()
 }
+func (dao *blockDAO) correct() (err error) {
```
CoderZhi: "err error" is defined, but not used

blockchain/blockdao/blockdao.go
```diff
@@ -174,9 +175,70 @@ func (dao *blockDAO) Start(ctx context.Context) error {
...
+	// have any error will rewrite block
+	tipHeight, err := dao.getTipHeight()
+	if err != nil {
```
CoderZhi: if there is any error, is "tipHeight" still a valid value?

lzxm160:
> Get the latest chainxx.db, and extract the largest block stored in this db. With the largest block, check the tip height stored in chain.db. If tip height == block height, everything is ok; else if tip height > block height, quit (reindexing is needed); else rebuild the index from tip height to block height.

There's only hash key in chainxx.db,it cannot get largest block

CoderZhi:
> > Get the latest chainxx.db, and extract the largest block stored in this db. With the largest block, check the tip height stored in chain.db. If tip height == block height, everything is ok; else if tip height > block height, quit (reindexing is needed); else rebuild the index from tip height to block height.
> 
> There's only hash key in chainxx.db,it cannot get largest block

That's what we should fix. If chainxxx.db is indexed by hash key, we need to store a tip hash for each db

lzxm160:
will have a new pr

PR #1649, opened 2019-11-14
use genesis.Genesis as parameter for height upgrade

CoderZhi:
1. Use genesis.Genesis for HeightUpgrade constructor parameter
2. Replace GenesisTimestamp API with Genesis API


config/heightupgrade.go
```diff
@@ -34,13 +36,13 @@ type (
 )
 
 // NewHeightUpgrade creates a height upgrade config
-func NewHeightUpgrade(cfg Config) HeightUpgrade {
+func NewHeightUpgrade(g genesis.Genesis) HeightUpgrade {
```
golangcibot: hugeParam: g is heavy (432 bytes); consider passing it by pointer (from `gocritic`)

PR #1651, opened 2019-11-14
move genesis into run action context

CoderZhi:
Pass the latest Genesis in Blockchain via RunActionsContext and ValidateActionsCtx.


blockchain/blockchain_test.go
```diff
@@ -1230,10 +1233,10 @@ func TestActions(t *testing.T) {
 		testutil.TimestampNow(),
 	)
 	val := &validator{sf: sf, validatorAddr: ""}
-	require.Nil(val.Validate(blk, 0, blk.PrevHash()))
+	require.Nil(val.Validate(ctx, blk, 0, blk.PrevHash()))
 }
 
-func addCreatorToFactory(sf factory.Factory) error {
+func addCreatorToFactory(cfg config.Config, sf factory.Factory) error {
```
golangcibot: hugeParam: cfg is heavy (1488 bytes); consider passing it by pointer (from `gocritic`)

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -60,20 +58,18 @@ type (
 		preimageSnapshot map[int]preimageMap
 		dao              db.KVStore
 		cb               db.CachedBatch
-		hu               config.HeightUpgrade
+		isPreAleutian    bool
```
dustinxie: suggest to keep hu, we may have a future height change that would affect StateDBAdapter behavior

blockchain/blockvalidator.go
```diff
@@ -94,18 +89,23 @@ func (v *validator) SetActPool(actPool ActPoolManager) {
...
+	actions := blk.Actions
 	// Verify transfers, votes, executions, witness, and secrets
 	errChan := make(chan error, len(actions))
 	accountNonceMap := make(map[string][]uint64)
 
 	if err := v.validateActions(
+		ctx,
+		blk.Height(),
+		producerAddr.String(),
 		actions,
```
dustinxie: directly use blk.Actions?

blockchain/blockvalidator.go
```diff
@@ -174,20 +174,14 @@ func (v *validator) validateActions(
...
-		ctx := protocol.WithValidateActionsCtx(
-			context.Background(),
-			protocol.ValidateActionsCtx{
-				BlockHeight:  height,
-				ProducerAddr: producerAddr.String(),
-				Caller:       caller,
-			},
-		)
+		vaCtx.Caller = caller
+		aCtx := protocol.WithValidateActionsCtx(ctx, vaCtx)
```
dustinxie: can use ctx = protocol.WithValidateActionsCtx(ctx, vaCtx)

CoderZhi: it is a multithread process, ctx will be shared. A local one is created on purpose

blockchain/blockvalidator.go
CoderZhi: actions is used in multiple places.

action/protocol/execution/evm/evmstatedbadapter.go
dustinxie: i just recalled, HeightUpgrade is created to address the wrong log topic at Aleutian height, b/c we think pass a bool "isPreAleutian" is not a general good solution
so better keep hu

CoderZhi: rename it to notFixTopicCopyBug

action/receipt.go
```diff
@@ -95,7 +95,7 @@ func (log *Log) ConvertToLogPb() *iotextypes.Log {
 	l.ContractAddress = log.Address
 	l.Topics = [][]byte{}
 	for _, topic := range log.Topics {
-		if log.PreAleutian {
+		if log.NotFixTopicCopyBug {
 			l.Topics = append(l.Topics, topic[:])
 		} else {
 			data := make([]byte, len(topic))
```
dustinxie: can you improve the fix?

for i, topic := range log.Topics {
    else {
         l.Topics = append(l.Topics, log.Topics[i][:])

so can avoid an extra make() and copy()

CoderZhi: not related to this pr

PR #1660, opened 2019-11-18
add blockchain/block/utils_test.go #1659

applemin:
For #1659

PR #1678, opened 2019-11-21
mock protocol

CoderZhi:
add back mockgen.sh which was deleted in https://github.com/iotexproject/iotex-core/commit/87395a6fc447b0ba8699d16e3747bb871f12a233#diff-ab33ddb588c4bbee3c9118346b3d9215


misc/scripts/mockgen.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
yutongp: not sure why this file is newly added, did we removed it somehow?

CoderZhi: it was removed accidentally in a previous PR #1660

Issue #1535, opened 2019-10-07
Nightly job to pick latest commit and sync from block 0


raullenchai:
We need an integration test that runs every N nights (depends on how long it has to run):
1. pick up the latest commit of the past N days
2. sync from block 0 to the current height
3. report errors if any.

Keep in mind, we want to expand 2) to other testing logics later.

PR #1708, opened 2019-11-24
sync block test

ququzone:
work for  #1535


test/sync/test.go
```diff
@@ -0,0 +1,138 @@
...
+	_ "go.uber.org/automaxprocs"
+	"go.uber.org/zap"
+	glog "log"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/iotexproject/iotex-core/config"
+	"github.com/iotexproject/iotex-core/pkg/log"
```
CoderZhi: wrong grouping

test/sync/test.go
```diff
@@ -0,0 +1,138 @@
...
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, os.Interrupt)
+	signal.Notify(stop, syscall.SIGTERM)
+	ctx, cancel := context.WithCancel(context.Background())
+	stopped := make(chan struct{})
+	livenessCtx, livenessCancel := context.WithCancel(context.Background())
+
+	genesisCfg, err := genesis.New()
+	if err != nil {
+		glog.Fatalln("Failed to new genesis config.", zap.Error(err))
```
CoderZhi: log.L().Fatal

test/sync/test.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	start := meta.ChainMeta.Height
+
+	for {
+		time.Sleep(time.Minute * 5)
+		meta, err = c.API().GetChainMeta(context.Background(), &iotexapi.GetChainMetaRequest{})
+		if err != nil {
+			glog.Fatal(err)
+		}
+		if start >= meta.ChainMeta.Height {
```
CoderZhi: start is not updated in the for loop. Therefore, if the first time it works, the if condition will always be true.

CoderZhi:
```
	start := 0
	for {
		time.Sleep(time.Minute * 5)
		meta, err = c.API().GetChainMeta(context.Background(), &iotexapi.GetChainMetaRequest{})
		if err != nil {
			log.L().Fatal(err)
		}
		if start >= meta.ChainMeta.Height {
			glog.Fatalf("blockchain stop sync at %d height", start)
		}
                start = meta.ChainMeta.Height
	}
```

yutongp:
we achieved this feature through other channel. so close this pr

Issue #1673, opened 2019-11-19
Unit Test for governanceChainCommitteeProtocol.CreatePostSystemActions


CoderZhi:
In action/protocol/poll/protocol.go:governanceChainCommitteeProtocol.CreatePostSystemActions

PR #1725, opened 2019-11-27
Unit Test for governanceChainCommitteeProtocol.CreatePostSystemActions

Frankonly:
#1673


action/protocol/poll/protocol_test.go
```diff
@@ -38,10 +40,13 @@ func initConstruct(ctrl *gomock.Controller) (Protocol, context.Context, protocol
 			BlockHeight: 0,
 		},
 	)
+	registry := protocol.NewRegistry()
+	err := registry.Register(rolldpos.ProtocolID, rolldpos.NewProtocol(36, 36, 20))
```
CoderZhi: if err != nil {
  return ...
}

PR #1745, opened 2019-12-03
Fix codecov and add unit test for blockchain.AddSubscriber and RemoveSubscriber

applemin:
Fix codecov and add unit test for blockchain.AddSubscriber and RemoveSubscriber


CoderZhi:
please split into two PRs

PR #1763, opened 2019-12-04
delete chain manager

CoderZhi:
ChainManager is not used any where


CoderZhi:
> can we also erase the mock_chainmanager?

addressed. keep the package name as mock_chainmanager to prevent modifying too many test files

PR #1764, opened 2019-12-05
introduce state reader

CoderZhi:
Split protocol.StateManager into protocol.StateReader and protocol.StateManager. In ReadState, using StateReader instead. Thus, in api, we can use state.Factory directly.


api/api_test.go
```diff
@@ -1475,7 +1475,7 @@ func TestServer_GetEpochMeta(t *testing.T) {
 
 			mbc.EXPECT().TipHeight().Return(uint64(4)).Times(2)
 			mbc.EXPECT().Factory().Return(msf).Times(2)
-			msf.EXPECT().NewWorkingSet().Return(nil, nil).Times(2)
+			// msf.EXPECT().NewWorkingSet().Return(nil, nil).Times(2)
```
golangcibot: commentedOutCode: may want to remove commented-out code (from `gocritic`)

state/factory/workingset.go
```diff
@@ -59,7 +59,7 @@ type (
 		RootHash() (hash.Hash256, error)
 		Digest() (hash.Hash256, error)
 		Version() uint64
-		Height() uint64
+		Height() (uint64, error)
```
koseoyoung: why change to return err ?

action/protocol/account/util/util.go
```diff
@@ -53,7 +53,7 @@ func LoadOrCreateAccount(sm protocol.StateManager, encodedAddr string) (*state.A
 }
 
 // LoadAccount loads an account state
-func LoadAccount(sm protocol.StateManager, addrHash hash.Hash160) (*state.Account, error) {
+func LoadAccount(sm protocol.StateReader, addrHash hash.Hash160) (*state.Account, error) {
 	var account state.Account
 	if err := sm.State(addrHash, &account); err != nil {
 		if errors.Cause(err) == state.ErrStateNotExist {
```
koseoyoung: in line 79, sm also can be stateReader

state/factory/workingset.go
CoderZhi: to be consistent with factory

PR #1786, opened 2019-12-07
remove BlockDAO() from blockchain interface

koseoyoung:
following to PR #1783, remove BlockDAO() from blockchain interface and use dao directly (not through bc)

### DIFF
--- a/blocksync/blocksync.go
+++ b/blocksync/blocksync.go
@@ -16,6 +16,7 @@ import (
 	"github.com/iotexproject/iotex-core/actpool"
 	"github.com/iotexproject/iotex-core/blockchain"
 	"github.com/iotexproject/iotex-core/blockchain/block"
+	"github.com/iotexproject/iotex-core/blockchain/blockdao"
 	"github.com/iotexproject/iotex-core/config"
 	"github.com/iotexproject/iotex-core/consensus"
 	"github.com/iotexproject/iotex-core/pkg/lifecycle"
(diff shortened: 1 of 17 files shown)


gasstation/gasstattion.go
```diff
@@ -22,14 +23,16 @@ import (
 type GasStation struct {
 	bc  blockchain.Blockchain
 	sf  factory.Factory
+	dao blockdao.BlockDAO
```
CoderZhi: let's define an interface here:
```
type BlockDAO interface {
  GetBlockHash,
  GetBlockByHeight,
}
```

blocksync/blocksync.go
```diff
@@ -71,6 +72,7 @@ type blockSyncer struct {
 	buf              *blockBuffer
 	worker           *syncWorker
 	bc               blockchain.Blockchain
+	dao              blockdao.BlockDAO
```
CoderZhi: define interface:
```
type BlockDAO interface {
  GetBlockByHeight,
}
```

gasstation/gasstattion.go
CoderZhi: similarly, we can define an interface
```
type ExecutionSimulator interface {
  SimulateExecution,
}
```
or pass in the function SimulateExecution as a parameter of NewGasStation

koseoyoung: fixed in latest commit

blocksync/blocksync.go
koseoyoung: fixed in latest commit

Issue #1782, opened 2019-12-06
Get RollDPoSProtocol from ctx for poll.Protocol


CoderZhi:
Similar to #1781, we can remove rp, 	getEpochHeight, and getEpochNum in poll.Protocol structs

PR #1794, opened 2019-12-10
try not to pass rolldpos protocol but context 

koseoyoung:
works on #1782  
I also tried to remove getEpochHeight/getEpochNum from protocol struct, but then I should change the CandidatesByHeight API to pass context which is used in consensus.


koseoyoung:
> I think we can delete these two functions, because the rolldpos protocol could be fetch from ctx->blockchain context->registry

Yes! But CandidateByHeight is also used in consensus which doesn't have registry

CoderZhi:
CandidatesByHeight in rolldpos protocol comes with rp:
https://github.com/iotexproject/iotex-core/blob/edbcde118fff9540068fe1c340bde3d5911ccb8d/consensus/consensus.go#L107-L108

action/protocol/poll/protocol.go
```diff
@@ -516,13 +494,12 @@ func (p *governanceChainCommitteeProtocol) ForceRegister(r *protocol.Registry) e
...
 
-func (p *governanceChainCommitteeProtocol) readDelegatesByEpoch(epochNum uint64) (state.CandidateList, error) {
-	epochHeight := p.getEpochHeight(epochNum)
+func (p *governanceChainCommitteeProtocol) readDelegatesByEpoch(epochHeight uint64) (state.CandidateList, error) {
 	return p.candidatesByHeight(epochHeight)
 }
 
-func (p *governanceChainCommitteeProtocol) readBlockProducersByEpoch(epochNum uint64) (state.CandidateList, error) {
-	delegates, err := p.readDelegatesByEpoch(epochNum)
+func (p *governanceChainCommitteeProtocol) readBlockProducersByEpoch(epochHeight uint64) (state.CandidateList, error) {
```
CoderZhi: rename to readBlockProducersByHeight

action/protocol/poll/protocol.go
```diff
@@ -516,13 +494,12 @@ func (p *governanceChainCommitteeProtocol) ForceRegister(r *protocol.Registry) e
 	return r.ForceRegister(protocolID, p)
 }
 
-func (p *governanceChainCommitteeProtocol) readDelegatesByEpoch(epochNum uint64) (state.CandidateList, error) {
-	epochHeight := p.getEpochHeight(epochNum)
+func (p *governanceChainCommitteeProtocol) readDelegatesByEpoch(epochHeight uint64) (state.CandidateList, error) {
```
CoderZhi: rename to readDelegatesByHeight

PR #1802, opened 2019-12-13
Use DelegatesByEpoch to unify the same logic at consensus

koseoyoung:
1. In roundCalculator, instead of using candidatesByHeightFunc, use DelegatesByEpochFunc from poll protocol so that we can unify same logic 

2. refactor poll protocol 
- change name from DelegatesByHeight to CalculateCandidatesByHeight which is used for putpollresult
- clarity method/variable name since delegate/candidate is confusing

### DIFF
--- a/action/protocol/poll/governance_protocol.go
+++ b/action/protocol/poll/governance_protocol.go
@@ -221,11 +221,11 @@ func (p *governanceChainCommitteeProtocol) ReadState(
 	args ...[]byte,
 ) ([]byte, error) {
 	switch string(method) {
-	case "DelegatesByEpoch":
+	case "CandidatesByEpoch":
 		if len(args) != 1 {
 			return nil, errors.Errorf("invalid number of arguments %d", len(args))
 		}
-		delegates, err := p.readDelegatesByEpoch(ctx, byteutil.BytesToUint64(args[0]))
+		delegates, err := p.readCandidatesByEpoch(ctx, byteutil.BytesToUint64(args[0]))
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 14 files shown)


action/protocol/poll/governance_protocol.go
```diff
@@ -312,8 +312,11 @@ func (p *governanceChainCommitteeProtocol) readActiveBlockProducersByEpoch(ctx c
 	epochHeight := rp.GetEpochHeight(epochNum)
 	crypto.SortCandidates(blockProducerList, epochHeight, crypto.CryptoSeed)
 
+	// TODO: kick-out unqualified delegates based on productivity
```
CoderZhi: I am thinking about soft-kick-out the unqualified delegates, by multiply a float weight in range [0-1), 0 means hard-kick-out. In this way, unqualified delegates will still be in the list but with a lower rank.

consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -163,7 +162,7 @@ func (r *RollDPoS) Calibrate(height uint64) {
 // ValidateBlockFooter validates the signatures in the block footer
 func (r *RollDPoS) ValidateBlockFooter(blk *block.Block) error {
 	height := blk.Height()
-	round, err := r.ctx.RoundCalc().NewRound(height, r.ctx.BlockInterval(height), blk.Timestamp(), nil)
+	round, err := r.ctx.roundCalc.NewRound(height, r.ctx.BlockInterval(height), blk.Timestamp(), nil)
```
CoderZhi: I was planning to make rolldposCtx an interface. That's why we call ctx.RoundCalc() instead of using the private variable directly.

action/protocol/poll/governance_protocol.go
koseoyoung: that's good idea! but this sort is random sort to pick 24 out of 36, so what will we multiply a float with? And does the float variable depend on historic misbehavior data? How about apply that idea on slashing the stakes?

koseoyoung: If we want to apply soft/hard concept into kick out, we should multiply a float with the weighted votes before we query candidates (when we calculate and merge with native staking), but at that time we don't know the productivity outcome yet

PR #1809, opened 2019-12-17
fix a bug in createPostSystemActions() at poll protocol

koseoyoung:
Previous we called governance protocol's createPostSystemActions() in staking_committee, but we should use its own createPostSystemActions() so that it'll call correct CalculateCandidatesByHeight. 

But I'm also curious if we had any breaking chain issue because of this bug?


raullenchai:
pls add tests for this

koseoyoung:
> pls add tests for this
I added unit tests but it doesn't test merging with native bucket so it mostly similiar with governance unit test. To fully test, we need to mock readBucket from contract function

Issue #1799, opened 2019-12-12
Slashing: kick out unqualified delegates for k epochs based on productivity 


koseoyoung:
**What would you like to be added**:
Currently we only skip the reward of unqualified delegates. According to discussion, we should kick out unqualified delegate for k epochs who didn't produce block.(underperformance node)

PR #1812, opened 2019-12-20
implement kick-out based on productivity 

koseoyoung:
works on #1799 
1. add kick-out period(k) into config 
2. when we read block producer from candidates(select 36 delegates), skip the blacklist candidates if someone is included in the low-productivity lists of last k epochs (only after EnglishHeight)
3. write activeDelegates at corresponding epoch into DB and read it because when calculating blacklist(N epoch), we also need active delegate list of N-1 ... N -k epoch


koseoyoung:
based on https://github.com/iotexproject/iips/blob/master/iip-1.md

because once we skip for next 3 epochs, it'll automatically skip the epoch reward for following 3 epochs too 
so we don't need reward slashing logic (otherwise it's double slashing)

action/protocol/poll/governance_protocol.go
```diff
@@ -268,23 +307,29 @@ func (p *governanceChainCommitteeProtocol) readActiveBlockProducersByEpoch(ctx c
...
-
 	var activeBlockProducers state.CandidateList
 	for i := 0; i < length; i++ {
 		activeBlockProducers = append(activeBlockProducers, blockProducerMap[blockProducerList[i]])
 	}
+	// write into DB
+	delegateKey := candidatesutil.ConstructDelegateKey(epochNum)
+	if err := p.sm.PutState(delegateKey, &activeBlockProducers); err != nil {
+		return nil, err
+	}
```
CoderZhi: this is a read function, so we shouldn't do sm.PutState here. Else, something weird will happen here: https://github.com/iotexproject/iotex-core/blob/d31309f628d34aea9d941026941f4a6a47ff4daa/action/protocol/poll/governance_protocol.go#L208

Back to the solution: do we need to store delegates into state db?

action/protocol/poll/governance_protocol.go
```diff
@@ -303,3 +348,26 @@ func (p *governanceChainCommitteeProtocol) getGravityHeight(ctx context.Context,
 	)
 	return p.electionCommittee.HeightByTime(blkTime)
 }
+
+func (p *governanceChainCommitteeProtocol) getBlackListDelegates(
+	ctx context.Context,
+	epochNum uint64,
+) (map[string]interface{}, error) {
```
CoderZhi: bool instead of interface{}?

action/protocol/rewarding/protocol.go
```diff
@@ -39,29 +39,24 @@ var (
...
-	productivityByEpoch ProductivityByEpoch
-	keyPrefix           []byte
-	addr                address.Address
+	keyPrefix []byte
+	addr      address.Address
 }
 
 // NewProtocol instantiates a rewarding protocol instance.
-func NewProtocol(productivityByEpoch ProductivityByEpoch) *Protocol {
+func NewProtocol() *Protocol {
```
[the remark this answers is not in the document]
koseoyoung: fixed in latest commit

action/protocol/poll/governance_protocol.go
koseoyoung: after off-line discussion, decide to store blacklist delegate into state DB through createPreState()

action/protocol/poll/governance_protocol.go
koseoyoung: fixed in latest commit

koseoyoung:
major change is that 
1. At the start of epoch, calculate blacklist for kick-out and write into DB 
2. Enable soft kick out - When we read ActiveBlockProducerByEpoch, after English Height, read blacklist from DB and change voting power of it based on kick-out intensity rate [0, 1) and re-sort and pick top 36 candidates 
3. Epoch Reward - Grant epoch reward based on decreased voting power after English Height

action/protocol/rewarding/reward.go
```diff
@@ -396,17 +415,25 @@ func (p *Protocol) splitEpochReward(
...
-			continue
-		}
-		var amountPerAddr *big.Int
-		if totalWeight.Cmp(big.NewInt(0)) == 0 {
-			amountPerAddr = big.NewInt(0)
+			if hu.IsPre(config.English, epochStartHeight) || p.kickOutIntensity == 0 {
+				// Before English or kick-out intensity is 0, skip the epoch reward
+				amounts = append(amounts, big.NewInt(0))
+				continue
+			} else {
```
CoderZhi: else is redundant

action/protocol/rewarding/reward.go
```diff
@@ -396,17 +415,25 @@ func (p *Protocol) splitEpochReward(
...
 	for _, candidate := range candidates {
-		// If not qualified, skip the epoch reward
 		if _, ok := uqd[candidate.Address]; ok {
-			amounts = append(amounts, big.NewInt(0))
-			continue
-		}
-		var amountPerAddr *big.Int
-		if totalWeight.Cmp(big.NewInt(0)) == 0 {
-			amountPerAddr = big.NewInt(0)
+			if hu.IsPre(config.English, epochStartHeight) || p.kickOutIntensity == 0 {
```
CoderZhi: do we need to special handling for `p.kickOutIntensity == 0`?

action/protocol/poll/governance_protocol.go
```diff
@@ -303,3 +379,38 @@ func (p *governanceChainCommitteeProtocol) getGravityHeight(ctx context.Context,
...
+		if err != nil {
+			return nil, err
+		}
+		expectedNumBlks := numBlks / uint64(len(produce))
+		for addr, actualNumBlks := range produce {
+			if actualNumBlks*100/expectedNumBlks < p.productivityThreshold {
+				unqualifiedDelegates = append(unqualifiedDelegates, addr)
+			}
+		}
+	}
```
CoderZhi: blacklist[N] = blacklists[N-1] - blacklists[N-K] + newly blocked, more efficient, and prevent two issues:
1. K is too big
2. if N = English, the blacklist is supposed to be empty, but it will recalculate the blacklist from the previous K epoch, and...

koseoyoung: If we change to this, the time complexity will be more efficient! From what I implemented, the calculation of blacklist has been going regardless of English height and kick-out after english height. But then the calculation of blacklist will start from EnglishHeight which is 0 at that English epoch, is that right?

CoderZhi: yes, does it make sense to you?

koseoyoung: yes, implement it on latest commit

koseoyoung:
Latest 3 commits are that 
1. Optimize calculation of blacklist logic (address zhi's comment) 
2. Calculation of blacklist and writing into DB should happen at the end of the epoch, so that next epoch can determine delegates based on it (address yutong's comment)

action/protocol/poll/governance_protocol.go
```diff
@@ -303,3 +381,91 @@ func (p *governanceChainCommitteeProtocol) getGravityHeight(ctx context.Context,
...
+				produce[blkCtx.Producer.String()]++
+			}
+			expectedNumBlks := numBlks / uint64(len(produce))
+			for addr, actualNumBlks := range produce {
+				if actualNumBlks*100/expectedNumBlks < p.productivityThreshold {
+					unqualifiedDelegates = append(unqualifiedDelegates, addr)
+				}
+			}
+			round--
+		}
```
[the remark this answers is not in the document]
koseoyoung: fixed in latest commit

action/protocol/poll/governance_protocol.go
```diff
@@ -303,3 +381,91 @@ func (p *governanceChainCommitteeProtocol) getGravityHeight(ctx context.Context,
...
+
+	// Blacklist[N] = Blacklist[N-1] - BlackList[N-K] + Low-productivity-list[N-1]
+	prevBlacklist, err := p.kickoutListByEpoch(p.sr, epochNum-1)
+	if err != nil {
+		return nil, err
+	}
+	prevKthBlacklist, err := p.kickoutListByEpoch(p.sr, epochNum-p.kickoutEpochPeriod)
+	if err != nil {
+		return nil, err
+	}
```
CoderZhi: Blacklist[N] = Blacklist[N - 1] - low_productivity_list[N-K] + low_productivity_list[N - 1]
where Blacklist[N] is a map[string]int (int indicates the number of low_productivity_list[N - k] == true, where k is in [0, K-1])

CoderZhi: K = 2
   low_p_list.  Blacklist
1: [A B C] -> [A:1 B:1 C:1]
2: [B D] -> [A:1 B:2 C:1 D:1]
3: [E F] -> [B D E F] vs [A:0 B:1 C:0 D:1 E:1 F:1]

koseoyoung: BlackList[N] = counter under-performance delegates between [N-K, N-1] 
So, BlackList[N] = BlackList[N-1] - low_productivity_list[N-K-1] + low_productivity_list[N-1]

koseoyoung:
fixed calculation bug & added a new struct blacklist to serialize/deserialize with DB & added unit tests

action/protocol/poll/governance_protocol.go
```diff
@@ -412,60 +400,85 @@ func (p *governanceChainCommitteeProtocol) calculateKickoutBlackList(
...
 	}
 
-	// Blacklist[N] = Blacklist[N-1] - BlackList[N-K] + Low-productivity-list[N-1]
-	prevBlacklist, err := p.kickoutListByEpoch(p.sr, epochNum-1)
+	// Blacklist[N] = Blacklist[N-1] - Low-productivity-list[N-K-1] + Low-productivity-list[N-1]
+	blackListMapping, err := p.kickoutListByEpoch(p.sr, epochNum-1)
+	if err != nil {
+		return nil, err
+	}
+	skipList, err := p.calculateUnproductiveDelegatesByEpoch(ctx, epochNum-p.kickoutEpochPeriod-1, false)
```
CoderZhi: if epochNum <= p.kickoutEpochPeriod, something may go wrong.

koseoyoung: line 395, handle ```if epochNum <= englishEpochNum+p.kickoutEpochPeriod ```case 
so, epochNum is always > englishEpochNum+p.kickoutEpochPeriod

action/protocol/poll/governance_protocol.go
```diff
@@ -412,60 +400,85 @@ func (p *governanceChainCommitteeProtocol) calculateKickoutBlackList(
...
 	if err != nil {
 		return nil, err
 	}
-	prevKthBlacklist, err := p.kickoutListByEpoch(p.sr, epochNum-p.kickoutEpochPeriod)
+	for _, addr := range skipList {
+		if _, ok := blackListMapping[addr]; !ok {
+			log.L().Info("It can't happen")
```
CoderZhi: log.L().Fatal(

PR #1817, opened 2019-12-29
[filedao] check legacy and V2 files in cfg.DbPath

dustinxie:
update regarding discussion with Zhi

1. the migration tool is the 2nd commit "code for migration"

2. storage improvement is 30%, using a snapshot of 10G, after migration it reduce to 7G

### DIFF
--- a/blockchain/filedao/filedao.go
+++ b/blockchain/filedao/filedao.go
@@ -18,7 +18,6 @@ import (
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/blockchain/block"
 	"github.com/iotexproject/iotex-core/config"
-	"github.com/iotexproject/iotex-core/db"
 	"github.com/iotexproject/iotex-core/pkg/lifecycle"
 	"github.com/iotexproject/iotex-core/pkg/log"
 )
@@ -70,7 +69,7 @@ type (
 
 // NewFileDAO creates an instance of FileDAO
 func NewFileDAO(compressLegacy bool, cfg config.DB) (FileDAO, error) {
-	legacyFd, err := NewFileDAOLegacy(db.NewBoltDB(cfg), compressLegacy, cfg)
+	legacyFd, err := NewFileDAOLegacy(compressLegacy, cfg)
 	if err != nil {
 		return nil, err
 	}
@@ -79,7 +78,7 @@ func NewFileDAO(compressLegacy bool, cfg config.DB) (FileDAO, error) {
 
 // NewFileDAOInMemForTest creates an in-memory FileDAO for testing
 func NewFileDAOInMemForTest(cfg config.DB) (FileDAO, error) {
-	legacyFd, err := NewFileDAOLegacy(db.NewMemKVStore(), false, cfg)
+	legacyFd, err := newFileDAOLegacyInMem(false, cfg)
 	if err != nil {
 		return nil, err
 	}
--- a/blockchain/filedao/filedao.go
+++ b/blockchain/filedao/filedao.go
@@ -69,7 +69,7 @@ type (
 )
 
 // NewFileDAO creates an instance of FileDAO
-func NewFileDAO(compressLegacy bool, cfg config.DB) (FileDAO, error) {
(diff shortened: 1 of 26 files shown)


blockchain/blockdao/block_migrate.go
```diff
@@ -0,0 +1,393 @@
...
+		return nil
+	}
+
+	heightValue := byteutil.Uint64ToBytes(blkHeight)
+	hashKey := hashKey(h)
+	dao.batch.Put(blockHashHeightMappingNS, hashKey, heightValue, "failed to put hash -> height mapping")
+	return nil
+}
+
+func (dao *blockDAO) getBlockHashLegacy(height uint64) (hash.Hash256, error) {
```
dustinxie: these code are all code for read legacy format DB, could be removed once migration is done

blockchain/blockdao/block_migrate.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	if dao.hashStore, err = db.NewCountingIndexNX(dao.kvstore, []byte(hashDataNS)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (dao *blockDAO) migrate() error {
+	// open the original chain.db
+	bakDbPath := path.Dir(dao.cfg.DbPath) + "/oldchain.db"
```
yutongp: use path.Join

blockchain/filedao/filedao.go
```diff
@@ -98,19 +129,19 @@ func (fd *fileDAO) Height() (uint64, error) {
...
-		err error
-	)
-	for _, file := range fd.v2Fd {
-		if h, err = file.GetBlockHash(height); err == nil {
-			return h, nil
+	if fd.v2Fd != nil {
+		if height == 0 {
+			return hash.ZeroHash256, nil
+		}
+		if v2 := fd.v2Fd.FileDAOByHeight(height); v2 != nil {
```
dustinxie, line 136: per Yutong suggestion,  use the start height to find out which DB owns the height, no need to run a for {} loop and try every file

blockchain/filedao/filedao_v2_util.go
```diff
@@ -57,13 +56,11 @@ func (fd *fileDAOv2) putTipHashHeightMapping(blk *block.Block) error {
 	fd.batch.Put(blockHashHeightMappingNS, hashKey(h), byteutil.Uint64ToBytesBigEndian(height), "failed to put hash -> height mapping")
 
 	// update file tip
-	if height > fd.tip.Height {
```
dustinxie, line 61: no need to check again, only height == tip.Height + 1 can proceed to here

blockchain/filedao/staging_buffer.go
```diff
@@ -22,49 +22,32 @@ func newStagingBuffer(size uint64) *stagingBuffer {
...
 	return s.buffer[pos], nil
 }
 
-func (s *stagingBuffer) Put(pos uint64, blk *block.Store) (bool, error) {
+func (s *stagingBuffer) Put(pos uint64, blkBytes []byte) (bool, error) {
 	if pos >= s.size {
 		return false, ErrNotSupported
 	}
+	blk := &block.Store{}
+	if err := blk.Deserialize(blkBytes); err != nil {
```
dustinxie, line 37: one test fails because a `[]byte` used in `blk` was reused after the block is committed, causing `blk` to become different later on when we retrieve `blk = Get(pos)` to use
in practice, `blk` to be comitted would be a new struct deserialized from network bytes, so this should not happen, but for safety remove that assumption store a new struct in the buffer

blockchain/filedao/filedao_v2_manager.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		end, err := fd.Height()
+		if err != nil {
+			return err
+		}
+		fm[i].start = start
+		fm[i].end = end
+	}
+	sort.Sort(fm)
```
dustinxie, line 52: open all V2 files, and sort them by start height

yutongp (approved):

PR #1821, opened 2020-01-04
Create blockminter 

koseoyoung:
Just like blockValidator, create blockMinter and move mintNewBlock logic from blockchain

### DIFF
--- a/blockchain/blockminter.go
+++ b/blockchain/blockminter.go
@@ -1,0 +1,115 @@
+// Copyright (c) 2019 IoTeX Foundation
+// This is an alpha (internal) release and is not suitable for production. This source code is provided 'as is' and no
+// warranties are given as to title or non-infringement, merchantability or fitness for purpose and, to the extent
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package blockchain
+
+import (
+	"context"
+
+	"github.com/iotexproject/go-pkgs/bloom"
+	"github.com/iotexproject/go-pkgs/crypto"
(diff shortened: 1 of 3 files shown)


blockchain/blockminter.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/iotexproject/iotex-core/action"
+	"github.com/iotexproject/iotex-core/action/protocol"
+	"github.com/iotexproject/iotex-core/blockchain/block"
+	"github.com/iotexproject/iotex-core/state/factory"
+)
+
+// Minter is the interface of block minters
+type Minter interface {
+	// Mint mints the block by given actionMap and timestamp
+	Mint(context.Context, map[string][]action.SealedEnvelope) (*block.Block, error)
```
CoderZhi: as you can see, minter is such a simple interface. So, I would recommend you to move the Mint implementation below as a function of state factory (thus, state factory implements Minter interface).
More accurately, minter is a "block builder factory", which produces BlockBuilder. Thus, the second parameter, minterPrivateKey, could be erased from minter. Here the pseudo code in blockchain.MintNewBlock():
```
    builder, err := factory.NewBlockBuilder(...)
    ...
    return builder.SignAndBuild(bc.producerPrivateKey)
```

You are going to implement it in this way after moving workingset out of block, right?

koseoyoung: if I move mint implementation into state factory, there is a import cycle because factory and block are calling each other

koseoyoung: Then after moving workingset out of block, i will implement this too !

CoderZhi:
if this one is working fine, could we merge it first and then have the improvement in a new PR?

PR #1827, opened 2020-01-15
put/commit block in order into indexer by using buffer 

koseoyoung:
because when we broadcast block we are using goroutine, the order can be different. So make a buffer to store pending blocks and make sure to put/commit block in order

### DIFF
--- a/blockchain/blockdao/indexbuilder.go
+++ b/blockchain/blockdao/indexbuilder.go
@@ -38,7 +38,7 @@ type addrIndex map[hash.Hash160]db.CountingIndex
 
 // IndexBuilder defines the index builder
 type IndexBuilder struct {
-	pendingBlks  map[uint64]*block.Block
+	pendingBlks  chan *block.Block
 	cancelChan   chan interface{}
 	timerFactory *prometheustimer.TimerFactory
 	dao          BlockDAO
@@ -46,7 +46,7 @@ type IndexBuilder struct {
 }
 
 // NewIndexBuilder instantiates an index builder
-func NewIndexBuilder(chainID uint32, dao BlockDAO, indexer blockindex.Indexer) (*IndexBuilder, error) {
+func NewIndexBuilder(chainID uint32, dao BlockDAO, indexer blockindex.Indexer, bufferSize uint64) (*IndexBuilder, error) {
 	timerFactory, err := prometheustimer.New(
 		"iotex_indexer_batch_time",
 		"Indexer batch time",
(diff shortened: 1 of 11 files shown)


yutongp:
@CoderZhi @dustinxie  for subscribe, shall we make it delivery in order to all subscribers or should we do like this subscribers themselves handle in block in order? I think first one may make everyone simpler

blockchain/blockdao/indexbuilder.go
```diff
@@ -100,25 +100,36 @@ func (ib *IndexBuilder) handler() {
 		select {
 		case <-ib.cancelChan:
 			return
-		case blk := <-ib.pendingBlks:
-			timer := ib.timerFactory.NewTimer("indexBlock")
-			if err := ib.indexer.PutBlock(blk); err != nil {
+		default:
```
dustinxie: define a func fetchBlockFromBuffer() (*block.Block, error), return next block (tipHeight+1) if it exists, otherwise nil

blk, err := fetchBlockFromBuffer()
if blk == nil || err != nil {
     continue
}

blockchain/blockdao/indexbuilder.go
```diff
@@ -91,7 +91,7 @@ func (ib *IndexBuilder) Indexer() blockindex.Indexer {
 
 // HandleBlock handles the block and create the indices for the actions and receipts in it
 func (ib *IndexBuilder) HandleBlock(blk *block.Block) error {
-	ib.pendingBlks <- blk
+	ib.pendingBlks[blk.Height()] = blk
```
dustinxie: define a func addBlockToBuffer()
add incoming block only if its height > current tip, and it does not exist in the map yet

dustinxie:
> @CoderZhi @dustinxie for subscribe, shall we make it delivery in order to all subscribers or should we do like this subscribers themselves handle in block in order? I think first one may make everyone simpler

that is also possible, we can have a buffered chan interface{} for each subscriber i think -- the block/event is sent into the channel and subscriber retrieve from channel in their Handler func

blockchain/blockdao/indexbuilder_test.go
```diff
@@ -71,7 +71,7 @@ func TestIndexer(t *testing.T) {
 		}()
 
 		ib := &IndexBuilder{
-			pendingBlks: make(chan *block.Block, 8),
+			pendingBlks: make(chan *block.Block, config.Default.BlockSync.BufferSize),
```
dustinxie: can use make(1) since the test here only handles 1 new block

PR #1830, opened 2020-01-16
move handler goroutine logic from subscriber to blockchain

koseoyoung:
Because each subscriber has same handler logic using goroutine, move this logic to blockchain.go
And when addsubscriber/removesubscriber, create the handler thread and kill it.

### DIFF
--- a/blockchain/blockcreationsubscriber.go
+++ b/blockchain/blockcreationsubscriber.go
@@ -6,16 +6,94 @@
 
 package blockchain
 
-import "github.com/iotexproject/iotex-core/blockchain/block"
+import (
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+
+	"github.com/iotexproject/iotex-core/blockchain/block"
+	"github.com/iotexproject/iotex-core/pkg/log"
+)
 
 // BlockCreationSubscriber is an interface which will get notified when a block is created
 type BlockCreationSubscriber interface {
 	ReceiveBlock(*block.Block) error
 }
 
-// pubSub includes Subscriber, buffered channel for storing the pending blocks and cancel channel to end the handler thread
+// PubSubManager is an interface which handles multi-thread publisher and subscribers
+type PubSubManager interface {
+	AddBlockListener(BlockCreationSubscriber) error
+	RemoveBlockListener(BlockCreationSubscriber) error
+	SendBlockToSubscribers(*block.Block)
+}
+
+// pubSubElem includes Subscriber, buffered channel for storing the pending blocks and cancel channel to end the handler thread
+type pubSubElem struct {
+	listener          BlockCreationSubscriber
+	pendingBlksBuffer chan *block.Block
+	cancel            chan interface{}
+}
+
+// pubSub defines array of blockListener to handle multi-thread publish/subscribe
 type pubSub struct {
-	Blocklistener       BlockCreationSubscriber
-	BlocklistenerBuffer chan *block.Block
-	BlocklistenerCancel chan interface{}
+	blocklisteners       []*pubSubElem
+	pendingBlkBufferSize uint64
+}
+
+// NewPubSub creates new pubSub struct with buffersize for pendingBlock buffer channel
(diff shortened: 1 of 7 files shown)


dustinxie:
Thx for noticing the same handler logic, and refactor effort.

I'm actually leaning towards the current programming for couple of thoughts:

1. Blockchain should just be the dispatcher role that funnels the incoming blocks to each subscriber, with this change blockchain needs to manage 2 []chan, one for sending block, and the other for signaling cancel/exit, which is kind of cumbersome. I feel it's better each subscriber maintain the channel on their own, that way is more flexible. Blockchain just send the new block to the channel -- that's it

2. the handler logic are the same today, but in the future there may come up  a new handler with special logic, so again having subscriber to maintain their own logic is more flexible

3. also more importantly, in the current code each subscriber's Stop() will exit the thread. With this change, that responsibility shifts to calling Blockchain's RemoveSubscriber -- which is an extra burden to take care. 

AND, I caught a bug when checking this: chainservice.go calls chain.AddSubscriber(indexBuilder), but it never calls RemoveSubscriber(). It is OK for now since indexBuilder.Stop() will exit the thread, with your change this bug will be exposed -- indexBuilder's handler thread would never exit. Could you pls fix this bug? :)

koseoyoung:
> 3\. tly, in the current code each subscriber's Stop() will exit the thread. With this change, that responsibility shifts to calling Blockchain's RemoveSubscriber -- which is an extra burden to take care.

Thanks for the comment! @CoderZhi  how do you think? 
For the last bug, I already fixed in this PR (chainservice.go 382 line)

koseoyoung:
> Thx for noticing the same handler logic, and refactor effort.
> 
> I'm actually leaning towards the current programming for couple of thoughts:
> 
> 1. Blockchain should just be the dispatcher role that funnels the incoming blocks to each subscriber, with this change blockchain needs to manage 2 []chan, one for sending block, and the other for signaling cancel/exit, which is kind of cumbersome. I feel it's better each subscriber maintain the channel on their own, that way is more flexible. Blockchain just send the new block to the channel -- that's it
> 2. the handler logic are the same today, but in the future there may come up  a new handler with special logic, so again having subscriber to maintain their own logic is more flexible
> 3. also more importantly, in the current code each subscriber's Stop() will exit the thread. With this change, that responsibility shifts to calling Blockchain's RemoveSubscriber -- which is an extra burden to take care.
> 
> AND, I caught a bug when checking this: chainservice.go calls chain.AddSubscriber(indexBuilder), but it never calls RemoveSubscriber(). It is OK for now since indexBuilder.Stop() will exit the thread, with your change this bug will be exposed -- indexBuilder's handler thread would never exit. Could you pls fix this bug? :)

Thanks for the comment! @CoderZhi  how do you think? 
For the last bug, I already fixed in this PR (chainservice.go 382 line)

dustinxie:
can you refactor the blocklistenerBuffer, blocklistenerBuffer, blocklistenerCancel into a new struct maybe PubSub, and add some util funcs, instead of directly operating these 3?

koseoyoung:
> can you refactor the blocklistenerBuffer, blocklistenerBuffer, blocklistenerCancel into a new struct maybe PubSub, and add some util funcs, instead of directly operating these 3?

Yes! Refactored it in latest commit

blockchain/blockcreationsubscriber.go
```diff
@@ -12,3 +12,10 @@ import "github.com/iotexproject/iotex-core/blockchain/block"
 type BlockCreationSubscriber interface {
 	ReceiveBlock(*block.Block) error
 }
+
+// pubSub includes Subscriber, buffered channel for storing the pending blocks and cancel channel to end the handler thread
+type pubSub struct {
+	Blocklistener       BlockCreationSubscriber
+	BlocklistenerBuffer chan *block.Block
```
dustinxie: can you move the [] into pubSub?

koseoyoung: yes, address on latest commit

Issue #1818, opened 2020-01-01
Unit test for save history feature


CoderZhi:
Save history feature was introduced in https://github.com/iotexproject/iotex-core/commit/dab29156364ada6dad2bd90aff9fc91892396806, but it hasn't been fully tested.

Unit test or end-to-end test for this feature, and verify that we are able to access historical data via factory.State function.

Test plan:
0. Init a blockchain: account A has X IOTX, B has Y IOTX, get the current root hash RH of account trie in factory
1. Put a block with a transfer with 10 IOTX from A->B, get the current root hash NRH
2. Verify that the balance of A is X - 10 IOTX, and the balance of B is Y + 10 IOTX
3. Reset account trie's root hash to RH
4. Verify that the balance of A is X, the balance of B is Y
5. Reset account trie's root hash to NRH
6. Verify that the balance of A is X - 10, and the balance of B is Y + 10

Eventually, we will have the test be applied to https://github.com/iotexproject/iotex-core/pull/1726, which is a replacement implementation of https://github.com/iotexproject/iotex-core/commit/dab29156364ada6dad2bd90aff9fc91892396806. The PR1726 hasn't been merged because it is not fully tested (this is definitely not a valid reason, because https://github.com/iotexproject/iotex-core/commit/dab29156364ada6dad2bd90aff9fc91892396806 hasn't be fully tested either, and moreover it modified the statetx. @dustinxie )

lzxm160:
need add unit test for account and contract seperately

lzxm160:
https://github.com/iotexproject/iotex-core/commits/addtestforsavehistory,base on https://github.com/iotexproject/iotex-core/commit/dab29156364ada6dad2bd90aff9fc91892396806, add TestHistoryForAccount and TestHistoryForContract in blockchain_test.go

lzxm160:
Before dab2915,account state use statetx without trie,now account state already change to use workingset with trie,need to rewrite test code for account

PR #1833, opened 2020-01-20
Unit test for save history feature 1818

lzxm160:
work on #1818


dustinxie:
i think we need to first adapt the AccountState() func, allow it to access state at diff height, by pass in height as an extra param, or using option (so can avoid changing existing code)

add test code for statetx.go (for account) and trie.go (for contract), that is unit test

then add test in blockchain.go (which is what this PR did), that is considered integration or e2e test

also suggest we can leverage existing test code instead of adding new code like this PR. for example, TestLoadBlockchainfromDB() calls addTestingTsfBlocks(), which already has code to commit couple of blocks to a local blockchain. You can maybe add another test into TestLoadBlockchainfromDB() with saveHistory option to test reading diff states at diff height

CoderZhi:
let's don't test against statetx, but workingset. statetx is not designed to store archive data.

dustinxie:
> let's don't test against statetx, but workingset. statetx is not designed to store archive data.

statetx needs to be tested as well, it stores account's state at diff block heights

@lzxm160  we'll need to add into statetx.go new API to read account's state at certain block height

lzxm160:
> o read account's state at certain block heigh

already change to workset,not using statetx

lzxm160:
> > let's don't test against statetx, but workingset. statetx is not designed to store archive data.
> 
> statetx needs to be tested as well, it stores account's state at diff block heights
> 
> @lzxm160 we'll need to add into statetx.go new API to read account's state at certain block height

already change to workset,not using statetx

lzxm160:
need to fix mock later,cannot install mockgen in local vm because of the gfw

lzxm160:
need to merge with latest master and test

state/factory/statetx.go
```diff
@@ -232,6 +232,23 @@ func (stx *stateTX) State(hash hash.Hash160, s interface{}) error {
...
+	}
+	data, err := ri.Get(height)
+	if err != nil {
+		if errors.Cause(err) == db.ErrNotExist {
+			return errors.Wrapf(state.ErrStateNotExist, "state of %x doesn't exist", addr)
+		}
+		return errors.Wrapf(err, "error when getting the state of %x", addr)
+	}
+	return state.Deserialize(s, data)
+}
```
CoderZhi: as commented previously, we don't need to implement this one. Instead, we should throw an error "not supported". StateTX and StateDB are not designed for archive mode.

state/factory/factory_test.go
```diff
@@ -303,6 +303,22 @@ func TestState(t *testing.T) {
 	testState(sf, t)
 }
 
+func TestHistoryState(t *testing.T) {
+	testTrieFile, _ := ioutil.TempFile(os.TempDir(), triePath)
+	cfg := config.Default
+	cfg.Chain.TrieDBPath = testTrieFile.Name()
+	cfg.Chain.EnableHistoryStateDB = true
```
CoderZhi: add a unit test: if this is false, we should expect an error, say ErrNoArchiveData

PR #1834, opened 2020-01-20
Detach workingset from block

koseoyoung:
Define BlockWorkingSet in factory.go which wraps block with workingset. And when we mintNewBlock/ ValidateBlock/ CommitBlock, use the BlockWorkingSet where it generates or commit the workingset. 

impact from this change 
- consensus: rolldposctx.go
- blocksync: blocksync/util.go


dustinxie:
can we try programming like

ws, err = bc.Mint/ValidateBlock(blk)
err = bc.CommitBlock(blk, ws)

this clearly conveys that workingset is the result of running block mint/validation

IMHO, combining Block + WorkingSet into a struct, and carrying that all around, is not a much better solution than having WorkingSet inside Block, in terms of serving the purpose of "detach workingset from block"

dustinxie:
or better yet, why not we absorb ValidateBlock() into CommitBlock()? then we don't need ws to store the intermediate WorkingSet before calling CommitBlock()

there seems no reason for ValidateBlock() to exist alone, it is always followed by CommitBlock() after validating the block

CoderZhi:
> or better yet, why not we absorb ValidateBlock() into CommitBlock()? then we don't need ws to store the intermediate WorkingSet before calling CommitBlock()
> 
> there seems no reason for ValidateBlock() to exist alone, CommitBlock() always follows after validating the block

check "consensus"

dustinxie:
> > or better yet, why not we absorb ValidateBlock() into CommitBlock()? then we don't need ws to store the intermediate WorkingSet before calling CommitBlock()
> > there seems no reason for ValidateBlock() to exist alone, CommitBlock() always follows after validating the block
> 
> check "consensus"

yes, 2 places in "consensus", for ctx.NewProposalEndorsement, we still need ValidateBlock(), but no need to store the Workingset, that is 

ValidateBlock() = current ValidateBlock(), but with ws removed

for ctx.Commit, it is just commit the block, that is

CommitBlock() = current ValidateBlock() + current CommitBlock()

this way we don't need to store the intermediate Workingset.

CoderZhi:
> > > or better yet, why not we absorb ValidateBlock() into CommitBlock()? then we don't need ws to store the intermediate WorkingSet before calling CommitBlock()
> > > there seems no reason for ValidateBlock() to exist alone, CommitBlock() always follows after validating the block
> > 
> > 
> > check "consensus"
> 
> yes, 2 places in "consensus", for ctx.NewProposalEndorsement, we still need ValidateBlock(), but no need to store the Workingset, that is
> 
> ValidateBlock() = current ValidateBlock(), but with ws removed
> 
> for ctx.Commit, it is just commit the block, that is
> 
> CommitBlock() = current ValidateBlock() + current CommitBlock()
> 
> this way we don't need to store the intermediate Workingset, and can just remove it from Block struct

duplicate validation

dustinxie:
> > > > or better yet, why not we absorb ValidateBlock() into CommitBlock()? then we don't need ws to store the intermediate WorkingSet before calling CommitBlock()
> > > > there seems no reason for ValidateBlock() to exist alone, CommitBlock() always follows after validating the block
> > > 
> > > 
> > > check "consensus"
> > 
> > 
> > yes, 2 places in "consensus", for ctx.NewProposalEndorsement, we still need ValidateBlock(), but no need to store the Workingset, that is
> > ValidateBlock() = current ValidateBlock(), but with ws removed
> > for ctx.Commit, it is just commit the block, that is
> > CommitBlock() = current ValidateBlock() + current CommitBlock()
> > this way we don't need to store the intermediate Workingset, and can just remove it from Block struct
> 
> duplicate validation

could you explain more where is duplicate? what i mean is keep Validate, and absorb Validate into Commit, like below:

ValidateBlock() {
    ws, err := validateBlock() // ws in generated, but no need to store it
    return err
}

CommitBlock() {
    ws, err := validateBlock()
    commitBlock() // it will use the ws generated
}

for Minter or Consensus, it calls ValidateBlock()
for blocksync, it calls CommitBlock()

all the code are same, just need slight modification (return ws from validate(), and pass it to commit() )

dustinxie:
> > > > or better yet, why not we absorb ValidateBlock() into CommitBlock()? then we don't need ws to store the intermediate WorkingSet before calling CommitBlock()
> > > > there seems no reason for ValidateBlock() to exist alone, CommitBlock() always follows after validating the block
> > > 
> > > 
> > > check "consensus"
> > 
> > 
> > yes, 2 places in "consensus", for ctx.NewProposalEndorsement, we still need ValidateBlock(), but no need to store the Workingset, that is
> > ValidateBlock() = current ValidateBlock(), but with ws removed
> > for ctx.Commit, it is just commit the block, that is
> > CommitBlock() = current ValidateBlock() + current CommitBlock()
> > this way we don't need to store the intermediate Workingset, and can just remove it from Block struct
> 
> duplicate validation

ws, err := ValidateBlock()
if you are proposer, save the ws in consensus, use it in ctx.Commit
otherwise

> > > > or better yet, why not we absorb ValidateBlock() into CommitBlock()? then we don't need ws to store the intermediate WorkingSet before calling CommitBlock()
> > > > there seems no reason for ValidateBlock() to exist alone, CommitBlock() always follows after validating the block
> > > 
> > > 
> > > check "consensus"
> > 
> > 
> > yes, 2 places in "consensus", for ctx.NewProposalEndorsement, we still need ValidateBlock(), but no need to store the Workingset, that is
> > ValidateBlock() = current ValidateBlock(), but with ws removed
> > for ctx.Commit, it is just commit the block, that is
> > CommitBlock() = current ValidateBlock() + current CommitBlock()
> > this way we don't need to store the intermediate Workingset, and can just remove it from Block struct
> 
> duplicate validation

it happens when the delegate is proposer, with frequency = 24/36 * (1/24) = 1/36 = 3%, i think we can take the hit, or if we don't want, then fallback to below?

ws, err = bc.Mint/ValidateBlock(blk)
err = bc.CommitBlock(blk, ws)

koseoyoung:
We discussed having a workingset cache inside of factory so that when we validate/mintNewBlock, put it into cache and commit the block, get from cache to prevent from validating twice

dustinxie:
> We discussed having a workingset cache inside of factory so that when we validate/mintNewBlock, put it into cache and commit the block, get from cache to prevent from validating twice

yeah, that will work

or you could put that cache in consensus, b/c only consensus's proposer would face this double validation

blockchain/blockchain.go
```diff
@@ -395,7 +395,7 @@ func (bc *blockchain) context(ctx context.Context, tipInfoFlag, candidateFlag bo
 func (bc *blockchain) MintNewBlock(
```
yutongp: from naming perspective, I think it not very trivial that `MintNewBlock` doesn't return a block.

koseoyoung:
will make another PR

PR #1835, opened 2020-01-20
adjust English Height

koseoyoung:
English height for kick out slashing


blockchain/genesis/genesis.go
```diff
@@ -53,7 +53,7 @@ func defaultConfig() Genesis {
 			BeringBlockHeight:       1512001,
 			CookBlockHeight:         1641601,
 			DardanellesBlockHeight:  1816201,
-			EnglishBlockHeight:      3000001,
+			EnglishBlockHeight:      3363298,
```
CoderZhi: 3363298 % 360 = 178, 3363298 % 720 = 178. Could you explain what this number stands for?

koseoyoung: after DardanellesBlockHeight, we can't calculate epoch start height by just dividing it

CoderZhi: it is still some number which %360 = 1 (not all of them are valid, but half of them are valid)

dustinxie: should it be some number == 720n + 1? since now each epoch is 720 blocks

as far as i understand, the intention is to make it happen on the first block of each epoch

koseoyoung: yes we want to make epoch start height
because DardanellesBlockHeight % 720 = 361, then the EnglishBlockHeight also should be % 720 == 361, is that right?

PR #1837, opened 2020-01-21
support sm2 for ioctl's account cmds

Frankonly:
Support sm2 in ioctl and create `xctl.go` in `tools\ioctl\xctl` who using sm2 as default.


ioctl/cmd/account/account.go
```diff
@@ -57,6 +59,13 @@ var (
 	ErrPasswdNotMatch = errors.New("password doesn't match")
 )
 
+var cryptoSm2 bool
+
+var flagSm2Usage = map[config.Language]string{
+	config.English: "sm2 cryptographic algorithm",
+	config.Chinese: "国密sm2加密算法",
```
dustinxie: 签名算法

ioctl/cmd/account/account.go
```diff
@@ -131,6 +137,19 @@ func KsAccountToPrivateKey(signer, password string) (crypto.PrivateKey, error) {
...
+
+	// find the account in pem files
+	pemFilePath := filepath.Join(config.ReadConfig.Wallet, "sm2sk-"+address.String()+".pem")
+	prvKey, err := crypto.ReadPrivateKeyFromPem(pemFilePath, password)
+	if err == nil {
+		return prvKey, nil
+	}
+
+	var pathError *os.PathError
+	if !errors.As(err, &pathError) {
```
dustinxie: we don't need this error?

it failed to find in keystore, and no sm2 exist, so it's just "not match all keys in keystore", not failed to read sm2 key

ioctl/cmd/account/account.go
dustinxie: also, can you resolve the 2 //TODO: in accountimport.go? delete they if not needed

ioctl/cmd/account/account.go
```diff
@@ -228,30 +274,41 @@ func newAccountByKeyStore(alias, passwordOfKeyStore, keyStorePath string, wallet
...
+}
+
 func storeKey(privateKey, walletDir, password string) (string, error) {
 	priKey, err := crypto.HexStringToPrivateKey(privateKey)
 	if err != nil {
 		return "", output.NewError(output.CryptoError, "failed to generate private key from hex string ", err)
 	}
 	defer priKey.Zero()
 
+	addr, err := address.FromBytes(priKey.PublicKey().Hash())
```
dustinxie: not necessary?
HexStringToPrivateKey() will return error if the string cannot make a valid private key

dustinxie: i see, move from below

ioctl/cmd/account/account_test.go
```diff
@@ -37,21 +38,37 @@ func TestAccount(t *testing.T) {
...
+	prvKey2, err := LocalAccountToPrivateKey(addr2.String(), passwd)
+	require.NoError(err)
+	msg2 := hash.Hash256b([]byte(nonce))
+	sig2, err := prvKey2.Sign(msg2[:])
 	require.NoError(err)
-	require.True(prvkey.PublicKey().Verify(msg[:], sig))
+	require.True(prvKey2.PublicKey().Verify(msg2[:], sig2))
 
 	// test import existing key
 	sk, err := crypto.GenerateKey()
```
dustinxie: add test to load from PEM file

ioctl/cmd/account/accountlist.go
```diff
@@ -8,15 +8,18 @@ package account
 
 import (
 	"fmt"
+	"io/ioutil"
 	"strings"
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/spf13/cobra"
```
dustinxie: delete line

ioctl/cmd/account/accountlist.go
```diff
@@ -67,6 +71,23 @@ func accountList() error {
 			Alias:   aliases[address.String()],
 		})
 	}
+	files, err := ioutil.ReadDir(config.ReadConfig.Wallet)
```
dustinxie: can we add test for these new code?

ioctl/cmd/account/accountupdate.go
```diff
@@ -62,33 +66,72 @@ func accountUpdate(arg string) error {
...
 				return output.NewError(output.KeystoreError, "failed to update keystore", err)
 			}
+
 			output.PrintResult(fmt.Sprintf("Account #%s has been updated.", account))
+			return nil
 		}
 	}
-	return output.NewError(output.KeystoreError, fmt.Sprintf("account #%s not found", account), nil)
+
+	// find the pem file and update
```
dustinxie: can we add test for these new code?

tools/ioctl/xctl/xctl.go
```diff
@@ -0,0 +1,17 @@
...
+// License 2.0 that can be found in the LICENSE file.
+
+package main
+
+import (
+	"github.com/iotexproject/iotex-core/ioctl/cmd"
+	"github.com/iotexproject/iotex-core/ioctl/cmd/account"
+)
+
+func main() {
```
dustinxie: add an entry in makefile to generate binary with diff name?

ioctl/cmd/account/account_test.go
Frankonly: The case that loading SM2 key from PEM file is included when running `LocalAccountToPrivateKey()`.

ioctl/cmd/account/accountlist.go
Frankonly: `io/ioutil` is used at line 74:
```
files, err := ioutil.ReadDir(config.ReadConfig.Wallet)
```

ioctl/cmd/account/accountcreate.go
```diff
@@ -61,15 +61,28 @@ type generatedAccount struct {
 }
 
 func init() {
-	accountCreateCmd.Flags().UintVarP(&numAccounts, "num", "n", 1, config.TranslateInLang(flagNumUsages, config.UILanguage))
+	accountCreateCmd.Flags().UintVarP(&numAccounts, "num", "n", 1,
+		config.TranslateInLang(flagNumUsages, config.UILanguage))
+	accountCreateCmd.Flags().BoolVar(&CryptoSm2, "sm2", false,
+		config.TranslateInLang(flagSm2Usage, config.UILanguage))
```
CoderZhi: why we need a flag?

ioctl/cmd/account/accountlist.go
```diff
@@ -44,7 +45,8 @@ var accountListCmd = &cobra.Command{
 }
 
 type listMessage struct {
-	Accounts []account `json:"accounts"`
+	Accounts    []account `json:"accounts"`
+	Sm2Accounts []account `json:"sm2Accounts"`
```
CoderZhi: why we need to distinguish?

ioctl/cmd/account/accountcreate.go
Frankonly: > why we need a flag?

I make ioctl support both two cryptography algorithms. This flag is to swtich them when using command like `account create`, avoiding more commands for similar functions.

ioctl/cmd/account/accountlist.go
Frankonly: > why we need to distinguish?

ioctl can handle different keys with different cryptography algorithms, if user stores those two kinds of key in same place (using same ioctl), ioctl need distinguish them when listing them for user.

ioctl/cmd/account/accountcreate.go
Frankonly: ioctl supports both two cryptography algorithms. xctl only supports sm2. How I implement xctl is to set `CryptoSm2` to `true` always.

PR #1839, opened 2020-01-21
whitelist signature scheme in config

dustinxie:
private/public key operation will verify against the whitelist schemes in config before continuing


blockchain/block/header.go
```diff
@@ -166,7 +165,7 @@ func (h *Header) HashHeaderCore() hash.Hash256 {
 func (h *Header) VerifySignature() bool {
 	hash := h.HashHeaderCore()
 
-	if h.pubkey == nil || len(h.blockSig) != action.SignatureLength {
+	if h.pubkey == nil {
```
dustinxie: this check will be done inside pubkey.Verify() below

ioctl/cmd/account/account.go
```diff
@@ -235,7 +234,7 @@ func storeKey(privateKey, walletDir, password string) (string, error) {
 			return "", output.NewError(output.KeystoreError, "failed to import private key into keystore ", err)
 		}
 	case *crypto.P256sm2PrvKey:
-		if _, err := sm2.WritePrivateKeytoPem(walletDir, sk.PrivateKey, []byte(password)); err != nil {
+		if err := crypto.WritePrivateKeyToPem(walletDir, sk, password); err != nil {
```
dustinxie: using crypto func to avoid directly importing sm2

config/config.go
```diff
@@ -499,9 +509,50 @@ func (cfg Config) ProducerPrivateKey() crypto.PrivateKey {
...
+	for _, e := range cfg.Chain.SignatureScheme {
+		if sigScheme == e {
+			// signature scheme is whitelisted
+			return true
+		}
+	}
+	return false
+}
+
+func (cfg Config) generateRandomKey() string {
```
CoderZhi: this function confused me. How could we randomly pick a scheme to generate a random key?

config/config.go
```diff
@@ -224,6 +229,7 @@ type (
 		ID              uint32           `yaml:"id"`
 		Address         string           `yaml:"address"`
 		ProducerPrivKey string           `yaml:"producerPrivKey"`
+		SignatureScheme []string         `yaml:"signatureScheme"`
```
CoderZhi: I thought we start with one valid scheme

dustinxie: if the chain wants only 1 valid scheme, then put only 1 in the config

the default is []string{SigP256k1}, which is the current scheme supported

config/config.go
dustinxie: this is to cover the corner case cfg.Chain.ProducerPrivKey is left empty, then generate a random key according to allowed schemes.

i think i'll leave it as empty, and not change the current code's behavior, please see the latest commit

config/config.go
```diff
@@ -423,9 +420,6 @@ func New(validates ...Validate) (Config, error) {
 		return Config{}, errors.Wrap(err, "failed to unmarshal YAML config to struct")
 	}
 
-	if cfg.Chain.ProducerPrivKey == "" {
```
dustinxie: remove it so not change original code's behavior

PR #1852, opened 2020-01-30
fix low gas limit for reading native staking contract

dustinxie:
test plan

1. verified the new gasLimit = 10000000 suffices to read 160 buckets (actually 2000000 is enough)
2. start a full-node, verified that votes on mainnet contract is correctly read and merged 

======= before:
address = io1xxx3drgja, votes = 110659195814776937774521455
address = io1xxx6v5vh, votes = 101483332311087281139889299
address = io1xxx7ylx0ts, votes = 96158054759058964589222967
address = io1xxx2gm63k, votes = 88549255314706513359871177
======= after:
address = io1xxx2gm63k, votes = 157322003764892821322558280
address = io1xxx3drgja, votes = 111586300298752708098734136
address = io1xxx6v5vh, votes = 101519051869216472051891381
address = io1xxx7ylx0ts, votes = 96327295782308694486429051


action/protocol/poll/nativestaking.go
```diff
@@ -94,55 +94,60 @@ func (ns *NativeStaking) Votes() (*VoteTally, time.Time, error) {
 		Buckets:    make([]*types.Bucket, 0),
 	}
 	prevIndex := big.NewInt(0)
-	limit := big.NewInt(256)
+	limit := big.NewInt(160)
```
yutongp: Let's also change it from 256 to 160 after daytonaBlockHeight, so we can make sure past behavior is the same.

blockchain/genesis/genesis.go
```diff
@@ -53,6 +53,7 @@ func defaultConfig() Genesis {
 			BeringBlockHeight:       1512001,
 			CookBlockHeight:         1641601,
 			DardanellesBlockHeight:  1816201,
+			DaytonaBlockHeight:      3168001,
```
yutongp: I think this is not at start of an epoch. You may use 360 block pre epoch to calculate, now we have 720 block pre epoch

action/protocol/poll/nativestaking.go
```diff
@@ -94,55 +94,60 @@ func (ns *NativeStaking) Votes() (*VoteTally, time.Time, error) {
...
 
 	for {
-		vote, err := ns.readBuckets(prevIndex, limit)
+		vote, index, err := ns.readBuckets(prevIndex, limit, correctGas)
```
yutongp: Can we test the case when we have more than 160 buckets? So we can make sure pagination is correct?

dustinxie: yes, i tested last night pagination is correct

we now have 161 buckets, i read using limit = 256 (result in 1 batch) vs. limit = 160 (result in 2 batch), the final vote count is same

blockchain/genesis/genesis.go
dustinxie: i used 720 to calculate at first, but found it is not multiple of 720, then change it. 

It's b/c Dardarnelle is not a multiple of 720, which i failed to realize

added comments in lastest push

action/protocol/poll/nativestaking.go
yutongp: let's add a unit test to reflect this case

action/protocol/poll/nativestaking.go
```diff
@@ -94,7 +94,10 @@ func (ns *NativeStaking) Votes(correctGas bool) (*VoteTally, time.Time, error) {
 		Buckets:    make([]*types.Bucket, 0),
 	}
 	prevIndex := big.NewInt(0)
-	limit := big.NewInt(160)
+	limit := big.NewInt(256)
+	if correctGas {
+		limit = big.NewInt(160)
+	}
```
CoderZhi: is this redundant?

dustinxie: removed in latest push

action/protocol/poll/nativestaking.go
dustinxie: added in latest push

blockchain/genesis/genesis.go
```diff
@@ -53,7 +53,7 @@ func defaultConfig() Genesis {
 			BeringBlockHeight:       1512001,
 			CookBlockHeight:         1641601,
 			DardanellesBlockHeight:  1816201,
-			DaytonaBlockHeight:      3168361,
+			DaytonaBlockHeight:      3238921,
```
dustinxie: target to hit US time 2/10/2010 2:30pm

e2etest/staking_test.go
```diff
@@ -0,0 +1,186 @@
...
+			require.Equal(unit.ConvertIotxToRau(100000000-int64(numBucket)*200), s.Balance)
+		}
+
+		// read from contract
+		ns, err := poll.NewNativeStaking(bc, func() (time.Time, error) {
+			return fixedTime, nil
+		})
+		require.NoError(err)
+		ns.SetContract(r.ContractAddress)
+		tally, now, err := ns.Votes(false)
```
dustinxie: read with low gas will fail

PR #1855, opened 2020-02-01
add namespace to general state API

dustinxie:
currently all data in stateDB are stored under default namespace "account"
we plan to store staking-related data into stateDB under a new namespace "staking"


action/protocol/managers.go
```diff
@@ -9,7 +15,7 @@ import (
 // StateReader defines an interface to read state db
 type StateReader interface {
 	Height() (uint64, error)
-	State(hash.Hash160, interface{}) error
+	State(string, hash.Hash160, interface{}) error
```
dustinxie: add namespace as first arg

Issue #1851, opened 2020-01-30
Failed to sync block from start in testnet 


lzxm160:
The following is the error when run latest commit 63f476844806ee58f5c5f932a95104fa1172dfbb:

`{"level":"info","ts":1580429903.0761547,"caller":"blockchain/pubsubmanager.go:68","msg":"Successfully unsubscribe block creation.","ioAddr":"io16lmwswsy59ea29fvejkma0vpdmj48m33fzyk3v","networkAddr":"0.0.0.0:4689"}
[...]

PR #1860, opened 2020-02-05
Failed to sync block from start in testnet

lzxm160:
work on #1851


action/protocol/vote/candidatesutil/candidatesutil.go
```diff
@@ -56,6 +56,9 @@ func CandidatesByHeight(sr protocol.StateReader, height uint64) ([]*state.Candid
 // KickoutListByEpoch returns array of unqualified delegate address in delegate pool for the given epochNum
 func KickoutListByEpoch(sr protocol.StateReader, epochNum uint64) (*vote.Blacklist, error) {
 	blackList := &vote.Blacklist{}
+	if epochNum == 1 {
```
CoderZhi: @koseoyoung this is a workaround. We can have a better solution on the db level.

PR #1865, opened 2020-02-07
New ioctl command for node reward and unit test

RuneAkko:


ioctl/client.go
```diff
@@ -48,6 +48,8 @@ type (
 		// doing
 		GetAddress(in string) (string, error)
 		// doing
+		Address(in string) (string, error)
```
Frankonly: What's the difference bewteen "GetAddress()" and "Address()" ?

RuneAkko: all comes from iotcl/util.go.`GetAddress():GetAddress get address from address or alias or context`,`Address():Address returns the address corresponding to alias. if 'in' is an IoTeX address, returns 'in'`.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/iotexproject/iotex-core
 go 1.13
 
 require (
+	bou.ke/monkey v1.0.2
```
yutongp: i don't see this dependence used anywhere. if we don't need it, please remove

ioctl/output/format.go
```diff
@@ -73,15 +73,21 @@ const (
 
 // Output is used for format output
 type Output struct {
-	MessageType MessageType `json:"messageType"`
-	Message     Message     `json:"message"`
+	MessageType            MessageType            `json:"messageType"`
+	Message                Message                `json:"message"`
+	MessageWithTranslation MessageWithTranslation `json:"MessageWithTranslation"`
```
yutongp: json:"messageWithTranslation

PR #1866, opened 2020-02-07
Fix a deadlock bug  

koseoyoung:
fixed a deadlock bug by only lock/unlock mutex when accessing workingsets cache

### DIFF
--- a/state/factory/statedb.go
+++ b/state/factory/statedb.go
@@ -331,7 +331,6 @@ func (sdb *stateDB) createGenesisStates(ctx context.Context) error {
 	return sdb.commit(ws)
 }
 
-
 func (sdb *stateDB) readFromWorkingSets(key hash.Hash256) (WorkingSet, error) {
 	sdb.mutex.RLock()
 	defer sdb.mutex.RUnlock()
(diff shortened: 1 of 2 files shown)


state/factory/statedb.go
```diff
@@ -332,3 +330,23 @@ func (sdb *stateDB) createGenesisStates(ctx context.Context) error {
 
 	return sdb.commit(ws)
 }
+
+func (sdb *stateDB) readFromWorkingSets(key hash.Hash256) (WorkingSet, error) {
+	sdb.mutex.RLock()
+	defer sdb.mutex.RUnlock()
+	if data, ok := sdb.workingsets.Get(key); ok {
```
dustinxie: so the root-cause is need to use mutex for get/putWorkingSets?

state/factory/statedb.go
```diff
@@ -332,3 +330,23 @@ func (sdb *stateDB) createGenesisStates(ctx context.Context) error {
 
 	return sdb.commit(ws)
 }
+
+func (sdb *stateDB) readFromWorkingSets(key hash.Hash256) (WorkingSet, error) {
```
dustinxie: can we change to getFrom, vs. putInto

state/factory/statedb.go
koseoyoung: The root cause is that in runActions(), we shouldn't lock the sf mutex. But while detaching workingset, get/put workingset needs mutex to be locked so I changed to lock/unlock the mutex seperately if needed

state/factory/statedb.go
koseoyoung: yes fixed in latest commit

Issue #1562, opened 2019-10-21
node bootstrap from state db Checkpoint instead of block 0


yutongp:
Currently we only support start from our snapshots or start from very beginning.

one ability we may need to not sync from block 0 but start syncing from some checkpoint. which means we start with a state.db at height X, and the node directly asking it's peers for block X+1 and above to keep up to the latest change and doesn't request for block less and equal to X.

This is also the fundamental feature if we will Re-genesis in the future.

lzxm160:
I think this will take two steps:
1. sync state to state.db,include account and trie node
2. modify code logic,implement sync data when only have state.db

lzxm160:
Start to research today

lzxm160:
After 1655 and 1652 ,I will restart this

lzxm160:
research and test code

lzxm160:
Branch 15621 is for this issue,now have the following error to fix:

[...]

PR #1868, opened 2020-02-10
node bootstrap from state db Checkpoint instead of block 0 1562

lzxm160:
work on #1562

Implements this through write two epochs blocks data to trie.db.
In checkpoint mode,we can only take trie.db and rewrite to chain.db when blockchain starts


lzxm160:
> I don't follow this implementation. Why checkpoint is in trie db? why checkpoint is a KVStore

trie.db have all the accounts and contracts status those have to save,so the balance and contract status will not change

lzxm160:
Temporarily close, and reopen if necessary

PR #1871, opened 2020-02-10
add stakingGet/Put/DelBucket API

dustinxie:
3 APIs for staking protocol for create, get, delete voting bucket

stakingGetBucket, stakingPutBucket, stakingDelBucket


action/protocol/staking/protocol.go
```diff
@@ -16,19 +16,17 @@ import (
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/action/protocol"
 	"github.com/iotexproject/iotex-core/pkg/log"
+	"github.com/iotexproject/iotex-core/state/factory"
 )
 
-// protocolID is the protocol ID
```
CoderZhi: they are just happened to be the same value.

dustinxie: yes, so just use the same value, instead of adding a new one

is this a concern?

Issue #1872, opened 2020-02-11
commit dd66f8a0fb4 "Move workingset into factory" broke test TestTransfer_Negative


dustinxie:
### What version of iotex-core are you using (`ioctl version`)?
master

### What operating system and processor architecture are you using?
iOS

### What did you do? If possible, provide a recipe for reproducing the error.
in TestTransfer_Negative(), if you comment out return at beginning, the test can actually pass. i.e. we should remove this "return" and let the test stand

### What did you expect to see?
at commit dd66f8a0fb4, the test should still pass, if remove "return"

### What did you see instead?
the test fails by removing "return"

PR #1883, opened 2020-02-12
fix commit logic not to regenerate workingset

koseoyoung:
works on #1872 

The issue is that even though we failed to validate, we are able to commit the block by regenerating the workingset. So I changed that if workingset doens't exist, not regenerate but return error. 

Impact on this change
1. StartExistingBlockchain() -- add validate before commit 
2. Consensus: if workingset cache has been reset because of offline -- error handling by revalidate/commit

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -602,7 +602,13 @@ func (bc *blockchain) commitBlock(ctx context.Context, blk *block.Block) error {
 	sfTimer := bc.timerFactory.NewTimer("sf.Commit")
 	err = bc.sf.Commit(ctx, blk)
 	sfTimer.End()
-	// detach working set so it can be freed by GC
+	if errors.Cause(err) == factory.ErrNotValidated {
+		// validate again and re-commit
+		log.L().Debug("Revalidate and generate workingset to commit.", zap.Uint64("height", blk.Height()))
+		if err = bc.validator.Validate(ctx, blk); err == nil {
+			err = bc.sf.Commit(ctx, blk)
+		}
+	}
 	if err != nil {
 		log.L().Panic("Error when committing states.", zap.Error(err))
 	}
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -534,6 +534,9 @@ func (bc *blockchain) startExistingBlockchain(ctx context.Context) error {
 			return err
 		}
 		ctx = bc.contextWithBlock(ctx, producer, blk.Height(), blk.Timestamp())
+		if err := bc.sf.Validate(ctx, blk); err != nil {
+			return err
+		}
 		if err := bc.sf.Commit(ctx, blk); err != nil {
 			return err
 		}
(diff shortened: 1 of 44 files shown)


consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -484,6 +485,14 @@ func (ctx *rollDPoSCtx) Commit(msg interface{}) (bool, error) {
...
 	case blockchain.ErrInvalidTipHeight:
 		return true, nil
+	case factory.ErrNotValidated:
+		// if the cache has been reset, we need to revalidate/commit the block
+		if err := ctx.chain.ValidateBlock(pendingBlock); err != nil {
+			return false, err
+		}
+		if err := ctx.chain.CommitBlock(pendingBlock); err != nil {
+			return false, err
+		}
```
CoderZhi: as discussed offline, these code won't be hit.

koseoyoung: yes! moved error handling logic into blockchain.go commitblock()

koseoyoung:
@dustinxie After discussion with zhi, the unit test seems not correct(bigint_test.go) We shouldn’t commit the block which failed to validate. And Commit() is ok to regenerate workingset when we assume the block is valid. Can we delete that part where it tries to commit again even though validation fails in the unit test?

blockchain/blockchain.go
```diff
@@ -599,7 +602,13 @@ func (bc *blockchain) commitBlock(ctx context.Context, blk *block.Block) error {
 	sfTimer := bc.timerFactory.NewTimer("sf.Commit")
 	err = bc.sf.Commit(ctx, blk)
 	sfTimer.End()
-	// detach working set so it can be freed by GC
+	if errors.Cause(err) == factory.ErrNotValidated {
```
dustinxie: this is kine of strange, Commit() --> ErrNotValidated --> then re-commit?
can there be a better way?

blockchain/blockchain.go
```diff
@@ -534,6 +534,9 @@ func (bc *blockchain) startExistingBlockchain(ctx context.Context) error {
 			return err
 		}
 		ctx = bc.contextWithBlock(ctx, producer, blk.Height(), blk.Timestamp())
+		if err := bc.sf.Validate(ctx, blk); err != nil {
```
dustinxie: why need to add this?
why the code can work now w/o it?

dustinxie:
> @dustinxie After discussion with zhi, the unit test seems not correct(bigint_test.go) We shouldn’t commit the block which failed to validate. And Commit() is ok to regenerate workingset when we assume the block is valid. Can we delete that part where it tries to commit again even though validation fails in the unit test?

yes, we should understand the test logic, and modify accordingly

basically, it need to be able to detect a negative transfer, and not let such block to be committed

blockchain/blockchain.go
dustinxie: can we nil the workingset, when validation fail?

factory can regenerate the workingset, but it will fail upon Commit()

koseoyoung:
I tried to have a validationFailFlag with workingset inside of state factory to set up the flag true if a block fails to be validated and remember the result when try to commit again. But it has a problem because negative transfer fails not in factory.Validate() but in blockvalidator side.

PR #1890, opened 2020-02-13
evm transfer proto

Frankonly:
`BlockEvmTransfers` and `ActionEvmTransfer` are designed for transporting evm transfers of certain block/action. I will add one or two APIs for APIServer. Requester like analytics can send grpc requests to query those data.
API design: `blockHeight -> BlockEvmTransfers` , `actionHash -> ActionEvmTransfer`

> how will this protobuf be used? can you add some description?

Ok, update above.


CoderZhi:
how will this protobuf be used? can you add some description?

lizhefeng:
If those protobufs are defined for API, please outline the API design.

PR #1892, opened 2020-02-14
Apply Two Layer Trie and Refactor WorkingSet

CoderZhi:
This is based on #1881 
1. Using two layer trie in factory to store archive data
2. Refactor workingset, and delete statetx


state/factory/factory.go
```diff
@@ -122,6 +134,32 @@ func InMemTrieOption() Option {
...
+	case db.ErrNotExist:
+		if !create {
+			return nil, err
+		}
+	case nil:
+		break
+	default:
+		return nil, err
+	}
+	dbForTrie, err := db.NewKVStoreForTrie(ArchiveTrieNamespace, dao)
```
yutongp: so if the kvstore exist already, this `NewKVStoreForTrie` doesn't have any side effect?

state/factory/factory.go
```diff
@@ -218,18 +251,102 @@ func (sf *factory) NewWorkingSet() (WorkingSet, error) {
...
+			flusher.KVStoreWithBuffer().MustDelete(ns, key)
+			nsHash := hash.Hash160b([]byte(ns))
+
+			return tlt.Delete(nsHash[:], key)
+		},
+		digestFunc: func() hash.Hash256 {
+			return hash.Hash256b(flusher.SerializeQueue())
+		},
+		finalizeFunc: func(h uint64) error {
+			if finalized {
```
yutongp: imo define all these callback/closure is not as clear as defining methods on a given struct.

state/factory/util.go
```diff
@@ -28,6 +28,18 @@ import (
 	"github.com/iotexproject/iotex-core/pkg/log"
 )
 
+func processOptions(opts ...protocol.StateOption) (bool, uint64, string, []byte, error) {
```
yutongp: i think just return cfg, make this function simpler to read and use

PR #1893, opened 2020-02-14
add native staking delegate struct

dustinxie:
define the delegate struct for native staking
WIP on another PR for storing/accessing these delegates


action/protocol/staking/protocol.go
```diff
@@ -39,8 +50,26 @@ func NewProtocol() *Protocol {
...
-	return nil, nil
+	// read native staking candidate list
+	if len(p.nativeCand) == 0 {
+		_, err := sm.State(
+			&p.nativeCand,
+			protocol.NamespaceOption(factory.StakingNameSpace),
+			protocol.KeyOption(factory.NativeCandKey))
+		if err != nil && errors.Cause(err) != state.ErrStateNotExist {
+			return nil, err
+		}
```
CoderZhi: put in Start func

action/protocol/staking/protocol.go
```diff
@@ -7,23 +7,34 @@
...
+var (
+	// ErrInvalidAmount indicates an invalid staking amount
+	ErrInvalidAmount = errors.New("invalid staking amount")
+)
+
 // Protocol defines the protocol of handling staking
 type Protocol struct {
-	addr address.Address
+	addr       address.Address
+	nativeCand state.CandidateList
```
CoderZhi: convert to a map if possible

state/candidate.go
```diff
@@ -29,13 +30,18 @@ var (
...
+type (
+	// CandName is the 12-byte candidate name
+	CandName [12]byte
+
+	// Candidate indicates the structure of a candidate
+	Candidate struct {
+		Address       string
+		Votes         *big.Int
+		RewardAddress string
+		CanName       []byte // used as identifier to merge with native staking result, not part of protobuf
```
CoderZhi: Should we restrict the type or it has been too late with backward compatible problem?

dustinxie: we could
but for safety (though the risk is low) would just leave it
we have func to12Bytes() to guarantee what we read from election must be 12-byte

action/protocol/staking/protocol.go
dustinxie: will fix in next PR

action/protocol/staking/protocol.go
dustinxie: fix in next PR

action/protocol/staking/bucket_index.go
```diff
@@ -8,14 +8,13 @@ package staking
 
 import (
 	"github.com/golang/protobuf/proto"
-	"github.com/iotexproject/iotex-core/state"
-	"github.com/pkg/errors"
-
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
+	"github.com/pkg/errors"
```
lizhefeng: Group imports

PR #1894, opened 2020-02-14
Delete NewWorkingSet and DeleteWorkingSet APIs

CoderZhi:
based on #1892 
After this PR, workingset is no longer a public struct, but an internal implementation of StateManager inside of factory package.


state/factory/workingset.go
```diff
@@ -32,105 +27,74 @@ var (
...
-		[]string{},
-	)
 )
 
 func init() {
 	prometheus.MustRegister(stateDBMtc)
-	prometheus.MustRegister(dbBatchSizelMtc)
 }
 
 type (
```
dustinxie: would suggest we avoid keep changing these interface radically ...

understand that the workingset implementation can be refactored using these get/put/delete/commitFunc(), and NewWorkingSet() replaced by workingSetCreator interface

but the whole team need to keep chasing and understand the new change ...

Issue #1773, opened 2019-12-05
New ioctl command for bc block and unit test


CoderZhi:
Refer to https://github.com/iotexproject/iotex-core/pull/1751 for example

PR #1898, opened 2020-02-15
Add ioctl bc unit test

kavonm:
Add unit test cases for ioctl bc command
https://github.com/iotexproject/iotex-core/issues/1773


ququzone:
should test all public method in `bc` package

kavonm:
> should test all public method in `bc` package

OK. I will add more testing cases for bc.go.
By the way, I just focused on the related functions for bc command.  I made the reference of existing account and alias commands. It seems that like unit testing cases are not complete.

PR #1899, opened 2020-02-16
add ioctl command for bc info and unit test

redxio:


ioctl/newcmd/bc/bcinfo.go
```diff
@@ -0,0 +1,82 @@
...
+	bcInfoCmdUse, _ := client.SelectTranslation(bcInfoCmdUses)
+
+	cmd := &cobra.Command{
+		Use:   bcInfoCmdUse,
+		Short: bcInfoCmdShort,
+		Args:  cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			apiServiceClient, err := client.APIServiceClient(ioctl.APIServiceConfig{
+				Endpoint: client.Config().Endpoint,
+				Insecure: (cmd.Flag("insecure") != nil && config.Insecure) || !client.Config().SecureConnect,
```
CoderZhi: the parameters should be optional. By default, it could look like
`apiServiceClient, err := client.APIServiceClient()` or
`apiServiceClient, err := client.InsecureAPIServiceClient()`

redxio: fixed

ioctl/newcmd/bc/bcinfo_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	apiServiceClient := mock_apiserviceclient.NewMockServiceClient(ctrl)
+	client.EXPECT().APIServiceClient(gomock.Any()).Return(apiServiceClient, nil).Times(1)
+
+	chainMetaResponse := &iotexapi.GetChainMetaResponse{ChainMeta: &iotextypes.ChainMeta{}}
+	apiServiceClient.EXPECT().GetChainMeta(gomock.Any(), gomock.Any()).Return(chainMetaResponse, nil).Times(1)
+
+	cmd := NewBCInfoCmd(client)
+	_, err := util.ExecuteCmd(cmd)
+	require.NoError(t, err)
```
CoderZhi: for unit test: not only the success case, but also all failure cases:
1. client.EXPECT().APIServiceClient(gomock.Any()).Return(nil, some error)
2. apiServiceClient.EXPECT().GetChainMeta(gomock.Any(), gomock.Any()).Return(nil, some error)

redxio: fixed

PR #1901, opened 2020-02-19
add self-stake and use owner address to index delegate

dustinxie:


action/protocol/staking/protocol.go
```diff
@@ -114,3 +147,55 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+
+	if _, err := address.FromString(d.RewardAddress); err != nil {
+		return errors.Wrap(err, "invalid reward address")
+	}
+
+	if d.Votes.Sign() <= 0 {
+		return ErrInvalidAmount
+	}
+	return nil
+}
```
CoderZhi: this looks like a static function, or a function of struct `Delegate`.

action/protocol/staking/protocol.go
```diff
@@ -114,3 +147,55 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+
+func (p *Protocol) addDelegate(d *Delegate) error {
+	if err := p.validateDelegate(d); err != nil {
+		return err
+	}
+
+	if !p.hasDelegate(d.CanName) {
```
CoderZhi: this function is not well defined. it should either "return error if canName is duplicate" or "replace the existing one"

CoderZhi: It is hard to review a private function because there is no context of how it will be used, and the functionality is conflict with the function name.

action/protocol/staking/protocol.go
```diff
@@ -114,3 +147,55 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+
+func (p *Protocol) delDelegate(name CandName) {
+	delete(p.delegates, name)
+}
+
+func (p *Protocol) addVote(name CandName, amount *big.Int) error {
+	if p.hasDelegate(name) {
+		return p.delegates[name].AddVote(amount)
+	}
+	return nil
```
CoderZhi: there will be a problem:
0. we have delegate [d1, d2, d3, d4]
1. voter votes delegate "d_not_exist"
2. someone registered as "d_not_exist"

action/protocol/staking/protocol.go
```diff
@@ -114,3 +147,55 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+}
+
+func (p *Protocol) subVote(name CandName, amount *big.Int) error {
+	if p.hasDelegate(name) {
+		return p.delegates[name].SubVote(amount)
+	}
+	return nil
```
CoderZhi: some problem here

action/protocol/staking/protocol.go
```diff
@@ -54,8 +62,18 @@ func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.St
...
 	case *action.Restake:
 		return p.handleRestake(ctx, act, sm)
+	default:
+		return nil, errors.New("invalid staking message")
 	}
-	return nil, nil
+
+	r := action.Receipt{}
+
+	// store the candidate list back
```
CoderZhi: I thought we are going to store delegates one by one as we discussed offline yesterday.

action/protocol/staking/protocol.go
```diff
@@ -80,6 +98,21 @@ func (p *Protocol) ForceRegister(r *protocol.Registry) error {
 	return r.ForceRegister(protocolID, p)
 }
 
+// CreateGenesisStates creates genesis states
+func (p *Protocol) CreateGenesisStates(ctx context.Context, sm protocol.StateManager) error {
+	// read native staking candidate list
+	if len(p.delegates) == 0 {
```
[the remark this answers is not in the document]
dustinxie: Start() does not have the StateManger needed to read from stateDB

will revise according to our latest discussion

action/protocol/staking/protocol.go
dustinxie: yes, you are right

action/protocol/staking/protocol.go
dustinxie: this needs to be change to use Owner address

action/protocol/staking/protocol.go
dustinxie: this need to be changed per latest discussion

first do for the in-memory map, then try entry in DB

action/protocol/staking/protocol.go
dustinxie: right, this need to be change per latest discussion

PR #1904, opened 2020-02-21
fix a map initialize bug 

koseoyoung:


action/protocol/rewarding/reward.go
```diff
@@ -157,7 +157,7 @@ func (p *Protocol) GrantEpochReward(
 		exemptAddrs[addr.String()] = nil
 	}
 
-	var uqd map[string]bool
+	uqd := make(map[string]bool)
```
yutongp: i think it is ok to define a nil map here.
we can make the map only in the else statement

koseoyoung: ok! changed in latest commit

PR #1910, opened 2020-02-24
Add systemLogIndexer into BlockDao

Frankonly:


blockchain/blockdao/blockdao.go
```diff
@@ -117,30 +119,33 @@ type (
...
-		headerCache   *cache.ThreadSafeLruCache
-		bodyCache     *cache.ThreadSafeLruCache
-		footerCache   *cache.ThreadSafeLruCache
-		cfg           config.DB
-		mutex         sync.RWMutex // for create new db file
-		tipHeight     uint64
+		compressBlock    bool
+		kvStore          db.KVStore
+		indexer          BlockIndexer
+		systemLogIndexer systemlog.Indexer
```
CoderZhi: Don't add this variable.
indexer could be a wrapper of a list of indexers, no need to modify blockdao.go at all

Frankonly: I consider this before, but the problem is it would affect many files. I will do it now.

PR #1911, opened 2020-02-24
add stakingGet/Put/DelDelegate API

dustinxie:


action/protocol/staking/delegate.go
```diff
@@ -247,7 +269,55 @@ func (m DelegateMap) Deserialize(buf []byte) error {
 	}
 
 	for _, v := range pb.Delegates {
-		m[hash.BytesToHash160(v.Owner)] = fromProto(v)
+		m[v.Owner] = fromProto(v)
 	}
 	return nil
 }
+
+func stakingGetDelegate(sr protocol.StateReader, name address.Address) (*Delegate, error) {
```
CoderZhi: why we need to add a prefix "staking" to the function name?

action/protocol/staking/stakingpb/staking.proto
```diff
@@ -20,7 +20,7 @@ message BucketIndices {
 }
 
 message Delegate {
-    bytes owner = 1;
+    string owner = 1;
     string address = 2;
     string rewardAddress = 3;
```
CoderZhi: nit: to be space efficient, it is ok to store owner, address, rewardadress as bytes.

action/protocol/staking/delegate.go
```diff
@@ -7,16 +7,28 @@
...
+)
+
+/*
+ * keys in namespace factory.DelegateNameSpace are prefixed with 1-byte tag
+ * to indicate the type of stored object. This enables to quickly scan the
+ * entire bucket and filter out objects of a certain type
+ */
+const (
+	inActiveDel = byte(iota)
+	activeDel   = inActiveDel + 1
```
CoderZhi: could you elaborate or comment on the usage of these two parameters?

action/protocol/staking/delegate.go
```diff
@@ -7,16 +7,28 @@
...
+	"github.com/iotexproject/iotex-core/state/factory"
+)
+
+/*
+ * keys in namespace factory.DelegateNameSpace are prefixed with 1-byte tag
+ * to indicate the type of stored object. This enables to quickly scan the
+ * entire bucket and filter out objects of a certain type
+ */
+const (
+	inActiveDel = byte(iota)
```
CoderZhi: inactive

action/protocol/staking/delegate.go
dustinxie: to name staking protocol internal API

already added other APIs like stakingGet/PutBucket() etc.

action/protocol/staking/stakingpb/staking.proto
dustinxie: yes, but lots of proto used string to store, so just use string

action/protocol/staking/delegate.go
dustinxie: prefix for storing inactive and active delegates, see stakingPutDelegate()

so on Start(), can quickly scan the entire bucket and filter out active delegates

action/protocol/staking/delegate.go
dustinxie: chagned

action/protocol/staking/delegate.go
CoderZhi: as they are internal, the prefix is redundant.

PR #1913, opened 2020-02-24
Move index builder to blockindex

CoderZhi:
Move IndexBuilder into blockindex package and delete Commit API

### DIFF
--- a/blockindex/indexbuilder.go
+++ b/blockindex/indexbuilder.go
@@ -4,20 +4,18 @@
 // permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
 // License 2.0 that can be found in the LICENSE file.
 
-package blockdao
+package blockindex
 
 import (
 	"strconv"
 
-	"github.com/iotexproject/go-pkgs/hash"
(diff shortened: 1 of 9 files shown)


blockindex/indexbuilder.go
```diff
@@ -129,19 +129,19 @@ func (ib *IndexBuilder) init() error {
...
 			return err
 		}
-		if err := ib.indexer.PutBlock(blk); err != nil {
-			return err
-		}
+		blks = append(blks, blk)
 		// commit once every 5000 blocks
 		if startHeight%5000 == 0 || startHeight == tipHeight {
-			if err := ib.indexer.Commit(); err != nil {
+			if err := ib.indexer.PutBlocks(blks); err != nil {
```
dustinxie: don't think this is a good design API

this could cause significant memory throttle and may slow down the full-node sync

raullenchai:
> can we avoid keep changing these common base interface/codeflow?
> 
> these codes are working well and stood the test of time, also everyone on team is familiar with it.
> 
> would suggest not change them unless there is a have-to reason

resolved offline.

Issue #1925, opened 2020-02-25
xctl cannot do action transfer


dustinxie:
### What version of iotex-core are you using (`ioctl version`)?
latest

### What operating system and processor architecture are you using?
macOS

### What did you do? If possible, provide a recipe for reproducing the error.
./bin/xctl account createadd name
./bin/xctl action transfer xxx 10 -s name

### What did you expect to see?
ask for password, then continue

### What did you see instead?
Enter private key #io1l2xgxxx

it should not ask for private key, but ask password

PR #1929, opened 2020-02-26
update IsSignerExist for sm2

Frankonly:
#1925


ioctl/cmd/account/account.go
```diff
@@ -178,6 +178,29 @@ func GetAccountMeta(addr string) (*iotextypes.AccountMeta, error) {
...
+	for _, ksAccount := range ks.Accounts() {
+		if address.Equal(addr, ksAccount.Address) {
+			return true
+		}
+	}
+	// find the account in pem files
+	pemFilePath := sm2KeyPath(addr)
+	_, err = os.Stat(pemFilePath)
+	if err == nil {
+		return true
```
dustinxie: return (err == nil)?

dustinxie: also, i recall seeing this code (check KeyStore then PEM file) in your SM2 check-in

why it is missed

Frankonly: I used to put all account related functions in account.go. I don't why someone put this function in action.go, which is odd and I missed it last time. Now, I have put this function in account.go.

ioctl/cmd/account/account.go
```diff
@@ -178,6 +178,25 @@ func GetAccountMeta(addr string) (*iotextypes.AccountMeta, error) {
...
 
+// IsSignerExist checks whether signer account is existed
+func IsSignerExist(signer string) bool {
+	addr, err := address.FromString(signer)
+	if err != nil {
+		return false
+	}
+	// find the account in keystore
+	ks := keystore.NewKeyStore(config.ReadConfig.Wallet,
+		keystore.StandardScryptN, keystore.StandardScryptP)
```
CoderZhi: format

ioctl/cmd/account/account.go
```diff
@@ -178,6 +178,25 @@ func GetAccountMeta(addr string) (*iotextypes.AccountMeta, error) {
...
+	for _, ksAccount := range ks.Accounts() {
+		if address.Equal(addr, ksAccount.Address) {
+			return true
+		}
+	}
+	// find the account in pem files
+	_, err = findSm2PemFile(addr)
```
CoderZhi: I still don't see why we need a mixture of sm2 and current pems. Sm2 account format is not determined by user input, https://github.com/iotexproject/iotex-core/blob/1e3895b75321f0ac8b946621958923c08d518fac/ioctl/cmd/account/accountcreateadd.go#L47. "xctl" always and only creates or uses sm2 accounts.

Overall, I think you implementation is misleading to users of "xctl", who shouldn't be aware of the exists of two different account systems or algorithms.

PR #1931, opened 2020-02-26
add delegate map into staking protocol struct

dustinxie:
and some API to manage Delegate, to unblock Zhefeng's handler work

### DIFF
--- a/action/protocol/staking/delegate_map.go
+++ b/action/protocol/staking/delegate_map.go
@@ -1,0 +1,74 @@
+// Copyright (c) 2020 IoTeX Foundation
+// This is an alpha (internal) release and is not suitable for production. This source code is provided 'as is' and no
+// warranties are given as to title or non-infringement, merchantability or fitness for purpose and, to the extent
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package staking
+
+import (
+	"sort"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/pkg/errors"
+
+	"github.com/iotexproject/iotex-core/action/protocol/staking/stakingpb"
+)
+
+type (
+	// DelegateMap is a map of delegates using owner's 12-byte name as key
+	DelegateMap map[CandName]*Delegate
+)
+
+// Contains returns true if the map contains the delegate by name
+func (m DelegateMap) Contains(name CandName) bool {
+	_, ok := m[name]
+	return ok
+}
+
+// Get returns the delegate by name
+func (m DelegateMap) Get(name CandName) *Delegate {
+	return m[name]
+}
+
+// Put writes the delegate into map, will overwrite if name already exists
+func (m DelegateMap) Put(d *Delegate) error {
+	if err := d.Validate(); err != nil {
+		return err
+	}
+
+	m[d.CanName] = d
+	return nil
+}
+
+// Delete deletes the delegate by name
+func (m DelegateMap) Delete(name CandName) {
+	delete(m, name)
+}
+
+// Serialize serializes a DelegateMap to bytes
+func (m DelegateMap) Serialize() ([]byte, error) {
+	l := make(DelegateList, 0, len(m))
+	for _, v := range m {
+		l = append(l, v)
+	}
+	sort.Sort(l)
+	return proto.Marshal(l.toProto())
+}
+
+// Deserialize deserializes bytes to DelegateMap
+func (m DelegateMap) Deserialize(buf []byte) error {
+	pb := &stakingpb.Delegates{}
+	if err := proto.Unmarshal(buf, pb); err != nil {
+		return errors.Wrap(err, "failed to unmarshal delegate list")
+	}
+
+	for k := range m {
+		delete(m, k)
+	}
+
+	for _, v := range pb.Delegates {
+		m[ToCandName(v.CanName)] = fromProto(v)
+	}
+	return nil
+}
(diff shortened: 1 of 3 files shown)


action/protocol/staking/protocol.go
```diff
@@ -114,3 +119,36 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+
+func (p *Protocol) addDelegate(d *Delegate) error {
+	if err := d.Validate(); err != nil {
+		return err
+	}
+
+	if !p.hasDelegate(d.Owner) {
+		p.delegates[d.Owner.String()] = d
+	}
+	return nil
```
CoderZhi: is it possible this function be called multiple times with the same key? if so, should we return error since we don't overwrite? if not, why we test "hasDelegate"

action/protocol/staking/protocol_test.go
```diff
@@ -0,0 +1,116 @@
...
+		require.NoError(stk.addVote(v.d.Owner, v.add))
+		require.Equal(v.final, stk.delegates[v.d.Owner.String()].Votes)
+	}
+
+	// test delegate that does not exist
+	noName, err := address.FromBytes(make([]byte, 20))
+	require.NoError(err)
+	require.False(stk.hasDelegate(noName))
+	require.Equal(ErrInvalidAmount, errors.Cause(stk.addVote(tests[0].d.Owner, big.NewInt(-1))))
+	exceed := big.NewInt(0).Add(tests[0].final, big.NewInt(1))
```
CoderZhi: it is hard to follow tests[0] which has been modified in line 85

action/protocol/staking/protocol.go
```diff
@@ -114,3 +119,36 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+		return p.delegates[name.String()].AddVote(amount)
+	}
+	return nil
+}
+
+func (p *Protocol) subVote(name address.Address, amount *big.Int) error {
+	if p.hasDelegate(name) {
+		return p.delegates[name.String()].SubVote(amount)
+	}
+	return nil
```
CoderZhi: return some error?

action/protocol/staking/protocol.go
```diff
@@ -114,3 +119,36 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+
+func (p *Protocol) delDelegate(name address.Address) {
+	delete(p.delegates, name.String())
+}
+
+func (p *Protocol) addVote(name address.Address, amount *big.Int) error {
+	if p.hasDelegate(name) {
+		return p.delegates[name.String()].AddVote(amount)
+	}
+	return nil
```
CoderZhi: return some error?

action/protocol/staking/protocol.go
```diff
@@ -114,3 +119,36 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
 	// TODO
 	return nil, nil
 }
+
+func (p *Protocol) hasDelegate(name address.Address) bool {
+	return p.delegates.Contains(name)
+}
+
+func (p *Protocol) addDelegate(d *Delegate) error {
```
CoderZhi: could this be a function of DelegateMap? same question for the rest

action/protocol/staking/protocol.go
```diff
@@ -114,3 +119,36 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
 	// TODO
 	return nil, nil
 }
+
+func (p *Protocol) hasDelegate(name address.Address) bool {
```
CoderZhi: if this is a private function, why not call p.delegates.Contains(name) directly?

action/protocol/staking/protocol.go
dustinxie: good point, thx

need to think the correct logic, since we discussed ysterday that we'll use 12-byte name. In this case, if an active delegate change the name, then we need to update this delegate, even though it's already in the map

action/protocol/staking/protocol_test.go
dustinxie: ok, line 85 is to get rid of the first 2 fail case, so the rest of test only use the remaining 3 valid case

action/protocol/staking/protocol.go
dustinxie: yes, need to think the correct logic and re-write this entire thing

action/protocol/staking/delegate_map.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Get returns the delegate by name
+func (m DelegateMap) Get(name CandName) *Delegate {
+	return m[name]
+}
+
+// Put writes the delegate into map, will overwrite if name already exists
+func (m DelegateMap) Put(d *Delegate) error {
+	if err := d.Validate(); err != nil {
```
CoderZhi: Validate is more like a function of DelegateMap

action/protocol/staking/delegate_map.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/pkg/errors"
+
+	"github.com/iotexproject/iotex-core/action/protocol/staking/stakingpb"
+)
+
+type (
+	// DelegateMap is a map of delegates using owner's 12-byte name as key
+	DelegateMap map[CandName]*Delegate
```
yutongp: I think you would need two maps. one is CandName to OwnerAddress mapping and one is OwnerAddress to Delegate instance mapping

PR #1932, opened 2020-02-26
add Filter() to KVStore

dustinxie:
this will be used to quickly scan the stateDB to filter out active delegates upon Start()

### DIFF
--- a/db/db_bolt.go
+++ b/db/db_bolt.go
@@ -112,7 +112,7 @@ func (b *boltDB) Filter(namespace string, cond Condition) ([][]byte, [][]byte, e
 	if err := b.db.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket([]byte(namespace))
 		if bucket == nil {
-			return errors.Wrapf(ErrNotExist, "bucket = %x doesn't exist", []byte(namespace))
+			return errors.Wrapf(ErrBucketNotExist, "bucket = %x doesn't exist", []byte(namespace))
 		}
 
 		return bucket.ForEach(func(k, v []byte) error {
@@ -128,6 +128,10 @@ func (b *boltDB) Filter(namespace string, cond Condition) ([][]byte, [][]byte, e
 		})
 	}); err != nil {
 		return nil, nil, err
+	}
+
+	if len(fk) == 0 {
+		return nil, nil, errors.Wrap(ErrNotExist, "filter returns no match")
 	}
 	return fk, fv, nil
 }
--- a/db/db_bolt.go
+++ b/db/db_bolt.go
@@ -106,6 +106,32 @@ func (b *boltDB) Get(namespace string, key []byte) ([]byte, error) {
 	return nil, errors.Wrap(ErrIO, err.Error())
 }
 
+// Filter returns <k, v> pair in a bucket that meet the condition
+func (b *boltDB) Filter(namespace string, cond Condition) ([][]byte, [][]byte, error) {
+	var fk, fv [][]byte
+	if err := b.db.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket([]byte(namespace))
+		if bucket == nil {
+			return errors.Wrapf(ErrNotExist, "bucket = %x doesn't exist", []byte(namespace))
+		}
+
+		return bucket.ForEach(func(k, v []byte) error {
+			if cond(k, v) {
+				key := make([]byte, len(k))
+				copy(key, k)
+				value := make([]byte, len(v))
+				copy(value, v)
+				fk = append(fk, key)
+				fv = append(fv, value)
+			}
+			return nil
+		})
+	}); err != nil {
+		return nil, nil, err
+	}
+	return fk, fv, nil
+}
+
 // Range retrieves values for a range of keys
 func (b *boltDB) Range(namespace string, key []byte, count uint64) ([][]byte, error) {
 	value := make([][]byte, count)
--- a/db/db_test.go
+++ b/db/db_test.go
@@ -7,18 +7,19 @@
 package db
 
 import (
+	"bytes"
 	"context"
 	"io/ioutil"
 	"os"
 	"testing"
 
-	"github.com/iotexproject/iotex-core/db/batch"
-	"github.com/iotexproject/iotex-core/testutil"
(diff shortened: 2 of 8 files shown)


misc/scripts/mockgen.sh
```diff
@@ -78,7 +78,7 @@ mockgen -destination=./action/protocol/mock_protocol_test.go  \
         -package=protocol \
         Protocol
 
-mockgen -destination=./db/mock_kvstore_test.go  \
+mockgen -destination=./db/mock_kvstore.go  \
```
dustinxie: modify file name, to be able to export NewMockKVStore() for use in other packages

db/db_bolt.go
```diff
@@ -106,6 +106,32 @@ func (b *boltDB) Get(namespace string, key []byte) ([]byte, error) {
 	return nil, errors.Wrap(ErrIO, err.Error())
 }
 
+// Filter returns <k, v> pair in a bucket that meet the condition
+func (b *boltDB) Filter(namespace string, cond Condition) ([][]byte, [][]byte, error) {
+	var fk, fv [][]byte
+	if err := b.db.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket([]byte(namespace))
+		if bucket == nil {
+			return errors.Wrapf(ErrNotExist, "bucket = %x doesn't exist", []byte(namespace))
```
CoderZhi: namespace not exist may be different from "ErrNotExist". It is possible that the bucket is empty or the entities in the bucket are all filtered by the "cond"

db/kvstore_impl.go
```diff
@@ -91,6 +91,11 @@ func (m *memKVStore) Delete(namespace string, key []byte) error {
 	return nil
 }
 
+// Filter returns <k, v> pair in a bucket that meet the condition
+func (m *memKVStore) Filter(namespace string, c Condition) ([][]byte, [][]byte, error) {
+	return nil, nil, nil
```
CoderZhi: return unsupported error

db/kvstorewithbuffer.go
```diff
@@ -168,6 +168,10 @@ func (kvb *kvStoreWithBuffer) MustDelete(ns string, key []byte) {
 	kvb.buffer.Delete(ns, key, "failed to delete %x in %s", key, ns)
 }
 
+func (kvb *kvStoreWithBuffer) Filter(ns string, c Condition) ([][]byte, [][]byte, error) {
+	return kvb.store.Filter(ns, c)
```
CoderZhi: need to filter the entities in buffer as well

db/db_test.go
```diff
@@ -315,7 +316,63 @@ func TestDeleteBucket(t *testing.T) {
...
+	testPath := testFile.Name()
+	cfg.DbPath = testPath
+	t.Run("Bolt DB", func(t *testing.T) {
+		testutil.CleanupPath(t, testPath)
+		defer testutil.CleanupPath(t, testPath)
+		testFunc(NewBoltDB(cfg), t)
+	})
+}
+
+func TestFilter(t *testing.T) {
```
CoderZhi: test bucket doesn't exist case. test all entities meet cond case.

db/db_bolt.go
dustinxie: don't quite get it, this is how we check if a bucket exist in the bolt DB ... 

even if the bucket is empty (no entries in it), returned "bucket" will not be nil if the name exist

db/kvstore_impl.go
dustinxie: ok

db/kvstorewithbuffer.go
dustinxie: ok

db/db_test.go
dustinxie: ok

db/db_bolt.go
CoderZhi: if we cannot return any entity from a bucket because of a filtering "cond", should we return ErrNotExist?

db/kvstorewithbuffer.go
```diff
@@ -168,6 +168,24 @@ func (kvb *kvStoreWithBuffer) MustDelete(ns string, key []byte) {
...
+func (kvb *kvStoreWithBuffer) Filter(ns string, cond Condition) ([][]byte, [][]byte, error) {
+	fk, fv, err := kvb.store.Filter(ns, cond)
+	if err != nil {
+		return fk, fv, err
+	}
+
+	// filter the entries in buffer
+	for i := 0; i < kvb.buffer.Size(); i++ {
+		entry, _ := kvb.buffer.Entry(i)
+		k, v := entry.Key(), entry.Value()
```
CoderZhi: the entries in a buffer are not naive key-value pair. For "Delete", we need to remove it from fk and fv, and only do append if it is a "Put" type.

db/kvstore_impl.go
```diff
@@ -91,6 +91,11 @@ func (m *memKVStore) Delete(namespace string, key []byte) error {
 	return nil
 }
 
+// Filter returns <k, v> pair in a bucket that meet the condition
+func (m *memKVStore) Filter(namespace string, c Condition) ([][]byte, [][]byte, error) {
+	return nil, nil, errors.New("in-memory KVStore does not support Filter()")
```
CoderZhi: Add a TODO, it is a good warm up task

db/db_bolt.go
dustinxie: i think this is not an error, but no entries match cond, in this case fk, fv will be nil, something like:

fk, fv, err := Filter()
if err != nil {
    return err
}

if len(fk) ==0 or len(fv) == 0 {
   // no match, do nothing
   return nil
}

process(fk, fv)
return nil

db/kvstorewithbuffer.go
```diff
@@ -168,6 +169,34 @@ func (kvb *kvStoreWithBuffer) MustDelete(ns string, key []byte) {
...
+		if cond(k, v) {
+			if entry.WriteType() == batch.Put {
+				fk = append(fk, k)
+				fv = append(fv, v)
+			} else if entry.WriteType() == batch.Delete {
```
CoderZhi: switch

PR #1941, opened 2020-02-27
Fix foundation bonus reward bug by refactoring CandidatesByHeight 

koseoyoung:
2 fixes 

1. Not only for epoch reward, but also for foundation bonus reward, we should re-calculate candidate list based on intensityRate/ blacklist to pick top 36 candidate 
2. In case of the hard kick out (intensityRate is 100), even though the total candidate count is less than 36, we shouldn't include in blockproducer(top 36). Therefore, always they won't be able to produce block/get any reward.


action/protocol/rewarding/reward.go
```diff
@@ -486,3 +493,32 @@ func (p *Protocol) assertLastBlockInEpoch(blkHeight uint64, epochNum uint64, rp
...
+		}
+	}
+	// sort again with updated voting power
+	sorted := util.Sort(updatedVotingPower, epochNum)
+	var verifiedCandidates state.CandidateList
+	for _, name := range sorted {
+		verifiedCandidates = append(verifiedCandidates, candidatesMap[name])
+	}
+	return verifiedCandidates
+}
```
CoderZhi: the code looks redundant. Let's figure out a better way to prevent the redundancy.

koseoyoung: To avoid code duplicate, refactor CandidatesByHeight/ReadBP~/ReadActiveBP~ to return filtered candidate list -- so reward, ioctl doesn't have to re-calculate

koseoyoung:
1. After this PR, candidatesByHeight/DelegatesByEpoch ... API will return filtered candidate list with decreased voting power, therefore reward calculation /ioctl /frontend doesn't have to recalculate candidate.
2. hard kick-out should exclude the delegate list

action/protocol/poll/governance_protocol.go
```diff
@@ -399,29 +378,41 @@ func (p *governanceChainCommitteeProtocol) readBlockProducersByEpoch(ctx context
...
 	intensityRate := float64(uint32(100)-unqualifiedList.IntensityRate) / float64(100)
 	for _, cand := range candidates {
-		candidatesMap[cand.Address] = cand
 		if _, ok := unqualifiedList.BlacklistInfos[cand.Address]; ok {
 			// if it is an unqualified delegate, multiply the voting power with kick-out intensity rate
 			votingPower := new(big.Float).SetInt(cand.Votes)
-			newVotingPower, _ := votingPower.Mul(votingPower, big.NewFloat(intensityRate)).Int(nil)
-			updatedVotingPower[cand.Address] = newVotingPower
-		} else {
+			cand.Votes, _ = votingPower.Mul(votingPower, big.NewFloat(intensityRate)).Int(nil)
```
CoderZhi: any side effect if we modify cand.Votes?

action/protocol/poll/governance_protocol.go
```diff
@@ -399,29 +378,41 @@ func (p *governanceChainCommitteeProtocol) readBlockProducersByEpoch(ctx context
...
+		}
+		if cand.Votes.Cmp(big.NewInt(0)) > 0 {
 			updatedVotingPower[cand.Address] = cand.Votes
 		}
+		candidatesMap[cand.Address] = cand
```
CoderZhi: if votes not larger than 0, shall we still add it to candidatesMap?

action/protocol/poll/governance_protocol.go
koseoyoung: just to make sure, fixed to make clone of it

action/protocol/poll/governance_protocol.go
koseoyoung: true! fixed on latest commit

PR #1942, opened 2020-02-27
Refactor staking actions

lizhefeng:


action/stake_create.go
```diff
@@ -30,28 +29,23 @@ const (
 type CreateStake struct {
 	AbstractAction
 
-	candAddress address.Address
-	amount      *big.Int
-	duration    uint32
-	autoStake   bool
-	payload     []byte
+	candName  string
```
dustinxie: i recall we discussed using CandName here?

then we can use it to directly access the delegate map

action/stake_create.go
```diff
@@ -64,11 +58,11 @@ func NewCreateStake(
 			gasLimit: gasLimit,
 			gasPrice: gasPrice,
 		},
-		candAddress: candAddr,
-		amount:      stake,
-		duration:    duration,
-		autoStake:   autoStake,
-		payload:     payload,
+		candName:  candidateName,
```
dustinxie: validate candName must be 12-byte in length

action/stake_move.go
```diff
@@ -30,13 +29,13 @@ const (
 type moveStake struct {
 	AbstractAction
 
-	candAddress address.Address
+	name        string
```
dustinxie: same as above, should use CandName

PR #1947, opened 2020-02-28
fix ioctl getKickoutList to handle Notfound 

koseoyoung:
the issue is that when call ReadState() API to get a kick-out list in ioctl, before Easter it'll return notFound error code, so need to handle this case


ioctl/cmd/node/nodedelegate.go
```diff
@@ -173,13 +173,16 @@ func delegates() error {
...
 	}
 	blacklist := &vote.Blacklist{}
-	if err := blacklist.Deserialize(kickoutListRes.Data); err != nil {
-		return output.NewError(output.SerializationError, "failed to deserialize kickout blacklist", err)
+	if kickoutListRes != nil {
+		if err := blacklist.Deserialize(kickoutListRes.Data); err != nil {
+			return output.NewError(output.SerializationError, "failed to deserialize kickout blacklist", err)
+		}
 	}
 	kickoutIntensityRate := float64(uint32(100)-blacklist.IntensityRate) / float64(100)
```
Frankonly: Personally, I think puting deserialization into function `GetKickoutList` and letting it return Blocklist are better. Thus, it can return an empty Blocklist before Easter.

koseoyoung: to be consistent with other interface such as ```GetEpochMeta```

PR #1948, opened 2020-02-28
rename delegate to candidate and some func interface refactoring

yutongp:
- rename delegate to candidate
- change some function to use more specific types
- move validation logic to validation funcs


action/protocol/staking/candidate.go
```diff
@@ -0,0 +1,204 @@
...
+		return nil, err
+	}
+
+	if _, err = sr.State(&d, protocol.NamespaceOption(factory.CandidateNameSpace), protocol.KeyOption(key)); err != nil {
+		return nil, err
+	}
+	return &d, nil
+}
+
+func stakingPutCandidate(sm protocol.StateManager, name address.Address, d *Candidate) error {
```
CoderZhi: same here

action/protocol/staking/candidate.go
```diff
@@ -0,0 +1,204 @@
...
+	key := make([]byte, len(name.Bytes()))
+	copy(key, name.Bytes())
+
+	_, err := sm.PutState(d, protocol.NamespaceOption(factory.CandidateNameSpace), protocol.KeyOption(key))
+	return err
+}
+
+func stakingDelCandidate(sm protocol.StateManager, name address.Address) error {
```
CoderZhi: and here

action/protocol/staking/candidate_map.go
```diff
@@ -0,0 +1,74 @@
...
+	l := make(CandidateList, 0, len(m))
+	for _, v := range m {
+		l = append(l, v)
+	}
+	sort.Sort(l)
+	return proto.Marshal(l.toProto())
+}
+
+// Deserialize deserializes bytes to CandidateMap
+func (m CandidateMap) Deserialize(buf []byte) error {
```
CoderZhi: Maybe we can change these two functions to
`ToCandidateList() CandidateList`
and
`FromList(l CandidateList) error`

state/factory/factory.go
```diff
@@ -42,8 +42,8 @@ const (
 	AccountKVNamespace = "Account"
 	// StakingNameSpace is the bucket name for staking state
 	StakingNameSpace = "Staking"
-	// DelegateNameSpace is the bucket name for delegate state
-	DelegateNameSpace = "Delegate"
+	// CandidateNameSpace is the bucket name for delegate state
```
CoderZhi: for candidate state

action/protocol/staking/validations.go
```diff
@@ -0,0 +1,128 @@
...
+		if !IsValidCandidateName(act.Name()) {
+			return ErrInvalidCanName
+		}
+	}
+	return nil
+}
+
+// IsValidCandidateName check if a candidate name string is valid.
+func IsValidCandidateName(s string) bool {
+	if len(s) == 0 || len(s) > 12 {
```
dustinxie: should we use len(s) != 12 here?

that is, we require front-end to send in 12-byte string

action/protocol/staking/candidate.go
```diff
@@ -0,0 +1,204 @@
...
+		c := &Candidate{}
+		if err := c.fromProto(v); err != nil {
+			return err
+		}
+		*l = append(*l, c)
+	}
+	return nil
+}
+
+func stakingGetCandidate(sr protocol.StateReader, name address.Address) (*Candidate, error) {
```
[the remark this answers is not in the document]
dustinxie: this is already explained in an earlier PR, the prefix is used on all func to access stateDB, for example, we have func named stakingGet/Put/DelBucket for bucket, and stakingGet/Put/DelBucketIndex. So this prefix is for the sake of naming consistency

IMHO, this is really just a naming preference, not sure why you keep thinking this is an issue ...

action/protocol/staking/validations.go
yutongp: I think that is just the limitation of contract. The intention is just limit name to less or equal to 12 bytes. Since now we can use string already, it makes no sense to make front-end to add any padding if the name is less than 12 bytes.

action/protocol/staking/stakingpb/staking.proto
```diff
@@ -36,9 +36,9 @@ message Candidate {
     string ownerAddress = 1;
     string operatorAddress = 2;
     string rewardAddress = 3;
-    bytes name = 4;
-    bytes votes = 5;
-    bytes selfStake = 6;
+    string name = 4;
+    string votes = 5;
```
dustinxie: i feel bytes are more natural for *big.Int
but it's OK

action/protocol/staking/validations.go
dustinxie: i see, then, should we consider remove the 12-byte restriction? 
candidate can register whatever name they like, as long as it is alpha-numeric string

action/protocol/staking/stakingpb/staking.proto
yutongp: https://golang.org/pkg/math/big/#Int.Bytes
`Bytes returns the absolute value of x as a big-endian byte slice.`
i understand we store uint there but I would like to keep the real number just in case

action/protocol/staking/validations.go
yutongp: I don't have opinion on if 12 is good limit, but we should have a limit to how long the name can be.

dustinxie: yeah, i think a reasonably longer length would suffice our needs, like 64?

PR #1951, opened 2020-02-28
candidate registration and update handle logic

yutongp:


action/protocol/staking/protocol.go
```diff
@@ -23,7 +23,8 @@ const protocolID = "staking"
 
 // Protocol defines the protocol of handling staking
 type Protocol struct {
-	addr address.Address
+	inMemCandidates CandidateMap
```
dustinxie: do you mind to add it after addr?

PR #1955, opened 2020-02-29
API to check whether owner or operator exist in candidate map

dustinxie:


action/protocol/staking/candidate_center.go
```diff
@@ -0,0 +1,127 @@
...
+// ContainsOperator returns true if the map contains the candidate by operator
+func (m CandidateCenter) ContainsOperator(operator address.Address) bool {
+	_, ok := m.operatorMap[operator]
+	return ok
+}
+
+// Get returns the candidate by name
+func (m CandidateCenter) Get(name string) *Candidate {
+	return m.nameMap[name]
+}
```
CoderZhi: we need more than GetByName, and also GetByOwner, GetByOperator

action/protocol/staking/candidate_center.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, v := range m.nameMap {
+		l = append(l, v)
+	}
+	sort.Sort(l)
+	lpb, err := l.toProto()
+	if err != nil {
+		return nil, err
+	}
+	return proto.Marshal(lpb)
+}
```
CoderZhi: This is a serialization for candidate list. I still think we could have ToCandidateList() and FromCandidateList() functions, instead of the confusing serialize and deserialize functions.

action/protocol/staking/handlers.go
```diff
@@ -55,13 +52,17 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
+
+	if p.inMemCandidates.ContainsOwner(act.OwnerAddress()) {
+		return nil, ErrAlreadyExist
 	}
 
-	if p.inMemCandidates.Contains(act.Name()) {
-		return nil, errors.New("name already in use")
+	if p.inMemCandidates.ContainsOperator(act.OperatorAddress()) {
+		return nil, ErrAlreadyExist
 	}
```
CoderZhi: duplicate checking with https://github.com/iotexproject/iotex-core/pull/1955/files#diff-6ee6adb0e34757ff08d9b44ca73da49eR67 ?

action/protocol/staking/candidate_center.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Get returns the candidate by name
+func (m CandidateCenter) Get(name string) *Candidate {
+	return m.nameMap[name]
+}
+
+// Put writes the candidate into map, overwrite an existing entry cause error
+func (m CandidateCenter) Put(d *Candidate) error {
+	if m.ContainsName(d.Name) || m.ContainsOwner(d.Owner) || m.ContainsOperator(d.Operator) {
```
yutongp: why we add this check? how to update existing candidate in men then?

yutongp: we have to support update candidate name/operatorAddress/rewardAddress

dustinxie: updated the logic

action/protocol/staking/candidate_center.go
dustinxie: Looking from caller perspective, this does serialize to []byte, and de-serialize from []byte, so i don't think the name is confusing.

it's just that it used CandidateList as an internal vehicle to perform the task

dustinxie: i think you might mean to filter out like top36 candidates from all? we might need to save that for each block

we can have another func like ToConsensusDelegates() to do that

action/protocol/staking/handlers.go
dustinxie: updated in latest push

action/protocol/staking/candidate_center.go
dustinxie: updated in latest push

action/protocol/staking/handlers.go
```diff
@@ -93,35 +87,55 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
...
+			GasConsumed:     actCtx.IntrinsicGas,
+			ContractAddress: p.addr.String(),
+		}, nil
 	}
 
+	// keep a copy of current candidate before updating
+	oldCand := c.Clone()
+	c.Name = act.Name()
+	c.Operator = act.OperatorAddress()
+	c.Reward = act.RewardAddress()
```
yutongp: these are all optional params, so user make just check one of these three params, no necessary all of them. plz refer to my original code.

action/protocol/staking/candidate_center.go
```diff
@@ -0,0 +1,132 @@
...
+// GetByOwner returns the candidate by owner
+func (m CandidateCenter) GetByOwner(owner address.Address) *Candidate {
+	return m.ownerMap[owner]
+}
+
+// Put writes the candidate into map, overwrite an existing entry cause error
+func (m CandidateCenter) Put(d *Candidate) error {
+	if m.ContainsOwner(d.Owner) {
+		return ErrAlreadyExist
+	}
```
yutongp: i still don't understand why we need this check, but allowing delete then put. it doesn't stop people making mistake if put an existing entry is their intention. So I don't know what this check prevent.

action/protocol/staking/handlers.go
yutongp: I think we should return a clone on get, instead letting user maintain clone themselves.

dustinxie: fixed in latest push

action/protocol/staking/candidate_center.go
dustinxie: as discussed remove this
candidate_update need to make sure call Del() before Put updated candidate

action/protocol/staking/handlers.go
dustinxie: updated in latest push

action/protocol/staking/handlers.go
```diff
@@ -55,20 +52,17 @@ func (p *Protocol) handleRestake(ctx context.Context, act action.Action, sm prot
...
 	owner := actCtx.Caller
 	if act.OwnerAddress() != nil {
 		owner = act.OwnerAddress()
 	}
 
+	// owner address is the unique ID for candidate
+	if p.inMemCandidates.ContainsOwner(act.OwnerAddress()) {
+		return nil, ErrAlreadyExist
+	}
+
```
yutongp: plz add back the name existing check
```
	if p.inMemCandidates.Contains(act.Name()) {	
		return nil, errors.New("name already in use")	
	}
```

action/protocol/staking/handlers.go
```diff
@@ -93,19 +87,41 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
...
 	blkCtx := protocol.MustGetBlockCtx(ctx)
-	c, err := getCandidate(sm, actCtx.Caller)
-	if err != nil {
-		return nil, err
+
+	// only owner can update candidate
+	if !p.inMemCandidates.ContainsOwner(actCtx.Caller) {
+		return nil, ErrInvalidOwner
+	}
+	c := p.inMemCandidates.GetByOwner(actCtx.Caller)
```
yutongp: can we can check if c is nil to check if owner exist?

action/protocol/staking/handlers.go
yutongp: also check operator address if is in used  or not

action/protocol/staking/handlers.go
dustinxie: ok

PR #1959, opened 2020-03-02
set 0 gas price when using sm2

Frankonly:


ioctl/cmd/action/action.go
```diff
@@ -135,6 +135,9 @@ func registerWriteCommand(cmd *cobra.Command) {
 
 // gasPriceInRau returns the suggest gas price
 func gasPriceInRau() (*big.Int, error) {
+	if account.CryptoSm2 {
```
CoderZhi: we need to pay attention not to use this flag everywhere.

Frankonly: Sure.

PR #1962, opened 2020-03-02
use string as key for candidate map

dustinxie:
1. also add self staking bucket index
2. remove serialize() per earlier comment

### DIFF
--- a/action/protocol/staking/candidate_center.go
+++ b/action/protocol/staking/candidate_center.go
@@ -6,31 +6,25 @@
 
 package staking
 
-import (
-	"sort"
-
-	"github.com/golang/protobuf/proto"
-	"github.com/iotexproject/iotex-address/address"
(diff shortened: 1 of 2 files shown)


action/protocol/staking/candidate_center.go
```diff
@@ -46,17 +38,23 @@ func (m CandidateCenter) ContainsName(name string) bool {
 }
 
 // ContainsOwner returns true if the map contains the candidate by owner
-func (m CandidateCenter) ContainsOwner(owner address.Address) bool {
+func (m CandidateCenter) ContainsOwner(owner string) bool {
```
yutongp: you don't need to change the apis, just change the internal map to use string and convert in the functions.

dustinxie: yeah, fixed in latest push

PR #1972, opened 2020-03-03
load active delegates from stateDB into candidate center upon protocol.Start()

dustinxie:


action/protocol/staking/protocol.go
```diff
@@ -109,3 +121,33 @@ func (p *Protocol) Register(r *protocol.Registry) error {
...
+	candPkHash, candValue, err := stateDB.Filter(factory.CandidateNameSpace, func(k, v []byte) bool {
+		return bytes.HasPrefix(k, []byte{candidate})
+	})
+	if err != nil {
+		return err
+	}
+
+	if len(candValue) == 0 {
+		return nil
+	}
```
yutongp: can be removed. len == 0 will just pass below for loop

action/protocol/staking/protocol.go
```diff
@@ -109,3 +121,33 @@ func (p *Protocol) Register(r *protocol.Registry) error {
...
+
+	// decode the candidate and put into candidate center
+	for i := range candPkHash {
+		c := &Candidate{}
+		if err := c.Deserialize(candValue[i]); err != nil {
+			return errors.Wrapf(err, "failed to deserialize candidate bytes %v", candValue[i])
+		}
+		p.inMemCandidates.Put(c)
```
yutongp: Put return error

action/protocol/staking/protocol.go
dustinxie: updated in latest push

action/protocol/staking/protocol.go
dustinxie: updated in latest push

action/protocol/staking/protocol.go
```diff
@@ -109,3 +121,31 @@ func (p *Protocol) Register(r *protocol.Registry) error {
...
+	for i := range candPkHash {
+		c := &Candidate{}
+		if err := c.Deserialize(candValue[i]); err != nil {
+			return errors.Wrapf(err, "failed to deserialize candidate bytes %v", candValue[i])
+		}
+		if err := p.inMemCandidates.Put(c); err != nil {
+			return err
+		}
+	}
+	return nil
```
CoderZhi: this is supposed to be in Start

action/protocol/staking/protocol.go
```diff
@@ -109,3 +121,31 @@ func (p *Protocol) Register(r *protocol.Registry) error {
...
+func (p *Protocol) CreateGenesisStates(ctx context.Context, sm protocol.StateManager) error {
+	if p.inMemCandidates.Size() > 0 {
+		return nil
+	}
+
+	// read all candidates from stateDB
+	stateDB := sm.GetDB()
+	candPkHash, candValue, err := stateDB.Filter(factory.CandidateNameSpace, func(k, v []byte) bool {
+		return bytes.HasPrefix(k, []byte{candidate})
+	})
```
CoderZhi: Let me think a way to not add back getdb

dustinxie: stateDB contains an underlying KVStore, i think it is OK to expose that for use

suggest we not spend time/effort on finding a way to not add back getdb

action/protocol/staking/protocol.go
dustinxie: yes, but Start() does not have state manager

action/protocol/staking/protocol.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/action/protocol"
 	"github.com/iotexproject/iotex-core/pkg/log"
+	"github.com/iotexproject/iotex-core/state/factory"
```
CoderZhi: we may need to move CandidateNamespace to this protocol as it is only used here.

PR #1978, opened 2020-03-04
fix startExistingBlockchain bug

koseoyoung:
In startExistingBlockchain, when we write to stateDB again, we need to fill context. 
Because we can not use tipInfo from blockchain, can not use bc.context()


blockchain/blockchain.go
```diff
@@ -544,6 +544,34 @@ func (bc *blockchain) startExistingBlockchain(ctx context.Context) error {
...
+			protocol.BlockchainCtx{
+				Registry: 		bc.registry, 
+				Genesis: 	 	bc.config.Genesis,
+				Tip:	  		protocol.TipInfo {
+					Height: i-1,
+				},
+			},
+		)
+		var candidates state.CandidateList
+		if pp := poll.FindProtocol(bc.registry); pp != nil {
```
CoderZhi: we don't need to do the checking or fetching multiple times in the for loop

blockchain/blockchain.go
```diff
@@ -544,6 +544,34 @@ func (bc *blockchain) startExistingBlockchain(ctx context.Context) error {
...
 			return err
 		}
+		candidateCtx := protocol.WithBlockchainCtx(
+			context.Background(),
+			protocol.BlockchainCtx{
+				Registry: 		bc.registry, 
+				Genesis: 	 	bc.config.Genesis,
+				Tip:	  		protocol.TipInfo {
+					Height: i-1,
+				},
```
CoderZhi: is this in use?

blockchain/blockchain.go
```diff
@@ -544,6 +544,34 @@ func (bc *blockchain) startExistingBlockchain(ctx context.Context) error {
...
+			ctx,
+			protocol.BlockchainCtx{
+				Registry: 		bc.registry, 
+				Genesis: 	 	bc.config.Genesis,
+				Tip:	  		protocol.TipInfo {
+					Height: i-1,
+				},
+				Candidates: candidates,
+			},
+		)
```
CoderZhi: duplicate code, can do better

blockchain/blockchain.go
```diff
@@ -452,7 +452,7 @@ func (bc *blockchain) CommitBlock(blk *block.Block) error {
 	defer bc.mu.Unlock()
 	timer := bc.timerFactory.NewTimer("CommitBlock")
 	defer timer.End()
-	ctx, err := bc.context(context.Background(), false, false)
+	ctx, err := bc.context(context.Background(), true, true)
```
CoderZhi: hmm, could you explain why we need this change?

Issue #1968, opened 2020-03-03
Fix test fail on windows


Frankonly:

lzxm160:
related to https://github.com/iotexproject/go-p2p/issues/13,will try to fix later

Frankonly:
> related to [iotexproject/go-p2p#13](https://github.com/iotexproject/go-p2p/issues/13)
> I will try to fix later

Ok. I am working on other cases who fail on windows.

Frankonly:
Only one fail case left now.
```
TestStartExistingBlockchain: local_test.go:595: 
        	Error Trace:	local_test.go:595
        	Error:      	Received unexpected error:
        	            	factory is higher than blockchain
        	            	github.com/iotexproject/iotex-core/blockchain.(*blockchain).startExistingBlockchain
        	            		C:/Users/Frank/go/src/github.com/iotexproject/iotex-core/blockchain/blockchain.go:575
        	            	github.com/iotexproject/iotex-core/blockchain.(*blockchain).Start
        	            		C:/Users/Frank/go/src/github.com/iotexproject/iotex-core/blockchain/blockchain.go:316
        	            	github.com/iotexproject/iotex-core/chainservice.(*ChainService).Start
        	            		C:/Users/Frank/go/src/github.com/iotexproject/iotex-core/chainservice/chainservice.go:362
        	            	github.com/iotexproject/iotex-core/server/itx.(*Server).Start
        	            		C:/Users/Frank/go/src/github.com/iotexproject/iotex-core/server/itx/server.go:97
        	            	github.com/iotexproject/iotex-core/e2etest.TestStartExistingBlockchain
        	            		C:/Users/Frank/go/src/github.com/iotexproject/iotex-core/e2etest/local_test.go:595
        	            	testing.tRunner
        	            		C:/Go/src/testing/testing.go:992
        	            	runtime.goexit
        	            		C:/Go/src/runtime/asm_amd64.s:1373
        	            	error when starting blockchain
        	            	github.com/iotexproject/iotex-core/chainservice.(*ChainService).Start
        	            		C:/Users/Frank/go/src/github.com/iotexproject/iotex-core/chainservice/chainservice.go:363
        	            	github.com/iotexproject/iotex-core/server/itx.(*Server).Start
        	            		C:/Users/Frank/go/src/github.com/iotexproject/iotex-core/server/itx/server.go:97
        	            	github.com/iotexproject/iotex-core/e2etest.TestStartExistingBlockchain
[...]

PR #1988, opened 2020-03-05
Fix test fail on windows

lzxm160:
work on #1968


e2etest/local_test.go
```diff
@@ -78,7 +80,12 @@ func TestLocalCommit(t *testing.T) {
 	// create client
 	cfg, err = newTestConfig()
 	require.NoError(err)
-	cfg.Network.BootstrapNodes = []string{svr.P2PAgent().Self()[0].String()}
+	for _, addr := range svr.P2PAgent().Self() {
```
dustinxie: make a func ValidNetworkAddr([]string) string

cfg.Network. BootstrapNodes = []string{ValidNetworkAddr(xx.Self())}

lzxm160: sure

Issue #1993, opened 2020-03-06
sanity check amount in stake/register message


dustinxie:
1. for NewDepositToStake(), change input amount to string, same as NewCreateStake()
2. for DepositToStake, CreateStake, CandidateRegister, need to check the amount is not 0 or negative
3. add test to cover 0/negative amount case

PR #1994, opened 2020-03-06
sanity check amount in stake/register message 1993

lzxm160:
work on #1993

### DIFF
--- a/action/protocol/staking/validations_test.go
+++ b/action/protocol/staking/validations_test.go
@@ -118,16 +118,6 @@ func TestProtocol_ValidateCreateStake(t *testing.T) {
 		},
 		{
 			candidateName,
-			"-1000000000000000000",
-			1,
-			false,
-			big.NewInt(unit.Qev),
-			10000,
-			1,
-			ErrInvalidAmount,
-		},
-		{
-			candidateName,
 			"1000000000000000000",
 			1,
 			false,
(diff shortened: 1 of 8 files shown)


action/stakecreate_test.go
```diff
@@ -28,6 +28,8 @@ var (
 	autoStake  = true
 	index      = uint64(10)
 	senderKey  = identityset.PrivateKey(27)
+	zero       = "0"
+	negtive    = "-10"
```
CoderZhi: again: it is a bad practice to have some global parameters with such common names for unit test purpose.

dustinxie: suggest to change this to a table-based test

var testActionParams = []struct{
  gaslimit   uint64
  gasprice *big.Int
  ...
  expected error
} {
  // valid test
  {
     ...
     expected: nil,
  }
  // invalid test
  {
     ...
     expected: ErrInvalidAmount,
  }
}

also with this you can do pass/fail test at same time, no need to add TestNotPositive()

lzxm160: done

action/protocol/staking/validations_test.go
```diff
@@ -84,7 +84,8 @@ func TestProtocol_ValidateCreateStake(t *testing.T) {
 		gasLimit  uint64
 		nonce     uint64
 		// expected results
-		errorCause error
+		newError      error
```
dustinxie: revert change in this file

this is to test validateCreateStake(), NewCreateStake() should return nil. In other words, we are not testing NewCreateStake() here

dustinxie: NewCreateStake() is tested in stakecreate_test.go

PR #1995, opened 2020-03-06
add CandidateCenter.Upsert()

dustinxie:


action/protocol/staking/candidate_center.go
```diff
@@ -81,8 +81,14 @@ func (m CandidateCenter) GetBySelfStakingIndex(index uint64) *Candidate {
...
 
-// Put writes the candidate into map
-func (m CandidateCenter) Put(d *Candidate) {
+// Upsert adds a candidate into map, overwrites if already exist
+func (m CandidateCenter) Upsert(d *Candidate) {
+	if c, ok := m.ownerMap[d.Owner.String()]; ok {
+		delete(m.nameMap, c.Name)
+		delete(m.operatorMap, c.Operator.String())
+		delete(m.selfStkBucketMap, c.SelfStakeBucketIdx)
+	}
```
CoderZhi: As nameMap/operatorMap/selfStkBucketMap are all indexers of d inside of CandidateCenter, if an upsert of d is intended, to maintain the consistency this struct, we are supposed to make sure (take name map for example):
m.nameMap[c.Name] is either empty or identical to d, if not, throw an error.

This sounds like a duplicate logic, but it may be better to keep the logic inside of this struct instead.
```
	if c, ok := m.ownerMap[d.Owner.String()]; ok {
		if nd, ok := m.nameMap[c.Name]; ok && nd.Owner.String() != d.Owner.String() {
                        return errors.New("duplicate name")
                 }
                 ... ...
	}

Comment (yutongp) [reply action/protocol/staking/candidate_center.go]: I think m.nameMap[c.Name]  should always link to c, since only CandidateCenter itself can insert to all maps it own

Comment (dustinxie) [reply action/protocol/staking/candidate_center.go]: updated in latest push

File: action/protocol/staking/protocol.go [hunk 2]
```diff
@@ -76,12 +76,18 @@ func (p *Protocol) Start(ctx context.Context) error {
...
 		if err := iter.Next(c); err != nil {
 			return errors.Wrapf(err, "failed to deserialize candidate")
 		}
-		p.inMemCandidates.Put(c)
+
+		if c.SelfStake.Cmp(thres) >= 0 {
+			if err := p.inMemCandidates.Upsert(c); err != nil {
+				return err
+			}
+		}
```
CoderZhi: I thought no matter what, we will keep the candidate in memory to maintain unique name/operator address and etc.

dustinxie: oh yeah, my bad
voter can still stake to inactive candidate (like being slashed)

PR #1996, opened 2020-03-06
fix test cleanup fail on Windows

Frankonly:
`os.TempFile()` creates and occupies new file on Windows. Many test cleanups fails on Windows due to those occupations. Thus, I close all those files after created in this PR.

### DIFF
--- a/action/protocol/execution/evm/contract_test.go
+++ b/action/protocol/execution/evm/contract_test.go
@@ -14,9 +14,10 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/golang/mock/gomock"
-	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
+
+	"github.com/iotexproject/go-pkgs/hash"
 
 	"github.com/iotexproject/iotex-core/action/protocol"
 	accountutil "github.com/iotexproject/iotex-core/action/protocol/account/util"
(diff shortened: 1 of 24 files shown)


Frankonly:
#1968

action/protocol/execution/evm/contract_test.go
```diff
@@ -32,8 +33,10 @@ func TestCreateContract(t *testing.T) {
 	require := require.New(t)
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
-	testTrieFile, _ := ioutil.TempFile(os.TempDir(), "trie")
+	testTrieFile, err := ioutil.TempFile(os.TempDir(), "trie")
+	require.NoError(err)
 	testTriePath := testTrieFile.Name()
+	require.NoError(testTrieFile.Close())
```
CoderZhi: so many duplicate code that it worths the effort to add a function into testutil
```
func PathOfTempFile(dirname string) (string, error) {
    tempFile, err := ioutil.TempFile(os.TempDir(), dirname)
    if err != nil {
        return "", err
    }
    return tempFile.Name(), tempFile.Close()
}

File: action/protocol/execution/protocol_test.go [hunk 2]
```diff
@@ -457,12 +457,18 @@ func TestProtocol_Handle(t *testing.T) {
 			delete(cfg.Plugins, config.GatewayPlugin)
 		}()
 
-		testTrieFile, _ := ioutil.TempFile(os.TempDir(), "trie")
+		testTrieFile, err := ioutil.TempFile(os.TempDir(), "trie")
+		require.NoError(err)
 		testTriePath := testTrieFile.Name()
-		testDBFile, _ := ioutil.TempFile(os.TempDir(), "db")
+		require.NoError(testTrieFile.Close())
```
dustinxie: the file is closed here, but 
sf, err := factory.NewStateDB(cfg, factory.DefaultStateDBOption())
then it will be opened again in
require.NoError(bc.Start(ctx))

action/protocol/execution/evm/contract_test.go
Frankonly: ok

PR #1999, opened 2020-03-07
Add a unit test for create_stake handler

lizhefeng:


action/protocol/staking/handlers_test.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	account, err := accountutil.LoadOrCreateAccount(sm, addr.String())
+	if err != nil {
+		return err
+	}
+	account.Balance = unit.ConvertIotxToRau(balance)
+	return accountutil.StoreAccount(sm, addr.String(), account)
+}
+
+func setupCandidate(
```
dustinxie: change to setup(p, sm, *Candidate)

and use one of testCandidates (already defined)

PR #2000, opened 2020-03-07
add FilterOption to state reader

dustinxie:
this is needed to read buckets from stateDB


state/factory/factory.go
```diff
@@ -497,22 +497,22 @@ func (sf *factory) Commit(ctx context.Context, blk *block.Block) error {
 func (sf *factory) State(s interface{}, opts ...protocol.StateOption) (uint64, error) {
 	sf.mutex.RLock()
 	defer sf.mutex.RUnlock()
-	archive, height, ns, key, err := processOptions(opts...)
+	cfg, err := processOptions(opts...)
```
dustinxie: with filter, now it needs to return 5 items, so change to return cfg

CoderZhi:
can we improve test coverage in case of any error

PR #2005, opened 2020-03-08
move two layer trie and trie kvstore implementation into trie package

CoderZhi:
Move TwoLayerTrie and KVStoreForTrie into trie package
Fix the bug of mixture usage of db.ErrNotExist and trie.ErrNotExist

### DIFF
--- a/state/factory/factory.go
+++ b/state/factory/factory.go
@@ -94,8 +94,8 @@ type (
 		cfg                config.Config
 		currentChainHeight uint64
 		saveHistory        bool
-		twoLayerTrie       *TwoLayerTrie // global state trie, this is a read only trie
-		dao                db.KVStore    // the underlying DB for account/contract storage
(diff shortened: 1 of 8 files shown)


state/factory/factory.go
```diff
@@ -581,7 +574,7 @@ func (sf *factory) stateAtHeight(height uint64, ns string, key []byte, s interfa
 	if !sf.saveHistory {
 		return ErrNoArchiveData
 	}
-	tlt, err := newTwoLayerTrie(fmt.Sprintf("%s-%d", ArchiveTrieRootKey, height), sf.dao, false)
+	tlt, err := newTwoLayerTrie(ArchiveTrieNamespace, sf.dao, fmt.Sprintf("%s-%d", ArchiveTrieRootKey, height), false)
```
dustinxie: nit:
feel it is more natural to use []byte
rootKey = append([]byte(ArchiveTrieRootKey), uint64tobytes(height))

PR #2015, opened 2020-03-10
[staking] add bootstrap candidates and use genesis config

yutongp:


action/protocol/staking/protocol_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/iotexproject/iotex-core/blockchain/genesis"
```
lizhefeng: Group imports

PR #2023, opened 2020-03-11
Add logs in staking action receipt if necessary

lizhefeng:


action/protocol/staking/handlers.go
```diff
@@ -62,7 +64,15 @@ func (p *Protocol) handleCreateStake(ctx context.Context, act *action.CreateStak
 		return nil, errors.Wrapf(err, "failed to store account %s", actionCtx.Caller.String())
 	}
 
-	receipt, err := p.settleAction(ctx, sm, gasFee)
+	log := &action.Log{
+		Address:     p.addr.String(),
+		Topics:      nil,
```
dustinxie: i am thinking we should also define diff topics for each type of staking/register action?
then from the log we can know what action it is
and data field may take diff meaning for diff action

action/protocol/staking/handlers.go
```diff
@@ -490,6 +509,7 @@ func (p *Protocol) settleAction(
 		ActionHash:      actionCtx.ActionHash,
 		GasConsumed:     actionCtx.IntrinsicGas,
 		ContractAddress: p.addr.String(),
+		Logs:            logs,
```
dustinxie: []*Log{logs}?

action/protocol/staking/handlers.go
lizhefeng: We can add when we need to. So far I think we can leave it blank.

dustinxie: and add logs for other actions?

like we discussed, a valid candidate try to update with another's existing name, this should end up with a receipt of log with status = failure, but the block can still be created

action/protocol/staking/handlers.go
```diff
@@ -62,7 +64,15 @@ func (p *Protocol) handleCreateStake(ctx context.Context, act *action.CreateStak
 		return nil, errors.Wrapf(err, "failed to store account %s", actionCtx.Caller.String())
 	}
 
-	receipt, err := p.settleAction(ctx, sm, gasFee)
+	log := &action.Log{
+		Address:     p.addr.String(),
+		Topics:      nil,
+		Data:        byteutil.Uint64ToBytes(bucketIdx),
```
CoderZhi: In rewarding protocol, data is struct with type, address, and amount. To be consistent with rewarding protocol, we may need to do the same thing in the staking protocol. However, we won't be able to do indexing or bloomfilter with the complex data structure. thoughts?

PR #2024, opened 2020-03-12
Add failure status in staking receipts

lizhefeng:


action/protocol/staking/handlers.go
```diff
@@ -31,13 +33,15 @@ func (p *Protocol) handleCreateStake(ctx context.Context, act *action.CreateStak
...
+		log.L().Debug("Error when fetching caller", zap.Error(err))
+		return p.settleAction(ctx, sm, uint64(iotextypes.ReceiptStatus_Failure), gasFee)
 	}
 
 	// Create new bucket and bucket index
 	candidate := p.inMemCandidates.GetByName(act.Candidate())
 	if candidate == nil {
-		return nil, errors.Wrap(ErrInvalidCanName, "cannot find candidate in candidate center")
+		log.L().Debug("Error when finding candidate in candidate center", zap.Error(ErrInvalidCanName))
+		return p.settleAction(ctx, sm, uint64(iotextypes.ReceiptStatus_Failure), gasFee)
```
dustinxie: add a log in this kind of receipt after PR2023

action/protocol/staking/handlers.go
```diff
@@ -186,25 +203,28 @@ func (p *Protocol) handleWithdrawStake(ctx context.Context, act *action.Withdraw
...
-	bucket, err := p.fetchBucket(ctx, sm, act.BucketIndex(), true, false)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to fetch bucket")
+	bucket, fetchErr := p.fetchBucket(ctx, sm, act.BucketIndex(), true, false)
+	if fetchErr != nil {
+		log.L().Debug("Error when fetching bucket", zap.Error(fetchErr.err))
+		return p.settleAction(ctx, sm, uint64(fetchErr.failureStatus), gasFee)
 	}
 
 	prevCandidate := p.inMemCandidates.GetByOwner(bucket.Candidate)
```
dustinxie: we should also settle this error?

action/protocol/staking/handlers.go
```diff
@@ -285,23 +307,27 @@ func (p *Protocol) handleTransferStake(ctx context.Context, act *action.Transfer
...
+		return p.settleAction(ctx, sm, uint64(fetchErr.failureStatus), gasFee)
 	}
 	if !bucket.AutoStake {
-		return nil, errors.New("deposit is only allowed on auto-stake bucket")
+		err := errors.New("deposit is only allowed on auto-stake bucket")
+		log.L().Debug("Error when depositing to stake", zap.Error(err))
+		return p.settleAction(ctx, sm, uint64(iotextypes.ReceiptStatus_ErrInvalidBucketType), gasFee)
 	}
 	candidate := p.inMemCandidates.GetByOwner(bucket.Candidate)
 	if candidate == nil {
```
dustinxie: should also settle this error?

action/protocol/staking/handlers.go
```diff
@@ -355,14 +381,16 @@ func (p *Protocol) handleDepositToStake(ctx context.Context, act *action.Deposit
...
-	bucket, err := p.fetchBucket(ctx, sm, act.BucketIndex(), true, true)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to fetch bucket")
+	bucket, fetchErr := p.fetchBucket(ctx, sm, act.BucketIndex(), true, true)
+	if fetchErr != nil {
+		log.L().Debug("Error when fetching bucket", zap.Error(fetchErr.err))
+		return p.settleAction(ctx, sm, uint64(fetchErr.failureStatus), gasFee)
 	}
 
 	candidate := p.inMemCandidates.GetByOwner(bucket.Candidate)
```
dustinxie: should also settle this error?

action/protocol/staking/handlers.go
```diff
@@ -551,18 +583,30 @@ func (p *Protocol) fetchBucket(
 	index uint64,
 	checkOwner bool,
 	allowSelfStaking bool,
-) (*VoteBucket, error) {
+) (*VoteBucket, *fetchError) {
 	actionCtx := protocol.MustGetActionCtx(ctx)
 	bucket, err := getBucket(sr, index)
 	if err != nil {
```
dustinxie: i think need to further differentiate the error

if err.Cause == state.ErrNotExist, then we should return ReceiptStatus_ErrInvalidBucketIndex, and settle the action as Failed, that is charge the gas, and Mint/CommitBlock() can proceed

otherwise, the error would be a real error (like DB failed), in that case, Mint/CommitBlock should fail/stop

and at the return of fetchBucket(), need to tell apart these 2 cases, and process accordingly

action/protocol/staking/handlers.go
```diff
@@ -551,18 +583,30 @@ func (p *Protocol) fetchBucket(
...
+		}
+		return nil, fetchErr
 	}
 	if checkOwner && !address.Equal(bucket.Owner, actionCtx.Caller) {
-		return nil, fmt.Errorf("bucket owner does not match action caller, bucket owner %s, action caller %s",
-			bucket.Owner.String(), actionCtx.Caller.String())
+		fetchErr := &fetchError{
+			err: fmt.Errorf("bucket owner does not match action caller, bucket owner %s, action caller %s",
+				bucket.Owner.String(), actionCtx.Caller.String()),
+			failureStatus: iotextypes.ReceiptStatus_ErrUnauthorizedOperator,
```
dustinxie: this error falls in first type, that is we charge gas, and let Mint/Commit() continue

action/protocol/staking/handlers.go
```diff
@@ -551,18 +583,30 @@ func (p *Protocol) fetchBucket(
...
+		}
+		return nil, fetchErr
 	}
 	if !allowSelfStaking && p.inMemCandidates.ContainsSelfStakingBucket(index) {
-		return nil, errors.New("self staking bucket cannot be processed")
+		fetchErr := &fetchError{
+			err:           errors.New("self staking bucket cannot be processed"),
+			failureStatus: iotextypes.ReceiptStatus_ErrInvalidBucketType,
+		}
```
dustinxie: this is also first type error

action/protocol/staking/handlers.go
lizhefeng: I didn't settle this error in receipt because the candidate is already associated with a bucket. If the candidate center cannot find this candidate, it is not the action caller's fault, but something wrong with the system.

action/protocol/staking/handlers.go
lizhefeng: Ditto.

action/protocol/staking/handlers.go
lizhefeng: Ditto.

action/protocol/staking/handlers.go
lizhefeng: agreed.

action/protocol/staking/handlers.go
lzxm160: can we return error with errors.Wrap(Errxxx,"xxxx"),so I can use errors.Cause(err) to check when write unit test

action/protocol/staking/handlers.go
```diff
@@ -551,18 +628,32 @@ func (p *Protocol) fetchBucket(
 	index uint64,
 	checkOwner bool,
 	allowSelfStaking bool,
-) (*VoteBucket, error) {
+) (*VoteBucket, *fetchError) {
 	actionCtx := protocol.MustGetActionCtx(ctx)
 	bucket, err := getBucket(sr, index)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to fetch bucket by index %d", index)
+		fetchErr := &fetchError{
```
dustinxie: failureStatus: iotextypes.ReceiptStatus_Failure

action/protocol/staking/handlers.go
```diff
@@ -46,19 +48,29 @@ const (
...
 func (p *Protocol) handleCreateStake(ctx context.Context, act *action.CreateStake, sm protocol.StateManager) (*action.Receipt, error) {
 	actionCtx := protocol.MustGetActionCtx(ctx)
 	blkCtx := protocol.MustGetBlockCtx(ctx)
 
-	staker, gasFee, err := fetchCaller(ctx, sm, act.Amount())
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to fetch caller")
+	staker, gasFee, fetchErr := fetchCaller(ctx, sm, act.Amount())
+	if fetchErr != nil {
+		if fetchErr.failureStatus == 0 {
```
dustinxie: == iotextypes.ReceiptStatus_Failure

action/protocol/staking/handlers.go
```diff
@@ -551,18 +628,32 @@ func (p *Protocol) fetchBucket(
...
-) (*VoteBucket, error) {
+) (*VoteBucket, *fetchError) {
 	actionCtx := protocol.MustGetActionCtx(ctx)
 	bucket, err := getBucket(sr, index)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to fetch bucket by index %d", index)
+		fetchErr := &fetchError{
+			err: errors.Wrapf(err, "failed to fetch bucket by index %d", index),
+		}
+		if errors.Cause(err) != state.ErrStateNotExist {
```
dustinxie: should be ==

action/protocol/staking/handlers.go
```diff
@@ -608,23 +699,29 @@ func putBucketAndIndex(sm protocol.StateManager, bucket *VoteBucket) (uint64, er
...
 }
 
-func fetchCaller(ctx context.Context, sm protocol.StateReader, amount *big.Int) (*state.Account, *big.Int, error) {
+func fetchCaller(ctx context.Context, sm protocol.StateReader, amount *big.Int) (*state.Account, *big.Int, *fetchError) {
 	actionCtx := protocol.MustGetActionCtx(ctx)
 
 	caller, err := accountutil.LoadAccount(sm, hash.BytesToHash160(actionCtx.Caller.Bytes()))
 	if err != nil {
-		return nil, nil, errors.Wrapf(err, "failed to load the account of caller %s", actionCtx.Caller.String())
+		return nil, nil, &fetchError{
```
dustinxie: failureStatus: iotextypes.ReceiptStatus_Failure

action/protocol/staking/handlers.go
lizhefeng: @lzxm160 This would be a nonerr case with a failure receipt. You may need to check the receipt status instead of error.

PR #2027, opened 2020-03-12
use rau string as token unit and format in config

yutongp:
use rau string as token unit and format in config to be consistent with others

### DIFF
--- a/blockchain/genesis/genesis.go
+++ b/blockchain/genesis/genesis.go
@@ -84,11 +84,11 @@ func defaultConfig() Genesis {
 				SelfStake:  1.05,
 			},
 			RegistrationConsts: RegistrationConsts{
-				Fee:          100,
-				MinSelfStake: 1200000,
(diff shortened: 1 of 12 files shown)


blockchain/genesis/genesis.go
```diff
@@ -84,11 +84,11 @@ func defaultConfig() Genesis {
...
 			RegistrationConsts: RegistrationConsts{
-				Fee:          100,
-				MinSelfStake: 1200000,
+				Fee:          unit.ConvertIotxToRau(100).String(),
+				MinSelfStake: unit.ConvertIotxToRau(1200000).String(),
 			},
-			WithdrawWaitingPeriod: 14,
-			MinStakeAmount:        100,
+			WithdrawWaitingPeriod: 14 * 24 * time.Hour,
+			MinStakeAmount:        unit.ConvertIotxToRau(100).String(),
```
CoderZhi: no impact on genesis hash, right?

PR #2016, opened 2020-03-10
add candidateIndexer to store candidate/kickoutlist for API call

koseoyoung:
Because archive mode takes too much time, add indexer to store candidate/blacklist for API call (ReadState and GetEpochMeta) 

1. implement candidateIndexer and if exists, read from indexer at ReadState() in poll.Protocol 
2. Refactor ```productivityByEpoch``` to distinguish API purpose and protocol purpose
(DelegatesByEpoch is only for protocol purpose -- related to tip) 

I will refactor ```productivityByEpoch``` in next PR to only keep current epoch data in state factory and move into poll protocol

Issue #1943, opened 2020-02-27
Add items in state db


CoderZhi:
Last N blocks {height, producer, hash, mintTime}

CoderZhi:
We have a query "ProductivityByEpoch" in blockchain.go, which is used in governance_protocol.go, should be included.

koseoyoung:
@CoderZhi do we need to store hash too?

PR #2029, opened 2020-03-12
introduce HistoryStateReader and refactor poll protocol 

koseoyoung:
Based on #2016 ,

1. introduce HistoryStateReader 
2. refactor ~ByHeight/ ~ByEpoch func in poll protocol
3. add "Height" field in ReadStateRequest in api.go 

I have to merge after #2016 because it has some dependency.

### DIFF
--- a/action/protocol/poll/protocol.go
+++ b/action/protocol/poll/protocol.go
@@ -63,8 +63,10 @@ type ProductivityByEpoch func(context.Context, uint64) (uint64, map[string]uint6
 type Protocol interface {
 	protocol.Protocol
 	protocol.GenesisStateCreator
-	DelegatesByEpoch(context.Context, protocol.StateReader, uint64) (state.CandidateList, error)
-	CandidatesByHeight(context.Context, protocol.StateReader, uint64) (state.CandidateList, error)
+	Delegates(context.Context, protocol.StateReader) (state.CandidateList, error)
+	NextDelegates(context.Context, protocol.StateReader) (state.CandidateList, error)
+	Candidates(context.Context, protocol.StateReader) (state.CandidateList, error)
+	NextCandidates(context.Context, protocol.StateReader) (state.CandidateList, error)
 	// CalculateCandidatesByHeight calculates candidate and returns candidates by chain height
 	CalculateCandidatesByHeight(context.Context, uint64) (state.CandidateList, error)
 }
--- a/action/protocol/poll/protocol.go
+++ b/action/protocol/poll/protocol.go
@@ -45,10 +45,10 @@ var ErrDelegatesNotExist = errors.New("delegates cannot be found")
 type CandidatesByHeight func(protocol.StateReader, uint64) ([]*state.Candidate, error)
 
 // GetCandidates returns the current candidates
-type GetCandidates func(protocol.StateReader, bool, ...protocol.StateOption) ([]*state.Candidate, uint64, error)
+type GetCandidates func(protocol.StateReader, bool) ([]*state.Candidate, uint64, error)
 
 // GetKickoutList returns current the blacklist
-type GetKickoutList func(protocol.StateReader, bool, ...protocol.StateOption) (*vote.Blacklist, uint64, error)
+type GetKickoutList func(protocol.StateReader, bool) (*vote.Blacklist, uint64, error)
 
 // GetUnproductiveDelegate returns unproductiveDelegate struct which contains a cache of upd info by epochs
 type GetUnproductiveDelegate func(protocol.StateReader) (*vote.UnproductiveDelegate, error)
@@ -63,8 +63,8 @@ type ProductivityByEpoch func(context.Context, uint64) (uint64, map[string]uint6
 type Protocol interface {
 	protocol.Protocol
 	protocol.GenesisStateCreator
-	DelegatesByEpoch(context.Context, uint64) (state.CandidateList, error)
-	CandidatesByHeight(context.Context, uint64) (state.CandidateList, error)
+	DelegatesByEpoch(context.Context, protocol.StateReader, uint64) (state.CandidateList, error)
+	CandidatesByHeight(context.Context, protocol.StateReader, uint64) (state.CandidateList, error)
 	// CalculateCandidatesByHeight calculates candidate and returns candidates by chain height
 	CalculateCandidatesByHeight(context.Context, uint64) (state.CandidateList, error)
 }
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -19,8 +19,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	"github.com/iotexproject/iotex-core/action"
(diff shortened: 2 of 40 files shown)


action/protocol/poll/protocol.go
```diff
@@ -62,8 +62,8 @@ type ProductivityByEpoch func(context.Context, uint64) (uint64, map[string]uint6
 type Protocol interface {
 	protocol.Protocol
 	protocol.GenesisStateCreator
-	DelegatesByEpoch(context.Context, uint64) (state.CandidateList, error)
-	CandidatesByHeight(context.Context, uint64) (state.CandidateList, error)
+	DelegatesByEpoch(context.Context, protocol.StateReader, uint64) (state.CandidateList, error)
```
CoderZhi: it is possible to kill the last parameter "epochNum"?

action/protocol/poll/protocol.go
```diff
@@ -62,8 +62,8 @@ type ProductivityByEpoch func(context.Context, uint64) (uint64, map[string]uint6
...
+	CandidatesByHeight(context.Context, protocol.StateReader, uint64) (state.CandidateList, error)
```
CoderZhi: same question here.

blockchain/blockchain.go
```diff
@@ -112,7 +112,7 @@ type (
 )
 
 // ProductivityByEpoch returns the map of the number of blocks produced per delegate in an epoch
-func ProductivityByEpoch(ctx context.Context, bc Blockchain, epochNum uint64) (uint64, map[string]uint64, error) {
+func ProductivityByEpoch(ctx context.Context, bc Blockchain, sr protocol.StateReader, epochNum uint64) (uint64, map[string]uint64, error) {
```
CoderZhi: to my understanding, we shouldn't modify this function

state/factory/factory.go
```diff
@@ -150,6 +158,33 @@ func newTwoLayerTrie(ns string, dao db.KVStore, rootKey string, create bool) (*t
...
+
+// States returns history states in the archive mode state factory 
+func (hReader *historyStateReader) States(opts ...protocol.StateOption) (uint64, state.Iterator, error) {
+	iterator, err := hReader.sf.StatesAtHeight(hReader.height, opts...)
+	if err != nil {
+		return 0, nil, err
+	}
+	return hReader.height, iterator, nil
+}
+
```
CoderZhi: move all related definition and implementation into a new file

action/protocol/poll/protocol.go
koseoyoung: yes it is possible, but we have to compare epochnumber with tipEpochnumber not inside of  DelegatesByEpoch but outside of it which happens in consensus/productivityByEpoch/ blockchain.context(). Is it ok?

blockchain/blockchain.go
koseoyoung: In the future after merge with #2016, this function will be only used for tip info. So I pass sr to be sf at the chainservice. Could you explain why I shouldn't modify this func?

CoderZhi: a generic productivityByEpoch API should only have either sr or epochNum as parameter, but not both. If this is an API provided by blockchain, we may need to skip sr as parameter. Correspondingly we don't need to know who are the delegates by epoch, but just return a static of map[string]int, where string is the producer address in string, int is the number of blocks the producer produced in the epoch. For the delegates who produced 0 blocks, they won't be in the return map.

blockchain/blockchain.go
```diff
@@ -129,7 +129,7 @@ func ProductivityByEpoch(ctx context.Context, bc Blockchain, epochNum uint64) (u
 	}
 	numBlks := epochEndHeight - epochStartHeight + 1
 
-	activeConsensusBlockProducers, err := poll.MustGetProtocol(bcCtx.Registry).DelegatesByEpoch(ctx, epochNum)
+	activeConsensusBlockProducers, err := poll.MustGetProtocol(bcCtx.Registry).DelegatesByEpoch(ctx, sr, epochNum)
```
CoderZhi: thus we don't need this line.

action/protocol/poll/protocol.go
CoderZhi: introduce a function "Delegates" to distinguish the historical read and tip read. If necessary, you can have one more function NextDelegates

koseoyoung: ok! I changed to have Delegates() and NextDelegates() function. Btw, we don't have to distinguish historical and tip read, because now DelegatesByEpoch is only for tip

blockchain/blockchain.go
koseoyoung: Oh I got it! but if we do that, we have to call Delegates() func in reward protocol and poll protocol to calculate unproductive delegate. For reward protocol, is it ok?

CoderZhi: For reward protocol, we may not need to call Delegates. You can try first.

koseoyoung: I'm trying right now, we do have candidate info from context, but not delegate info. before easter, we need to know unqualified delegate.

koseoyoung: I just tried to do that, but we need to Delegates() in reward protocol, because we need to calculate unproductive list before easter. they do have candidate from context but not delegate list.

blockchain/blockchain.go
```diff
@@ -111,37 +109,30 @@ type (
 	}
 )
 
-// ProductivityByEpoch returns the map of the number of blocks produced per delegate in an epoch
-// TODO: move to poll protocol and implement reading current epoch meta from state factory -- now only reading current epoch productivity
-func ProductivityByEpoch(ctx context.Context, bc Blockchain, sr protocol.StateReader, epochNum uint64) (uint64, map[string]uint64, error) {
+// CurrentEpochProductivity returns the map of the number of blocks produced per delegate in current epoch
+// TODO: implement reading current epoch meta from state factory
+func CurrentEpochProductivity(ctx context.Context, bc Blockchain, sr protocol.StateReader, epochNum uint64) (uint64, map[string]uint64, error) {
```
CoderZhi: we shouldn't introduce "sr" in this function.
blockchain.ProductivityByEpoch is something relying on the data stored in blockchain db. If we have all blocks, we should be able to return the productivity for any epoch.

koseoyoung: According to  #1943 , we are going to read productivity info(producers) from state DB, is it not true?

state/factory/factory.go
```diff
@@ -420,6 +422,7 @@ func (sf *factory) SimulateExecution(
 	caller address.Address,
 	ex *action.Execution,
 	getBlockHash evm.GetBlockHash,
+	depositGas evm.DepositGas,
```
CoderZhi: delete

state/factory/factory.go
```diff
@@ -428,7 +431,7 @@ func (sf *factory) SimulateExecution(
 		return nil, nil, errors.Wrap(err, "failed to obtain working set from state factory")
 	}
 
-	return simulateExecution(ctx, ws, caller, ex, getBlockHash)
+	return simulateExecution(ctx, ws, caller, ex, getBlockHash, depositGas)
```
CoderZhi: because it is a simulation, we can put a dummy depositGas function here.

koseoyoung: yes! address in latest commit

PR #2034, opened 2020-03-13
check condition to enable native staking V2 or not

dustinxie:


action/protocol/poll/protocol.go
```diff
@@ -122,7 +122,9 @@ func NewProtocol(
 	if cfg.Consensus.Scheme != config.RollDPoSScheme {
 		return nil, nil
 	}
-	if !genesisConfig.EnableGravityChainVoting || electionCommittee == nil || genesisConfig.GravityChainStartHeight == 0 {
+
+	if !genesisConfig.EnableGravityChainVoting || genesisConfig.GravityChainStartHeight == 0 ||
```
yutongp: remove `genesisConfig.GravityChainStartHeight == 0`

dustinxie: so why remove it? it's in current code

PR #2050, opened 2020-03-19
move kickout code to slasher

dustinxie:
think of poll protocol as 2 parts:
1. governance (with staking V1 after Cook) or staking V2 (after Fairbank) gives out candidates list
2. slasher gives out kick-out list (after Easter)

staking_committee (and new staking_command) can then use slasher

### DIFF
--- a/action/protocol/poll/governance_protocol.go
+++ b/action/protocol/poll/governance_protocol.go
@@ -20,7 +20,6 @@ import (
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/action/protocol"
 	"github.com/iotexproject/iotex-core/action/protocol/rolldpos"
-	"github.com/iotexproject/iotex-core/action/protocol/vote"
 	"github.com/iotexproject/iotex-core/config"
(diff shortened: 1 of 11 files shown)


action/protocol/poll/governance_protocol.go
```diff
@@ -81,24 +60,16 @@ func NewGovernanceChainCommitteeProtocol(
...
 		initGravityChainHeight:    initGravityChainHeight,
 		getBlockTime:              getBlockTime,
-		numCandidateDelegates:     numCandidateDelegates,
-		numDelegates:              numDelegates,
 		addr:                      addr,
 		initialCandidatesInterval: initialCandidatesInterval,
-		productivityByEpoch:       productivityByEpoch,
-		productivityThreshold:     productivityThreshold,
-		kickoutEpochPeriod:        kickoutEpochPeriod,
 		kickoutIntensity:          kickoutIntensity,
```
koseoyoung: can we entirely remove kickoutIntensity field here by also moving the part related with kick-out in CreateGenesisStates()?

koseoyoung:
Also for naming, Slasher is broader meaning for now, but we can include slash-the-stake at this module in future

PR #2051, opened 2020-03-19
consortium committee

yutongp:


action/protocol/poll/consortium.go
```diff
@@ -0,0 +1,219 @@
...
+	return cc.readDelegates(ctx)
+}
+
+func (cc *consortiumCommittee) CandidatesByHeight(ctx context.Context, height uint64) (state.CandidateList, error) {
+	return cc.readDelegates(ctx)
+}
+
+func (cc *consortiumCommittee) readDelegates(ctx context.Context) (state.CandidateList, error) {
+	bcCtx := protocol.MustGetBlockchainCtx(ctx)
+	if cc.bufferHeight == bcCtx.Tip.Height && cc.bufferResult != nil {
```
CoderZhi: not sure whether we should distinguish by height or epoch.

PR #2053, opened 2020-03-20
move util function to workingSet

CoderZhi:
These util functions are part of the working set implementation. Put them back to the right place.

### DIFF
--- a/state/factory/statedb.go
+++ b/state/factory/statedb.go
@@ -229,7 +229,7 @@ func (sdb *stateDB) Validate(ctx context.Context, blk *block.Block) error {
 	if isExist {
 		return nil
 	}
-	if err = validateWithWorkingset(ctx, ws, blk); err != nil {
+	if err = ws.ValidateBlock(ctx, blk); err != nil {
 		return errors.Wrap(err, "failed to validate block with workingset in statedb")
 	}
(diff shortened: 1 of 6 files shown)


state/factory/statedb.go
```diff
@@ -302,7 +302,7 @@ func (sdb *stateDB) Commit(ctx context.Context, blk *block.Block) error {
 		return err
 	}
 	if !isExist {
-		_, ws, err = runActions(ctx, ws, blk.RunnableActions().Actions())
+		_, err = ws.Process(ctx, blk.RunnableActions().Actions())
```
dustinxie: can we still use the name runActions or RunActions?

state/factory/workingset.go
```diff
@@ -227,3 +223,248 @@ func (ws *workingSet) DelState(opts ...protocol.StateOption) (uint64, error) {
...
+					srcAddr,
+					confirmedState.Nonce,
+				)
+			}
+		}
+	}
+	return nil
+}
+
+func (ws *workingSet) Process(ctx context.Context, actions []action.SealedEnvelope) ([]*action.Receipt, error) {
```
dustinxie: why we need both Process() and process(), they are identical func?
i mean, there is no lock/unlock in Process()

Issue #2045, opened 2020-03-17
"ioctl action deposit" should display IOTX unit not Raul


raullenchai:
```
➜  ~ ioctl action deposit 200000 -s zebra
Enter password #io1uqhmnttmv0pg8prugxxn7d8ex9angrvfjfthxa:


version: 1  nonce: 103  gasLimit: 10000  gasPrice: 1000000000000 Rau
senderAddress: io1uqhmnttmv0pg8prugxxn7d8ex9angrvfjfthxa (rrrrrrr)
version: 1
nonce: 103
gasLimit: 10000
gasPrice: "1000000000000"
depositToRewardingFund: <
  amount: "200000000000000000000000"
>➜  ~ ioctl action deposit 200000 -s zebra
Enter password #io1uqhmnttmv0pg8prugxxn7d8ex9angrvfjfthxa:


version: 1  nonce: 103  gasLimit: 10000  gasPrice: 1000000000000 Rau
senderAddress: io1uqhmnttmv0pg8prugxxn7d8ex9angrvfjfthxa (rrrrrrr)
version: 1
nonce: 103
gasLimit: 10000
gasPrice: "1000000000000"
depositToRewardingFund: <
  amount: "200000000000000000000000"
>
```

Prefer to use IOTX as the unit, not Raul.

raullenchai:
Pls check for other cmds

PR #2054, opened 2020-03-20
ioctl action display IOTX unit instead of Raul

RuneAkko:
#2045


ioctl/cmd/action/actionhash.go
```diff
@@ -167,10 +167,12 @@ func printActionProto(action *iotextypes.Action) (string, error) {
 	if err != nil {
 		return "", output.NewError(output.ConvertError, "failed to convert bytes into address", err)
 	}
+	//ioctl action should display IOTX unit instead Raul
+	gasPriceUnitIOTX, err := util.StringToIOTX(action.Core.GasPrice)
```
dustinxie: check err
or use gasPriceUnitIOTX, _ := if we are sure action.Core.GasPrice is correct

PR #2056, opened 2020-03-20
get rid of state reader in governance protocol struct

koseoyoung:
1. fix a bug that we used wrong state reader 
2. get rid of the state reader inside of governance protocol struct


dustinxie:
calculateUnproductiveDelegatesByEpoch() can use sr StateReader instead of sm StateManager
can you also change that?

PR #2058, opened 2020-03-20
add poll mode and enable consortium poll

yutongp:


action/protocol/poll/protocol.go
```diff
@@ -123,48 +132,62 @@ func NewProtocol(
...
-	if !genesisConfig.EnableGravityChainVoting || electionCommittee == nil {
+
+	switch genesisConfig.PollMode {
+	case _modeLifeLong:
 		delegates := genesisConfig.Delegates
 		if uint64(len(delegates)) < genesisConfig.NumDelegates {
 			return nil, errors.New("invalid delegate address in genesis block")
 		}
 		return NewLifeLongDelegatesProtocol(delegates), nil
+	case _modeGovernance:
```
dustinxie: fallthrough means governance only won't be used
we can remove this mode?

blockchain/genesis/genesis.go
```diff
@@ -61,6 +61,7 @@ func defaultConfig() Genesis {
 			InitBalanceMap: make(map[string]string),
 		},
 		Poll: Poll{
+			PollMode:                         "governanceMix",
```
dustinxie: certain tests used lifelong
i think you'll need to set "lifelong" in those tests?

action/protocol/poll/protocol.go
yutongp: currently it is controlled within `staking committee`  if we don't enable native staking contract, it is governance only mode

dustinxie: i mean, both the "_modeGovernance" and "mixed" mode create the exactly same staking committee struct, so those 2 modes are actually identical

yutongp: i think eventhrough the code here is the same, we still have two different mode. They are not the same mode on our chain, not sure it is correct to combine them just because the code here is the same.

PR #2061, opened 2020-03-23
Make state factory an implementation of block indexer

CoderZhi:
1. Delete candidates from blockchain Context
2. Change state factory API to be compatible with blockdao.BlockIndexer


blockchain/blockdao/blockdao.go
```diff
@@ -210,25 +213,80 @@ func (dao *blockDAO) initStores() error {
...
-			if err := indexer.PutBlock(blk); err != nil {
+			producer, err := address.FromBytes(blk.PublicKey().Hash())
+			if err != nil {
+				return err
+			}
+			ctx, err = dao.fillWithBlockInfoAsTip(ctx, i-1)
+			if err != nil {
+				return err
+			}
+			fmt.Println("===========", tipHeight, dao.tipHeight, i, blk.Height())
```
yutongp: remove testing log

state/factory/factory.go
```diff
@@ -200,8 +201,29 @@ func (sf *factory) Start(ctx context.Context) error {
...
+			return errors.Wrap(err, "failed to init factory's height")
+		}
+		if bcCtx, ok := protocol.GetBlockchainCtx(ctx); ok {
+			for _, p := range bcCtx.Registry.All() {
+				if s, ok := p.(lifecycle.Starter); ok {
+					if err := s.Start(ctx); err != nil {
+						return errors.Wrap(err, "failed to start protocol")
+					}
+				}
+			}
```
yutongp: so we don't need to do anything if err is not nil or  db.ErrNotExist?

state/factory/factory.go
```diff
@@ -200,8 +201,29 @@ func (sf *factory) Start(ctx context.Context) error {
...
+		}
 		break
```
yutongp: do we need this break?

PR #2062, opened 2020-03-23
add candidate indexer unit test

koseoyoung:


action/protocol/poll/candidateindexer_test.go
```diff
@@ -0,0 +1,74 @@
...
+			Votes:         big.NewInt(10),
+			RewardAddress: "rewardAddress4",
+		},
+	}
+	require.NoError(indexer.PutCandidateList(uint64(1), &candidates))
+	candidates2, err := indexer.CandidateList(uint64(1))
+	require.NoError(err)
+	for i, cand := range candidates {
+		require.True(cand.Equal(candidates2[i]))
+	}
```
CoderZhi: even through the data reading from the indexer can not have extra fields, but we still need to check
```require.Equal(len(candidates2), len(candidates))```

File: action/protocol/poll/candidateindexer_test.go [hunk 2]
```diff
@@ -0,0 +1,74 @@
...
+			Votes:         big.NewInt(20),
+			RewardAddress: "rewardAddress3",
+		},
+		{
+			Address:       identityset.Address(4).String(),
+			Votes:         big.NewInt(10),
+			RewardAddress: "rewardAddress4",
+		},
+	}
+	require.NoError(indexer.PutCandidateList(uint64(1), &candidates))
```
CoderZhi: let's also try to call PutCandidateList twice with the same key and value, as well as with a new pair of key and value

koseoyoung: ok, but why do we have to test same key and value twice?

CoderZhi: in case that multiple writes may have different behavior. Actually, it is possible with current code base that it will be written twice with the same value.

PR #2064, opened 2020-03-24
unload/load candidate center from/to StateReader

dustinxie:
you can focus on last commit "Dock incremental change"

1. now only the changes to candidate center are pushed to Dock, not the entire in-mem struct
2. also, this is implemented in such a way that only blocks containing staking actions will trigger these changes. For a no-staking-action block, it will not generate any data exchange to/from StateReader

===========
added CandidateStateReader/Manager 
1. all handlers only access this, no more need to write to sm + in-mem center
2. putCandidate() and Upsert() are combined together
3. getAllCandidates() directly reads all from center


state/factory/factory.go
```diff
@@ -220,7 +227,22 @@ func (sf *factory) Start(ctx context.Context) error {
...
+		for _, p := range bcCtx.Registry.All() {
+			if s, ok := p.(lifecycle.Starter); ok {
+				if err := s.Start(ctx); err != nil {
+					return errors.Wrap(err, "failed to start protocol")
+				}
+			}
+		}
+	}
+	// push private data from protocols
+	return sf.dock.Push()
```
dustinxie: protocol.Start() may read data from DB and generates their initial view, so call Push() here to commit the initial view

misc/scripts/mockgen.sh
```diff
@@ -102,6 +95,13 @@ mockgen -destination=./test/mock/mock_chainmanager/mock_chainmanager.go  \
         -package=mock_chainmanager \
         StateManager
 
+mkdir -p ./test/mock/mock_blocksync
+mockgen -destination=./test/mock/mock_blocksync/mock_blocksync.go  \
+        -source=./blocksync/blocksync.go \
+        -self_package=github.com/iotexproject/iotex-core/blocksync \
+        -package=mock_blocksync \
+        BlockSync
+
```
dustinxie: make mockgen complains about error in block_sync
move it here will get rid of those error

action/protocol/staking/handlers_test.go
```diff
@@ -46,8 +46,11 @@ func TestProtocol_HandleCreateStake(t *testing.T) {
 	require.NoError(err)
 
 	// set up candidate
+	ctx := protocol.WithStateReaderCtx(context.Background(), protocol.StateReaderCtx{sm})
 	candidate := testCandidates[0].d.Clone()
-	require.NoError(setupCandidate(p, sm, candidate))
+	require.NoError(putCandidate(sm, candidate))
+	require.NoError(p.Start(ctx))
+	require.NoError(sm.Push())
```
dustinxie: this is to simulate factory.Start()

state/factory/factory.go
```diff
@@ -220,7 +227,22 @@ func (sf *factory) Start(ctx context.Context) error {
 	default:
 		return err
 	}
-	return sf.lifecycle.OnStart(ctx)
+	sf.mutex.Unlock()
+	needUnlock = false
+
+	// start the protocols
+	ctx = protocol.WithStateReaderCtx(ctx, protocol.StateReaderCtx{SR: sf})
```
CoderZhi: as yutong mentioned previously, we shouldn't pass so many parameters via ctx.

action/protocol/staking/protocol.go
```diff
@@ -178,6 +179,27 @@ func (p *Protocol) Start(ctx context.Context) error {
 			return err
 		}
 	}
+
+	// load the candidate center to StateReader
+	srCtx := protocol.MustGetStateReaderCtx(ctx)
+	return srCtx.SR.Load(protocolID, p.inMemCandidates)
```
CoderZhi: instead the "Load" could be moved into a function like "CreatePreStates", where you can pass sr as extra parameter.

action/protocol/staking/protocol.go
```diff
@@ -178,6 +179,27 @@ func (p *Protocol) Start(ctx context.Context) error {
...
+		return err
+	}
+
+	center, ok := v.(*CandidateCenter)
+	if !ok {
+		return ErrUnload
+	}
+
+	p.inMemCandidates = nil
+	p.inMemCandidates = center.Clone()
```
CoderZhi: Moreover, inMemCandidates is not a variable of protocol, but something stored in sr.

action/protocol/staking/protocol.go
dustinxie: Load means write the cand center to SR
in CreatePreStates, we call Unload, which is to read from SR

action/protocol/staking/protocol.go
dustinxie: yes, it is stored in SR, we call Unload() above to read it from SR

dustinxie: i think SR should only provide a means for protocol to store their data in general, that is why the data is defined as interface{}

it is up to each protocol to interpret the actual format of that data, and here it is *CandidateCenter

state/factory/factory.go
dustinxie: ok, i find a way such that we don't need to pass SR to protocol at Start()

action/protocol/staking/protocol.go
dustinxie: modified such that this is not needed
see new commit "address comment"

action/protocol/staking/handlers.go
```diff
@@ -684,6 +682,16 @@ func (p *Protocol) createLog(
 	}
 }
 
+func (p *Protocol) checkReceiptAndLoadCandCenter(r *action.Receipt, sm protocol.StateManager) (*action.Receipt, error) {
+	if r.Status != uint64(iotextypes.ReceiptStatus_Success) {
+		return r, nil
+	}
+	if err := sm.Load(protocolID, p.inMemCandidates); err != nil {
```
dustinxie: here we write the updated cand center to sm, to be committed in his Commit()

action/protocol/staking/protocol.go
CoderZhi: My previous explanation may not be clear. Let me rephrase it:
0. CreatePreStates is not used for this purpose, it just happens to be called before validating or committing a block
1. CandidateCenter is cloned when workingset is created, it is a variable stored in a working set (I can understand that we may only have one workingset at a time, that in current implementation, it behaves the same, but the way of cloning it is not appropriate)
2. Whenever we need to use CandidateCenter, it should be derived from sm. It is the same way as what you do here, but it is per sm. e.g., in function handleUnstake:
`func (p *Protocol) Hanle(ctx context.Context, act *action.Unstake, sm protocol.StateManager) (*action.Receipt, error) {`
we can convert the "sm" into a CandidateCenter, and then pass it to the sub handlers, that the sub handlers do not need to know the existence of "sm".
3. With the above two points, we don't really need a struct level variable inMemCandidates

dustinxie: well, isn't CreatePreStates() called before ws.runActions() to prepare the protocol for upcoming Validate() and Handle() calls? so i think it is the place for this purpose.

i know what you mean, and i think we have an issue/bug here. The block may have more than 1 staking actions, say voting for 3 diff candidates, we cannot have each handler read the (same) CandidateCenter, and push the dirty CandidateCenter (with 1 change) back to sm.

we still need a protocol-level inMemCandidates to collect all 3 changes, and push the dirty CandidateCenter (with 3 changes), after ws.runActions().

as we discussed earlier, CreatePostSystemActions() is only called in MintNewBlock(). So for this Dock idea to work, we'll need to add a similar interface like PostHandle(), and call that after ws.runActions(). Or we fallback to read all candidates from DB and construct the CandidateCenter on-the-fly in CreatePreStates()

dustinxie: thx, your comments opened more thoughts and insights

i structured it a bit such that this loading/syncing of CandidateCenter would only be needed when the protocol is actually called, that is, if the block contains staking actions. This avoids the unnecessary preparation of the CandidateCenter for irrelevant blocks.

pls check the latest commit in this PR

action/protocol/staking/protocol.go
```diff
@@ -182,8 +182,13 @@ func (p *Protocol) Start(ctx context.Context) error {
...
 
-// CreatePreStates creates states before Handle
-func (p *Protocol) CreatePreStates(ctx context.Context, sm protocol.StateManager) error {
+// ActionPostHandler performs post-handler action
+func (p *Protocol) ActionPostHandler(ctx context.Context, sm protocol.StateManager) error {
+	return sm.Load(protocolID, p.inMemCandidates)
+}
+
+// ActionPostCommit performs post-commit action
+func (p *Protocol) ActionPostCommit(ctx context.Context, sm protocol.StateManager) error {
```
dustinxie: Upon factory.Commit() success, this is called to update the CandidateCenter, to prepare it for next time

dustinxie: so that we only update when necessary, instead of update for each block in CreatePreStates()

action/protocol/staking/candidate_test.go
```diff
@@ -146,54 +197,92 @@ func TestCandCenter(t *testing.T) {
...
-	// test delete
-	for i, v := range testCandidates {
-		m.Delete(v.d.Owner)
-		r.False(m.ContainsOwner(v.d.Owner))
-		r.False(m.ContainsName(v.d.Name))
-		r.False(m.ContainsOperator(v.d.Operator))
-		r.False(m.ContainsSelfStakingBucket(v.d.SelfStakeBucketIdx))
-		r.Equal(len(testCandidates)-i-1, m.Size())
+	// test update existing and add new
+	// each time pick 2 random candidates to update, and add 2 new cand
```
dustinxie: add more rigorous test for update/add

action/protocol/staking/candidate_center.go
```diff
@@ -11,139 +11,347 @@ import (
...
+	}
+
 	// CandidateCenter is a struct to manage the candidates
 	CandidateCenter struct {
 		nameMap          map[string]*Candidate
 		ownerMap         map[string]*Candidate
 		operatorMap      map[string]*Candidate
 		selfStkBucketMap map[uint64]*Candidate
+		size             int
+		change           *candChange
```
CoderZhi: I didn't go into every detail. This may be an issue, because the changes are kept in candidate center as a struct level variable, instead of in dock.

dustinxie: Dock is not meant to keep the dynamic changes at any time, but the final pending changes at the end of running Handler, which call ActionPostHandler(), which will capture all the changes and load into Dock, basically:

delta = all the changes to CandidateCenter
dock.Load( serialization of delta)

upon success of factory.Commit(), it will unload the confirmed change, and commit back

b = dock.Unload()
delta2 = b.Deserialize(), so it is identical to "delta" earlier
and then:
CandidateCenter.Commit(delta2)

tests are added in candidate_test.go to verify that CandidateCenter at time of generating "delta", and CandidateCenter after committing "delta2" are identical

CoderZhi: I have different understanding:
changes should be in dock, because dock goes with sm. In other word, each sm has its own changes, not the "inMemCC".

action/protocol/staking/candidate_center.go
```diff
@@ -7,143 +7,432 @@
...
+
+// Commit writes the list into center
+func (m *CandidateCenter) Commit(c CandidateList) error {
+	if len(c) == 0 {
+		// nothing to commit, simply return
+		return nil
+	}
+
+	m.change = nil
+	m.change = newCandChange()
```
CoderZhi: 1. since "change" is in Candidate Center, why pass in "c"?
2. even we need "c", len(c) is 0 or not has different behavior, which is not expected.

action/protocol/staking/candidate_center.go
```diff
@@ -7,143 +7,432 @@
...
 		ownerMap         map[string]*Candidate
 		operatorMap      map[string]*Candidate
 		selfStkBucketMap map[uint64]*Candidate
 	}
+
+	// CandidateCenter is a struct to manage the candidates
+	CandidateCenter struct {
+		*candBase
+		size   int
+		change CandidateView
```
CoderZhi: if lock is on "candBase" level, it shouldn't be exposed to CandidateCenter or used in the CandidateCenter level.

action/protocol/staking/candidate_center.go
```diff
@@ -7,143 +7,432 @@
...
-	for _, c := range m.nameMap {
-		list = append(list, c)
+func (m *CandidateCenter) All() CandidateList {
+	list := m.change.All()
+	if list == nil {
+		list = make(CandidateList, 0, len(m.ownerMap))
+	}
+
+	m.lock.RLock()
+	defer m.lock.RUnlock()
```
CoderZhi: This is an example we shouldn't do that, lock is in the middle, and it is so hard to understand or remember that lock is not ownerMap.

action/protocol/staking/candidate_center.go
```diff
@@ -7,143 +7,432 @@
...
+		if _, ok := m.ownerMap[v.Owner.String()]; !ok {
+			m.size++
+		}
 	}
-	return list, nil
+	return nil
+}
+
+// Commit writes the list into center
+func (m *CandidateCenter) Commit(c CandidateList) error {
```
CoderZhi: Commit is not for CandidateCenter, but CandidateBase

action/protocol/staking/protocol.go
```diff
@@ -96,8 +100,8 @@ func NewProtocol(depositGas DepositGas, sr protocol.StateReader, cfg genesis.Sta
 	}
 
 	return &Protocol{
-		addr:            addr,
-		inMemCandidates: NewCandidateCenter(),
+		addr:        addr,
+		candCenters: cache.NewThreadSafeLruCache(8),
```
CoderZhi: we should have lru for historical height, and a view for current height which should always be kept in memory.

action/protocol/staking/candidate_center.go
dustinxie: "change" is the final confirmed change, it comes from workingset's Dock

i can reformat it to a call of SetDelta(), then Commit(), this would be easier to understand

action/protocol/staking/candidate_center.go
dustinxie: fixed in latest push

action/protocol/staking/candidate_center.go
dustinxie: fixed in latest push

action/protocol/staking/candidate_center.go
dustinxie: fixed in latest push

action/protocol/staking/protocol.go
dustinxie: not sure if i understand your comment correctly, but this is the LRU for heights, the latest 8 heights will be kept in this lru cache

view for current height is always in the LRU, so i think not necessary to keep a separate one in memory

for example, if you call ActiveCandidates(ctx context.Context, height uint64) with tip height, getCandCenter() will hit and pull up the correct view for you

the same for ReadState(sr) if sr.Height() = tip height

action/protocol/staking/candidate_center.go
```diff
@@ -7,143 +7,499 @@
...
 type (
-	// CandidateCenter is a struct to manage the candidates
-	CandidateCenter struct {
+	// CandidateCenter is the candidate center
+	CandidateCenter interface {
+		CandidateView
+		Base() CandidateCenter
+		Delta() CandidateList
+		SetDelta(CandidateView) error
+		Commit() error
```
yutongp, line 22: use committer interface

action/protocol/staking/candidate_statemanager.go
```diff
@@ -0,0 +1,137 @@
...
+		Size() int
+		ContainsName(string) bool
+		ContainsOwner(address.Address) bool
+		ContainsOperator(address.Address) bool
+		ContainsSelfStakingBucket(uint64) bool
+		GetByName(string) *Candidate
+		GetByOwner(address.Address) *Candidate
+		GetBySelfStakingIndex(uint64) *Candidate
+		Upsert(*Candidate) error
+		Commit() error
```
yutongp, line 32: use committer interface

action/protocol/staking/protocol.go
```diff
@@ -153,55 +164,71 @@ func (p *Protocol) CreateGenesisStates(
...
-		}
+	csm, err := NewCandidateStateManager(sm, center)
+	if err != nil {
+		return errors.Wrap(err, "failed to commit candidate change in sm")
 	}
+	return csm.Commit()
+}
+
+// Abort aborts the last change
+func (p *Protocol) Abort(ctx context.Context, sm protocol.StateManager) error {
```
yutongp, line 190: don't need this

action/protocol/protocol.go
```diff
@@ -31,18 +31,29 @@ type Protocol interface {
...
 	CreatePreStates(context.Context, StateManager) error
 }
 
+// Commit performs commit action of the protocol
+type Commit interface {
+	Commit(context.Context, StateManager) error
+}
+
+// Abort performs abort action of the protocol
+type Abort interface {
```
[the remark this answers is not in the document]
CoderZhi: if no one is using it now, can we add this later when it is needed?

action/protocol/staking/candidate_statemanager.go
dustinxie: this needs to wait until we move CandCenter into sf, at that time CandidateStateManager can obtain the center from sm

PR #2065, opened 2020-03-24
fix blockdao deadlock bug

koseoyoung:
fix 2 bugs 
1. deadlock issue
2. In GetTopDB() in case of ```index the height --> file index mapping``` it should put index(top+1) 

For dead lock issue, the blockdao mutex is not for general usage but for creating new files/ rangeIndex files. But while previous refactoring, in the ```putBlock(), DeleteBlockToTarget()```, we lock/unlock mutex wrongly.

### DIFF
--- a/blockchain/blockdao/blockdao.go
+++ b/blockchain/blockdao/blockdao.go
@@ -88,8 +88,6 @@ type (
 		GetReceipts(uint64) ([]*action.Receipt, error)
 		PutBlock(*block.Block) error
 		DeleteBlockToTarget(uint64) error
-		IndexFile(uint64, []byte) error
-		GetFileIndex(uint64) ([]byte, error)
 		KVStore() db.KVStore
 		HeaderByHeight(uint64) (*block.Header, error)
 		FooterByHeight(uint64) (*block.Footer, error)
@@ -327,8 +325,6 @@ func (dao *blockDAO) GetReceipts(blkHeight uint64) ([]*action.Receipt, error) {
 
 func (dao *blockDAO) PutBlock(blk *block.Block) error {
 	err := func() error {
-		dao.mutex.Lock()
-		defer dao.mutex.Unlock()
 		if err := dao.putBlock(blk); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


blockchain/blockdao/blockdao.go
```diff
@@ -327,8 +325,6 @@ func (dao *blockDAO) GetReceipts(blkHeight uint64) ([]*action.Receipt, error) {
 
 func (dao *blockDAO) PutBlock(blk *block.Block) error {
 	err := func() error {
-		dao.mutex.Lock()
-		defer dao.mutex.Unlock()
 		if err := dao.putBlock(blk); err != nil {
```
dustinxie: i think our agreed coding practice is public interface should use lock, and private not
so we should keep this, see comments in FileIndex() below

blockchain/blockdao/blockdao.go
```diff
@@ -381,7 +375,7 @@ func (dao *blockDAO) DeleteBlockToTarget(targetHeight uint64) error {
 	return nil
 }
 
-func (dao *blockDAO) IndexFile(height uint64, index []byte) error {
+func (dao *blockDAO) indexFile(height uint64, index []byte) error {
 	dao.mutex.Lock()
 	defer dao.mutex.Unlock()
```
dustinxie: we should remove the lock here
and move the below init code block to Start()

blockchain/blockdao/blockdao.go
```diff
@@ -395,8 +389,8 @@ func (dao *blockDAO) IndexFile(height uint64, index []byte) error {
 	return dao.htf.Insert(height, index)
 }
 
-// GetFileIndex return the db filename
-func (dao *blockDAO) GetFileIndex(height uint64) ([]byte, error) {
+// getFileIndex return the db filename
+func (dao *blockDAO) getFileIndex(height uint64) ([]byte, error) {
 	dao.mutex.RLock()
 	defer dao.mutex.RUnlock()
```
dustinxie: we should remove the lock here
and move the below init code block to Start()

PR #2074, opened 2020-03-27
[staking] support using stakingV2 in poll protocol

yutongp:


action/protocol/poll/staking_command.go
```diff
@@ -81,29 +85,22 @@ func (sc *stakingCommand) Start(ctx context.Context) error {
...
-	}
-
 	// transition to V2 starting Fairbank
 	height, err := sm.Height()
 	if err != nil {
 		return nil, err
 	}
-	if sc.hu.IsPost(config.Fairbank, height) {
+	bcCtx := protocol.MustGetBlockchainCtx(ctx)
+	hu := config.NewHeightUpgrade(&bcCtx.Genesis)
```
dustinxie: keep sc.hu, which can be generated one-time at New()
then no need to re-generate here each time?

action/protocol/poll/staking_command.go
```diff
@@ -81,29 +85,22 @@ func (sc *stakingCommand) Start(ctx context.Context) error {
...
+	if sc.stakingV1 == nil || hu.IsPost(config.Fairbank, height) {
```
dustinxie: might be an issue to simplify the logic like this

for stakingV1 == nil, and height < Fairbank, it will skip the if(), and call stakingV1.Handle(ctx, act, sm), that will panic

action/protocol/poll/staking_command.go
```diff
@@ -114,47 +111,56 @@ func (sc *stakingCommand) Validate(ctx context.Context, act action.Action) error
...
-	if sc.stakingV2 == nil {
-		return sc.stakingV1.CalculateCandidatesByHeight(ctx, height)
-	}
-
 	// transition to V2 starting Fairbank
-	if sc.hu.IsPost(config.Fairbank, height) {
-		return sc.stakingV2.ActiveCandidates(ctx)
+	bcCtx := protocol.MustGetBlockchainCtx(ctx)
+	hu := config.NewHeightUpgrade(&bcCtx.Genesis)
+	if sc.stakingV1 == nil || hu.IsPost(config.Fairbank, height) {
```
dustinxie: same

action/protocol/poll/staking_command.go
```diff
@@ -114,47 +111,56 @@ func (sc *stakingCommand) Validate(ctx context.Context, act action.Action) error
...
 // NextCandidates returns candidate list from state factory of next epoch
 func (sc *stakingCommand) NextCandidates(ctx context.Context, sr protocol.StateReader) (state.CandidateList, error) {
-	// TODO: handle V2
-	return sc.stakingV1.NextCandidates(ctx, sr)
+	return sc.slasher.GetCandidates(ctx, sr, true)
 }
 
 func (sc *stakingCommand) ReadState(ctx context.Context, sr protocol.StateReader, method []byte, args ...[]byte) ([]byte, error) {
-	// TODO: handle V2
+	if sc.stakingV1 == nil {
```
dustinxie: this is the right order, it returns for stakingV1 == nil
then no need to worry about nil stakingV1 passing this point

action/protocol/poll/staking_command.go
dustinxie: i think the way in ReadState() is correct/better

action/protocol/poll/staking_command.go
```diff
@@ -57,18 +61,18 @@ func NewStakingCommand(
...
-		if err := sc.stakingV2.CreateGenesisStates(ctx, sm); err != nil {
-			return err
-		}
+	cands, err := sc.stakingV2.ActiveCandidates(ctx)
+	if err != nil {
+		return err
 	}
-	return nil
+	bcCtx := protocol.MustGetBlockchainCtx(ctx)
+	cands = sc.filterAndSortCandidatesByVoteScore(cands, bcCtx.Tip.Timestamp)
```
dustinxie: if this part is missing in current code, is it doable to move this into stakingV2.CreateGenesis(), and handle there?

and here still directly call stakingV2.CreateGenesis(), StakingCommand is just combining V1 and V2, and call their respective func, feel that would be better

yutongp: then we need define a stakingv2 in poll package. staking package shouldn't do this part of work.

action/protocol/poll/staking_command.go
```diff
@@ -114,47 +111,56 @@ func (sc *stakingCommand) Validate(ctx context.Context, act action.Action) error
...
-		return sc.stakingV2.ActiveCandidates(ctx)
+	bcCtx := protocol.MustGetBlockchainCtx(ctx)
+	hu := config.NewHeightUpgrade(&bcCtx.Genesis)
+	if sc.stakingV1 == nil || hu.IsPost(config.Fairbank, height) {
+		cands, err := sc.stakingV2.ActiveCandidates(ctx)
+		if err != nil {
+			return cands, err
+		}
+		bcCtx := protocol.MustGetBlockchainCtx(ctx)
+		return sc.filterAndSortCandidatesByVoteScore(cands, bcCtx.Tip.Timestamp), nil
```
dustinxie: can we absorb filterAndSortCandidatesByVoteScore() into V2.ActiveCandidates() ?

action/protocol/poll/slasher.go
```diff
@@ -70,6 +71,142 @@ func NewSlasher(
...
+	blkCtx := protocol.MustGetBlockCtx(ctx)
+	rp := rolldpos.MustGetProtocol(protocol.MustGetRegistry(ctx))
+	epochNum := rp.GetEpochNum(blkCtx.BlockHeight) // tip
+	epochStartHeight := rp.GetEpochHeight(epochNum)
+	switch string(method) {
+	case "CandidatesByEpoch":
+		if len(args) != 0 {
+			epochNum = byteutil.BytesToUint64(args[0])
+			epochStartHeight = rp.GetEpochHeight(epochNum)
+		}
```
CoderZhi: duplicated in all cases

action/protocol/poll/protocol.go
```diff
@@ -124,7 +124,7 @@ func NewProtocol(
 	getkickoutList GetKickoutList,
 	getUnproductiveDelegate GetUnproductiveDelegate,
 	electionCommittee committee.Committee,
-	stakingV2 *staking.Protocol,
+	stakingProto *staking.Protocol,
```
dustinxie: nit: rename to nativeStk?
proto might lead to think protobuf

yutongp: i think it is fine. staking.Protocol -> stakingProto

yutongp: nativeStk may confused with poll.nativestaking and poll.nativestakingv2 as well.

PR #2083, opened 2020-04-01
fix poll result action issue

CoderZhi:
This may cause an issue for consortium poll protocol, because its NextCandidates always returns a candidate list with no error, which may cause a failure of creating poll action result post system action.


koseoyoung:
This handling doesn't work with current implementation. 
B/c shiftCandidates function is not erasing next candidate, but only shifting from next to current candidate.

The thing we want to handle is after mid of epoch, if we already sent out putpollresult action, not return the error but just skipping making action.

then we can fix this bug in either way  
1) check if it's only mid of epoch, send out action 
2) or change ShiftCandidate function to erase nextCandidate when shifting, and if the nextCandidate() == state.NotExist, not returning error but just skipping

action/protocol/poll/util.go
```diff
@@ -114,6 +114,9 @@ func createPostSystemActions(ctx context.Context, p Protocol) ([]action.Envelope
 	if blkCtx.BlockHeight < epochHeight+(nextEpochHeight-epochHeight)/2 {
 		return nil, nil
 	}
+	if _, err := p.NextCandidates(ctx, sr); errors.Cause(err) != state.ErrStateNotExist {
+		return nil, err
```
koseoyoung, line 118: not return error but just skip

Issue #2017, opened 2020-03-10
ioctl to support native staking/registration


dustinxie:
refer to /action/stake_xxx.go and /action/candidate_xxx.go, add support in the following order:

1. CreateStake
2. CandidateRegister
3. Unstake
4. Withdraw
5. ChangeCandidate
6. TransferStake
7. DepositToStake
8. Restake
9. CandidateUpdate

dustinxie:
corresponding to commands:
1. ioctl stake2 register
2. ioctl stake2 create
3. ioctl stake2 release
4. ioctl stake2 withdraw
5. ioctl stake2 change
6. ioctl stake2 transfer
7. ioctl stake2 add
8. ioctl stake2 renew
9. ioctl stake2 update

dustinxie:
thx @RuneAkko finishing first PR for CreateStake

can you continue to add new cmds in the order listed, one PR for one command?

PR #2084, opened 2020-04-01
add native stake command: ioctl stake2 create

RuneAkko:
#2017
@dustinxie


ioctl/cmd/action/stake2.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/iotexproject/iotex-core/ioctl/config"
+)
+
+// Multi-language support
+var (
+	stake2CmdUses = map[config.Language]string{
+		config.English: "stake2",
+		config.Chinese: "stake2",
```
dustinxie, line 19: add usage like current v1

ioctl/cmd/action/stake2.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+)
+
+var stake2AutoRestake bool
```
dustinxie, line 23: autoRestakeV2?

ioctl/cmd/action/stake2.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/iotexproject/iotex-core/ioctl/config"
+)
+
+// Multi-language support
+var (
+	stake2CmdUses = map[config.Language]string{
+		config.English: "stake2",
+		config.Chinese: "stake2",
+	}
```
dustinxie, line 20: do we need stakeCmdShorts, flagEndpointUsages, flagInsecureUsages like current code?

ioctl/cmd/action/stake2create.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func init() {
+	registerWriteCommand(stake2CreateCmd)
+	stake2CreateCmd.Flags().BoolVar(&stake2AutoRestake, "auto-restake", false,
+		config.TranslateInLang(stake2FalgAutoRestakeUsages, config.UILanguage))
+}
+
+func stake2Create(args []string) error {
+	var amount = args[0]
```
dustinxie, line 55: this should also need StringToRau?

ioctl/cmd/action/stake2create.go
```diff
@@ -0,0 +1,100 @@
...
+	var candidateName = args[1]
```
dustinxie, line 56: we also need to validate the name, will let u know the detail method soon

dustinxie: refer to staking/validations.go

func IsValidCandidateName(s string) bool

ioctl/cmd/action/stake2create.go
RuneAkko: action.NewCreateStake function accepts the "amount" as string?

RuneAkko: > this should also need StringToRau?

action.NewCreateStake function accepts the "amount" as string?

ioctl/cmd/action/stake2create.go
RuneAkko: > refer to staking/validations.go
> 
> func IsValidCandidateName(s string) bool

ok, I will change ioctl/validator/validator.go:func ValidateCandidateName() to this func.

ioctl/cmd/action/stake2create.go
dustinxie: yes, but as far as i remember unit should be Rau?

For example, staking 100 token, args[0] from input is "100", but amount to call NewCreateStake () should be "100000000000000000000"

can you double-check that's the case?

ioctl/validator/validator.go
```diff
@@ -71,9 +71,13 @@ func ValidateStakeDuration(stakeDuration *big.Int) error {
 
 // ValidateCandidateName validates candidate name for native staking
 func ValidateCandidateName(candidateName string) error {
-	if len(candidateName) > 12 {
-		return ErrLongCandidateName
+	if len(s) == 0 || len(s) > 12 {
```
dustinxie, line 74: current ValidateCandidateName() is also used for V1, should not change it

i think we'll need a new ValidateCandidateNameV2() for staking V2

ioctl/validator/validator.go
```diff
@@ -24,7 +24,7 @@ var (
 	// ErrInvalidStakeDuration indicates error for invalid stake duration
 	ErrInvalidStakeDuration = errors.New("stake duration must be within 0 and 1050 and in multiples of 7")
 	// ErrInvalidCandidateName indicates error for invalid candidate name
-	ErrLongCandidateName = errors.New("invalid length of candidate name that is more than 12 ")
+	ErrCandidateName = errors.New("the candidate name string is not valid")
```
dustinxie, line 27: i think define a new Err for V2, see comment below

ioctl/cmd/action/stake2create.go
RuneAkko: ok，but it seems that codes in line 49  in action.NewCreateStake function will be duplicated.

ioctl/cmd/action/stake2create.go
```diff
@@ -0,0 +1,111 @@
...
+	gasPriceRau, err := gasPriceInRau()
+	if err != nil {
+		return output.NewError(0, "failed to get gas price", err)
+	}
+	nonce, err := nonce(sender)
+	if err != nil {
+		return output.NewError(0, "failed to get nonce ", err)
+	}
+
+	s2c, err := action.NewCreateStake(nonce, candidateName, amountStringInRau, duration, stake2AutoRestake, data, gasLimit, gasPriceRau)
```
dustinxie, line 100: no need for amountStringInRau := strconv.FormatUint(amountInRau.Uint64(), 10) in above
just use amountInRau.String()

PR #2093, opened 2020-04-02
Modify handleUnstake and handleWithrawStake in staking protocol

lizhefeng:


koseoyoung:
from my understanding, if autoStake = true, we can't unstake. If someone change to autoStake = false, then after the duration from that point, they are able to unstake. Is it right?

lizhefeng:
@koseoyoung I think you are right.

lizhefeng:
@koseoyoung I think you are right.

action/protocol/staking/handlers.go
```diff
@@ -127,16 +127,24 @@ func (p *Protocol) handleUnstake(ctx context.Context, act *action.Unstake, csm C
...
+		return p.settleAction(ctx, csm, uint64(iotextypes.ReceiptStatus_ErrUnstakeBeforeMaturity), gasFee)
 	}
 
 	candidate := csm.GetByOwner(bucket.Candidate)
 	if candidate == nil {
 		return nil, errors.Wrap(ErrInvalidOwner, "cannot find candidate in candidate center")
 	}
+
+	// update bucket
+	bucket.UnstakeStartTime = blkCtx.BlockTimeStamp
```
dustinxie, line 143: add bucket.AutoStake = false

action/protocol/staking/handlers.go
```diff
@@ -127,16 +127,24 @@ func (p *Protocol) handleUnstake(ctx context.Context, act *action.Unstake, csm C
...
-	if err := updateBucket(csm, act.BucketIndex(), bucket); err != nil {
-		return nil, errors.Wrapf(err, "failed to update bucket for voter %s", bucket.Owner)
+	if blkCtx.BlockTimeStamp.Before(bucket.StakeStartTime.Add(bucket.StakedDuration)) {
+		err := fmt.Errorf("bucket is not ready to be unstaked, current time %s, required time %s",
+			blkCtx.BlockTimeStamp, bucket.StakeStartTime.Add(bucket.StakedDuration))
+		log.L().Debug("Error when withdrawing bucket", zap.Error(err))
+		return p.settleAction(ctx, csm, uint64(iotextypes.ReceiptStatus_ErrUnstakeBeforeMaturity), gasFee)
 	}
 
 	candidate := csm.GetByOwner(bucket.Candidate)
```
dustinxie, line 137: can you move this before check unstake time?

in other handlers, we always check candidate exist/conflict first, before we check bucket, let's be consistent

PR #2094, opened 2020-04-03
reconcile common code into ws.Commit()

dustinxie:


state/factory/factory.go
```diff
@@ -329,8 +329,11 @@ func (sf *factory) newWorkingSet(ctx context.Context, height uint64) (*workingSe
 			if err := flusher.Flush(); err != nil {
 				return errors.Wrap(err, "failed to Commit all changes to underlying DB in a batch")
 			}
+			if err := sf.twoLayerTrie.SetRootHash(tlt.RootHash()); err != nil {
+				return err
+			}
```
dustinxie, line 334: should wait SetRootHash() success, before update sf's height

action/protocol/managers.go
```diff
@@ -96,7 +96,6 @@ type (
 		ProtocolDirty(string) bool
 		Load(string, interface{}) error
 		Unload(string) (interface{}, error)
-		Push() error
```
dustinxie, line 99: Push() is designed to have multiple workingsets put their items into the Dock, whoever called Push(), his item will be the final value to be committed to Dock

but now with each workingset having its own Dock, Push() is not needed

state/factory/statedb.go
```diff
@@ -458,20 +458,6 @@ func (sdb *stateDB) state(ns string, addr []byte, s interface{}) error {
 	return nil
 }
 
-func (sdb *stateDB) commit(ctx context.Context, ws *workingSet) error {
```
dustinxie, line 461: common code, move to workingset

state/factory/factory.go
```diff
@@ -637,20 +640,6 @@ func (sf *factory) stateAtHeight(height uint64, ns string, key []byte, s interfa
 	return readState(tlt, ns, key, s)
 }
 
-func (sf *factory) commit(ctx context.Context, ws *workingSet) error {
```
dustinxie, line 640: common code, move to workingset

Issue #2087, opened 2020-04-02
refactor settleAction() in order to simplify CandidateStateManager interface


dustinxie:
**What would you like to be added**:
1. move settleAction() out of sub-handler and into Handle()
2. so that CandidateStateManager does not need to implement/expose protocol.StateManager

**Why is this needed**:
CandidateStateManager no need to expose more than needed

PR #2095, opened 2020-04-03
store protocol's view in factory

dustinxie:
1. read cand center from state factory/SR, and write the updated value upon commit
2. delete LRU cache
3. staking protocol code is now also same as before, only used 2 more funcs: NewCandidateStateManager(), and getOrCreateCandCenter()

### DIFF
--- a/action/protocol/managers.go
+++ b/action/protocol/managers.go
@@ -74,18 +74,19 @@ type (
 		Height() (uint64, error)
 		State(interface{}, ...StateOption) (uint64, error)
 		States(...StateOption) (uint64, state.Iterator, error)
+		ReadView(string) (interface{}, error)
 	}
 
 	// StateManager defines the stateDB interface atop IoTeX blockchain
 	StateManager interface {
 		StateReader
-		ConfirmedHeight() uint64
 		// Accounts
 		Snapshot() int
 		Revert(int) error
 		// General state
 		PutState(interface{}, ...StateOption) (uint64, error)
 		DelState(...StateOption) (uint64, error)
+		WriteView(string, interface{}) error
 		Dock
 	}
 
--- a/action/protocol/poll/staking_command.go
+++ b/action/protocol/poll/staking_command.go
@@ -8,7 +8,6 @@ package poll
 
 import (
 	"context"
-
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
 	"github.com/pkg/errors"
@@ -16,7 +15,6 @@ import (
 	"github.com/iotexproject/iotex-core/action"
 	"github.com/iotexproject/iotex-core/action/protocol"
 	"github.com/iotexproject/iotex-core/config"
-	"github.com/iotexproject/iotex-core/pkg/lifecycle"
 	"github.com/iotexproject/iotex-core/state"
 )
 
@@ -53,16 +51,16 @@ func (sc *stakingCommand) CreateGenesisStates(ctx context.Context, sm protocol.S
 	return sc.stakingV2.CreateGenesisStates(ctx, sm)
 }
 
-func (sc *stakingCommand) Start(ctx context.Context) error {
+func (sc *stakingCommand) Start(ctx context.Context, sr protocol.StateReader) (interface{}, error) {
 	if sc.stakingV1 != nil {
-		if starter, ok := sc.stakingV1.(lifecycle.Starter); ok {
-			return starter.Start(ctx)
+		if starter, ok := sc.stakingV1.(protocol.Starter); ok {
+			if _, err := starter.Start(ctx, sr); err != nil {
+				return nil, err
+			}
 		}
 	}
-	if starter, ok := sc.stakingV2.(lifecycle.Starter); ok {
-		return starter.Start(ctx)
-	}
-	return nil
+	// stakingV2 Start() do nothing now
+	return nil, nil
 }
 
 func (sc *stakingCommand) CreatePreStates(ctx context.Context, sm protocol.StateManager) error {
--- a/action/protocol/poll/staking_command.go
+++ b/action/protocol/poll/staking_command.go
@@ -60,7 +60,12 @@ func (sc *stakingCommand) Start(ctx context.Context, sr protocol.StateReader) (i
 			}
 		}
 	}
-	// stakingV2 Start() do nothing now
+
+	if sc.stakingV2 != nil {
+		if starter, ok := sc.stakingV2.(protocol.Starter); ok {
+			return starter.Start(ctx, sr)
+		}
+	}
 	return nil, nil
 }
 
--- a/action/protocol/poll/staking_command.go
+++ b/action/protocol/poll/staking_command.go
@@ -8,6 +8,7 @@ package poll
 
 import (
 	"context"
+
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
 	"github.com/pkg/errors"
--- a/action/protocol/registry.go
+++ b/action/protocol/registry.go
@@ -11,7 +11,6 @@ import (
 	"reflect"
 	"sync"
 
-	"github.com/iotexproject/iotex-core/pkg/lifecycle"
 	"github.com/pkg/errors"
 )
 
@@ -92,17 +91,26 @@ func (r *Registry) all() []Protocol {
 }
 
 // StartAll starts all protocols which are startable
-func (r *Registry) StartAll(ctx context.Context) error {
+func (r *Registry) StartAll(ctx context.Context, sr StateReader, dock Dock) error {
 	if r == nil {
 		return nil
 	}
 	r.mu.RLock()
 	defer r.mu.RUnlock()
 	for _, p := range r.all() {
-		if s, ok := p.(lifecycle.Starter); ok {
-			if err := s.Start(ctx); err != nil {
-				return errors.Wrapf(err, "failed to start protocol %s", reflect.TypeOf(s))
-			}
+		s, ok := p.(Starter)
+		if !ok {
+			continue
+		}
+		view, err := s.Start(ctx, sr)
+		if err != nil {
+			return errors.Wrapf(err, "failed to start protocol %s", reflect.TypeOf(p))
+		}
+		if view == nil {
+			continue
+		}
+		if err := dock.Load(p.Name(), view); err != nil {
+			return errors.Wrapf(err, "failed to start protocol %s", reflect.TypeOf(p))
 		}
 	}
 	return nil
--- a/action/protocol/registry.go
+++ b/action/protocol/registry.go
@@ -91,12 +91,13 @@ func (r *Registry) all() []Protocol {
 }
 
 // StartAll starts all protocols which are startable
-func (r *Registry) StartAll(ctx context.Context, sr StateReader, dock Dock) error {
+func (r *Registry) StartAll(ctx context.Context, sr StateReader) (map[string]interface{}, error) {
 	if r == nil {
-		return nil
+		return nil, nil
 	}
 	r.mu.RLock()
 	defer r.mu.RUnlock()
+	allView := make(map[string]interface{})
 	for _, p := range r.all() {
 		s, ok := p.(Starter)
 		if !ok {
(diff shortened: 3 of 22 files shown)


action/protocol/managers.go
```diff
@@ -74,18 +74,19 @@ type (
 		Height() (uint64, error)
 		State(interface{}, ...StateOption) (uint64, error)
 		States(...StateOption) (uint64, state.Iterator, error)
+		ReadView(string) (interface{}, error)
 	}
 
 	// StateManager defines the stateDB interface atop IoTeX blockchain
 	StateManager interface {
 		StateReader
-		ConfirmedHeight() uint64
```
dustinxie, line 82: no longer needed

state/factory/historyfactory.go
```diff
@@ -43,3 +43,8 @@ func (hReader *historyStateReader) States(opts ...protocol.StateOption) (uint64,
 	}
 	return hReader.height, iterator, nil
 }
+
+// ReadView reads the view
+func (hReader *historyStateReader) ReadView(name string) (interface{}, error) {
+	return hReader.sf.ReadView(name)
+}
```
CoderZhi, line 50: wrong implementation?

state/factory/workingset.go
```diff
@@ -46,6 +46,8 @@ type (
 		finalized    bool
 		dock         protocol.Dock
 		commitFunc   func(uint64) error
+		readviewFunc func(name string) (interface{}, error)
+		writviewFunc func(name string, v interface{}) error
```
CoderZhi, line 50: typo?

action/protocol/staking/protocol.go
```diff
@@ -169,26 +166,35 @@ func (p *Protocol) CreateGenesisStates(
...
-	if err != nil {
-		return errors.Wrap(err, "failed to commit candidate change in sm")
+	// commit updated view
+	if err := csm.Commit(); err != nil {
+		return errors.Wrap(err, "failed to commit candidate change")
 	}
-	return csm.Commit()
+
+	// write update view back to state factory
+	return csm.WriteView(protocolID, csm.CandCenter())
```
CoderZhi, line 192: WriteView should be called in csm.Commit

state/factory/statedb.go
```diff
@@ -42,6 +42,7 @@ type stateDB struct {
 	dao                db.KVStore // the underlying DB for account/contract storage
 	timerFactory       *prometheustimer.TimerFactory
 	workingsets        *lru.Cache // lru cache for workingsets
+	protocolView       protocol.Dock
```
CoderZhi, line 45: view should be loaded on start for both statedb and factory

action/protocol/staking/candidate_statemanager.go
```diff
@@ -19,6 +19,7 @@ type (
 	// CandidateStateManager is candidate manager on top of StateMangaer
 	CandidateStateManager interface {
 		protocol.StateManager
+		CandCenter() CandidateCenter
```
CoderZhi, line 22: CandidateStateManager  shouldn't inherit protocol.StateManager

dustinxie: right, this is your earlier comment, already captured in https://github.com/iotexproject/iotex-core/issues/2087 couple of days ago

it will be the next step after finish this PR

state/factory/historyfactory.go
dustinxie: can you be more specific what is wrong?

action/protocol/staking/protocol.go
dustinxie: yeah, can do

state/factory/statedb.go
dustinxie: not sure if i understand correctly, i think view should be loaded by the protocol upon receiving the workingset? b/c the value stored is a generic interface{}, it's up to each protocol to load from ws and convert it

also if the block does not contain staking actions there's actually no need for statedb/factory to load the view, so it should be loaded by its consumer, which is protocol. that is my understanding

state/factory/historyfactory.go
dustinxie: for now historyfactory.go seems just calling sf.Method() so just follow suite, do u mean for historyfactory the view is different for different height?

state/factory/workingset.go
dustinxie: fixed in latest push

state/factory/historyfactory.go
CoderZhi: You can leave it as `return errors.New("not implemented")`

state/factory/statedb.go
CoderZhi: factory always stays in memory.  if the view is loaded in the fly, it may take a long time.
Two cases for a protocol's view:
1. It is used in a block: we need to load it any way
2. it is not used in a block: the view won't be updated, it doesn't hurt to keep it as it was in the factory.

dustinxie: yes, factory always in-mem, i think the "load" we discuss here refers to constructing the cand center from candidates in the factory, and keep it in the factory's Dock, right?

2 cases for our staking protocol:
1. bootstrap cand list != nil, then CreateGenesisState() will do the "load", and store the cand center into factory's Dock, Dock["staking"] = center
2. bootstrap cand list == nil, "load" will be done by staking protocol on the 1st block that contains staking action

and after that, read view (in-the-fly) is just an in-mem map access: center = factory.Dock["staking"], won't take long time

since "load" is only a one-time action, i think can let each protocol handle its own. factory just provide the Dock, each protocol will only "load" once and later it's a quick memory access to load the view

if we want to do "load" in factory, i feel there are 2 concerns:
1. factory needs to call staking protocol's method to construct the cand center, in our case createCandCenter(), but factory shouldn't understand these details
2. if we have more protocols in the future, factory needs to know all the protocols and "load" all of them in Start(), again factory no need to know all these protocols, it just provide a Dock for protocols to read/write a generic interface{}, that's it

state/factory/historyfactory.go
dustinxie: fixed in latest push

action/protocol/staking/protocol.go
```diff
@@ -55,11 +54,10 @@ var (
...
 // Protocol defines the protocol of handling staking
 type Protocol struct {
-	addr        address.Address
-	candCenters *cache.ThreadSafeLruCache
-	depositGas  DepositGas
-	sr          protocol.StateReader
-	config      Configuration
+	addr       address.Address
+	depositGas DepositGas
+	sr         protocol.StateReader
```
yutongp, line 59: why we have this sr stored in protocol?

action/protocol/staking/protocol.go
```diff
@@ -255,25 +251,10 @@ func (p *Protocol) Validate(ctx context.Context, act action.Action) error {
 
 // ActiveCandidates returns all active candidates in candidate center
 func (p *Protocol) ActiveCandidates(ctx context.Context, height uint64) (state.CandidateList, error) {
```
yutongp, line 253: should this `ActiveCandidates` take in a sr?

action/protocol/staking/protocol.go
dustinxie: yes, this is also a lingering issue. b/c we still have ActiveCandidates(ctx context.Context, height uint64), we need to read the view from sr

ActiveCandidates is for the oldest governance protocol, so comes with a height that you can read the cand list (from election service) at that Ether height

we may need to clearly re/define ActiveCandidates(), for staking v2 it should always be current height, and we need to pass in StateReader, so we can remove p.sr from protocol

action/protocol/staking/protocol.go
dustinxie: exactly, then we can remove p.sr

action/protocol/poll/staking_command.go
```diff
@@ -8,15 +8,13 @@ package poll
 
 import (
 	"context"
-
```
CoderZhi, line 11: revert

action/protocol/poll/staking_command.go
```diff
@@ -53,16 +51,16 @@ func (sc *stakingCommand) CreateGenesisStates(ctx context.Context, sm protocol.S
...
+			if _, err := starter.Start(ctx, sr); err != nil {
+				return nil, err
+			}
 		}
 	}
-	if starter, ok := sc.stakingV2.(lifecycle.Starter); ok {
-		return starter.Start(ctx)
-	}
-	return nil
+	// stakingV2 Start() do nothing now
```
dustinxie, line 62: i think it's cleaner not to call V2 Start(), since it does nothing now

action/protocol/staking/protocol.go
```diff
@@ -173,7 +188,12 @@ func (p *Protocol) CreateGenesisStates(
 
 // Commit commits the last change
 func (p *Protocol) Commit(ctx context.Context, sm protocol.StateManager) error {
-	csm, err := NewCandidateStateManager(sm)
+	center, err := getCandCenter(sm)
```
dustinxie, line 191: we must have view loaded here, so call get()

action/protocol/staking/protocol.go
```diff
@@ -184,7 +204,12 @@ func (p *Protocol) Commit(ctx context.Context, sm protocol.StateManager) error {
 
 // Handle handles a staking message
 func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.StateManager) (*action.Receipt, error) {
-	csm, err := NewCandidateStateManager(sm)
+	center, err := getCandCenter(sm)
```
dustinxie, line 207: we must have view loaded here, so call get()

action/protocol/staking/candidate_statemanager.go
```diff
@@ -94,6 +94,56 @@ func (csm *candSM) Upsert(d *Candidate) error {
...
+
+func getOrCreateCandCenter(sr protocol.StateReader) (CandidateCenter, error) {
+	c, err := getCandCenter(sr)
+	if err != nil {
+		if errors.Cause(err) == protocol.ErrNoName {
+			// the view does not exist yet, create it
+			cc, err := createCandCenter(sr)
+			return cc, err
+		}
+		return nil, err
```
CoderZhi, line 114: after changing the latest change, do we still need this logic or panic here?

action/protocol/staking/candidate_statemanager.go
```diff
@@ -94,6 +94,56 @@ func (csm *candSM) Upsert(d *Candidate) error {
...
+	}
+
+	center := NewCandidateCenter()
+	if err := center.SetDelta(all); err != nil {
+		return nil, err
+	}
+	if err := center.Commit(); err != nil {
+		return nil, err
+	}
```
CoderZhi, line 143: why commit here?

action/protocol/staking/candidate_statemanager.go
dustinxie: i think so, this is getOrCreateCandCenter(), so if get() did not return a view, need to create it.

however, for createGenesis() and Start(), we can use create() for sure

i added a TODO in other 2 places ActiveCandidates() and ReadState(), we can think it thru clearly whether could use create() when pass in sr for those 2

action/protocol/staking/candidate_statemanager.go
dustinxie: center := NewCandidateCenter() at line 137 is an empty struct

say there are 5 candidates at beginning, they are read out from sf/DB, and add to it as the Delta, Commit() writes the 5 candidates into the center, then WriteView save it to sf as the starting point

action/protocol/registry.go
```diff
@@ -92,16 +91,20 @@ func (r *Registry) all() []Protocol {
 }
 
 // StartAll starts all protocols which are startable
-func (r *Registry) StartAll(ctx context.Context) error {
+func (r *Registry) StartAll(ctx context.Context, sr StateReader, dock Dock) error {
```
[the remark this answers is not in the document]
dustinxie: fixed

action/protocol/poll/staking_command.go
dustinxie: fixed

action/protocol/staking/candidate_statemanager.go
CoderZhi: only two places use "create", start and read state if not the latest height.

action/protocol/staking/candidate_statemanager.go
CoderZhi: I got what you want to do
`return NewCandidateCenter(all)`

action/protocol/poll/staking_command.go
CoderZhi: v2 is a protocol, which could be a starter. As a container, sc does not know the details whether it does nothing or not. leave the code as it is.

action/protocol/staking/protocol.go
```diff
@@ -124,7 +131,12 @@ func (p *Protocol) CreateGenesisStates(
 		return nil
 	}
 
-	csm, err := p.createCandidateStateManager(sm)
+	center, err := getOrCreateCandCenter(sm)
```
CoderZhi: creategenesisstates is called after protocol started, so this should be get instead of create

dustinxie: fixed in latest push

PR #2097, opened 2020-04-04
clean up protocol interface 1

CoderZhi:
delete aborter

### DIFF
--- a/state/factory/workingset.go
+++ b/state/factory/workingset.go
@@ -186,7 +186,7 @@ func (ws *workingSet) Commit(ctx context.Context) error {
 	if err := ws.commitFunc(ws.height); err != nil {
 		return err
 	}
-	if !ws.dock.Dirty() {
+	if ws.dock.Dirty() {
 		if err := protocolCommit(ctx, ws); err != nil {
 			return err
 		}
--- a/state/factory/workingset.go
+++ b/state/factory/workingset.go
@@ -183,17 +183,16 @@ func (ws *workingSet) Revert(snapshot int) error {
 
 // Commit persists all changes in RunActions() into the DB
 func (ws *workingSet) Commit(ctx context.Context) error {
-	err := ws.commitFunc(ws.height)
-	if !ws.Dirty() {
+	if err := ws.commitFunc(ws.height); err != nil {
 		return err
 	}
-	if err == nil {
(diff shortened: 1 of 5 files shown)


state/factory/workingset.go
```diff
@@ -183,17 +183,16 @@ func (ws *workingSet) Revert(snapshot int) error {
...
-	err := ws.commitFunc(ws.height)
-	if !ws.Dirty() {
+	if err := ws.commitFunc(ws.height); err != nil {
 		return err
 	}
-	if err == nil {
-		protocolCommit(ctx, ws)
-	} else {
-		protocolAbort(ctx, ws)
+	if !ws.dock.Dirty() {
```
yutongp, line 189: i am confused here.
i thought it should be if dirty() commit

state/factory/workingset.go
```diff
@@ -183,17 +183,16 @@ func (ws *workingSet) Revert(snapshot int) error {
 
 // Commit persists all changes in RunActions() into the DB
 func (ws *workingSet) Commit(ctx context.Context) error {
-	err := ws.commitFunc(ws.height)
-	if !ws.Dirty() {
+	if err := ws.commitFunc(ws.height); err != nil {
```
yutongp, line 186: also seems in previous logic, ws.Reset is always called. but now it doesn't call ws.Reset on error

state/factory/workingset.go
```diff
@@ -183,17 +183,16 @@ func (ws *workingSet) Revert(snapshot int) error {
...
 		return err
 	}
-	if err == nil {
-		protocolCommit(ctx, ws)
-	} else {
-		protocolAbort(ctx, ws)
+	if !ws.dock.Dirty() {
+		if err := protocolCommit(ctx, ws); err != nil {
```
yutongp, line 190: also seems in previous logic, ws.Reset is called. but now it doesn't call ws.Reset on error

state/factory/workingset.go
CoderZhi: this is a bug, fixing it.

state/factory/workingset.go
CoderZhi: We actually do not need to call "Reset". The data stored in ws.dock should be similar to ws itself, which doesn't need to be explicitly cleaned up.

PR #2103, opened 2020-04-06
clean up protocol interface 2

CoderZhi:
Move validate into Handle and delete Validate API.


state/factory/workingset.go
```diff
@@ -377,27 +363,14 @@ func (ws *workingSet) pickAndRunActions(
...
 		if err != nil {
-			if errors.Cause(err) == action.ErrHitGasLimit {
-				// hit block gas limit, we should not process actions belong to this user anymore since we
-				// need monotonically increasing nonce. But we can continue processing other actions
-				// that belong other users
-				actionIterator.PopAccount()
-				continue
-			}
-			return nil, nil, errors.Wrapf(err, "Failed to update state changes for selp %x", nextAction.Hash())
+			actionIterator.PopAccount()
```
dustinxie, line 372: why we remove this check?
would this cause potential backward compatibility?

action/protocol/poll/staking_command.go
```diff
@@ -83,22 +83,21 @@ func (sc *stakingCommand) CreatePostSystemActions(ctx context.Context, sr protoc
 }
 
 func (sc *stakingCommand) Handle(ctx context.Context, act action.Action, sm protocol.StateManager) (*action.Receipt, error) {
+	// no height here,  v1 v2 has the same validate method, so directly use common one
```
dustinxie, line 86: on one hand: sc is a container, sc does/should not know the details of v1, v2

on the other hand: code should finally represent/obey to the actual situation, not bound by above

i am with the 2nd thought, this is a good comment that clearly explains the actual case, and give good readability

action/protocol/protocol.go
```diff
@@ -49,21 +48,11 @@ type Committer interface {
 	Commit(context.Context, StateManager) error
 }
 
-// Aborter performs abort action of the protocol
-type Aborter interface {
-	Abort(context.Context, StateManager) error
-}
-
```
dustinxie, line 56: this is for protocol to perform necessary actions (if there's any) in case attempted commit block failed, don't think we need to remove it

PR #2107, opened 2020-04-07
Cache mp tree

CoderZhi:
1. move implementation to merklepatriciatree
2. apply the new trie implementation for historical trie. In previous implementation, the most consuming part of trie is DB operations. The reason we had a large number of DB operations is that each trie update needs multiple DB operations. For example, to update the value of a leaf node whose depth is 3, it used to need at least 6 DB operations (1 delete and 1 insert for each node in the path). If there were multiple trie updates, we can eliminate some of the DB operations because they will cancel out each other. For example, if we update the value of a leaf node twice, it used to need 12 DB operations (2 deletes and 2 inserts for each node in the path); but if we merge the DB operations of the two trie updates, the total DB operation number will be reduced to 6 (1 delete and 1 insert for each node in the path).

test plan: tested on mainnet in archive mode, catching up to 4 million


dustinxie:
as a general note, one reason that trie operation becomes slower as it grows is that the key to access DB is the hash of node, which is a random 20-bytes, causing an all-scattered pattern in the underlying storage (for bolt it is a B+ tree)

one improvement, that i think worth trying, is to speed-up the DB access by taking the common prefix path into the key, that is
key := append(prefix leading to this node, hash(node)[:10])  (more on why 10 below)

a simple example: say there are 2 nodes on path to account "1234567890abcdefghij", one is a branch B after "123456", the other is an extension E after "1234567890ab":
key of B = "123456" + hash(B)[:10]
key of E = "1234567890ab" + hash(E)[:10]
and actually, all nodes under B will have common prefix "123456" in their key, and all nodes under E will have common prefix "1234567890ab" in their key

the common prefix would cause a clustering effect, now all nodes under the same prefix are stored closer in the underlying storage, this will make access faster

now onto why take 10-byte from actual hash: 10-byte has 5-byte collision-resistance power, which is 2^40 = 1 trillion, so the nodes in the trie can be overwritten 1 trillion times (like a hot contract) before a collision is expected to happen, i think that should be good enough

dustinxie:
also, this change should not be too complicated, just need to construct

key := append(prefix, hash(node)[:10])

and return key (instead of the hash()) to parent node upon updating the nodes

db/trie/merklepatriciatree/branchnode.go
```diff
@@ -17,24 +17,21 @@ import (
 const radix = 256
 
 type branchNode struct {
-	mpt    *merklePatriciaTree
-	hashes map[byte][]byte
-	isRoot bool
-	ser    []byte
+	cacheNode
+	children map[byte]node
+	isRoot   bool
```
dustinxie, line 22: so saving the nodes instead of the hash
does this cache help speed up the trie operation, and how?

db/trie/merklepatriciatree/extensionnode.go
```diff
@@ -50,21 +50,18 @@ func (e *extensionNode) Delete(key keyType, offset uint8) (node, error) {
 		return nil, err
 	}
 	if newChild == nil {
-		return nil, e.mpt.deleteNode(e)
+		return nil, e.delete()
 	}
 	switch node := newChild.(type) {
 	case *extensionNode:
-		if err := e.mpt.deleteNode(e); err != nil {
```
dustinxie: why this can be removed for extension node?

db/trie/merklepatriciatree/branchnode.go
```diff
@@ -170,7 +176,7 @@ func (b *branchNode) child(key byte) (node, error) {
 	return c, nil
 }
 
-func (b *branchNode) updateChild(key byte, child node) (*branchNode, error) {
+func (b *branchNode) updateChild(key byte, child node, h bool) (node, error) {
```
dustinxie, line 200: h means isLeaft?

db/trie/merklepatriciatree/extensionnode.go
```diff
@@ -135,25 +138,35 @@ func (e *extensionNode) commonPrefixLength(key []byte) uint8 {
 	return commonPrefixLength(e.path, key)
 }
 
-func (e *extensionNode) updatePath(path []byte) (*extensionNode, error) {
+func (e *extensionNode) updatePath(path []byte, h bool) (node, error) {
```
dustinxie, line 160: h means isLeaf?

db/trie/merklepatriciatree/extensionnode.go
```diff
@@ -135,25 +138,35 @@ func (e *extensionNode) commonPrefixLength(key []byte) uint8 {
...
 		return nil, err
 	}
+	if h {
+		return hn, nil
+	}
 	return e, nil
 }
 
-func (e *extensionNode) updateChild(newChild node) (*extensionNode, error) {
+func (e *extensionNode) updateChild(newChild node, h bool) (node, error) {
```
dustinxie, line 179: h means isLeaf?

db/trie/merklepatriciatree/twolayertrie.go
```diff
@@ -70,81 +84,109 @@ func (tlt *twoLayerTrie) Start(ctx context.Context) error {
 	layerOne, err := New(
 		KVStoreOption(tlt.kvStore),
 		RootHashOption(rootHash),
+		AsyncOption(),
```
dustinxie, line 87: the idea is to enable async, so all the nodes hash/serialize will be done in one stroke at the time of Stop() and write into DB, is that right?

can we run TestPressure() in trie_test.go, and measure the improvement?

and 2 more questions:
1. why we need 2-layer trie, what is the first layer used for?
2. how do we keep the Root hash same as before?

dustinxie: also where is this "cancel out" done in your change?
i mean, if there are 2 DB op on the same key, the later 2nd value overwrites the 1st value

CoderZhi: > the idea is to enable async, so all the nodes hash/serialize will be done in one stroke at the time of Stop() and write into DB, is that right?
> 
> can we run TestPressure() in trie_test.go, and measure the improvement?
> 
> and 2 more questions:
> 
> 1. why we need 2-layer trie, what is the first layer used for?
> 2. how do we keep the Root hash same as before?

it is for historical trie, to store different namespaces. The first layer is for namespace, the second layer is for the values in the namespace.

dustinxie: we talked about earlier, can use a map to cache the namespaces, this can skip the first layer look-up and gain some speed-up

do you think can add into this PR, or a new PR?

CoderZhi: Since you have some new idea to improve it, would you mind implement this part?

dustinxie: ok

dustinxie: 2 unresolved question:

1. how to keep the Root hash (of layer-2 trie) same as before?

2. can you point out, where is this "cancel out" done in your change?
i mean, if there are 2 DB op on the same key, the later 2nd value overwrites the 1st value

db/trie/merklepatriciatree/branchnode.go
```diff
@@ -17,95 +17,88 @@ import (
...
-			continue
-		}
-		bnode.hashes[i] = mpt.nodeHash(n)
+	if children == nil {
+		children = make(map[byte]node)
 	}
-	if len(bnode.hashes) != 0 {
-		if err := mpt.putNode(bnode); err != nil {
+	bnode := &branchNode{cacheNode: cacheNode{mpt: mpt}, children: children}
+	bnode.cacheNode.serializable = bnode
```
koseoyoung, line 33: why cachenode needs to embed seralizable interface?

db/trie/merklepatriciatree/extensionnode.go
koseoyoung: node delete happens inside of the updatePath, updateChild

db/trie/merklepatriciatree/extensionnode.go
koseoyoung: i think h is a flag whether the return value is hashnode or extension node itself

Issue #2113, opened 2020-04-08
Refactor unstake/withdraw staking handlers


lizhefeng:
**What would you like to be added**:
Unstake should only be processable with non-autostake bucket. If a user wants to unstake an autostake bucket, he/she has to restake it first (set autostake= false + reset stakeStartTime), and then wait for the stakeDuration to decay. In withdrawStake handler, it is sufficient to just check the withdrawWaitingPeriod. 
**Why is this needed**:

PR #2115, opened 2020-04-09
change restake/unstake/withdraw handle logic regarding autostake

koseoyoung:
works on #2113 
For now, 
1. Unstake: can not unstake in case of auto-stake bucket/ compare duration with stake start time 
2. Restake: in case of switching off auto-stake, update stake start time 
+ in case of reducing duration, it's possible only when autoStake = Off && mature enough which means it's after stake start time + original duration 
2. withdraw: compare 14 days with unstake start time


action/protocol/staking/handlers.go
```diff
@@ -426,6 +430,10 @@ func (p *Protocol) handleRestake(ctx context.Context, act *action.Restake, csm C
 	prevWeightedVotes := p.calculateVoteWeight(bucket, csm.ContainsSelfStakingBucket(act.BucketIndex()))
 	// update bucket
 	bucket.StakedDuration = time.Duration(act.Duration()) * 24 * time.Hour
+	if bucket.AutoStake && !act.AutoStake() {
+		// in case of switching off autostake, update bucket StakeStartTime
+		bucket.StakeStartTime = blkCtx.BlockTimeStamp.UTC()
```
lizhefeng, line 435: blkCtx.BlockTimeStamp

action/protocol/staking/handlers.go
```diff
@@ -426,6 +430,10 @@ func (p *Protocol) handleRestake(ctx context.Context, act *action.Restake, csm C
 	prevWeightedVotes := p.calculateVoteWeight(bucket, csm.ContainsSelfStakingBucket(act.BucketIndex()))
```
dustinxie, line 430: move prevWeightdVotes to before using it

action/protocol/staking/handlers.go
```diff
@@ -426,6 +430,10 @@ func (p *Protocol) handleRestake(ctx context.Context, act *action.Restake, csm C
...
 	// update bucket
 	bucket.StakedDuration = time.Duration(act.Duration()) * 24 * time.Hour
```
dustinxie, line 432: in case of switching off autostake, stakedDuration should be same as before

otherwise, user can create autostake bucket for duration=30 days, switch it off with duration = 7 days, then i can withdraw after 7 + 14 days, while we expect user need to wait 30 + 14 days before withdraw

action/protocol/staking/handlers.go
koseoyoung: To where? you mean before here?```candidate.SubVote(prevWeightedVotes) ```
we can't, bucket info will be changed

action/protocol/staking/handlers.go
koseoyoung: we changed implementation that when user withdraw, they only wait for 14days. But after switching off autostake, user have to wait for duration and be able to unstake after it.

koseoyoung: Also for the changing duration, from what I understand, duration is not to make user to hold for the duration but to affect voting power, because we designed to update for user to restake and update duration "anytime". This issue is not only applied to this case which is updating autostake.  

For example, we can think that if user autoStake = false, he set up 10 days for duration. So we expect that he has to wait 10 days to unstake, but he can restake to change it to 5 days, and he will be able to unstake it after 5 days, but this is how we designed. Changing duration can affect his voting power tho.

dustinxie: i see, thx for the explanation

my question is: user can stake with duration = 3 month, autostake = ON, then enjoy a higher reward %
after 3 month, he restake to duration = 7 days, autostake = OFF, then he only need to wait 7 days to unstake, then he can withdraw (wait another 14 days)

while our design for autostake should be: to enjoy the higher reward, you need to wait the original duration at the time of unstake, that is, anytime he wanted to unstake, he need to wait 3 months from the time of unstake

but now if we change that to 7 days, he only needs to wait 7 days

koseoyoung:
After discussion, we decided that Restake action will fail, if it's trying to reduce duration when autostake ON status or OFF but maturity is not enough after comparing with original duration.

action/protocol/staking/handlers.go
```diff
@@ -425,7 +429,28 @@ func (p *Protocol) handleRestake(ctx context.Context, act *action.Restake, csm C
...
+		if !bucket.AutoStake && blkCtx.BlockTimeStamp.Before(bucket.StakeStartTime.Add(bucket.StakedDuration)) {
+			// if auto-stake off and maturity is not enough
+			err := fmt.Errorf("bucket is not ready to be able to reduce duration, current time %s, required time %s",
+				blkCtx.BlockTimeStamp, bucket.StakeStartTime.Add(bucket.StakedDuration))
+			log.L().Debug("Error when restaking bucket", zap.Error(err))
+			return p.settleAction(ctx, csm, uint64(iotextypes.ReceiptStatus_ErrReduceDurationBeforeMaturity), gasFee)
+		}
+	}
+	bucket.StakedDuration = actDuration
+	if bucket.AutoStake && !act.AutoStake() {
```
lizhefeng, line 450: Shouldn't we update the stakeStartTime no matter what?

action/protocol/staking/handlers.go
```diff
@@ -425,7 +429,28 @@ func (p *Protocol) handleRestake(ctx context.Context, act *action.Restake, csm C
...
+	actDuration := time.Duration(act.Duration()) * 24 * time.Hour
+	if bucket.StakedDuration.Hours() > actDuration.Hours() {
+		// in case of reducing the duration
+		if bucket.AutoStake {
+			// if auto-stake on, user can't reduce duration
+			err := errors.New("AutoStake should be disabled first in order to reduce duration")
+			log.L().Debug("Error when restaking bucket", zap.Error(err))
+			return p.settleAction(ctx, csm, uint64(iotextypes.ReceiptStatus_ErrInvalidBucketType), gasFee)
+		}
+		if !bucket.AutoStake && blkCtx.BlockTimeStamp.Before(bucket.StakeStartTime.Add(bucket.StakedDuration)) {
```
lizhefeng, line 441:
```
else if blkCtx.BlockTimeStamp.Before(bucket.StakeStartTime.Add(bucket.StakedDuration))
```

dustinxie, line 441: can remove !bucket.AutoStake b/c above if{} returns

action/protocol/staking/handlers.go
dustinxie: i have the same question, we should, right?

koseoyoung: It can be 3 cases, 
1. not change autostake (only changing duration) 
2. switch-on 
3. switch-off 
All of those case, we want to restart the staking time elapse? 

such as duration was 7 days, but he changed to 30 days at certain point. 
then he have to wait another 30 days is that right?

koseoyoung: make sense, address on latest commit

action/protocol/staking/handlers.go
koseoyoung: address on latest commit

koseoyoung: address on latest commit

PR #2117, opened 2020-04-10
refactor settleAction() in order to simplify CandidateStateManager in…

lzxm160:
…terface 2087

work on #2087

### DIFF
--- a/action/protocol/managers.go
+++ b/action/protocol/managers.go
@@ -76,16 +76,26 @@ type (
 		States(...StateOption) (uint64, state.Iterator, error)
 		ReadView(string) (interface{}, error)
 	}
+	
+	// StateWriter defines an interface to write stateDB
+	StateWriter interface {
+		PutState(interface{}, ...StateOption) (uint64, error)
+		DelState(...StateOption) (uint64, error)
+	}
 
+	// StateReadWriter defines an interface to read and write stateDB
+	StateReadWriter interface {
+		StateReader
+		StateWriter
+	}
+	
 	// StateManager defines the stateDB interface atop IoTeX blockchain
 	StateManager interface {
-		StateReader
+		StateReadWriter
 		// Accounts
 		Snapshot() int
 		Revert(int) error
 		// General state
-		PutState(interface{}, ...StateOption) (uint64, error)
-		DelState(...StateOption) (uint64, error)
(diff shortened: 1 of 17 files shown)


action/protocol/managers.go
```diff
@@ -76,16 +76,26 @@ type (
...
 		ReadView(string) (interface{}, error)
 	}
+	
+	// StateWriter defines an interface to write stateDB
+	StateWriter interface {
+		PutState(interface{}, ...StateOption) (uint64, error)
+		DelState(...StateOption) (uint64, error)
+	}
 
+	// StateReadWriter defines an interface to read and write stateDB
```
CoderZhi, line 86: bad naming

action/protocol/managers.go
```diff
@@ -76,16 +76,26 @@ type (
 		States(...StateOption) (uint64, state.Iterator, error)
 		ReadView(string) (interface{}, error)
 	}
+	
+	// StateWriter defines an interface to write stateDB
+	StateWriter interface {
```
CoderZhi, line 81: it is better to have
```
StateWriter interface {
    StateReader
    PutState...
    DelState...
}
```

but I don't see the value distinguishing the statewriter and statemanager.

action/protocol/staking/candidate_statemanager.go
```diff
@@ -30,6 +29,9 @@ type (
 		GetBySelfStakingIndex(uint64) *Candidate
 		Upsert(*Candidate) error
 		Commit() error
+		// bucket-related
+		protocol.StateReadWriter
+		Load(string, interface{}) error
```
CoderZhi, line 34: if this is the purpose you create new interface, I think you are not on the right track. Our goal is to remove protocol.StateXXX from this interface definition.

lzxm160: There's some interfaces cannot be removed in protocol.StateReadWriter,like State,PutState,DelState,used in handleXXX

action/protocol/managers.go
lzxm160: sure,I can change it to this

lzxm160:
so many state manager's interface to add,like putstate,delstate,and depositGas used protocol.StateManager,LoadAccount used protocol.StateReader

PR #2119, opened 2020-04-13
[staking] remove state reader from struct

dustinxie:


action/protocol/staking/protocol.go
```diff
@@ -272,7 +270,6 @@ func (p *Protocol) handle(ctx context.Context, act action.Action, csm CandidateS
 
 // ActiveCandidates returns all active candidates in candidate center
 func (p *Protocol) ActiveCandidates(ctx context.Context, sr protocol.StateReader, height uint64) (state.CandidateList, error) {
-	// TODO: should use createCandCenter()?
```
CoderZhi, line 275: this shouldn't be "create"

dustinxie: right

PR #2121, opened 2020-04-14
[db] fix concurrent mapwrite in indexer

dustinxie:


blockindex/indexer.go
```diff
@@ -179,16 +179,7 @@ func (x *blockIndexer) DeleteTipBlock(blk *block.Block) error {
...
-	index, err := db.GetCountingIndex(x.kvStore, totalBlocksBucket)
-	if err != nil {
-		if errors.Cause(err) == db.ErrBucketNotExist || errors.Cause(err) == db.ErrNotExist {
-			// counting index does not exist yet
-			return 0, nil
-		}
-		return 0, err
-	}
-	return index.Size() - 1, nil
+	return x.tbk.Size() - 1, nil
```
dustinxie, line 182: x.tbk is the index below, and is done in Start()

blockindex/indexer.go
```diff
@@ -251,12 +242,7 @@ func (x *blockIndexer) GetActionIndex(h []byte) (*actionIndex, error) {
 func (x *blockIndexer) GetTotalActions() (uint64, error) {
 	x.mutex.RLock()
 	defer x.mutex.RUnlock()
-
-	total, err := db.GetCountingIndex(x.kvStore, totalActionsBucket)
-	if err != nil {
-		return 0, err
-	}
-	return total.Size(), nil
+	return x.tac.Size(), nil
```
dustinxie, line 245: x.tac is the "total" below, and is done in Start()

db/counting_index.go
```diff
@@ -100,13 +100,6 @@ func GetCountingIndex(kv KVStore, name []byte) (CountingIndex, error) {
 	if errors.Cause(err) == ErrNotExist || total == nil {
 		return nil, errors.Wrapf(err, "counting index 0x%x doesn't exist", name)
 	}
-	if kvFillPercent, ok := kv.(KVStoreWithBucketFillPercent); ok {
```
dustinxie, line 103: GetCountingIndex() only returns the index for read, so no need to set the fill percent, which is only for write operation

db/db_bolt.go
```diff
@@ -34,7 +35,7 @@ func NewBoltDB(cfg config.DB) KVStoreWithBucketFillPercent {
 		db:          nil,
 		path:        cfg.DbPath,
 		config:      cfg,
-		fillPercent: make(map[string]float64),
+		fillPercent: cache.NewThreadSafeLruCache(256),
```
dustinxie, line 38: choice of 256 depends on the number of diff address involved in a block
given our current block traffic, 256 should be way enough

PR #2128, opened 2020-04-17
delete abstract action

CoderZhi:
1. Remove actionPayload and AbstractAction. Nonce/GasLimit/GasPrice and other informations will be passed via ActionCtx
2. Change action.Cost logic to return the cost, excluding gas. Gas will be calculated in envelope level. 

even though it seems 91 files have been changed, but the real change is limited to 15 actions, envelope and sealed envelope. Other than that, 38 files are fixing unit tests, 14 files are ioctl cmd, 7 files are protocol related, 6 files are tools in which actions are assembled.
 
Further changes may be needed, will leave them to future.


action/envelope.go
```diff
@@ -120,109 +131,115 @@ func (elp *Envelope) LoadProto(pbAct *iotextypes.ActionCore) error {
...
+	builder.SetVersion(pbAct.GetVersion())
+	builder.SetNonce(pbAct.GetNonce())
+	builder.SetGasLimit(pbAct.GetGasLimit())
+	if len(pbAct.GetGasPrice()) != 0 {
+		gasPrice, ok := new(big.Int).SetString(pbAct.GetGasPrice(), 10)
+		if !ok {
+			return errors.Errorf("invalid gas price '%s'", pbAct.GetGasPrice())
+		}
+		builder.SetGasPrice(gasPrice)
+	}
```
dustinxie, line 144: do we need else{} to return error? b/c pb does not have valid price

action/stakerestake_test.go
```diff
@@ -37,7 +32,7 @@ func TestRestake(t *testing.T) {
 	require.Equal(uint64(10700), gas)
 	cost, err := stake.Cost()
 	require.NoError(err)
-	require.Equal("107000", cost.Text(10))
+	require.Equal("0", cost.Text(10))
```
dustinxie, line 35: does this mean stake.Cost() change to 0 now? is Cost() used in calculating gas charge?

action/transfer.go
```diff
@@ -77,12 +61,12 @@ func (tsf *Transfer) Destination() string { return tsf.recipient }
 
 // TotalSize returns the total size of this Transfer
 func (tsf *Transfer) TotalSize() uint32 {
-	size := tsf.BasicActionSize()
+	size := uint32(len(tsf.payload))
```
dustinxie, line 64: BasicActionSize() is not 0, TotalSize() has changed as a result.
if TotalSize() is used in calculating gas charge, might be an issue

action/action.go
```diff
@@ -24,15 +24,8 @@ var (
...
-	SetEnvelopeContext(SealedEnvelope)
-	SanityCheck() error
-}
-
-type actionPayload interface {
-	Serialize() []byte
 	Cost() (*big.Int, error)
 	IntrinsicGas() (uint64, error)
-	SetEnvelopeContext(SealedEnvelope)
 	SanityCheck() error
```
dustinxie, line 29: do we still need SanityCheck()? seems they all return nil now

action/action.go
```diff
@@ -24,15 +24,8 @@ var (
 
 // Action is the action can be Executed in protocols. The method is added to avoid mistakenly used empty interface as action.
 type Action interface {
-	SetEnvelopeContext(SealedEnvelope)
-	SanityCheck() error
-}
-
-type actionPayload interface {
-	Serialize() []byte
```
dustinxie, line 32: we can keep Serialize()?

dustinxie: i mean, keep in Action interface

dustinxie:
on the other hand, AbstractAction has everything needed in ActionCtx, except for Caller.

we can let AbstractAction expose Caller(), and set Caller (and also Nonce, GasPrice, Hash) at the time of envelope.LoadProto(), then we can remove ActionCtx from the code, and directly use Action instead

action/action.go
```diff
@@ -24,15 +24,8 @@ var (
 
 // Action is the action can be Executed in protocols. The method is added to avoid mistakenly used empty interface as action.
 type Action interface {
```
yutongp, line 26: the reason why `actionPyaload` is internal type is to give the authority to create action payload to this package, so other packages cannot create actionPyaload

yutongp: and the reason `Action` exist, is because i want when other packages take in an action as input, like 
```
func (p *Protocol) handleTransfer(ctx context.Context, act action.Action, sm protocol.StateManager) (*action.Receipt, error)
```
It takes a concrete type as function param instead use `interface{}` to confuse user.  And because `actionPyaload` is internal type because the reason above, i defined `Action` just to avoid using `interface{}` so people can make mistakes.

PR #2131, opened 2020-04-17
fix flooding of error messages

CoderZhi:
to stop the flooding of error messages caused by the same problem.

### DIFF
--- a/consensus/scheme/rolldpos/roundcalculator.go
+++ b/consensus/scheme/rolldpos/roundcalculator.go
@@ -146,8 +146,11 @@ func (c *roundCalculator) roundInfo(
 		}
 	}
 	if !lastBlockTime.Before(now) {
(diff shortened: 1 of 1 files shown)


consensus/scheme/rolldpos/roundcalculator.go
```diff
@@ -146,8 +146,11 @@ func (c *roundCalculator) roundInfo(
 		}
 	}
 	if !lastBlockTime.Before(now) {
+		// TODO: if this is the case, the system time is far behind the time of other nodes.
+		// the code below is just to mute the warning, but "panic" may be a better choice.
+		time.Sleep(lastBlockTime.Sub(now))
```
dustinxie, line 151: roundInfo() is called by couple of places in consensus
i think we should test/verify the delay of return err due to this Sleep() would not affect the normal functionality, like consensus can still be reached, full-node can correctly sync

PR #2140, opened 2020-04-27
add back validator interface

CoderZhi:
1. Add back action validator interface
2. Move intrinsic gas check into working set (except EVM, which is too complicated to this PR)

### DIFF
--- a/action/protocol/staking/handlers.go
+++ b/action/protocol/staking/handlers.go
@@ -581,7 +581,8 @@ func (p *Protocol) handleCandidateUpdate(ctx context.Context, act *action.Candid
 	}
 
 	if err := csm.Upsert(c); err != nil {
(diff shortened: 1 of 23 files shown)


action/protocol/staking/handlers.go
```diff
@@ -581,7 +581,8 @@ func (p *Protocol) handleCandidateUpdate(ctx context.Context, act *action.Candid
 	}
 
 	if err := csm.Upsert(c); err != nil {
-		return nil, errors.Wrapf(err, "failed to put state of candidate %s", c.Owner.String())
+		log.L().Debug("failed to put state of candidate", zap.Error(err))
+		return p.settleAction(ctx, csm, uint64(iotextypes.ReceiptStatus_ErrCandidateConflict), gasFee)
```
CoderZhi, line 585: https://github.com/iotexproject/iotex-core/issues/2141
More cases in this file need to be addressed.

PR #2143, opened 2020-04-28
extend foundation bonus from fairbank to 1 yr later

koseoyoung:


koseoyoung:
@yutongp How do you think about renaming FoundationBonusLastEpoch in genesis.go to FoundationBonusOneYearEpoch?

action/protocol/rewarding/reward.go
```diff
@@ -215,7 +215,8 @@ func (p *Protocol) GrantEpochReward(
 	}
 
 	// Reward additional bootstrap bonus
-	if epochNum <= a.foundationBonusLastEpoch {
+	fairBankEpochNum := rp.GetEpochNum(hu.FairbankBlockHeight()) // extend foundation bonus from fairbank to fairbank + 1 year
+	if epochNum <= a.foundationBonusLastEpoch || (epochNum >= fairBankEpochNum && epochNum <= fairBankEpochNum+a.foundationBonusLastEpoch) {
```
dustinxie, line 219: should calculate numEpochIn1Year = xxx
and <= fairBankEpochNum+numEpochIn1Year

a.foundationBonusLastEpoch just happens to be equal to 1 year now, but should not count on that

koseoyoung: yes, ```numEpochIn1Year``` should be in genesis? or const variable?

yutongp: i think it is more clear to have a new start and new end to indicate this is the second wave of foundation bonus.
I think an end time is better than a start time + a last period.  the reason is if we want to extend that end time, it is easier. if we use `numEpochIn1Year`, we may introduce a bug by changing that

koseoyoung: yes, I changed to add that 2 fields(start, end epoch) in admin struct in the last commit. 
But still we need to specify ```numEpochIn1Year``` to set up end epoch value, however in the future, we always can change it I guess

yutongp: we can calculate end epoch offline

action/protocol/rewarding/admin.go
```diff
@@ -26,6 +32,8 @@ type admin struct {
 	foundationBonus                *big.Int
 	numDelegatesForFoundationBonus uint64
 	foundationBonusLastEpoch       uint64
+	foundationBonusP2StartEpoch    uint64
+	foundationBonusP2EndEpoch      uint64
```
dustinxie, line 36: need to also put these 2 into protobuf

action/protocol/rewarding/admin.go
```diff
@@ -130,6 +138,12 @@ func (p *Protocol) CreateGenesisStates(
 	numDelegatesForFoundationBonus := bcCtx.Genesis.NumDelegatesForFoundationBonus
 	foundationBonusLastEpoch := bcCtx.Genesis.FoundationBonusLastEpoch
 	productivityThreshold := bcCtx.Genesis.ProductivityThreshold
+	hu := config.NewHeightUpgrade(&bcCtx.Genesis)
+	rp := rolldpos.FindProtocol(protocol.MustGetRegistry(ctx))
+	foundationBonusP2StartEpoch := uint64(0)
+	if rp != nil {
+		foundationBonusP2StartEpoch = rp.GetEpochNum(hu.FairbankBlockHeight()) // extend foundation bonus from fairbank to fairbank + 1 year
+	}
```
dustinxie, line 146: this won't work? since you are not starting a new chain from height 0

action/protocol/rewarding/admin.go
dustinxie: also need to add these 2 into genesis, same as foundationBonusLastEpoch

action/protocol/rewarding/reward.go
```diff
@@ -215,7 +215,7 @@ func (p *Protocol) GrantEpochReward(
 	}
 
 	// Reward additional bootstrap bonus
-	if epochNum <= a.foundationBonusLastEpoch {
+	if epochNum <= a.foundationBonusLastEpoch || (epochNum >= a.foundationBonusP2StartEpoch && epochNum <= a.foundationBonusP2EndEpoch) {
```
dustinxie, line 218: somewhere in reward protocol, we need to do
if epoch is Fairbank {
    a.foundationBonusP2StartEpoch = genesis.xxx
    a.foundationBonusP2EndEpoch = genesis.yyy
    p.state(sm, adminKey, &a)
}
that is to write the 2 new values at Fairbank height

then this logic can follow

dustinxie: or better off
if epoch == Fairbank - 1
to write it in advance

then at Fairbank, this is ready to hit

action/protocol/rewarding/admin.go
koseoyoung: sure!

koseoyoung:
we can't add any field to admin struct becuase of backward incompatibility, so just added to genesis

action/protocol/rewarding/protocol_test.go
```diff
@@ -180,7 +183,7 @@ func testProtocol(t *testing.T, test func(*testing.T, context.Context, protocol.
 		},
 	)
 	ctx = protocol.WithBlockchainCtx(
-		ctx,
+		protocol.WithRegistry(ctx, registry),
```
dustinxie, line 186: we don't need this?

PR #2147, opened 2020-04-30
move actpool into factory package

CoderZhi:
Delete actpool dependency in blockchain, consensus, and add it to state/factory.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -16,7 +16,6 @@ import (
 	"github.com/iotexproject/iotex-core/action/protocol"
 	"github.com/iotexproject/iotex-core/action/protocol/poll"
 	rp "github.com/iotexproject/iotex-core/action/protocol/rolldpos"
-	"github.com/iotexproject/iotex-core/actpool"
 	"github.com/iotexproject/iotex-core/blockchain"
 	"github.com/iotexproject/iotex-core/blockchain/block"
 	"github.com/iotexproject/iotex-core/config"
@@ -85,7 +84,6 @@ func NewConsensus(
 	cfg config.Config,
 	bc blockchain.Blockchain,
 	sf factory.Factory,
-	ap actpool.ActPool,
 	opts ...Option,
 ) (Consensus, error) {
 	var ops optionParams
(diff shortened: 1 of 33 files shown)


state/factory/workingset.go
```diff
@@ -373,48 +375,49 @@ func (ws *workingSet) pickAndRunActions(
...
-		// TODO: if the action is not handled by any handler, receipt will be nil.
-		//  in this case, should this action be appended?
-		if receipt != nil {
-			blkCtx.GasLimit -= receipt.GasConsumed
-			ctx = protocol.WithBlockCtx(ctx, blkCtx)
-			receipts = append(receipts, receipt)
-		}
-		executedActions = append(executedActions, nextAction)
+			if err != nil {
+				ap.DeleteAction(nextAction)
```
CoderZhi, line 395: this is the place where an invalid action is deleted during preparing block.

consensus/consensus.go
```diff
@@ -167,8 +162,6 @@ func NewConsensus(
 			if err != nil {
 				log.Logger("consensus").Info("Failed to commit the block.", zap.Error(err), zap.Uint64("height", blk.Height()))
 			}
-			// Remove transfers in this block from ActPool and reset ActPool state
-			ap.Reset()
```
dustinxie, line 171: this is removed, does not to be called somewhere else?

state/factory/workingset.go
dustinxie: would suggest pass a callback func DeleteAction() into factory, if this is the fix needed, to unblock the FB release

then we can do a better refactor of actpool as we planned, passing actpool into factory (and create a wrapper struct minter{f: f, ap: ap}) feels more like a hack for me

consensus/consensus.go
dustinxie: i mean, i only see it being called in _test.go

state/factory/workingset.go
yutongp: maybe delete in batch to avoid lock and unlock?

yutongp: > would suggest pass a callback func DeleteAction() into factory, if this is the fix needed, to unblock the FB release
> 
> then we can do a better refactor of actpool as we planned, passing actpool into factory (and create a wrapper struct minter{f: f, ap: ap}) feels more like a hack to me

would prefer pass into struct or interface instead of callback function

consensus/consensus.go
yutongp: I have same question as Dustin. Reset checks current account nonce and clears out processed actions.
i think maybe we can combine reset pool with delete invalid action

CoderZhi: checking

PR #2157, opened 2020-05-05
Store current epochmeta and refactor unproductive delegate calculation logic in poll/reward protocol 

koseoyoung:
works on #1943 

1. add epochmeta proto/struct/unit test 
2. define new height(Greenland height)
3. refactor poll/reward protocol by defining CalculateUnproductiveDelegates interface in poll protocol and use epochmeta to read current productivity


action/protocol/vote/epochmetapb/epochmeta.proto
```diff
@@ -0,0 +1,24 @@
...
+//      protoc --go_out=plugins=grpc:. *.proto
+
+syntax = "proto3";
+package epochmetapb;
+
+import "google/protobuf/timestamp.proto";
+
+message EpochMeta {
+  uint64 epochNumber = 1;
+  repeated BlockMeta blockMetas = 2;
```
CoderZhi, line 17: what's your plan of "EpochMeta"?
I can see that you may want to define more fields in the struct.I have two ideas:
1. shall they be grouped together? For example, you can have a list of length 720 * E to store the last 720 * E block meta, which including block producer, block time, block hash and etc. So, it may not be appropriate to be placed in epoch meta. I am not saying grouping by block is a better way, just be sure you already compared them
2. why do we have epochNumber field? we may want to store last E epoch meta info and they can be stored in an array (fixed size) with a pointer to current epoch. In this case, if we know the current epoch number and the pointer, we can tell the epoch number of each meta in the array. Moreover, in this way, we don't need to rotate (write a large field) or write multiple fields in the state db.

action/protocol/poll/slasher.go
```diff
@@ -548,3 +594,26 @@ func filterCandidates(
...
+// currentEpochProductivity returns the map of the number of blocks produced per delegate of current epoch
+func currentEpochProductivity(sr protocol.StateReader, start uint64, end uint64) (map[string]uint64, error) {
+	stats := make(map[string]uint64)
+	epochMeta, err := epochMetaFromDB(sr)
+	if err != nil {
+		return nil, err
+	}
+	for _, blockmeta := range epochMeta.BlockMetas {
+		if blockmeta.Height < start {
+			return nil, errors.New("block meta height is out of range")
```
CoderZhi, line 607: how could this happen?

action/protocol/poll/slasher.go
```diff
@@ -548,3 +594,26 @@ func filterCandidates(
 	}
 	return verifiedCandidates, nil
 }
+
+// currentEpochProductivity returns the map of the number of blocks produced per delegate of current epoch
+func currentEpochProductivity(sr protocol.StateReader, start uint64, end uint64) (map[string]uint64, error) {
```
CoderZhi, line 599: this is why grouping-by-block is a strategy we may consider

action/protocol/poll/slasher.go
```diff
@@ -548,3 +594,26 @@ func filterCandidates(
...
+	if err != nil {
+		return nil, err
+	}
+	for _, blockmeta := range epochMeta.BlockMetas {
+		if blockmeta.Height < start {
+			return nil, errors.New("block meta height is out of range")
+		}
+		if blockmeta.Height > end {
+			break
+		}
```
CoderZhi, line 611: if all block metas are of height less than end-1, should we throw an error?

action/protocol/vote/epochmetapb/epochmeta.proto
koseoyoung: I couldn't fully understand about what you are saying. Just to confirm my understanding, 
1. "Group-by-block" means not having epoch concept as epochmeta does, but just storing latest 720 block meta information, is that correct? And just retrieve current productivity from start to end, by comparing blockmeta's height, right? But then we have to shift to append latest block, right? Currently, we don't have to shift but only append or initialize to be a new struct at epoch start height. 
2. We only have to store current epoch meta for this now productivity usage. I can remove epoch number field because we don't need it. What do you mean by having a pointer to get an epoch number? You are saying is that for example, having a fixed size array(720) and having two array indexer(epoch start point and erase point) and replace the oldest block meta? 

If my understanding is correct, in order to satisfy current usage, because current productivity is calculated on epoch unit, just having only current epoch's block meta and initialize those at epoch start height don't need shifting and indexing, so it looks fine. I can get rid of epochNumber field in epochmeta (since we are not using), IMO it's okay to keep having current epoch's block meta for current usage. But do you think will there be other usages that require keeping latest 720 blocks? If so, I can change as "Group-by-block" way for future extension.

action/protocol/poll/slasher.go
koseoyoung: it shouldn't happen, just in case I wrote it

action/protocol/poll/slasher.go
koseoyoung: good point, I think we should return error at those case

blockchain/genesis/genesis.go
```diff
@@ -163,6 +164,8 @@ type (
 		EasterBlockHeight uint64 `yaml:"easterHeight"`
 		// FairbankBlockHeight is the start height to switch to native staking V2
 		FairbankBlockHeight uint64 `yaml:"fairbankHeight"`
+		// GoldengateBlockHeight is the start height of storing current epoch meta and TODO
+		GoldengateBlockHeight uint64 `yaml:"goldengateHeight"`
```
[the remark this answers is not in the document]
koseoyoung: I will change to Greenland

action/protocol/vote/epochmetapb/epochmeta.proto
CoderZhi: thanks for the details, now I have a full picture what you want to do. Initially, I thought you want to store multiple epoch metas.

about block meta part, storing metas of the latest 720 blocks by 720 state db keys.
  pros:
      a. This will minimize writes to state db. To tell what's the next block meta to write to, we need an extra index in state db, NEXT_BLOCK_META_INDEX. To append, we just need to write the block meta into state db by key 'BLOCK_META_KEY_' + value_in_state_db(NEXT_BLOCK_META_INDEX).
      b. Benefit other usage of block meta, say in consensus, we need to get block time and etc.
  cons:
      a. Significantly increase the read cost if we cannot read in batch.

blockchain/genesis/genesis.go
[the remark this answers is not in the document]
dustinxie: +1 for Greenland

action/protocol/poll/slasher.go
```diff
@@ -440,23 +445,45 @@ func (sh *Slasher) CalculateProbationList(
...
-		return nil, err
-	}
-	// The current block is not included, so add it
-	numBlks++
-	if _, ok := produce[blkCtx.Producer.String()]; ok {
-		produce[blkCtx.Producer.String()]++
+	hu := config.NewHeightUpgrade(&bcCtx.Genesis)
+	var numBlks uint64
+	var produce map[string]uint64
+	if hu.IsPre(config.Goldengate, blkCtx.BlockHeight) {
```
dustinxie, line 460: can you explain a little bit, what is the new change for this new height?
like calculating productivity in a different way?

koseoyoung: According to what I discussed with zhi offline, we are thinking about storing latest 720 blockmeta, so that in future it can be used for implementing lightnode as well (not using chainDB but only using state DB). Currently it can replace calculating current productivity logic. So this new change is mainly about writing latest block meta(producer, block time, so on) into stateDB and it will be used in productivity calculation(current epoch) for now, but in future it may have more use cases.

action/protocol/poll/slasher.go
koseoyoung: fixed in latest commit

koseoyoung:
Latest commit changes from storing epoch meta to storing blockmeta 
BlockMeta's key is prefix + blockheight % 720 

1. need to figure out namespace issue (because system namespace is shared with candidate/probation) 
2. need to define ```NumOfBlocksByEpoch``` config variable or can have ```MaxBlockMetasCount ```

action/protocol/poll/slasher.go
```diff
@@ -474,6 +501,12 @@ func (sh *Slasher) calculateUnproductiveDelegatesByEpoch(ctx context.Context, sr
 	return unqualified, nil
 }
 
+func (sh *Slasher) updateCurrentEpochMeta(ctx context.Context, sm protocol.StateManager) error {
```
CoderZhi, line 504: updateCurrentBlockMeta

action/protocol/poll/util.go
```diff
@@ -304,3 +306,47 @@ func shiftProbationList(sm protocol.StateManager) (uint64, error) {
...
+	return err
+}
+
+// allBlockMetasFromDB returns all latest block meta structs
+func allBlockMetasFromDB(sr protocol.StateReader) ([]*BlockMeta, error) {
+	maxKey := []byte{_blockmeta + 1}
+	stateHeight, iter, err := sr.States(
+		protocol.NamespaceOption(protocol.SystemNamespace),
+		protocol.FilterOption(func(k, v []byte) bool {
+			return bytes.HasPrefix(k, []byte{_blockmeta})
```
CoderZhi, line 327: why is the prefix []byte{_blockmeta} unique?

action/protocol/vote/candidatesutil/candidatesutil.go
```diff
@@ -41,6 +41,9 @@ const NxtProbationKey = "NextKickoutKey."
 // UnproductiveDelegateKey is the key of unproductive Delegate struct
 const UnproductiveDelegateKey = "UnproductiveDelegateKey."
 
+// EpochMetaKey is the key of epoch meta struct
+const EpochMetaKey = "EpochMetaKey."
+
```
CoderZhi, line 46: in use?

action/protocol/poll/util.go
koseoyoung: I fixed this issue by having unique prefix by hashing blockMetaPrefix as other creating keys. Prefix == Hash(blockMetaPrefix), so it has no overlap with other keys

PR #2158, opened 2020-05-05
[staking] maintain receipt.Log backward compatible

dustinxie:


action/protocol/staking/protocol.go
```diff
@@ -227,39 +232,61 @@ func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.St
...
+		topics, preFairbankLog, err = p.handleCandidateUpdate(ctx, act, csm)
 	default:
 		return nil, nil
 	}
 
+	blkCtx := protocol.MustGetBlockCtx(ctx)
+	postFairbankMigration := true
+	if p.hu.FairbankBlockHeight() >= 12*720 {
+		postFairbankMigration = blkCtx.BlockHeight >= p.hu.FairbankBlockHeight()-12*720
+	}
```
dustinxie, line 268: here set migration time = 12 hours before actual fairbank block time

so all the candidates/buckets we import during migration will have correct topics in the log

wanted to know what you think about this, or maybe 8 hours is good enough

action/protocol/staking/handlers.go
```diff
@@ -67,85 +67,92 @@ func (h *handleError) ReceiptStatus() uint64 {
 }
 
 func (p *Protocol) handleCreateStake(ctx context.Context, act *action.CreateStake, csm CandidateStateManager,
-) (action.Topics, error) {
-	topics := action.Topics{hash.BytesToHash256([]byte(HandleCreateStake))}
+) (*receiptLog, error) {
 	actionCtx := protocol.MustGetActionCtx(ctx)
 	blkCtx := protocol.MustGetBlockCtx(ctx)
+	postFairbankMigration := blkCtx.BlockHeight >= p.hu.FbkMigrationBlockHeight()
+	log := newReceiptLog(p.addr.String(), HandleCreateStake, postFairbankMigration)
```
yutongp, line 74: maybe put 
```
postFairbankMigration := blkCtx.BlockHeight >= p.hu.FbkMigrationBlockHeight()
```
in `newReceiptLog`?

dustinxie: then need to pass height and FbkMigrationBlockHeight() into newReceiptLog(), feel that's too much

receiptLog only need  pre/post, don't need the actual height

PR #2170, opened 2020-05-13
add getBucketByIndex API and ioctl commands to get vote bucket data

koseoyoung:
works on #2162 
add ```ReadStakingDataMethod_BUCKET_BY_INDEX ``` method and bucket/bucketlist ioctl commands

PR #2175, opened 2020-05-14
add getBucketsByIndexes API and ioctl commands to get vote bucket data

yutongp:
based on #2170

### DIFF
--- a/ioctl/cmd/bc/bucket.go
+++ b/ioctl/cmd/bc/bucket.go
@@ -8,6 +8,7 @@ package bc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 	"strconv"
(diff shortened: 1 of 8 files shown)


ioctl/cmd/bc/bucket.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+// bcBucketCmd represents the bc Bucket command
+var bcBucketCmd = &cobra.Command{
+	Use:   config.TranslateInLang(bcBucketUses, config.UILanguage),
+	Short: config.TranslateInLang(bcBucketCmdShorts, config.UILanguage),
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		err := getBucket(args[0])
```
dustinxie, line 51: we'll need to parse out a list of index from args[i]
like 47 56 77 86 100 372 7328 72727

ioctl/cmd/bc/bucket.go
```diff
@@ -0,0 +1,196 @@
...
+		config.English: "bucket [BUCKET_INDEX]",
+		config.Chinese: "bucket [票索引]",
+	}
+)
+
+// bcBucketCmd represents the bc Bucket command
+var bcBucketCmd = &cobra.Command{
+	Use:   config.TranslateInLang(bcBucketUses, config.UILanguage),
+	Short: config.TranslateInLang(bcBucketCmdShorts, config.UILanguage),
+	Args:  cobra.ExactArgs(1),
```
dustinxie, line 48: should be a range (1, inf)

ioctl/cmd/bc/bucket.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	message := bucketMessage{
+		Node:   config.ReadConfig.Endpoint,
+		Bucket: bucket,
+	}
+	fmt.Println(message.String())
+	return nil
+}
+
+func getBucketByIndex(index uint64) (*iotextypes.VoteBucket, error) {
```
dustinxie, line 142: should take []uint64 (parsed from args[]) as input

ioctl/cmd/bc/bucket.go
yutongp: i am not going to change ioctl for now. it is a client side changes which is easier if we need to this. API is harder to change later on, so it is better to change it now

Issue #2173, opened 2020-05-14
add EstimateActionGasConsumption for NSV2 actions


yutongp:
currently EstimateActionGasConsumption only support 2 native actions: `transfer` and `execution`.

should also support native staking actions.

PR #2180, opened 2020-05-17
add EstimateActionGasConsumption for NSV2 actions 2173

lzxm160:
work on #2173


yutongp:
do you need to update iotex-proto dependency as well?

lzxm160:
> do you need to update iotex-proto dependency as well?

using latest master

PR #2221, opened 2020-06-10
[ioctl] fix generating payload for stake2

dustinxie:


ioctl/cmd/action/stake2Transfer.go
```diff
@@ -63,8 +63,10 @@ func stake2Transfer(args []string) error {
 
 	var payload []byte
 	if len(args) == 3 {
-		payload = make([]byte, 2*len([]byte(args[2])))
-		hex.Encode(payload, []byte(args[2]))
+		payload, err = hex.DecodeString(args[2])
```
dustinxie, line 66: payload entered on command line is in hex-encoded format

should use same processing as in ioctl/cmd/action/actiontransfer.go

Issue #2223, opened 2020-06-11
nightly-2 cluster failed to produce block at height 720


dustinxie:
### What version of iotex-core are you using (`ioctl version`)?
latest master

### What operating system and processor architecture are you using?
docker, nightly-2 cluster

### What did you do? If possible, provide a recipe for reproducing the error.
start nightly-2 cluster 

### What did you expect to see?
cluster continue to produce blocks

### What did you see instead?
it failed to produce new block at height 720, with error in log:

"failed to create block builder at new block height 720: failed to calculate current epoch upd 1: block metas from stateDB doesn't have enough data for given start, end height","errorVerbose":"block metas from stateDB doesn't have enough data for given start, end height\ngithub.com/iotexproject/iotex-core/action/protocol/poll.currentEpochProductivity

PR #2225, opened 2020-06-12
Fix a bug in reading current productivity (greenland) 

koseoyoung:
works on #2223 
The root cause was that it didn't have 720 block meta data in stateDB, but only has 1-719 data when we try to read it when calculating the unproductive list. (Probably since we didn't commit 720 into stateDB yet) 
So, change to use tip height as before.


action/protocol/poll/slasher.go
```diff
@@ -465,27 +465,24 @@ func (sh *Slasher) calculateUnproductiveDelegates(ctx context.Context, sr protoc
...
-		if _, ok := produce[blkCtx.Producer.String()]; ok {
-			produce[blkCtx.Producer.String()]++
-		} else {
-			produce[blkCtx.Producer.String()] = 1
-		}
 	} else {
 		numBlks, produce, err = rp.ProductivityByEpoch(
 			epochNum,
-			bcCtx.Tip.Height+1,
+			bcCtx.Tip.Height,
```
dustinxie, line 471: i saw v1.0.0 used Tip.Height, and change to +1 in https://github.com/iotexproject/iotex-core/commit/8bc77d3e24360bd8959eee49bae7ec2efc9975f9 20 days again

is it possible to add test case to cover this?

action/protocol/poll/slasher.go
```diff
@@ -465,27 +465,24 @@ func (sh *Slasher) calculateUnproductiveDelegates(ctx context.Context, sr protoc
...
 		)
-		if err != nil {
-			return nil, err
-		}
+	}
+	if err != nil {
+		return nil, err
+	}
+	// The current block is not included, so add it
+	numBlks++
```
dustinxie, line 481: so this applies to both if{} and else{}? 
can we also add test case to cover this?

koseoyoung:
@dustinxie In short, that commit is about changing productivity function to read recent 720 block metas from stateDB. Previously we get that block metas until 719 and calculate current one on the fly -- line 480 (because we are reading from blockdao which only contains committed data). After my PR, I assumed that on last epoch(720th), because we already put 720 block meta into stateDB, I thought we can read all data including 720 which is Tip + 1. However, in fact, I'm guessing because of stateDB commit is happening on batch, it still can't read 720 (Current), so this fix is about changing as before. (read until 719 and calculate on the fly 720th)

dustinxie:
i see, thx for the explanation
so actually we need only to choose the productivity func by Greenland, like:
if isPost(Greenland) {
    prodFunc = xxx
} else {
    prodFunc = sh.productivity
}
numBlks, produce, err := rp.ProductivityByEpoch(
			epochNum,
			bcCtx.Tip.Height,
			prodFunc,
		)

koseoyoung:
> i see, thx for the explanation
> so actually we need only to choose the productivity func by Greenland, like:
> if isPost(Greenland) {
> prodFunc = xxx
> } else {
> prodFunc = sh.productivity
> }
> numBlks, produce, err := rp.ProductivityByEpoch(
> epochNum,
> bcCtx.Tip.Height,
> prodFunc,
> )

yes!

Issue #2228, opened 2020-06-16
add ioctl did command to generate did document


lzxm160:

lzxm160:
done

PR #2229, opened 2020-06-17
add ioctl did command to generate did document 2228

lzxm160:
work on #2228 

using:
`ioctl did generate -s yy`
`Enter password #io1ladc7y0scpcs768xvusqn6r7thavafe8wyxypa:`
```
{
  "@context": "https://www.w3.org/ns/did/v1",
  "id": "did:io:0xff5b8f11f0c0710F68e6672009e87e5DFacea727",
  "authentication": [
    {
      "id": "did:io:0xff5b8f11f0c0710F68e6672009e87e5DFacea727#owner",
      "type": "EcdsaSecp256k1VerificationKey2019",
      "controller": "did:io:0xff5b8f11f0c0710F68e6672009e87e5DFacea727",
      "publicKeyHex": "033f8dc3224ed5d2212e0b89093044294c55a286a29417e20393a82a682517d6f1"
    }
  ]
}
```

### DIFF
--- a/ioctl/cmd/action/action.go
+++ b/ioctl/cmd/action/action.go
@@ -108,6 +108,7 @@ func decodeBytecode() ([]byte, error) {
 	return hex.DecodeString(util.TrimHexPrefix(bytecodeFlag.Value().(string)))
 }
 
+// Signer returns signer's address
 func Signer() (address string, err error) {
 	return util.GetAddress(signerFlag.Value().(string))
 }
--- a/ioctl/cmd/action/action.go
+++ b/ioctl/cmd/action/action.go
@@ -108,7 +108,7 @@ func decodeBytecode() ([]byte, error) {
 	return hex.DecodeString(util.TrimHexPrefix(bytecodeFlag.Value().(string)))
 }
 
-func signer() (address string, err error) {
+func Signer() (address string, err error) {
 	return util.GetAddress(signerFlag.Value().(string))
 }
 
@@ -311,7 +311,7 @@ func Execute(contract string, amount *big.Int, bytecode []byte) error {
 	if err != nil {
 		return output.NewError(0, "failed to get gas price", err)
 	}
-	signer, err := signer()
+	signer, err := Signer()
 	if err != nil {
 		return output.NewError(output.AddressError, "failed to get signer address", err)
 	}
@@ -343,7 +343,7 @@ func Execute(contract string, amount *big.Int, bytecode []byte) error {
 
 // Read reads smart contract on IoTeX blockchain
 func Read(contract address.Address, amount *big.Int, bytecode []byte) (string, error) {
-	caller, err := signer()
+	caller, err := Signer()
 	if err != nil {
 		caller = address.ZeroAddress
 	}
--- a/ioctl/cmd/did/diddoc.go
+++ b/ioctl/cmd/did/diddoc.go
@@ -1,0 +1,33 @@
+// Copyright (c) 2020 IoTeX Foundation
+// This is an alpha (internal) release and is not suitable for production. This source code is provided 'as is' and no
+// warranties are given as to title or non-infringement, merchantability or fitness for purpose and, to the extent
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package did
+
+const (
+	DIDPrefix   = "did:io:"
+	DIDAuthType = "Secp256k1VerificationKey2018"
+	DIDOwner    = "#owner"
+)
+
+type (
+	authenticationStruct struct {
+		Id           string `json:"id,omitempty"`
+		Type         string `json:"type,omitempty"`
+		Controller   string `json:"controller,omitempty"`
+		PublicKeyHex string `json:"publicKeyHex,omitempty"`
+	}
+	DIDDoc struct {
+		Context        string                 `json:"@context,omitempty"`
+		Id             string                 `json:"id,omitempty"`
+		Authentication []authenticationStruct `json:"authentication,omitempty"`
+	}
+)
+
+func newDIDDoc() *DIDDoc {
+	return &DIDDoc{
+		Context: "https://www.w3.org/ns/did/v1",
+	}
+}
--- a/ioctl/cmd/did/diddoc.go
+++ b/ioctl/cmd/did/diddoc.go
@@ -7,9 +7,12 @@
 package did
 
 const (
-	DIDPrefix   = "did:io:"
+	// DIDPrefix is the prefix string
+	DIDPrefix = "did:io:"
+	// DIDAuthType is the authentication type
 	DIDAuthType = "Secp256k1VerificationKey2018"
-	DIDOwner    = "#owner"
+	// DIDOwner is the suffix string
+	DIDOwner = "#owner"
 )
 
 type (
@@ -19,6 +22,7 @@ type (
 		Controller   string `json:"controller,omitempty"`
 		PublicKeyHex string `json:"publicKeyHex,omitempty"`
 	}
+	// DIDDoc is the DID document struct
 	DIDDoc struct {
 		Context        string                 `json:"@context,omitempty"`
 		Id             string                 `json:"id,omitempty"`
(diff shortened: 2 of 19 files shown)


ioctl/cmd/did/diddoc.go
```diff
@@ -0,0 +1,37 @@
...
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package did
+
+const (
+	// DIDPrefix is the prefix string
+	DIDPrefix = "did:io:"
+	// DIDAuthType is the authentication type
+	DIDAuthType = "Secp256k1VerificationKey2018"
```
dustinxie, line 13: from https://www.w3.org/ns/did/v1
it should be: EcdsaSecp256k1VerificationKey2019?

lzxm160: > from https://www.w3.org/ns/did/v1
> it should be: EcdsaSecp256k1VerificationKey2019?
I am not sure,because https://www.w3.org/TR/did-core/ only show Secp256k1VerificationKey2018,I didn't find any more definitions.
In here,https://uniresolver.io/,did:ethr:0xE6Fe788d8ca214A080b0f6aC7F48480b2AEfa9a6 also used Secp256k1VerificationKey2018

yutongp: @xinxin-crypto can you take a look?

ioctl/cmd/did/diddoc.go
```diff
@@ -0,0 +1,37 @@
...
+	// DIDOwner is the suffix string
+	DIDOwner = "#owner"
+)
+
+type (
+	authenticationStruct struct {
+		ID           string `json:"id,omitempty"`
+		Type         string `json:"type,omitempty"`
+		Controller   string `json:"controller,omitempty"`
+		PublicKeyHex string `json:"publicKeyHex,omitempty"`
```
yutongp, line 23: `Secp256k1 Koblitz public key values MUST either be encoded as a JWK or be encoded as the raw 33-byte public key value in Base58 Bitcoin format using the publicKeyBase58 property.`

@xinxin-crypto what encoding shall we use? `hex` or `base58`?

ioctl/cmd/did/diddoc.go
xinxin-crypto: The specification has been constantly updated during past years. It looks like we should use EcdsaSecp256k1VerificationKey2019 according to the lates spec at https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/.

ioctl/cmd/did/diddoc.go
xinxin-crypto: We may use publicKeyHex (i.e., a hex encoded compressed secp256k1 public key) and this format can be converted to publicKeyJwk if needed (i.e., verify a JSON Web Signature).

ioctl/cmd/did/diddoc.go
lzxm160: ok

ioctl/cmd/did/didgenerate.go
```diff
@@ -0,0 +1,111 @@
...
+	Args:  cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		err := generate()
+		return output.PrintError(err)
+	},
+}
+
+func init() {
+	generateCmd.Flags().StringVarP(&signer, "signer", "s", "", config.TranslateInLang(flagSignerUsages, config.UILanguage))
```
dustinxie, line 53: we can use RegisterWriteCommand(), just like all other commands that has '-s' ?

ioctl/cmd/did/didgenerate.go
```diff
@@ -0,0 +1,111 @@
...
+		config.Chinese: "generate [-s 签署人]",
+	}
+	flagSignerUsages = map[config.Language]string{
+		config.English: "choose a signing account",
+		config.Chinese: "选择一个签名账户",
+	}
+)
+
+// generateCmd represents the generate command
+var generateCmd = &cobra.Command{
```
dustinxie, line 41: i think 'didGenerateCmd' is more clear?
in the future we can have other didXxxxCmd

ioctl/cmd/did/didgenerate.go
lzxm160: let me check this

ioctl/cmd/did/didgenerate.go
lzxm160: sure

ioctl/cmd/did/didgenerate.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+func generateFromSigner(signer, password string) (generatedMessage string, err error) {
+	pri, err := account.LocalAccountToPrivateKey(signer, password)
+	if err != nil {
+		return
+	}
+	doc := newDIDDoc()
+	ethAddress, err := util.IoAddrToEvmAddr(signer)
```
dustinxie, line 80: was wondering, since our DID domain is iotex, that is, it starts with did:io:
should we use the ioAddr, like
`did:io:io1ladc7y0scpcs768xvusqn6r7thavafe8wyxypa`
instead of still using 20-byte Ether addr?
`did:io:0xff5b8f11f0c0710F68e6672009e87e5DFacea727`

lzxm160: I have done this before,it's not difficult to do this,but if users not use this ioctl command,they may generate did:io:0xff5b8f11f0c0710F68e6672009e87e5DFacea727,that may generate different did doc from same private key
If we do this,https://uniresolver.io/ also can resolve io1ladc7y0scpcs768xvusqn6r7thavafe8wyxypa,I think it's the better way

ioctl/cmd/did/didgenerate.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return "", output.NewError(output.AddressError, "", err)
+	}
+	doc.ID = DIDPrefix + ethAddress.String()
+	uncompressed := pri.PublicKey().HexString()
```
dustinxie, line 85: to be safe, i think should
uncompressed := pri.PublicKey().Bytes()
if len() == 33 && uncompressed[0] == 2 or 3, this is already compressed format, can directly use it
if len() == 65 && uncompressed[0] == 4, this is uncompressed format, use the following logic to determine prefix = 2 or 3

dustinxie: also use PublicKey().Bytes() can avoid the ParseInt() below

lzxm160: I'll change to this logic

ioctl/cmd/did/didgenerate.go
yutongp: @dustinxie  we have decided to use did:io: 0xff5b8f11f0c0710F68e6672009e87e5DFacea727 because it is consistent when directly calling contract (our contract only understand eth address).

ioctl/cmd/action/action.go
```diff
@@ -108,7 +108,7 @@ func decodeBytecode() ([]byte, error) {
 	return hex.DecodeString(util.TrimHexPrefix(bytecodeFlag.Value().(string)))
 }
 
-func signer() (address string, err error) {
+func Signer() (address string, err error) {
```
lzxm160, line 111: should be export,so we can use this to get signer in outside this package

ioctl/cmd/did/didgenerate.go
dustinxie: i see, thx!

ioctl/cmd/did/didgenerate.go
```diff
@@ -0,0 +1,109 @@
...
+		authentication.PublicKeyHex = hex.EncodeToString(uncompressed)
+	} else if len(uncompressed) == 65 && uncompressed[0] == 4 {
+		lastNum := uncompressed[64]
+		authentication.PublicKeyHex = hex.EncodeToString(uncompressed[1:33])
+		if lastNum%2 == 0 {
+			authentication.PublicKeyHex = "02" + authentication.PublicKeyHex
+		} else {
+			authentication.PublicKeyHex = "03" + authentication.PublicKeyHex
+		}
+	}
```
dustinxie, line 100: else {
   return output.NewError("invalid public key")
}

PR #2237, opened 2020-06-22
feat: Add unit test for rolldpos protocol

### LINKED ISSUE #1686: Unit test for action/protocol/rolldpos/epoch.go
Add unit test for rolldpos protocol

WhoAmI233:
Issue #1686
Closes #1686


WhoAmI233:
I wrote it with reference to the test case above

CoderZhi:
> I wrote it with reference to the test case above

Besides test coverage, readable code and easy to update are also important. For the unit test, if it is not straightforward to tell what it is doing, we'd better have some comments.

Issue #2142, opened 2020-04-28
Clean up some static functions in staking protocol


CoderZhi:
**What would you like to be added**:
We have some static functions like getTotalBucketCount (not used anywhere), getBucket, updateBucket, putBucket, delBucket, getAllBuckets, getBucketsWithIndices, putBucketAndIndex, which could all be moved in to CandidateStateManager. Also, function settleAction could be moved into CandidateStateManager as well.

**Why is this needed**:
Because of these functions, we have a mixture usage of CandidateStateManager and StateManager throughout this package. By moving them into CandidateStateManager, we don't need to expose StateManager APIs, and CandidateStateManager could be the only interface interacting with this protocol internally.

LuckyPigeon:
I can take this, but needs to create separate PRs.

PR #2238, opened 2020-06-22
DRAFT: move some static functions to CandidateStateManager

ktlshya:
Issues: #2142
Hi, I don't know if this is the right way to do this , could you give me a feedback of this draft.

### DIFF
--- a/action/protocol/staking/protocol.go
+++ b/action/protocol/staking/protocol.go
@@ -129,8 +129,12 @@ func (p *Protocol) Start(ctx context.Context, sr protocol.StateReader) (interfac
 	bcCtx := protocol.MustGetBlockchainCtx(ctx)
 	p.hu = config.NewHeightUpgrade(&bcCtx.Genesis)
 
+	csr, err := NewCandidateStateReader(sr)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create candidate state reader")
+	}
 	// load view from SR
-	c, err := createCandCenter(sr)
+	c, err := csr.createCandCenter()
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to start staking protocol")
 	}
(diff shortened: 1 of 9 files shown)


yutongp:
from my understanding, what @CoderZhi means is in function like `readStateBuckets` instead of passing in a `StateReader`, pass in a `CandidateStateManager` or better a `CandidateStateReader`, so you don't need to do the type assertion in all following functions everywhere, but just top level functions like `ReadState`

action/protocol/staking/read_state.go
```diff
@@ -19,7 +19,7 @@ import (
 
 func readStateBuckets(ctx context.Context, sr protocol.StateReader,
```
CoderZhi, line 20: as @yutongp commented, we should change "sr protocol.StateReader" to "csr CandidateStateReader" or "csm CandidateStateManager". This is the key part of the issue. With this kind of changes, you can directly call csr.allBuckets() and etc.

ktlshya: Is that to say I don't need to change the param ```StateReader``` of ```ReadState()``` to csm or csr @CoderZhi @yutongp

action/protocol/staking/protocol.go
```diff
@@ -331,21 +354,25 @@ func (p *Protocol) ReadState(ctx context.Context, sr protocol.StateReader, metho
 		return nil, errors.Wrap(err, "failed to unmarshal request")
 	}
 
-	center, err := getOrCreateCandCenter(sr)
```
CoderZhi, line 334: this is not necessary for some cases, such as readStateBuckets and readStateBucketsByVoter. From which we can tell that it is not necessary to create candidate center here (which may be time consuming).

action/protocol/staking/read_state.go
```diff
@@ -54,21 +53,21 @@ func readStateBucketsByVoter(ctx context.Context, sr protocol.StateReader,
...
-func readStateBucketsByCandidate(ctx context.Context, sr protocol.StateReader, cc CandidateCenter,
+func readStateBucketsByCandidate(ctx context.Context, csr CandidateStateReader, cc CandidateCenter,
 	req *iotexapi.ReadStakingDataRequest_VoteBucketsByCandidate) (*iotextypes.VoteBucketList, error) {
 	c := cc.GetByName(req.GetCandName())
 	if c == nil {
 		return &iotextypes.VoteBucketList{}, nil
 	}
 
-	indices, err := getCandBucketIndices(sr, c.Owner)
+	indices, err := csr.getCandBucketIndices(c.Owner)
```
CoderZhi, line 63: the reason we need to create candidate center outside is that we need to interact with candidate center to get candidate by name. What if candidate center is part of csr?

ktlshya:
fork a new one

PR #2246, opened 2020-06-30
Apply new trie improvement to contract trie  

koseoyoung:
Apply new trie change(cache/hash node + async option) to contract trie  
@yutongp if the Greenland hard fork is too tight, I will make another height for this change.

### DIFF
--- a/action/protocol/execution/evm/contract_test.go
+++ b/action/protocol/execution/evm/contract_test.go
@@ -68,7 +68,7 @@ func TestCreateContract(t *testing.T) {
 	_, err = accountutil.LoadOrCreateAccount(sm, addr.String())
 	require.NoError(err)
 	hu := config.NewHeightUpgrade(&cfg.Genesis)
-	stateDB := NewStateDBAdapter(sm, 0, hu.IsPre(config.Aleutian, 0), hash.ZeroHash256)
+	stateDB := NewStateDBAdapter(sm, 0, hu.IsPre(config.Aleutian, 0), hu.IsPost(config.Fairbank, 0), hash.ZeroHash256)
 	contract := addr.Bytes()
 	var evmContract common.Address
 	copy(evmContract[:], contract[:])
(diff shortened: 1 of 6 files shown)


action/protocol/execution/evm/contract_test.go
```diff
@@ -68,7 +68,7 @@ func TestCreateContract(t *testing.T) {
 	_, err = accountutil.LoadOrCreateAccount(sm, addr.String())
 	require.NoError(err)
 	hu := config.NewHeightUpgrade(&cfg.Genesis)
-	stateDB := NewStateDBAdapter(sm, 0, hu.IsPre(config.Aleutian, 0), hash.ZeroHash256)
+	stateDB := NewStateDBAdapter(sm, 0, hu.IsPre(config.Aleutian, 0), hu.IsPost(config.Fairbank, 0), hash.ZeroHash256)
```
CoderZhi, line 71: let's use greenland in the unit tests as well

action/protocol/execution/evm/evm_test.go
```diff
@@ -130,13 +130,13 @@ func TestConstantinople(t *testing.T) {
 		require.NoError(err)
 
 		hu := config.NewHeightUpgrade(&bcCtx.Genesis)
-		stateDB := NewStateDBAdapter(sm, e.height, hu.IsPre(config.Aleutian, e.height), ex.Hash())
+		stateDB := NewStateDBAdapter(sm, e.height, hu.IsPre(config.Aleutian, e.height), hu.IsPost(config.Fairbank, e.height), ex.Hash())
```
CoderZhi, line 133: same here

yutongp:
i think it's fine to use greenland height. we can always adjust the hard fork height before release

dustinxie:
i have 2 questions:
1. i don't see where the "cancel-out" is done
2. i saw AsyncOption{} added to NewStateAdaptor() and newContract(). What about the account trie, do we need to apply AsyncOption{} to the account trie as well?

koseoyoung:
@dustinxie both are explained in https://github.com/iotexproject/iotex-core/pull/2243 
1. I think that async flush gives an effect as "cancel-out" 
2. https://github.com/iotexproject/iotex-core/pull/2243/files#diff-fa792709355c2cdf01dc799a779bdcf7R46 Async option was enabled for twolayertrie, I thought this is for account trie.

dustinxie:
> 2. https://github.com/iotexproject/iotex-core/pull/2243/files#diff-fa792709355c2cdf01dc799a779bdcf7R46 Async option was enabled for twolayertrie, I thought this is for account trie.

i see, we don't need to enable it at Greenland height? like this PR does?

koseoyoung:
> > 1. https://github.com/iotexproject/iotex-core/pull/2243/files#diff-fa792709355c2cdf01dc799a779bdcf7R46 Async option was enabled for twolayertrie, I thought this is for account trie.
> 
> i see, we don't need to enable it at Greenland height? like this PR does?

Yes, because contract trie db will write to workingset, so affect delta state digest. But account trie will write to underlying kv store, so won't affect digest.

PR #2257, opened 2020-07-03
store systemlog (evm transfer + withdraw bucket) into chain DB

dustinxie:


action/protocol/staking/handlers_test.go
```diff
@@ -1316,26 +1270,30 @@ func TestProtocol_HandleWithdrawStake(t *testing.T) {
...
+		require.NoError(err)
 		if r != nil {
 			require.Equal(uint64(test.status), r.Status)
 		} else {
 			require.Equal(test.status, iotextypes.ReceiptStatus_Failure)
 		}
 
-		if test.err == nil && test.status == iotextypes.ReceiptStatus_Success {
+		if test.status == iotextypes.ReceiptStatus_Success {
+			// check the special withdraw bucket log
```
dustinxie, line 1281: check the special withdraw log is generated correctly

blockchain/blockdao/blockdao.go
```diff
@@ -719,9 +720,13 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
 	}
 	batchForBlock := batch.NewBatch()
 	hash := blk.HashBlock()
+	heightKey := heightKey(blkHeight)
 	batchForBlock.Put(blockHeaderNS, hash[:], serHeader, "failed to put block header")
 	batchForBlock.Put(blockBodyNS, hash[:], serBody, "failed to put block body")
 	batchForBlock.Put(blockFooterNS, hash[:], serFooter, "failed to put block footer")
+	if sysLog := action.SystemLog(blk.Receipts); len(sysLog) != 0 {
+		batchForBlock.Put(systemLogNS, heightKey, sysLog.Serialize(), "failed to put system log")
```
dustinxie, line 728: put system log into a separate bucket

systemlog/systemlog.go
```diff
@@ -121,7 +121,7 @@ func (x *Indexer) PutBlock(_ context.Context, blk *block.Block) error {
 
 		evmTransferList := &iotextypes.EvmTransferList{}
 		for _, l := range receipt.Logs {
-			if action.IsSystemLog(l) {
+			if action.IsEvmTransfer(l) {
```
dustinxie, line 124: do i need to write the withdraw bucket systemlog, or rosetta client will switch to directly read the new bucket in chain.db?

action/receipt.go
```diff
@@ -12,24 +12,67 @@ import (
...
+func IsWithdrawBucket(l *Log) bool {
+	if l == nil || len(l.Topics) < 3 {
+		return false
+	}
+
+	h := hash.Hash160b([]byte(StakingProtocolID))
+	addr, _ := address.FromBytes(h[:])
+	if l.Address != addr.String() {
+		return false
+	}
```
CoderZhi, line 47: if `l.Topics[0] == hash.BytesToHash256([]byte(WithdrawAmountTopic))` is true, then `l.Address != addr.String()` is a fatal error.

action/receipt.go
```diff
@@ -12,24 +12,67 @@ import (
...
 // IsSystemLog checks whether a log is system log
-// lowerBound chooses the largest system log topic, which is InContractTransfer currently
 func IsSystemLog(l *Log) bool {
-	if len(l.Topics) == 0 {
+	return IsEvmTransfer(l) || IsWithdrawBucket(l)
+}
+
+// IsWithdrawBucket checks withdraw bucket log
+func IsWithdrawBucket(l *Log) bool {
+	if l == nil || len(l.Topics) < 3 {
```
CoderZhi, line 39: if `l.Topics[0] == hash.BytesToHash256([]byte(WithdrawAmountTopic))` is true, `len(l.Topics) < 3` should be a fatal error

action/protocol/staking/handlers.go
```diff
@@ -206,34 +207,34 @@ func (p *Protocol) handleWithdrawStake(ctx context.Context, act *action.Withdraw
...
-
 	log.AddAddress(actionCtx.Caller)
-	if p.hu.IsPost(config.Greenland, blkCtx.BlockHeight) {
-		log.AddTopics(bucket.StakedAmount.Bytes())
-	}
-	return log, nil
+
+	// generate withdraw amount log
+	amountlog := newReceiptLog(p.addr.String(), HandleWithdrawAmount, true)
+	amountlog.AddTopics(byteutil.Uint64ToBytesBigEndian(bucket.Index), bucket.StakedAmount.Bytes())
```
CoderZhi, line 236: why should it be an receiptLog, but not an action.Log?

CoderZhi, line 236: usually, we don't set data like "staked amount" as topic, but data

action/receipt.go
```diff
@@ -12,24 +12,67 @@ import (
...
+	"github.com/iotexproject/iotex-address/address"
 	"github.com/iotexproject/iotex-proto/golang/iotextypes"
 
 	"github.com/iotexproject/iotex-core/pkg/log"
 )
 
+// constants
+const (
+	StakingProtocolID   = "staking"
+	WithdrawAmountTopic = "withdrawAmount"
```
CoderZhi, line 26: shall we move this as a var after InContractTransfer?

action/receipt.go
```diff
@@ -146,9 +192,18 @@ func (log *Log) ConvertFromLogPb(pbLog *iotextypes.Log) {
...
 	log.Index = uint(pbLog.GetIndex())
+
+	if len(log.Topics) > 1 {
+		for _, l := range log.Topics[1:] {
+			if log.Topics[0] == l {
+				log.NotFixTopicCopyBug = true
+				break
+			}
+		}
+	}
```
CoderZhi, line 203: invalid statement, revert it

action/receipt.go
```diff
@@ -162,3 +217,34 @@ func (log *Log) Deserialize(buf []byte) error {
...
+		log.ConvertFromLogPb(pbLog)
+		logs = append(logs, log)
+	}
+	return logs, nil
+}
+
+// Serialize returns a serialized byte stream for Logs
+func (ls Logs) Serialize() []byte {
+	return byteutil.Must(proto.Marshal(ls.toProto()))
+}
```
CoderZhi, line 250: If you have concern on "NotFixTopicCopyBug", rename "Logs" to "SystemLogs"

blockchain/blockdao/blockdao.go
```diff
@@ -719,9 +720,13 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
...
+	}
```
CoderZhi, line 729: I thought we only write system log only if the feature is turned on.

action/receipt.go
dustinxie: i think it is a quick check, if topics < 3, we don't need to check further, it cannot be a withdraw bucket log

action/receipt.go
dustinxie: i think it is a quick check, if `l.Address != addr.String()` this log is not generated by staking protocol, we don't need to check further

action/protocol/staking/handlers.go
dustinxie: yeah, can directly use action.Log

action/receipt.go
dustinxie: we could, and is there a reason we want to use var instead of const? just want to learn
i always prefer to use const whenever possible.

action/receipt.go
dustinxie: i add this b/c i saw the newly added TestLogs() would fail with `log.NotFixTopicCopyBug = true`

so add this check, if all topics are same --> the log is earlier than this fix, so set this flag = true

it is not absolutely necessary, but makes the code more accurate

blockchain/blockdao/blockdao.go
dustinxie: yeah, let's discuss what's the best plan

action/protocol/staking/handlers.go
```diff
@@ -35,6 +35,7 @@ const (
 	HandleRestake           = "restake"
 	HandleCandidateRegister = "candidateRegister"
 	HandleCandidateUpdate   = "candidateUpdate"
+	HandleWithdrawAmount    = "withdrawAmount"
```
CoderZhi, line 38: We may just use the one defined in receipt.go

action/receipt.go
CoderZhi: As we design SystemLog, the topic, like "InContractTransfer", is more like an enum, which starts with 0x0...0. It is not saying this is a better way, but it is better to be consistent.

action/receipt.go
CoderZhi: As we discussed offline, the first topic is the one we used to tell whether it is an WithdrawBucket system log.

dustinxie: fixed

action/systemlog.go
```diff
@@ -0,0 +1,186 @@
...
+func (log *TokenTxRecord) toProto() *iotextypes.ActionSystemLog_Transaction {
+	return &iotextypes.ActionSystemLog_Transaction{
+		Amount:    log.amount,
+		Sender:    log.sender,
+		Recipient: log.recipient,
+	}
+}
+
+// SystemLogFromReceipt returns system logs in the receipt
+func SystemLogFromReceipt(receipts []*Receipt) *BlockSystemLog {
```
dustinxie, line 121: generate all system logs from a block's receipt

systemlog/systemlog.go
yutongp: not sure what you mean. rosetta client will only read through the new API you expose. So it can only read whatever  and however you expose through those APIs

action/systemlog.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/iotexproject/iotex-address/address"
+	"github.com/iotexproject/iotex-proto/golang/iotextypes"
+
+	"github.com/iotexproject/iotex-core/pkg/util/byteutil"
+)
+
+// system log definitions
+type (
+	// TokenTxRecord is a token transaction record
+	TokenTxRecord struct {
```
dustinxie, line 24: these 3 corresponds to 
https://github.com/iotexproject/iotex-proto/blob/master/proto/types/action.proto#L306-L320

blockchain/blockdao/blockdao.go
```diff
@@ -719,9 +743,15 @@ func (dao *blockDAO) putBlock(blk *block.Block) error {
 	}
 	batchForBlock := batch.NewBatch()
 	hash := blk.HashBlock()
+	heightKey := heightKey(blkHeight)
 	batchForBlock.Put(blockHeaderNS, hash[:], serHeader, "failed to put block header")
 	batchForBlock.Put(blockBodyNS, hash[:], serBody, "failed to put block body")
 	batchForBlock.Put(blockFooterNS, hash[:], serFooter, "failed to put block footer")
+	if dao.ContainsSystemLog() {
```
dustinxie, line 750: if the DB file supports system log
1. generate all systemlog from a block's receipt
2. store if systemlog exists

action/protocol/staking/handlers.go
```diff
@@ -206,34 +206,48 @@ func (p *Protocol) handleWithdrawStake(ctx context.Context, act *action.Withdraw
...
-	}
-	return log, nil
+		log.SetData(bucket.StakedAmount.Bytes())
+	}
+
+	// generate withdraw amount log
+	amountLog := action.Log{
+		Address: p.addr.String(),
+		Topics: action.Topics{
+			action.BucketWithdraAmount,
```
lzxm160, line 241: BucketWithdrawAmount

blockchain/blockdao/blockdao.go
```diff
@@ -944,12 +973,24 @@ func (dao *blockDAO) openDB(idx uint64) (kvStore db.KVStore, index uint64, err e
 	defer dao.mutex.Unlock()
 	// open or create this db file
 	cfg.DbPath = path.Dir(cfg.DbPath) + "/" + name
+	var newFile bool
+	_, err = os.Stat(cfg.DbPath)
+	if err != nil && os.IsNotExist(err) {
+		newFile = true
+	}
```
lzxm160, line 980: if err != nil && !os.IsNotExist(err) {
		return
	}else if err!=nil{
		newFile=true
	}

blockchain/blockdao/blockdao.go
lzxm160: is this for old db?if old db is not sync from height 1,so they will not have systemlog forever?

dustinxie: that's right
the plan is after this PR merge, Yutong will restart to sync from block 1

api/api.go
```diff
@@ -809,6 +809,58 @@ func (api *Server) GetEvmTransfersByBlockHeight(ctx context.Context, in *iotexap
...
+	in *iotexapi.GetSystemLogByBlockHeightRequest) (*iotexapi.GetSystemLogByBlockHeightResponse, error) {
+	tip, err := api.dao.Height()
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if in.BlockHeight < 1 || in.BlockHeight >= tip {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid block height = %d", in.BlockHeight)
+	}
+
+	sysLog, err := api.dao.GetSystemLog(in.BlockHeight)
```
yutongp, line 857: shall we return a different error code (`not supported` maybe) when blockdao doesn't contain system log?

action/protocol/staking/handlers.go
yutongp: good catch

action/systemlog.go
yutongp: I probably put these three types in the wrong place in `iotex-proto`. I think these three are not actions. They are data holder which used on API we present to outside.

api/api.go
```diff
@@ -809,6 +809,58 @@ func (api *Server) GetEvmTransfersByBlockHeight(ctx context.Context, in *iotexap
...
+	}
+
+	for _, log := range sysLog.ActionSystemLog {
+		if in.ActionHash == hex.EncodeToString(log.ActionHash) {
+			return &iotexapi.GetSystemLogByActionHashResponse{
+				ActionSystemLog: log,
+			}, nil
+		}
+	}
+	return nil, status.Errorf(codes.NotFound, "system log not found for action %s", in.ActionHash)
```
yutongp, line 842: I am not sure if this is the desired way to implement this on core end.

dustinxie: yeah, i was originally think adding system log as part an action index `/blockindex/actionindex.go`
then GetByActionHash() becomes a direct DB access

action/protocol/staking/protocol.go
```diff
@@ -261,13 +262,17 @@ func (p *Protocol) handle(ctx context.Context, act action.Action, csm CandidateS
 		return nil, nil
 	}
 
+	logs := []*action.Log{rLog.Build(ctx, err)}
 	if err == nil {
-		return p.settleAction(ctx, csm, uint64(iotextypes.ReceiptStatus_Success), rLog.Build(ctx, err))
+		if aLog != nil && aLog.IsWithdrawBucket() {
```
CoderZhi, line 267: if aLog is not nil, why we still want to check its type?

action/receipt.go
```diff
@@ -13,22 +13,24 @@ import (
...
 	"github.com/iotexproject/iotex-proto/golang/iotextypes"
 
 	"github.com/iotexproject/iotex-core/pkg/log"
 )
 
-// InContractTransfer is topic for system log of evm transfer
-var InContractTransfer = common.Hash{} // 32 bytes with all zeros
+// constants
+const (
+	StakingProtocolID = "staking"
```
CoderZhi, line 24: we may use staking.Name() if there is not conflict

api/api.go
```diff
@@ -765,48 +760,77 @@ func (api *Server) GetActionByActionHash(h hash.Hash256) (action.SealedEnvelope,
...
 }
 
-// GetEvmTransfersByBlockHeight returns evm transfers by block height
-func (api *Server) GetEvmTransfersByBlockHeight(ctx context.Context, in *iotexapi.GetEvmTransfersByBlockHeightRequest) (*iotexapi.GetEvmTransfersByBlockHeightResponse, error) {
-	if !api.hasActionIndex || api.systemLogIndexer == nil {
-		return nil, status.Error(codes.Unavailable, "evm transfer index not supported")
+// GetSystemLogByBlockHeight returns system log by block height
+func (api *Server) GetSystemLogByBlockHeight(
```
CoderZhi, line 811: for external, is "SystemLog" a confusing name?

blockchain/block/systemlog.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+// system log definitions
+type (
+	// TokenTxRecord is a token transaction record
+	TokenTxRecord struct {
+		amount    string
+		sender    string
+		recipient string
+	}
```
CoderZhi, line 29: shall we add a topic here to distinguish evm transfer, bucket withdraw and etc.

api/api.go
dustinxie: fixed

action/receipt.go
dustinxie: tried but unfortunately this will create import cycle

blockchain/block/systemlog.go
dustinxie: good suggestion

action/protocol/execution/protocol_test.go
```diff
@@ -437,7 +437,7 @@ func (sct *SmartContractTest) run(r *require.Assertions) {
 		if receipt.Status == uint64(iotextypes.ReceiptStatus_Success) {
 			numLog := 0
 			for _, l := range receipt.Logs {
-				if !action.IsSystemLog(l) {
+				if !l.IsEvmTransfer() {
```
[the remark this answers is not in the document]
dustinxie: since this is contract execution, IsSystemLog() is equivalent to IsEvmTransfer() here right?
so think using IsEvmTransfer() is more accurate

api/api.go
```diff
@@ -765,48 +759,80 @@ func (api *Server) GetActionByActionHash(h hash.Hash256) (action.SealedEnvelope,
...
+// GetEvmTransfersByBlockHeight returns evm transfers by block height
+func (api *Server) GetEvmTransfersByBlockHeight(ctx context.Context, in *iotexapi.GetEvmTransfersByBlockHeightRequest) (*iotexapi.GetEvmTransfersByBlockHeightResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "evm transfer index is deprecated, call GetSystemLogByBlockHeight instead")
+}
+
+// GetImplicitTransferLogByActionHash returns implict transfer log by action hash
+func (api *Server) GetImplicitTransferLogByActionHash(
+	ctx context.Context,
+	in *iotexapi.GetImplicitTransferLogByActionHashRequest) (*iotexapi.GetImplicitTransferLogByActionHashResponse, error) {
+	if !api.hasActionIndex || api.indexer == nil {
```
lzxm160, line 774: is this check still need?

api/api.go
```diff
@@ -765,48 +759,80 @@ func (api *Server) GetActionByActionHash(h hash.Hash256) (action.SealedEnvelope,
...
+	sysLog, err := api.dao.GetImplicitTransferLog(actIndex.BlockHeight())
+	if err != nil {
+		if errors.Cause(err) == db.ErrNotExist {
+			return nil, status.Error(codes.NotFound, err.Error())
+		}
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	for _, log := range sysLog.ImplicitTransferLog {
+		if in.ActionHash == hex.EncodeToString(log.ActionHash) {
```
lzxm160, line 803: how about use the above h :
if bytes.compare(h,log.ActionHash)==0

action/protocol/staking/protocol.go
dustinxie: fixed

api/api.go
dustinxie: yes, by action hash first needs to get the action index (which is the block height), then get log by height and match against action hash.
w/o action index it won't be to able to know the height, and cannot proceed

blockchain/block/systemlog.go
```diff
@@ -0,0 +1,208 @@
...
+		txRecord.sender = from.String()
+		to, _ := address.FromBytes(log.Topics[2][12:])
+		txRecord.recipient = to.String()
+		return &txRecord
+	case log.IsWithdrawBucket():
+		txRecord.topic = make([]byte, len(log.Topics[0]))
+		copy(txRecord.topic, log.Topics[0][:])
+		txRecord.amount = new(big.Int).SetBytes(log.Data).String()
+		txRecord.sender = log.Address
+		to, _ := address.FromBytes(log.Topics[2][12:])
```
lzxm160, line 202: for stakewithdraw,sender is "staking",Recipient is the action caller,right?

api/api.go
dustinxie: ok

blockchain/block/systemlog.go
dustinxie: right, sender is staking protocol addr: io1qnpz47hx5q6r3w876axtrn6yz95d70cjl35r53

api/api.go
lzxm160: I see,that's why GetImplicitTransferLogByBlockHeight does not need this check

action/protocol/staking/protocol.go
```diff
@@ -261,13 +262,17 @@ func (p *Protocol) handle(ctx context.Context, act action.Action, csm CandidateS
...
+		if aLog != nil {
+			logs = append(logs, aLog)
+		}
+		return p.settleAction(ctx, csm, uint64(iotextypes.ReceiptStatus_Success), logs)
 	}
 
 	if receiptErr, ok := err.(ReceiptError); ok {
 		log.L().Debug("Non-critical error when processing staking action", zap.Error(err))
-		return p.settleAction(ctx, csm, receiptErr.ReceiptStatus(), rLog.Build(ctx, err))
+		return p.settleAction(ctx, csm, receiptErr.ReceiptStatus(), logs)
```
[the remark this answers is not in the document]
dustinxie: a valid aLog is only generated for (err == nil) above (that is, action success and bucket is withdrawn)
  
this is non-critical error (for example caller is not owner of bucket) happened, aLog must be == nil here

raullenchai:
Congrats!

PR #2267, opened 2020-07-09
unify two layer trie layer two key format

CoderZhi:
In current implementation, one namespace may have keys of different lengths. To resolve this issue, unify the key to Hash160


state/factory/factory.go
```diff
@@ -652,6 +653,11 @@ func readState(tlt trie.TwoLayerTrie, ns string, key []byte, s interface{}) erro
 	return state.Deserialize(s, data)
 }
 
+func toLayerTwoKey(input []byte) []byte {
```
dustinxie, line 656: add if( len(input) != 20 )?
since most keys now (account, contract) are 20-bytes long

dustinxie: also maybe name it toLegacyKey()? we use Legacy to mean 20-bytes key, like:

func LegacyKeyOption(key hash.Hash160) StateOption

PR #2270, opened 2020-07-09
fix digest issue in contract

CoderZhi:
https://github.com/iotexproject/iotex-core/pull/2259 was supposed to improve "async mode", but was applied to "sync mode" by accident. This is a fix for that.

### DIFF
--- a/action/protocol/execution/evm/contract.go
+++ b/action/protocol/execution/evm/contract.go
@@ -89,7 +89,7 @@ func (c *contract) SetState(key hash.Hash256, value []byte) error {
 	if err := c.trie.Upsert(key[:], value); err != nil {
 		return err
 	}
-	if ！c.async {
+	if !c.async {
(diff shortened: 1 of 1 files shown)


action/protocol/execution/evm/contract.go
```diff
@@ -85,7 +86,20 @@ func (c *contract) SetState(key hash.Hash256, value []byte) error {
 		c.GetState(key)
 	}
 	c.dirtyState = true
-	return c.trie.Upsert(key[:], value)
+	if c.async {
+		return c.trie.Upsert(key[:], value)
+	}
+	if err := c.trie.Upsert(key[:], value); err != nil {
+		return err
+	}
```
dustinxie, line 94: delete line 89~91, add here
if c.async {
   return nil
}

yutongp: +1

PR #2280, opened 2020-07-11
iterate branch node children in order

CoderZhi:
Looping branch node's children in some order to have consistent result.


db/trie/mptrie/branchnode.go
```diff
@@ -80,8 +80,10 @@ func (b *branchNode) MarkAsRoot() {
 func (b *branchNode) Children() []node {
 	trieMtc.WithLabelValues("branchNode", "children").Inc()
 	children := []node{}
-	for _, child := range b.children {
-		children = append(children, child)
+	for index := 0; index < radix; index++ {
```
dustinxie, line 83: we better understand why it works in old code (before mptrie refactor)

the fix should make it work by same logic as before, otherwise this might be a breaking change

PR #2282, opened 2020-07-11
add staking action amount into implicit transfer log

dustinxie:


action/protocol/staking/handlers.go
```diff
@@ -62,56 +62,71 @@ func (h *handleError) ReceiptStatus() uint64 {
...
 	log.SetData(byteutil.Uint64ToBytesBigEndian(bucketIdx))
-	return log, nil
+
+	// generate create amount log
+	cLog := action.Log{
+		Address: p.addr.String(),
+		Topics: action.Topics{
+			action.BucketCreateAmount,
+			hash.BytesToHash256(actionCtx.Caller.Bytes()),
+			hash.BytesToHash256(p.addr.Bytes()),
```
dustinxie, line 121: from action caller to staking protocol address

action/protocol/staking/handlers.go
```diff
@@ -391,79 +407,93 @@ func (p *Protocol) handleConsignmentTransfer(
...
 	log.AddAddress(actionCtx.Caller)
-	return log, nil
+
+	// generate deposit amount log
+	dLog := action.Log{
+		Address: p.addr.String(),
+		Topics: action.Topics{
+			action.BucketDepositAmount,
+			hash.BytesToHash256(actionCtx.Caller.Bytes()),
+			hash.BytesToHash256(p.addr.Bytes()),
```
dustinxie, line 488: from action caller to staking protocol address

action/protocol/staking/handlers.go
```diff
@@ -597,30 +627,46 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
...
-	return log, nil
+
+	// generate candidate register log
+	rewardingAddr := hash.Hash160b([]byte(action.RewardingProtocolID))
+	cLog := action.Log{
+		Address: p.addr.String(),
+		Topics: action.Topics{
+			action.CandidateRegistrationFee,
+			hash.BytesToHash256(actCtx.Caller.Bytes()),
+			hash.BytesToHash256(rewardingAddr[:]),
```
dustinxie, line 661: from action caller to rewarding protocol address

action/protocol/staking/handlers.go
```diff
@@ -239,8 +254,9 @@ func (p *Protocol) handleWithdrawStake(ctx context.Context, act *action.Withdraw
 		Address: p.addr.String(),
 		Topics: action.Topics{
 			action.BucketWithdrawAmount,
-			hash.BytesToHash256(byteutil.Uint64ToBytesBigEndian(bucket.Index)),
+			hash.BytesToHash256(p.addr.Bytes()),
 			hash.BytesToHash256(actionCtx.Caller.Bytes()),
+			hash.BytesToHash256(byteutil.Uint64ToBytesBigEndian(bucket.Index)),
```
dustinxie, line 259: withdraw: from staking protocol address to action caller

dustinxie: unify the topics order, so it is the same for all:
topic[1]: from
topic[2]: to

blockchain/block/systemlog.go
```diff
@@ -160,27 +159,17 @@ func ReceiptImplicitTransferLog(r *action.Receipt) *ImplictTransferLog {
...
 	}
+
+	txRecord := TokenTxRecord{}
+	txRecord.topic = make([]byte, len(log.Topics[0]))
+	copy(txRecord.topic, log.Topics[0][:])
+	txRecord.amount = new(big.Int).SetBytes(log.Data).String()
+	from, _ := address.FromBytes(log.Topics[1][12:])
+	txRecord.sender = from.String()
+	to, _ := address.FromBytes(log.Topics[2][12:])
+	txRecord.recipient = to.String()
```
dustinxie, line 173: topics order are unified to be same:
topic[0]: type
topic[1]: from
topic[2]: to

action/protocol/staking/handlers.go
yutongp: also need one from action caller to staking protocol for bucket creation

action/protocol/staking/protocol.go
```diff
@@ -267,8 +270,17 @@ func (p *Protocol) handle(ctx context.Context, act action.Action, csm CandidateS
 		logs = append(logs, l)
 	}
 	if err == nil {
-		if aLog != nil {
-			logs = append(logs, aLog)
+		if createLog != nil {
```
yutongp, line 273: maybe just use aLog?

dustinxie: we now have 4 possible logs, so name them explicitly

CoderZhi: I think they are the same, for this function "handle", it doesn't need to understand what's the meaning of the logs. We may further unify handleXXX return values to (*receiptLog, []*action.Log, error)

action/receipt.go
```diff
@@ -30,6 +31,15 @@ var (
...
 	BucketWithdrawAmount = hash.BytesToHash256([]byte("withdrawAmount"))
+
+	// BucketCreateAmount is topic for bucket create
+	BucketCreateAmount = hash.BytesToHash256([]byte("createAmount"))
+
+	// BucketDepositAmount is topic for bucket deposit
+	BucketDepositAmount = hash.BytesToHash256([]byte("depositAmount"))
+
+	// CandidateRegistrationFee is topic for candidate register
+	CandidateRegistrationFee = hash.BytesToHash256([]byte("registrationFee"))
```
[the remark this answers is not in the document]
dustinxie: 1. used hash.Hash256 b/c that is the type of log.Topics, easy for programming
2. use text like "createAmount" so can clearly tell by Topics[0], are there any benefits using 1, 2, 3, 4?

action/receipt.go
```diff
@@ -182,14 +192,32 @@ func (log *Log) IsWithdrawBucket() bool {
...
-	if log.Topics[0] != BucketWithdrawAmount {
+	if log.Topics[0] != topic {
 		return false
 	}
 
-	if len(log.Topics) < 3 || log.Index != 1 {
-		panic("withdraw bucket log is corrupted")
+	index := uint(1)
+	if topic == CandidateRegistrationFee {
+		index = 2
```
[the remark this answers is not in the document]
dustinxie: currently taking index = order of log in this receipt. if it's meaning index in entire block, then yes, it is wrong

any suggestions how to handle this? or we can leave it as is

action/protocol/staking/protocol.go
dustinxie: handleRegister() will return 2 extra logs, handleCreate/Deposit/Withraw() 1 extra log, others no extra log

would leave it as is, feel unifying to (*receiptLog, []*action.Log, error) does not make the program better/clear

action/receipt.go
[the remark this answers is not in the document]
CoderZhi: the new format was introduced by your PR, InContractTransfer which is common.Hash{} (equals to 0x0000...000

action/receipt.go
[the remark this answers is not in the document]
CoderZhi: if it is not too late, we should do it right.

action/protocol/staking/protocol.go
CoderZhi: we have different understanding of programming

action/receipt.go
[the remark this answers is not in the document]
dustinxie: staking is native protocol, the natural type should be hash.Hash256, InContractTransfer is defined as common.Hash{} only b/c it is used inside evm. I don't see it as a must-to-do to unify these ...

if use common.Hash{}, code like log.Topics[1] == CreateBucket need to change to 
log.Topics[1] == hash.BytesToHash256(CreateBucket[:]), i don't see why should bother to do that while using the natural type is very natural

action/receipt.go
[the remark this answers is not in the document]
dustinxie: if this is an issue, i think we need to tackle in another PR, b/c the change would involve all protocols

for example, a block has 1 transfer, 1 execution, 1 stake create, then should have:
transfer.Log.Index = 0
execution.Log.Index = 1
staking.Log.Index = 2
staking.CreateLog.Index = 3
currently, all protocol use Index = 0 (staking use 1, 2 for extra log)

action/receipt.go
```diff
@@ -182,14 +192,32 @@ func (log *Log) IsWithdrawBucket() bool {
...
 
-	if log.Topics[0] != BucketWithdrawAmount {
+	if log.Topics[0] != topic {
 		return false
 	}
 
-	if len(log.Topics) < 3 || log.Index != 1 {
-		panic("withdraw bucket log is corrupted")
+	index := uint(1)
+	if topic == CandidateRegistrationFee {
```
lzxm160, line 200: is this right?should use bytes.Equal?

dustinxie: it is [32]byte, so can use ==

Issue #2279, opened 2020-07-11
ioctl contract prepare cmd needs to install solidity version 0.5+


RuneAkko:
Using brew install solidity v0.4 has some issues needed to fix manually, but it works fine for v0.5.
Now ioctl contract commands has supported v0.5.
So it's time to change ioctl contract prepare commnad to avoid this problem.

PR #2283, opened 2020-07-12
ioctl contract prepare cmd needs to install solidity version 0.5+

RuneAkko:
#2279


install-solc.sh
```diff
@@ -103,7 +103,7 @@ initOS
 
 # assemble expected release artifact name
 if [ "$OS" = "darwin" ]; then
-    sudo brew install solidity@4
+    sudo brew install solidity@5
```
raullenchai, line 106: Need `brew tap ethereum/ethereum`? or not?

RuneAkko: Yes, modified.

PR #2289, opened 2020-07-14
fix contract determine bug for ioctl

Frankonly:


raullenchai:
Which issue this PR is fixing?

PR #2291, opened 2020-07-14
correctly check reclaim bucket after greenland

dustinxie:


action/protocol/staking/handlers.go
```diff
@@ -343,21 +343,18 @@ func (p *Protocol) handleTransferStake(ctx context.Context, act *action.Transfer
...
 	newOwner := act.VoterAddress()
-	if !address.Equal(bucket.Owner, actionCtx.Caller) {
-		// check if the payload contains a valid consignment transfer
-		if consignment, ok := p.handleConsignmentTransfer(blkCtx, actionCtx, act, bucket); ok {
-			newOwner = consignment.Transferee()
-		} else {
-			return log, &handleError{
-				err:           errors.New("bucket owner does not match action caller"),
-				failureStatus: iotextypes.ReceiptStatus_ErrUnauthorizedOperator,
+	bucket, fetchErr := p.fetchBucket(csm, actionCtx.Caller, act.BucketIndex(), true, false)
```
dustinxie, line 347: need to set checkOwner = true, like before

otherwise, for self-staking bucket, fetchBucket() will return ReceiptStatus_ErrInvalidBucketType

PR #2296, opened 2020-07-15
early quit for action pick

CoderZhi:
If the action's gas limit is larger than remaining gas limit, pop the account from action iterator.

Test Plan:

1. Pressure test: loop through actions from 402,837 accounts takes 1.57s (3909 ns/op)
2. Normal test: loop through actions from 20,000 accounts takes around 0.05s (2682 ns/op).

Conclusion:
Instead of doing some fancy improvement in action iterator, the "early quit" strategy is sufficient for daily usage.


state/factory/workingset.go
```diff
@@ -382,6 +382,10 @@ func (ws *workingSet) pickAndRunActions(
 			if !ok {
 				break
 			}
+			if nextAction.GasLimit() > blkCtx.GasLimit {
```
yutongp, line 385: just wonder what if there are many actions in actpool belong to different account? it will still loop up one action per account, right?

CoderZhi: yes, that's exactly the benchmark is testing against. We should be able to tolerant at least 20,000 accounts at the same time.

PR #2298, opened 2020-07-16
add number of block actions

CoderZhi:
Add back this metric which was deleted.


state/factory/workingset.go
```diff
@@ -35,10 +36,20 @@ var (
...
 func init() {
 	prometheus.MustRegister(stateDBMtc)
+	var err error
+	if timerFactory, err = prometheustimer.New(
+		"workingset_perf",
+		"Performance of workingset",
+		[]string{"topic"},
+		[]string{"default"},
+	); err != nil {
+		log.L().Panic("Failed to generate prometheus timer factory for workingset.", zap.Error(err))
```
dustinxie, line 51: nit: saw 3 other places, 1 used Panic() and 2 used Error(), maybe can unify?

PR #2307, opened 2020-07-20
add test for api/blocklistener.go and api/listener.go

newyear6565:


api/blocklistener_test.go
```diff
@@ -0,0 +1,70 @@
...
+		{
+			BlockHeight: 2,
+		},
+	}
+	builder := block.NewTestingBuilder()
+	builder.SetHeight(1)
+	builder.SetVersion(111)
+	builder.SetTimeStamp(time.Now())
+	builder.SetReceipts(receipts)
+	testBlock, _ := builder.SignAndBuild(identityset.PrivateKey(0))
```
koseoyoung, line 49: catch error here?

api/blocklistener_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	server.EXPECT().Send(gomock.Any()).Return(nil).Times(1)
+	err := responder.Respond(&testBlock)
+	require.NoError(t, err)
```
koseoyoung, line 53: ```require.NoError(t, responder.Respond(&testBlock)) ```

koseoyoung: same comments on below as well

newyear6565: > `require.NoError(t, responder.Respond(&testBlock)) `

ok ,it has been done

api/blocklistener_test.go
newyear6565: > catch error here?

ok ,it has been done

PR #2320, opened 2020-07-23
minor change for ioctl action hash

Frankonly:


raullenchai:
Pls demo the change in the description section, next time

PR #2321, opened 2020-07-23
Clean up static functions in staking protocol

### LINKED ISSUE #2142: Clean up some static functions in staking protocol

newyear6565:
Closes #2142

### DIFF
--- a/action/protocol/staking/candidate_statemanager.go
+++ b/action/protocol/staking/candidate_statemanager.go
@@ -46,6 +46,7 @@ type (
 	CandidateStateManager interface {
 		protocol.StateManager
 		BucketSet
+		BucketGetByIndex
 		CandidateSet
 		settleAction(ctx context.Context, p *Protocol, status uint64, logs []*action.Log, tLogs []*action.TransactionLog) (*action.Receipt, error)
 		// candidate and bucket pool related
@@ -165,11 +166,11 @@ func (csm *candSM) Upsert(d *Candidate) error {
 }
 
 func (csm *candSM) CreditBucketPool(amount *big.Int) error {
-	return csm.bucketPool.CreditPool(csm.StateManager, amount)
+	return csm.bucketPool.CreditPool(csm, amount)
 }
 
 func (csm *candSM) DebitBucketPool(amount *big.Int, newBucket bool) error {
-	return csm.bucketPool.DebitPool(csm.StateManager, amount, newBucket)
+	return csm.bucketPool.DebitPool(csm, amount, newBucket)
 }
 
 func (csm *candSM) Commit() error {
@@ -177,7 +178,7 @@ func (csm *candSM) Commit() error {
 		return err
 	}
 
-	if err := csm.bucketPool.Commit(csm.StateManager); err != nil {
+	if err := csm.bucketPool.Commit(csm); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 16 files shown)


action/protocol/staking/candidate_statemanager.go
```diff
@@ -173,3 +180,129 @@ func (csm *candSM) Commit() error {
 	// write update view back to state factory
 	return csm.WriteView(protocolID, ConvertToViewData(csm))
 }
+
+func (csm *candSM) updateBucket(index uint64, bucket *VoteBucket) error {
```
dustinxie, line 184: think these private methods are to be added to CandidateStateManager interface
so protocol.StateManager can be removed

dustinxie: according to  #2142 description

dustinxie: also, latest code has changed quite a bit, pls rebase the PR

action/protocol/staking/candidate_statemanager.go
```diff
@@ -22,9 +27,28 @@ const (
...
+		delBucketIndex(key []byte, index uint64) error
+	}
+	// CandidateSet related to setting candidates
+	CandidateSet interface {
+		delCandidate(name address.Address) error
+		putCandidate(d *Candidate) error
+	}
 	// CandidateStateManager is candidate state manager on top of StateManager
 	CandidateStateManager interface {
 		protocol.StateManager
```
dustinxie, line 47: #2142 wants CandidateStateManager not to expose StateManager interface 
so we need to implement methods like PutState(), DelState() for (*candSM)
then we can delete "protocol.StateManager" here
@CoderZhi  can confirm

dustinxie: should look like:
candSM struct {
		sm protocol.StateManager
		candCenter *CandidateCenter
		bucketPool *BucketPool
	}

and 
func (csm *candSM) PutState(s interface{}, options ...StateOption) (uint64, error) {
    return csm.sm.PutState(s, options...)
}

newyear6565: After hide `StateManager ` interface , I think we may not modify `candSM` structure. If `candSM` is not changed , it has `StateManager` interface , so there is no need to implement functions like `PutState()`, `DelState()` for `candSM`.The code will be more simplified.

newyear6565: just like `candSR` in candidate_statereader.go

action/protocol/staking/candidate_statemanager.go
```diff
@@ -128,26 +162,193 @@ func (csm *candSM) Upsert(d *Candidate) error {
...
+	return err
+}
+
+func (csm *candSM) delCandidate(name address.Address) error {
+	_, err := csm.DelState(protocol.NamespaceOption(CandidateNameSpace), protocol.KeyOption(name.Bytes()))
+	return err
+}
+
+// settleAccount deposits gas fee and updates caller's nonce
+func (csm *candSM) settleAction(
```
dustinxie, line 320: this should be staking protocol's func, not csm's
it's kinda counter-intuitive that csm needs to take p *Protocol as input

newyear6565: ok , i will revert it

newyear6565: reverted @dustinxie

PR #2327, opened 2020-07-24
add test for stateCheckLegacy

dustinxie:


yutongp:
maybe also do an e2e test?

PR #2328, opened 2020-07-24
create bucket stats state in prestates of greenland

CoderZhi:
After greenland, there will alway be a state of bucket stats


action/protocol/staking/bucket_pool_test.go
```diff
@@ -152,7 +148,53 @@ func TestBucketPool(t *testing.T) {
 			r.Equal(count, pool.Count())
 		}
 	}
+	c, err = ConstructBaseView(sm)
```
dustinxie, line 151: var testGreenland bool
if !testGreenland && v.postGreenland {
    c, err = ConstructBaseView(sm)
    _, err = sm.PutState()
    testGreenland = true
}

then don't need to copy the test code below

PR #2332, opened 2020-07-25
Refactor Blockchain.commitBlock

### LINKED ISSUE #1788: refactor Blockchain.commitBlock
As a follow up of #1787, move bc.sf.Commit into dao.PutBlock. After the change, the commitBlock function will be like
```
    bc.dao.PutBlock(blk)
    bc.emitToSubscribers(blk)
```

newyear6565:
Closes #1788


blockchain/blockdao/blockdao.go
```diff
@@ -398,7 +398,19 @@ func (dao *blockDAO) GetReceipts(blkHeight uint64) ([]*action.Receipt, error) {
 }
 
 func (dao *blockDAO) PutBlock(ctx context.Context, blk *block.Block) error {
-	err := func() error {
+	// early exit if block already exists
+	blkHeight := blk.Height()
+	blkHash, err := dao.GetBlockHash(blkHeight)
```
dustinxie, line 403: our coding practice is that, when a struct has lock/mutex, exported method should only call internal un-exported method, but not other exported method (to avoid potential deadlock)

so use getBlockHash() here

blockchain/blockdao/blockdao.go
```diff
@@ -420,6 +432,12 @@ func (dao *blockDAO) PutBlock(ctx context.Context, blk *block.Block) error {
 		}
 	}
 
+	blkHash, err = dao.GetBlockHash(blkHeight)
```
dustinxie, line 435: putBlock() success, this block has become the tip, so directly use blkHeight/blkHash here

PR #2333, opened 2020-07-25
Allow Dock interface take multiple stash values and fix ProtocolDirty

dustinxie:


action/protocol/staking/candidate_statemanager.go
```diff
@@ -67,19 +72,14 @@ func NewCandidateStateManager(sm protocol.StateManager, enableSMStorage bool) (C
 	}
 
 	// TODO: remove CandidateCenter interface, convert the code below to candCenter.SyncCenter()
-	ser, err := protocol.UnloadAndAssertBytes(sm, protocolID)
-	switch errors.Cause(err) {
+	delta := CandidateList{}
+	switch err = sm.Unload(protocolID, stakingCandCenter, &delta); errors.Cause(err) {
```
CoderZhi, line 76: default and nil cases

state/factory/util.go
```diff
@@ -76,17 +76,15 @@ func generateWorkingSetCacheKey(blkHeader block.Header, producerAddr string) has
...
+	if v, hit := view[name]; hit {
+		return v, nil
 	}
+	return nil, protocol.ErrNoName
+}
+
+func writeView(view protocol.ProtocolView, name string, v interface{}) error {
+	view[name] = v
 	return nil
 }
```
CoderZhi, line 89: why not move these as functions of protocol view?

action/protocol/dock.go
```diff
@@ -17,57 +22,47 @@ var (
...
+	stash batch.KVStoreCache
+	dirty map[string]bool
 }
 
 // NewDock returns a new dock
 func NewDock() Dock {
 	return &dock{
-		value:     make(map[string]interface{}),
-		protocols: make(map[string]bool),
+		stash: batch.NewKVCache(),
```
CoderZhi, line 32: it is just a map of map.

action/protocol/dock.go
```diff
@@ -17,57 +22,47 @@ var (
...
+		dirty: make(map[string]bool),
```
CoderZhi, line 33: with map of map, we don't even need "dirty"

PR #2341, opened 2020-07-27
Add TransactionLog to capture all asset-transferring tx

dustinxie:


action/receipt_test.go
```diff
@@ -23,7 +23,7 @@ func TestConvert(t *testing.T) {
 		hash.Hash256b([]byte("Pacific")),
 		hash.Hash256b([]byte("Aleutian")),
 	}
-	log := &Log{"1", topics, []byte("cd07d8a74179e032f030d9244"), 1, hash.ZeroHash256, 1, true, "", ""}
+	log := &Log{"1", topics, []byte("cd07d8a74179e032f030d9244"), 1, hash.ZeroHash256, 1, true, false, "", ""}
```
yutongp, line 26: strongly suggest add field name here. 
otherwise, very confusing

action/receipt_test.go
```diff
@@ -124,7 +118,7 @@ func TestSerDerLog(t *testing.T) {
...
```
yutongp, line 121: same here

action/receipt_test.go
```diff
@@ -92,7 +86,7 @@ func TestConvertLog(t *testing.T) {
...
```
yutongp, line 89: same here

action/receipt_test.go
dustinxie: fixed

PR #2347, opened 2020-07-28
add native transfer transaction log

dustinxie:


yutongp:
plz use AddLogs to add log

Issue #2231, opened 2020-06-18
Improve test coverage for state/factory/workingset.go


raullenchai:
https://codecov.io/gh/iotexproject/iotex-core/src/master/state/factory/workingset.go

Pls add unit tests

PR #2350, opened 2020-07-28
workingset test(#2231)

JonSnow47:


state/factory/workingset_test.go
```diff
@@ -0,0 +1,232 @@
...
+				makeBlock(t, 1, hash.Hash256b([]byte("test")), digestHash),
+				block.ErrReceiptRootMismatch,
+			},
+			{
+				makeBlock(t, 1, hash.ZeroHash256, hash.Hash256b([]byte("test"))),
+				block.ErrDeltaStateMismatch,
+			},
+		}
+	)
+	sf, err := NewFactory(config.Default, InMemTrieOption())
```
CoderZhi, line 177: same here

state/factory/workingset_test.go
```diff
@@ -0,0 +1,232 @@
...
+				makeBlock(t, 3, hash.ZeroHash256, digestHash),
+				action.ErrNonce,
+			},
+			{
+				makeBlock(t, 1, hash.Hash256b([]byte("test")), digestHash),
+				block.ErrReceiptRootMismatch,
+			},
+			{
+				makeBlock(t, 1, hash.ZeroHash256, hash.Hash256b([]byte("test"))),
+				block.ErrDeltaStateMismatch,
```
CoderZhi, line 173: we also need to test one success case

JonSnow47: the success case is in line 191

Issue #2351, opened 2020-07-28
add AddLogs method in action.Receipt


yutongp:
https://github.com/iotexproject/iotex-core/blob/master/action/receipt.go#L34

1. add an AddLogs(...*Log) method.
2. this method should filter out nil *Log input.
3. change currently all  existing code where directly set Receipt.Logs field to use the method(in order to avoid mistakenly add nil log)

PR #2352, opened 2020-07-28
panic on negative amount transactionlog, add addlog on receipt

yutongp:


yutongp:
fix #2351

action/receipt.go
```diff
@@ -119,6 +119,18 @@ func (receipt *Receipt) Hash() hash.Hash256 {
 	return hash.Hash256b(data)
 }
 
+// AddLogs add log to receipt and filter out nil log.
+func (receipt *Receipt) AddLogs(logs ...*Log) {
```
dustinxie, line 123: directly use []*Log? so no need to do logs... outside

yutongp: i like this. this is more flexible to add single log

action/receipt.go
```diff
@@ -119,6 +119,18 @@ func (receipt *Receipt) Hash() hash.Hash256 {
...
+	if receipt.Logs == nil {
+		receipt.Logs = make([]*Log, 0, len(logs))
+	}
```
CoderZhi, line 126: append works on nil slices, we can delete these three lines

Issue #2353, opened 2020-07-28
[ioctl] add 'ioctl bc bucket count/max' to show total number of buckets and max bucket index


dustinxie:
**What would you like to be added**:
1. ioctl bc bucket max --> the max bucket index
2. ioctl bc bucket count  --> total number of active buckets (this will be smaller than max, due to withdrawn buckets)

**Why is this needed**:
better usability/visibility for staking

dustinxie:
for max: call func getTotalBucketCount(sr protocol.StateReader) (uint64, error)

for count: refer to func getTotalStakedAmount(ctx context.Context, csr CandidateStateReader) (*big.Int, error)
1. total.count is what you need
2. add a new method ActiveBucketsCount() to csr

JonSnow47:
How to calculate the number of active buckets?
```go
func (c *candSR) ActiveBucketsCount() uint64 {
	// TODO: return c.view.bucketPool.Count() - withdrawnBucketsCount
}
```

JonSnow47:
> How to calculate the number of active buckets?
> 
> ```go
> func (c *candSR) ActiveBucketsCount() uint64 {
> 	// TODO: return c.view.bucketPool.Count() - withdrawnBucketsCount
> }
> ```

@dustinxie

dustinxie:
return c.view.bucketPool.Count()

that is the active count

PR #2362, opened 2020-07-30
ioctl bc bucket count/max(#2353)

JonSnow47:
ioctl bc bucket count/max
when iotex-proto merge into master, I'll update `github.com/iotexproject/iotex-proto` version in `go.mod`


action/protocol/staking/read_state.go
```diff
@@ -101,6 +101,21 @@ func readStateBucketByIndices(ctx context.Context, sr protocol.StateReader,
...
+	total, err := getTotalBucketCount(csr.SR())
+	if errors.Cause(err) == state.ErrStateNotExist {
+		return &iotextypes.BucketsCount{}, csr.Height(), nil
+	}
+	if err != nil {
+		return nil, 0, err
+	}
+	return &iotextypes.BucketsCount{
+		Total:  total,
+		Active: csr.ActiveBucketsCount(),
```
dustinxie, line 115: for active count, similar to how to handle total amount in 
`func getTotalStakedAmount(ctx context.Context, csr CandidateStateReader) (*big.Int, error)`

JonSnow47: is that means we should have a func like `getActiveBucketsCount(ctx context.Context, csr CandidateStateReader) uint64` to return active count?

action/protocol/staking/read_state.go
```diff
@@ -216,3 +231,7 @@ func getTotalStakedAmount(ctx context.Context, csr CandidateStateReader) (*big.I
 	// otherwise read from bucket pool
 	return csr.TotalStakedAmount(), nil
 }
+
+func getActiveBucketsCount(csr CandidateStateReader) uint64 {
```
dustinxie, line 235: i mean, similar to the code in getTotalStakedAmount()

if after greenland {
     // read state from db
    return total.count
}

return csr.ActiveBucketsCount(), nil

action/protocol/staking/read_state.go
```diff
@@ -232,6 +236,17 @@ func getTotalStakedAmount(ctx context.Context, csr CandidateStateReader) (*big.I
...
 }
 
-func getActiveBucketsCount(csr CandidateStateReader) uint64 {
-	return csr.ActiveBucketsCount()
+func getActiveBucketsCount(ctx context.Context, csr CandidateStateReader) (uint64, error) {
+	chainCtx := protocol.MustGetBlockchainCtx(ctx)
+	hu := config.NewHeightUpgrade(&chainCtx.Genesis)
+	if hu.IsPost(config.Greenland, csr.Height()) {
+		// after Greenland, read state from db
+		buckets, _, err := getAllBuckets(csr.SR())
```
dustinxie, line 244: var total totalAmount
		_, err := csr.SR().State(&total, protocol.NamespaceOption(StakingNameSpace), protocol.KeyOption(bucketPoolAddrKey))

return total.count, nil

action/protocol/staking/read_state.go
```diff
@@ -243,7 +243,7 @@ func getActiveBucketsCount(ctx context.Context, csr CandidateStateReader) (uint6
 		// after Greenland, read state from db
 		var total totalAmount
 		_, err := csr.SR().State(&total, protocol.NamespaceOption(StakingNameSpace), protocol.KeyOption(bucketPoolAddrKey))
-		if err != nil {
+		if err != nil && errors.Cause(err) != state.ErrStateNotExist {
```
dustinxie, line 246: why need this check?
if pass Greenland and state does not exist, it is a real error that should not be escaped

PR #2363, opened 2020-07-30
use alpine base image

yutongp:


Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM golang:1.14.4-stretch as build
+FROM golang:1.14.4-alpine as build
```
raullenchai, line 1: https://hub.docker.com/_/golang shows 1.14.4-alpine doesn't exist, 14.6?

yutongp: https://hub.docker.com/layers/golang/library/golang/1.14.4-alpine/images/sha256-2de21f4fe5c3557c31b716291a7986c0d54abbd282cd7fb43aaa484672b0c855?context=explore

Dockerfile
```diff
@@ -14,7 +14,9 @@ COPY . .
 RUN mkdir -p $GOPATH/pkg/linux_amd64/github.com/iotexproject/ && \
     make clean build-all
 
-FROM debian:stretch
+FROM alpine:latest
+
+RUN apk add --no-cache ca-certificates
```
dustinxie, line 19: this cannot be combined with above?

yutongp: https://docs.docker.com/develop/develop-images/multistage-build/

PR #2368, opened 2020-08-01
prevent operate on unstaked bucket

yutongp:


dustinxie:
better to check-in r1.0 branch?
the fix is deployed, i am thinking better not to have this on master?

PR #2370, opened 2020-08-02
add vote reviser

yutongp:


action/protocol/staking/receipt_log.go
```diff
@@ -55,7 +55,10 @@ func (r *receiptLog) SetData(data []byte) {
 	r.data = data
 }
 
-func (r *receiptLog) Build(ctx context.Context, err error) *action.Log {
+func (r *receiptLog) Build(ctx context.Context, err error, skipLogOnErr bool) *action.Log {
+	if err != nil && skipLogOnErr {
```
dustinxie, line 59: why add this? don't think it's correct?

for instance, it returns ErrCandidateNotExist (not critical), in this case we should return a receipt with corresponding status

dustinxie: BTW, this (skip log on err) is the behavior prior to FbkMigration height

action/protocol/staking/vote_reviser.go
```diff
@@ -0,0 +1,110 @@
...
+	cands, _, err := getAllCandidates(sm)
+	if err != nil {
+		return nil, err
+	}
+	candm := make(map[string]*Candidate)
+	selfStakingBuckets := make(map[uint64]bool)
+	for _, cand := range cands {
+		candm[cand.Owner.String()] = cand.Clone()
+		candm[cand.Owner.String()].Votes = new(big.Int)
+		candm[cand.Owner.String()].SelfStake = new(big.Int)
```
yutongp, line 72: Doesn't seem like self stake amount is set to some none zero value anywhere else

action/protocol/staking/vote_reviser.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, err
+	}
+
+	for _, bucket := range buckets {
+		if bucket.UnstakeStartTime.Unix() != 0 && bucket.UnstakeStartTime.After(bucket.StakeStartTime) {
+			continue
+		}
+		cand, ok := candm[bucket.Candidate.String()]
+		if !ok {
+			return nil, errors.Errorf("invalid candidate %s of bucket %d", bucket.Candidate.String(), bucket.Index)
```
yutongp, line 86: I would suggest not to return error here. first of all, this shouldn't happen. Second, even this kind of vote exist, it is not a meaningful vote, we can just not count it. I think the last thing we want to do is let it stuck our chain.
I would suggest just log error here

action/protocol/staking/vote_reviser.go
```diff
@@ -0,0 +1,110 @@
...
+		candm[cand.Owner.String()].SelfStake = new(big.Int)
+		selfStakingBuckets[cand.SelfStakeBucketIdx] = true
+	}
+	buckets, _, err := getAllBuckets(sm)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, bucket := range buckets {
+		if bucket.UnstakeStartTime.Unix() != 0 && bucket.UnstakeStartTime.After(bucket.StakeStartTime) {
```
yutongp, line 81: Would suggest we unify a isunstake function

yutongp:
Plz also fix the test. Thanks

action/protocol/staking/vote_reviser.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return cands, nil
+}
+
+func (vr *VoteReviser) flush(height uint64, csm CandidateStateManager) error {
+	cands, ok := vr.cache[height]
+	if !ok {
+		return nil
+	}
+	for _, cand := range cands {
```
dustinxie, line 104: should sort the map before Upsert()?
otherwise the order of entries may change in the batch and affect the digest?

action/protocol/staking/vote_reviser.go
CoderZhi: let's do it after merging with dustin's PR

action/protocol/staking/vote_reviser.go
CoderZhi: this is something which should not happen. If we just log this error, we may miss some critical issue. we need to debate which one is better.

action/protocol/staking/vote_reviser.go
CoderZhi: fixed

action/protocol/staking/receipt_log.go
CoderZhi: fixing the inconsistency of number of topics. skip only after greenland which is larger than fairbankmigrationheight

yutongp:
Approved

action/protocol/staking/protocol.go
```diff
@@ -219,6 +224,18 @@ func (p *Protocol) CreatePreStates(ctx context.Context, sm protocol.StateManager
 		}
 	}
 
+	if !p.voteReviser.NeedRevise(blkCtx.BlockHeight) {
+		return nil
+	}
```
yutongp, line 229: this is incorrect. should just skip revise not the whole function. that is why i was to put this as an internal function.

yutongp:
wait on nightly testing to merge

PR #2372, opened 2020-08-02
disable flag parsing in deprecated cmd

JonSnow47:
disable flag parsing in deprecated cmd
Now `ioctl action deploy/invoke contract <flag> <arguments>` will not show useless information, they will return new cmd examples.

### DIFF
--- a/ioctl/cmd/action/actioninvoke.go
+++ b/ioctl/cmd/action/actioninvoke.go
@@ -28,9 +28,9 @@ var (
 
 // actionInvokeCmd represents the action invoke command
 var actionInvokeCmd = &cobra.Command{
-	Use:   config.TranslateInLang(invokeCmdUses, config.UILanguage),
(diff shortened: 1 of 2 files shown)


raullenchai:
Should remove "ioctl action deploy/invoke" entirely? ```ioctl contract``` is the new command to deal with contract

JonSnow47:
> Should remove "ioctl action deploy/invoke" entirely? `ioctl contract` is the new command to deal with contract

we can remove that in code, but they are helpful for use.

ioctl/cmd/action/actioninvoke.go
```diff
@@ -28,9 +28,10 @@ var (
 
 // actionInvokeCmd represents the action invoke command
 var actionInvokeCmd = &cobra.Command{
```
raullenchai, line 30: pls add a comment here - 
"notify users to use the new ioctl contract command; 
TODO: this command will be deprecated soon"

same above

PR #2374, opened 2020-08-05
enable API for standalone mode

dustinxie:
can be useful for local testing/repro


dustinxie:
default 0.0.0.0 can work as well

Issue #2380, opened 2020-08-06
Provide default accounts when running in STANDALONE mode


simonerom:
**What would you like to be added**:
When running the node in STANDALONE mode, provide a set of pre-generated accounts with some amount of IOTX 

**Why is this needed**:
STANDALONE mode is convenient for developers to easily simulate a full testnet locally on their, pc to be used as a development endpoint. Having at least one (or better a few) default accounts with a IOTX balance is required to test transactions.

raullenchai:
IMO this is possible via configuration -- @dustinxie pls confirm

dustinxie:
i happen to be working on an issue using standalone mode, will have a PR for this soon

@simonerom  can you pls update docs about standalone mode after PR merged

PR #2381, opened 2020-08-06
add test account with initial balance in standalone mode

dustinxie:
for https://github.com/iotexproject/iotex-core/issues/2380

### DIFF
--- a/config/standalone-genesis.yaml
+++ b/config/standalone-genesis.yaml
@@ -1,0 +1,9 @@
+account:
+  initBalances:
+    # overwrite with your test address
+    io1mflp9m6hcgm2qcghchsdqj3z3eccrnekx9p0ms: "100000000000000000000000000000000000"
+blockchain:
+  blockInterval: 5s
+  numSubEpochs: 15
+  timeBasedRotation: true
+  # overwrite heights for testing needs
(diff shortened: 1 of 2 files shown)


config/standalone-genesis.yaml
```diff
@@ -0,0 +1,9 @@
+account:
+  initBalances:
+    # overwrite with your test address
+    io1mflp9m6hcgm2qcghchsdqj3z3eccrnekx9p0ms: "100000000000000000000000000000000000"
```
raullenchai, line 4: shall document the privkey of this addr? so easier for ppl to use

simonerom: We totally need the private key of the test account to be included, possibly in the same file, just as a comment above the address itself?

dustinxie: this is just a placeholder, user should overwrite with their test address

simonerom: Yes makes sense. Yet I would still provide a very first test account with private key included so that I can include it in the docs and The user can get started with sending a transaction immediately, without having to browse fders and edit a config file. As long as it’s a Testnet, should be safe to have a test account with Private key included? Or it isn’t? Then for more complex scenarios the user can learn where to add custom initial accounts.

PR #2262, opened 2020-07-07
Add unit test workingset

WhoAmI233:
Issue #2231
Closes #2231

PR #2392, opened 2020-08-12
add unit test for workingset

dustinxie:
this is to re-create #2262, which cannot merge due to wrong test config


dustinxie:
already had 2350

PR #2394, opened 2020-08-12
[BlockDAO] add FileDAO interface to manage chain db file

dustinxie:
1. this PR only adds the FileDAO interface and move current block read/write code to filedao_legacy.go
2. next PR will handle the new file implementation
3. last PR will take care how to handle chain db file depending on sf.Height(), and starting chain only on statedb

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -465,18 +465,14 @@ func (bc *blockchain) tipInfo() (*protocol.TipInfo, error) {
 			Timestamp: time.Unix(bc.config.Genesis.Timestamp, 0),
 		}, nil
 	}
-	tipHash, err := bc.dao.GetBlockHash(tipHeight)
-	if err != nil {
-		return nil, err
-	}
-	header, err := bc.dao.Header(tipHash)
+	header, err := bc.dao.HeaderByHeight(tipHeight)
 	if err != nil {
 		return nil, err
 	}
 
 	return &protocol.TipInfo{
 		Height:    tipHeight,
-		Hash:      tipHash,
+		Hash:      header.HashBlock(),
 		Timestamp: header.Timestamp(),
 	}, nil
 }
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -26,7 +26,6 @@ import (
 	"github.com/iotexproject/iotex-core/blockchain/blockdao"
 	"github.com/iotexproject/iotex-core/blockchain/genesis"
 	"github.com/iotexproject/iotex-core/config"
-	"github.com/iotexproject/iotex-core/db"
 	"github.com/iotexproject/iotex-core/pkg/lifecycle"
 	"github.com/iotexproject/iotex-core/pkg/log"
 	"github.com/iotexproject/iotex-core/pkg/prometheustimer"
@@ -154,12 +153,7 @@ func BoltDBDaoOption(indexers ...blockdao.BlockIndexer) Option {
 			return nil
 		}
 		cfg.DB.DbPath = cfg.Chain.ChainDBPath // TODO: remove this after moving TrieDBPath from cfg.Chain to cfg.DB
-		bc.dao = blockdao.NewBlockDAO(
-			db.NewBoltDB(cfg.DB),
-			indexers,
-			cfg.Chain.CompressBlock,
-			cfg.DB,
-		)
+		bc.dao = blockdao.NewBlockDAO(indexers, cfg.Chain.CompressBlock, cfg.DB)
 		return nil
 	}
 }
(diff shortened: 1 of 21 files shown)


blockindex/indexbuilder.go
```diff
@@ -151,30 +151,6 @@ func (ib *IndexBuilder) init() error {
 	if startHeight >= tipHeight {
 		// successfully migrated to latest block
 		zap.L().Info("Finished migrating DB", zap.Uint64("height", startHeight))
-		return ib.purgeObsoleteIndex()
```
dustinxie, line 154: code used for index db separation (about 1 year ago), not needed now

blockchain/blockchain.go
```diff
@@ -278,7 +278,7 @@ func (bc *blockchain) TipHash() hash.Hash256 {
 func (bc *blockchain) TipHeight() uint64 {
 	tipHeight, err := bc.dao.Height()
 	if err != nil {
-		log.L().Panic("failed to get tip height", zap.Error(err))
+		log.L().Error("failed to get tip height", zap.Error(err))
```
yutongp, line 281: what can we do instead of panic if we can not get tip height?

yutongp:
I assume this diff only does two things:
1. create a new fileDao interface
2. move existing code to fileDaolegacy implemention.

if that's the case, I think the diff should be safe. but the diff is big, so anything in this diff which is not 1 and 2, i don't know if that part of code is safe or not.

blockchain/blockchain.go
dustinxie: changed it back

there is a wrong e2etest, it is root-caused and fixed

dustinxie:
> I assume this diff only does two things:
> 
> 1. create a new fileDao interface
> 2. move existing code to fileDaolegacy implemention.
> 
> if that's the case, I think the diff should be safe. but the diff is big, so anything in this diff which is not 1 and 2, i don't know if that part of code is safe or not.

that is right, the diff has only 1 and 2

server/itx/server.go
```diff
@@ -221,6 +217,11 @@ func StartServer(ctx context.Context, svr *Server, probeSvr *probe.Server, cfg c
 	if err := adminserv.Shutdown(ctx); err != nil {
 		log.L().Error("Error when serving metrics data.", zap.Error(err))
 	}
+	if cfg.System.HeartbeatInterval > 0 {
+		if err := task.Stop(ctx); err != nil {
+			log.L().Panic("Failed to stop heartbeat routine.", zap.Error(err))
+		}
+	}
```
dustinxie, line 224: Stop() should be called in the LIFO order of Start()

task starts after svr (and used svr), so should stop before svr

e2etest/local_transfer_test.go
```diff
@@ -312,28 +311,29 @@ func TestLocalTransfer(t *testing.T) {
...
 	defer func() {
-		time.Sleep(10 * time.Second)
 		require.NoError(probeSvr.Stop(ctx))
-		require.NoError(svr.Stop(ctx))
 	}()
 
+	// Start server
+	ctx, cancel := context.WithTimeout(ctx, 32*time.Second)
+	defer cancel()
+	go itx.StartServer(ctx, svr, probeSvr, cfg)
```
dustinxie, line 328: should use a timeout context, which calls svr.Stop() when done

currently, the svr is never stopped

e2etest/local_transfer_test.go
```diff
@@ -312,28 +311,29 @@ func TestLocalTransfer(t *testing.T) {
...
+
 	// target address for grpc connection. Default is "127.0.0.1:14014"
 	grpcAddr := fmt.Sprintf("127.0.0.1:%d", apiPort)
 	grpcctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 	defer cancel()
 	conn, err := grpc.DialContext(grpcctx, grpcAddr, grpc.WithBlock(), grpc.WithInsecure())
 	require.NoError(err)
+	defer conn.Close()
```
dustinxie, line 336: close the connection

server/itx/server.go
```diff
@@ -174,6 +174,11 @@ func StartServer(ctx context.Context, svr *Server, probeSvr *probe.Server, cfg c
 		log.L().Fatal("Failed to start server.", zap.Error(err))
 		return
 	}
+	defer func() {
+		if err := svr.Stop(ctx); err != nil {
+			log.L().Panic("Failed to stop server.", zap.Error(err))
+		}
+	}()
```
dustinxie, line 181: stop() should be called in the LIFO order of start(), so use defer to do that

yutongp: Maybe change task to not to use defer to make it more clear? Just stop them in order at end of the function

dustinxie: i think this way is also clear?

```
a.Start()
defer a.Stop()

b.Start()
defer b.Stop()

c.Start()
defer c.Stop()

<- channel to receive exit or cancel signal
```
then we don't need to worry about the exact order. defer is exactly designed to handle this kind of usage.

dustinxie:
with the upcoming change, BlockDAO will handle multiple KVStore, so remove db.KVStore input from NewBlockDAO(), let FileDAO handle that

PR #2396, opened 2020-08-13
reorg staking/protocol.go

CoderZhi:
Load parameters in real time.


action/protocol/staking/protocol.go
```diff
@@ -64,84 +64,55 @@ type (
...
+		RewardAddress     address.Address
+		Name              string
+		SelfStakingAmount *big.Int
+	}
+
 	// Protocol defines the protocol of handling staking
 	Protocol struct {
 		addr               address.Address
-		depositGas         DepositGas
-		config             Configuration
```
dustinxie, line 71: are we expecting the config to dynamically change (possibly on a per-block basis)?

the config are constants that would affect the staking behavior, any change to the config would be a hard-fork, and hence any such change will only take effect on next fork height

so i don't see the benefits of passing in the config thru ctx on a per-block basis, the current way of one-time read at NewProtocol() is both proper and good enough

dustinxie: and same reasoning goes for hu

PR #2404, opened 2020-08-20
range limit api.GetLogs() to 1000 blocks

dustinxie:


api/logfilter.go
```diff
@@ -14,9 +14,9 @@ import (
 
 // LogFilter contains options for contract log filtering.
 type LogFilter struct {
-	stream  iotexapi.APIService_StreamLogsServer
-	errChan chan error
-	*iotexapi.LogsFilter
+	stream   iotexapi.APIService_StreamLogsServer
+	errChan  chan error
+	pbFilter *iotexapi.LogsFilter
```
dustinxie, line 19: fix earlier comment, that embed struct is not clear

PR #2418, opened 2020-08-25
correctly apply FillPercent at bucket level

dustinxie:


blockindex/indexer.go
```diff
@@ -354,9 +352,6 @@ func (x *blockIndexer) commit() error {
 	if err := x.tac.AddTotalSize(); err != nil {
 		return err
 	}
-	if err := db.CommitWithFillPercent(x.kvStore, x.cBatch, 1.0); err != nil {
-		return err
-	}
```
dustinxie, line 359: we can use x.batch, no need for a separate x.cBatch

db/batch/batch_impl.go
```diff
@@ -133,6 +136,17 @@ func (b *baseKVStoreBatch) Translate(wit WriteInfoTranslate) KVStoreBatch {
 	return c
 }
 
+func (b *baseKVStoreBatch) CheckFillPercent(ns string) (float64, bool) {
+	p, ok := b.fill[ns]
+	return p, ok
```
yutongp, line 141: why read doesn't need rlock?

db/counting_index.go
```diff
@@ -201,6 +204,7 @@ func (c *countingIndex) UseBatch(b batch.KVStoreBatch) error {
 		return ErrInvalid
 	}
 	c.batch = b
+	c.batch.AddFillPercent(c.bucket, 1.0)
```
yutongp, line 207: this changes previous logic

db/batch/batch_impl.go
dustinxie: read is actually called in WriteBatch(), where the lock is obtained before writing all entries to DB
so cannot rlock here, otherwise it is deadlock

db/counting_index.go
dustinxie: before the batch is committed, AddTotalSize() will be called to update the total size, that is a more proper place to set the FillPercent. so change AddTotalSize() to Finalize() and move it there

db/batch/batch_impl.go
yutongp: in that case, i think it could be more clear to have a separate lock to lock different resources? It is very weird that `CheckFillPercent` and `AddFillPercent` access same resource, one has lock, one doesn't

dustinxie: yeah, fixed in latest push

Issue #1873, opened 2020-02-11
fix TestBlockEpochReward


dustinxie:
**What would you like to be added**:
1. i pushed a commit to origin/storage branch, with it the test can run, but balance did not match
2. continue to root-cause and fix the test

**Why is this needed**:
improve test coverage

raullenchai:
@RuneAkko what is the status of this one?

raullenchai:
Close this as it seems not important

PR #2419, opened 2020-08-25
fixed: TestBlockEpochReward #1873

JonSnow47:
1. fixed the main process
2. comment 2 extra reward, plz confirm(line#379 and line#385)


e2etest/rewarding_test.go
```diff
@@ -340,62 +354,64 @@ func TestBlockEpochReward(t *testing.T) {
...
+						// require.NoError(err)
+						// foundationBonus, err := rps[0].FoundationBonus(ctx, sfs[0])
+						// require.NoError(err)
+						// if epochNum <= foundationBonusLastEpoch {
+						// 	for i := 0; i < numNodes; i++ {
+						// 		rewardAddrStr := identityset.Address(i + numNodes).String()
+						// 		expectAfterEpochBonus := big.NewInt(0).Add(exptUnclaimed[rewardAddrStr], foundationBonus)
+						// 		exptUnclaimed[rewardAddrStr] = expectAfterEpochBonus
+						// 	}
+						// }
```
JonSnow47, line 399: exclude normal epoch reward and foundation bonus

dustinxie: this test is about block/epochReward, i think this just mask the issue rather than fix it?
did you get to the root-cause of why the test is failing (i recall it failed that the reward balance not match)

yutongp: any update on reenable this test?

JonSnow47: yes, this test needs to be updated

PR #2426, opened 2020-09-01
GetAccount return IsContract field

JonSnow47:


dustinxie:
here's a test deploying a contract: https://github.com/iotexproject/iotex-core/blob/master/action/protocol/execution/protocol_test.go#L575-L577

add test code below to assert deployed contract address.IsContract() = true

PR #2436, opened 2020-09-03
avoid multiple mint in one round

JonSnow47:
1. How about avoid multiple endrose in one round
2. I tought may should think about round.IsLocked() state, I'm not clear on this state.
3. Maybe should calibrate something if already minted in this round.
4. TODO: update lastEndorse after sync

### DIFF
--- a/consensus/scheme/rolldpos/endorsementmanager.go
+++ b/consensus/scheme/rolldpos/endorsementmanager.go
@@ -31,7 +31,7 @@ var (
 	statusKey             = []byte("status")
 )
 
-//EndorsedByMajorityFunc defines a function to give an information of consensus status
+// EndorsedByMajorityFunc defines a function to give an information of consensus status
 type EndorsedByMajorityFunc func(blockHash []byte, topics []ConsensusVoteTopic) bool
 
 type endorserEndorsementCollection struct {
@@ -109,6 +109,7 @@ func (ee *endorserEndorsementCollection) Endorsement(
 }
 
 type blockEndorsementCollection struct {
+	isMinted  bool
 	blk       *block.Block
 	endorsers map[string]*endorserEndorsementCollection
 }
@@ -138,6 +139,7 @@ func (bc *blockEndorsementCollection) fromProto(blockPro *endorsementpb.BlockEnd
 		}
 		bc.endorsers[endorsement.Endorser] = ee
 	}
+	// TODO: bc.isMinted = blockPro.IsMinted
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -149,6 +152,7 @@ func newRollDPoSCtx(
 		roundCalc:         roundCalc,
 		eManagerDB:        eManagerDB,
 		toleratedOvertime: toleratedOvertime,
+		lastMinted:        chain.TipHeight(),
```
CoderZhi, line 155: lastMintHeight: 0,

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -578,11 +582,15 @@ func (ctx *rollDPoSCtx) Active() bool {
...
 }
 
-///////////////////////////////////////////
+// /////////////////////////////////////////
 // private functions
-///////////////////////////////////////////
+// /////////////////////////////////////////
 
 func (ctx *rollDPoSCtx) mintNewBlock() (*EndorsedConsensusMessage, error) {
+	if ctx.round.Height() != ctx.lastMinted+1 {
```
CoderZhi, line 590: I would suggest we do the checking in https://github.com/iotexproject/iotex-core/pull/2436/files#diff-35b2b97589895506bc967676ee6f4ca9L333, and the parameter should be stored in a local db, like eManagerDB

CoderZhi: moreover, the condition could be a boolean `minted`, because it is per round.

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -109,6 +109,7 @@ func (ee *endorserEndorsementCollection) Endorsement(
 }
 
 type blockEndorsementCollection struct {
+	isMinted  bool
```
CoderZhi, line 112: you may have some misunderstanding of what endorsement collection is.

consensus/scheme/rolldpos/roundctx.go
```diff
@@ -43,6 +43,7 @@ type roundCtx struct {
 	blockInLock []byte
 	proofOfLock []*endorsement.Endorsement
 	status      status
+	lastMinted  *block.Block
```
JonSnow47, line 46: maybe this field should move into endorsementManager

JonSnow47:
@CoderZhi plz review the last commit

Issue #1770, opened 2019-12-05
New ioctl command for alias set and unit test


CoderZhi:
Refer to https://github.com/iotexproject/iotex-core/pull/1751 for example

CoderZhi:
or refer to https://github.com/iotexproject/iotex-core/pull/1840

PR #2449, opened 2020-09-15
new ioctl cmd alias set and unit test

RuneAkko:
#1770


ioctl/newcmd/alias/aliasset.go
```diff
@@ -0,0 +1,72 @@
...
+			for aliases[addr] != "" {
+				delete(c.Config().Aliases, aliases[addr])
+				aliases = GetAliasMap()
+			}
+			c.Config().Aliases[alias] = addr
+			out, err := yaml.Marshal(c.Config())
+			if err != nil {
+				return output.NewError(output.SerializationError, "failed to marshal config", err)
+			}
+			if err := c.WriteFile(config.DefaultConfigFile, out, 0600); err != nil {
```
yutongp, line 63: do we support changing config file path? if so, here we shall write to user defined config file as well?

ioctl/newcmd/alias/aliasset.go
```diff
@@ -0,0 +1,72 @@
...
+		config.English: "Set alias for address",
+		config.Chinese: "设定地址的别名",
+	}
+	setCmdUses = map[config.Language]string{
+		config.English: "set ALIAS ADDRESS",
+		config.Chinese: "set 别名 地址",
+	}
+)
+
+// NewAliasSet represents the alias set command
```
CoderZhi, line 33: wrong comment

ioctl/client.go
```diff
@@ -186,6 +190,10 @@ func (c *client) GetAliasMap() map[string]string {
 	return aliases
 }
 
+func (c *client) WriteFile(filename string, data []byte, perm os.FileMode) error {
+	return ioutil.WriteFile(filename, data, perm)
+}
```
CoderZhi, line 195: For client, the function should be `func (c *client) WriteConfig(cfg Config) error`, the config file path and the file mode should be defined and used inside of `client`

ioctl/newcmd/alias/aliasset.go
RuneAkko: yes, some ioctl commands can change config file. The old ioctl command doesn't seem to support user defined config dir. Do I need to add this feature?

yutongp: i think for newcmd, we should use user defined config if that's possible. for legacy ioctl cmd if it doesn't support this feature, it is ok.

ioctl/client.go
RuneAkko: changed.
Now the "cfg Config" is from cmd/config.go, do I need to migrate config.go?

ioctl/newcmd/alias/aliasset.go
RuneAkko: sure. I will add a new config cmd flag about custom dir. Maybe I can do it when migrate cmd/config.go cmd?

CoderZhi (approved):

millken:
can be closed

PR #2452, opened 2020-09-21
[staking] use correct SR height in ConstructBaseView

dustinxie:


yutongp:
discussed offline with @dustinxie  will read height from state instead of from user input

yutongp (approved):

Issue #2375, opened 2020-08-05
Implement self-stake slashing for double producing blocks 


koseoyoung:
Along with already implemented first slashing which is probation for unproductive delegates, we have to design the second slashing mechanism in order to punish double producing delegates. 

1. We need a detection mechanism that detects submitted blocks which have different block hash by the same producers at the same height. 

2. Once detected, such delegate's self-stake should be confiscated depending on predefined percentage or amount.
2-1. should also put on probation status 
2-2. need to decide the exact number for percentage or amount

raullenchai:
cc @CoderZhi

raullenchai:
More realistically, we could implement the detection part first.

raullenchai:
per @CoderZhi, here are the steps:

 - [ ] Make sure our code won't double sign
 - [ ] Detection of DS
 - [ ] Report of DS to the protocol
 - [ ] Slash double-signers
 - [ ] PR'd but not right way

PR #2453, opened 2020-09-22
[Slashing] store minted block in emanagerDB to avoid multiple block proposals in a round 

koseoyoung:
Works on #2375  , taking over PR #2436.
To cover edge cases(e.g., crash in the middle of the round and restart), make sure our protocol does not produce multiple block proposals in the same round. 

1. Add cachedMintedBlk in eManager/change the proto file.
2. Modify MintNewBlock() to use cached minted block.


consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -367,6 +380,21 @@ func (m *endorsementManager) AddVoteEndorsement(
...
+func (m *endorsementManager) MintBlock(blk *block.Block) error {
+	m.cachedMintedBlk = blk
+	if m.eManagerDB != nil {
+		return m.PutEndorsementManagerToDB()
+	}
+	return nil
+}
+
+func (m *endorsementManager) CachedMintedBlock() *block.Block {
+	if m.cachedMintedBlk != nil {
```
yutongp, line 392: just return `m.cachedMintedBlk`?

consensus/scheme/rolldpos/endorsementmanager.go
```diff
@@ -375,6 +400,7 @@ func (m *endorsementManager) Cleanup(timestamp time.Time) error {
 	} else {
 		m.collections = map[string]*blockEndorsementCollection{}
 	}
+	m.cachedMintedBlk = nil
```
CoderZhi, line 403: every time, we "Prepare" a new round, the cachedMintedBlk will be clean up. So, this implementation may not work.

koseoyoung: Thanks for pointing it out! I fixed this issue by adding conditions when we clean up endorsementmanager. (Clean up only when the cached block timestamp is before round start time)

CoderZhi (commented):
Finally, let's add some unit test to cover different cases

consensus/scheme/rolldpos/rolldposctx_test.go
```diff
@@ -265,6 +260,85 @@ func TestCheckBlockProposer(t *testing.T) {
...
+
+	res, err := rctx.Proposal()
+	require.NoError(err)
+	ecm, ok := res.(*EndorsedConsensusMessage)
+	require.True(ok)
+	hash1, err := ecm.Document().Hash()
+	require.NoError(err)
+	height1 := ecm.Height()
+
+	res2, err := rctx.Proposal()
```
CoderZhi, line 331: is it possible that the proposer has been switched between these two ".Proposal()" calls?

CoderZhi (approved):
The others look good, just one concern of the unit test

consensus/scheme/rolldpos/rolldposctx_test.go
koseoyoung: In this unit test, I assumed the case that the same proposer's node crashes and restart, so call Proposal() again. And make sure using cache from eManager.

koseoyoung: So didn't consider proposer switching or round update

koseoyoung: If the proposer has been switched in the middle, the changed proposer will not be the proposer of this round, so will skip it.

PR #2455, opened 2020-09-23
[filedao] add FileTip to avoid writing entire header to chain db every time

dustinxie:


blockchain/filedao/filedao_header.go
```diff
@@ -23,13 +23,17 @@ type (
 		Compressor     string
 		BlockStoreSize uint64
 		Start          uint64
-		End            uint64
-		TopHash        hash.Hash256
+	}
+
+	// FileTip is tip info of chain
+	FileTip struct {
```
dustinxie, line 29: only these 2 need to be updated on each block commit

the header should only be written once in the entire lifetime of chain db file (first time Start() after new file is created)

blockchain/filedao/filedao_v2_util.go
```diff
@@ -57,13 +57,13 @@ func (fd *fileDAOv2) putTipHashHeightMapping(blk *block.Block) error {
 	heightValue := byteutil.Uint64ToBytesBigEndian(height)
 	fd.batch.Put(blockHashHeightMappingNS, hashKey(h), heightValue, "failed to put hash -> height mapping")
 
-	// update file header
-	if height > fd.header.End {
-		ser, err := fd.header.createHeaderBytes(height, h)
+	// update file tip
+	if height > fd.tip.Height {
+		ser, err := (&FileTip{Height: height, Hash: h}).Serialize()
```
dustinxie, line 62: only need to update new height/hash
no need to write header again

yutongp (approved):

Issue #2458, opened 2020-09-24
factory.State() 


dustinxie:
**What would you like to be added**:
noticed this when checking archive mode cannot read pre-greenland heights issue
- [x] factory.State() should access two layer trie to read a state, instead of directly reading <k, v> from sf.dao
- [x] refer to https://github.com/dustinxie/iotex-core/commit/8bcae165ff14901d0afec37b9e65053a5e7492ad, this commit passed `make test`

**Why is this needed**:
if this is a bug, we should 
- [ ] fix before it is exposed
- [ ] understand why it escaped tests, add a test to capture this

dustinxie:
after checking, this is not a bug, a copy of current state is stored in sf.dao as a direct <k, v>

```go
putStateFunc: func(ns string, key []byte, s interface{}) error {
			ss, err := state.Serialize(s)
			if err != nil {
				return errors.Wrapf(err, "failed to convert account %v to bytes", s)
			}
			flusher.KVStoreWithBuffer().MustPut(ns, key, ss)
			nsHash := hash.Hash160b([]byte(ns))

			return tlt.Upsert(nsHash[:], toLegacyKey(key), ss)
		},
```

PR #2459, opened 2020-09-25
fix State func in factory to read from not sf.dao but trie 

koseoyoung:


dustinxie:
after checking, this is not a bug, refer to https://github.com/iotexproject/iotex-core/issues/2458

PR #2464, opened 2020-10-03
[filedao] add cfg.DB.V2BlocksToSplitDB to split new file

dustinxie:


blockchain/filedao/filedao.go
```diff
@@ -248,30 +248,106 @@ func (fd *fileDAO) PutBlock(ctx context.Context, blk *block.Block) error {
...
 	fds := []*fileDAOv2{}
-	_, v2Files := checkAuxFiles(cfg.DbPath, FileV2)
+	v2Top, v2Files := checkAuxFiles(cfg.DbPath, FileV2)
 	if legacy {
 		legacyFd, err := newFileDAOLegacy(cfg)
 		if err != nil {
 			return nil, err
 		}
 		fd.legacyFd = legacyFd
+		fd.topIndex, _ = checkAuxFiles(cfg.DbPath, FileLegacyAuxiliary)
```
dustinxie, line 348: for example, 2 legacy files  chain-00000001.db, chain-00000002.db
then fd.topIndex = 2

blockchain/filedao/filedao.go
```diff
@@ -281,9 +357,14 @@ func CreateFileDAO(legacy bool, gen genesis.Genesis, cfg config.DB) (FileDAO, er
...
-		cfg.DbPath = name
-		fds = append(fds, openFileDAOv2(cfg))
+	if len(v2Files) > 0 {
+		for _, name := range v2Files {
+			cfg.DbPath = name
+			fds = append(fds, openFileDAOv2(cfg))
+		}
+
+		// v2 file's top index overrides v1's top
+		fd.topIndex = v2Top
```
dustinxie, line 367: for example, we also have 2 v2 files chain-00000003.db, chain-00000004.db
then fd.topIndex = 4

blockchain/filedao/filedao_v2_manager.go
```diff
@@ -90,6 +90,24 @@ func (fm *FileV2Manager) FileDAOByHeight(height uint64) FileDAO {
...
+func (fm *FileV2Manager) AddFileDAO(fd *fileDAOv2, start uint64) error {
+	// update current top's end
+	top := fm.Indices[len(fm.Indices)-1]
+	end, err := top.fd.Height()
+	if err != nil {
+		return err
+	}
+	if start != end+1 {
+		// new file's first block != current tip height
+		return ErrInvalidTipHeight
```
dustinxie, line 103: check and add next new v2 file, for example:
current top file start = 1000, end = 1500, then only `AddFileDAO(1501)` will be accepted

blockchain/filedao/testing.go
```diff
@@ -0,0 +1,227 @@
...
+type (
+	// testInMemFd is an in-memory FileDAO
+	testInMemFd struct {
+		*fileDAOv2
+	}
+
+	// testSplitFile will split new db file at certain heights
+	testSplitFile struct {
+		*fileDAO
+		splitHeight []uint64
```
dustinxie, line 35: you can input a slice []heights to let it split at these heights, used in testing

config/config.go
```diff
@@ -193,7 +193,7 @@ var (
 			SQLITE3: SQLITE3{
 				SQLite3File: "./explorer.db",
 			},
-			SplitDBSizeMB:         0,
+			SplitDBSizeMB:         5000,
```
dustinxie, line 196: by default, split file once its size reaches 5GB

blockchain/filedao/filedao.go
```diff
@@ -248,30 +248,102 @@ func (fd *fileDAO) PutBlock(ctx context.Context, blk *block.Block) error {
...
+	if fd.v2Fd != nil {
+		if err = fd.v2Fd.AddFileDAO(v2, height); err != nil {
+			return err
+		}
+		err = v2.Start(ctx)
+		return err
+	}
+
+	// create v2 manager
+	fd.v2Fd, _ = newFileV2Manager([]*fileDAOv2{v2})
```
yutongp, line 320: why ignore the error here?

blockchain/filedao/filedao.go
```diff
@@ -248,30 +248,102 @@ func (fd *fileDAO) PutBlock(ctx context.Context, blk *block.Block) error {
...
+	if fd.topIndex > 0 {
+		filename = kthAuxFileName(fd.cfg.DbPath, fd.topIndex)
+	}
+
+	dat, err := os.Stat(filename)
+	if err != nil {
+		return false, err
+	}
+	return uint64(dat.Size()) > fd.cfg.SplitDBSize(), nil
+}
```
yutongp, line 291: I think there are better approach for this. it adds additional disk io to check file status for every put which is quite unnecessary

blockchain/filedao/filedao_v2_manager.go
```diff
@@ -90,6 +90,24 @@ func (fm *FileV2Manager) FileDAOByHeight(height uint64) FileDAO {
...
+	}
+
+	top.end = end
+	fm.Indices = append(fm.Indices, &fileV2Index{fd: fd, start: start, end: end})
```
yutongp, line 107: end is set to start -1 now

blockchain/filedao/filedao.go
dustinxie: `newFileV2Manager ` will only err when input is nil, which is not the case here

blockchain/filedao/filedao_v2_manager.go
dustinxie: yes, `end` is never used, until being updated next time a new file is added

see at beginning of
```go
func (fm *FileV2Manager) FileDAOByHeight(height uint64) FileDAO {
	right := len(fm.Indices) - 1
	if height >= fm.Indices[right].start {
		return fm.Indices[right].fd
	}
```
if input height >= [right].start, it knows height belongs to top fd and returns, `[right].end` is not used for comparison

blockchain/filedao/filedao.go
dustinxie: changed to count blocks

blockchain/filedao/filedao_v2_manager.go
```diff
@@ -60,7 +74,12 @@ func (fm *FileV2Manager) Start(ctx context.Context) error {
 		fm.Indices[i].start = start
 		fm.Indices[i].end = end
 	}
-	sort.Slice(fm.Indices, func(i, j int) bool { return fm.Indices[i].start < fm.Indices[j].start })
+
+	if len(fm.Indices) > 0 {
+		sort.Slice(fm.Indices, func(i, j int) bool { return fm.Indices[i].start < fm.Indices[j].start })
+		start, _ := fm.TopFd().(*fileDAOv2).Bottom()
+		fm.splitHeight = start
```
dustinxie, line 81: if there's v2 file, last split height is top v2 file's start height

blockchain/filedao/filedao_v2_manager.go
```diff
@@ -19,24 +26,31 @@ type (
 
 	// FileV2Manager manages collection of v2 files
 	FileV2Manager struct {
-		lock    sync.RWMutex
-		Indices []*fileV2Index
+		lock        sync.RWMutex
+		topIndex    uint64
```
dustinxie, line 30: it is more natural to handle topIndex here, since new file is created and added to Indices     []*fileV2Index

blockchain/integrity/integrity_test.go
```diff
@@ -1174,7 +1174,7 @@ func TestLoadBlockchainfromDB(t *testing.T) {
 	cfg.ActPool.MinGasPriceStr = "0"
 
 	t.Run("load blockchain from DB w/o explorer", func(t *testing.T) {
-		// testValidateBlockchain(cfg, t)
+		testValidateBlockchain(cfg, t)
```
dustinxie, line 1177: this test can pass, not sure why it's disabled

blockchain/filedao/filedao.go
```diff
@@ -264,24 +261,13 @@ func (fd *fileDAO) prepNextDbFile(height uint64) error {
...
-	return fd.addNewV2File(height)
-}
-
-func (fd *fileDAO) needToSplitDB(height uint64) (bool, error) {
-	filename := fd.cfg.DbPath
-	if fd.topIndex > 0 {
-		filename = kthAuxFileName(fd.cfg.DbPath, fd.topIndex)
-	}
+	fd.lock.Lock()
+	defer fd.lock.Unlock()
```
dustinxie, line 265: only need lock starting here

blockchain/filedao/filedao.go
```diff
@@ -322,7 +309,7 @@ func (fd *fileDAO) DeleteTipBlock() error {
 
 // CreateFileDAO creates FileDAO according to master file
 func CreateFileDAO(legacy bool, cfg config.DB) (FileDAO, error) {
-	fd := fileDAO{cfg: cfg}
+	fd := fileDAO{splitHeight: 1, cfg: cfg}
```
dustinxie, line 312: will be updated in Start() if there are v2 files

blockchain/filedao/filedao.go
```diff
@@ -243,45 +242,109 @@ func (fd *fileDAO) PutBlock(ctx context.Context, blk *block.Block) error {
...
+	}
+
+	// create v2 manager
+	fd.v2Fd, _ = newFileV2Manager([]*fileDAOv2{v2})
+	err = fd.v2Fd.Start(ctx)
+	return err
+}
+
+func (fd *fileDAO) DeleteTipBlock() error {
+	return fd.currFd.DeleteTipBlock()
```
yutongp, line 307: delete need lock"?

yutongp (approved):

PR #2470, opened 2020-10-15
Check LogsBloomFilter before look up receipts in getLogs API

koseoyoung:
As we are doing in https://github.com/iotexproject/iotex-core/blob/master/api/logfilter.go#L45 here, in getLogs API, look up LogsBloomFilter first, and then if it exists, look up the logs in the receipts.


CoderZhi (approved):
After this PR, we need to fix the matching at the receipt/log level. Right now, it is a coarse-grained matching (bloom filter). A better choice is address/topic exact matching.

Issue #2471, opened 2020-10-16
Modify not to get all block info to get header in HeaderByHeight  


koseoyoung:
Previously in HeaderByHeight function (blockdao.go), we only deserialized block header. 
After filedao, we modified to read all block info(header, body, footer) from DB to get block header. 

It might affect performance when getting multiple headers to get logsBloom. (for the API node) 
It would be better if we can somehow change back to read block header only in HeaderByHeight (at least for API node or filedao_legacy)

koseoyoung:
Have we compared the performance of using filedao_legacy and filedao_v2? 
What do you guys think of either making API node to just use filedao_legacy or finding another way to improve filedaoV2 headerByHeight as well? @CoderZhi @yutongp @dustinxie

yutongp:
I think @dustinxie has some data on read performance comparison.
Let's just step back a little bit, do we have the data to support that if we reading the entire block to build the bloom filter, the API node performance(commit blocks, serve requests etc) will be impacted?

P.S. I agree worst case, we just use legacy file format in API nodes

koseoyoung:
In terms of "getLogs" API, by using bloomfilter indexer, API performance does not rely on dao.HeaderByHeight anymore.

PR #2472, opened 2020-10-17
[filedao] improve HeaderFooterByHeight() efficiency

dustinxie:
address #2471


blockchain/filedao/filedao.go
```diff
@@ -204,6 +212,57 @@ func (fd *fileDAO) GetBlockByHeight(height uint64) (*block.Block, error) {
...
+		err error
+	)
+	if fd.v2Fd != nil {
+		if blk, err = fd.v2Fd.GetBlock(hash); err == nil {
+			return &blk.Header, nil
+		}
+	}
+
+	if fd.legacyFd != nil {
+		return fd.legacyFd.Header(hash)
```
dustinxie, line 227: this would only get/deserialize header, instead of entire block

blockchain/filedao/filedao.go
```diff
@@ -204,6 +212,57 @@ func (fd *fileDAO) GetBlockByHeight(height uint64) (*block.Block, error) {
...
+			blk, err := v2.GetBlockByHeight(height)
+			if err != nil {
+				return nil, err
+			}
+			return &blk.Header, nil
+		}
+	}
+
+	if fd.legacyFd != nil {
+		return fd.legacyFd.HeaderByHeight(height)
```
dustinxie, line 244: this would only get/deserialize header, instead of entire block

yutongp:
I don't think this addresses the issue in V2 file format

dustinxie:
> I don't think this addresses the issue in V2 file format

right, V2 file store the entire block together, to get the header you'll need to get/deserialize the entire block.
in other words, this issue is specific only for legacy file

yutongp:
then in that case we may have performance concern on v2 file for building bloom filter

dustinxie:
> then in that case we may have performance concern on v2 file for building bloom filter

we did comprehensive tests to try out diff storage options for v2 file (whether or not to store multiple blocks together, how many blocks, use Snappy or Gzip, etc.), and reach the conclusion to use current v2 file implementation, which has a better performance in terms of read/write access time than legacy file

again, this issue is specific to legacy file, that HeaderByHeight() is not implemented with the best way in first place, so this PR is to correct that

koseoyoung:
Can we make API node to only use filedao_legacy? 
I think there is an issue regarding splitting filedao and querying old data.
If a node splits the DB and has both legacy(e.g., 0-7M) and filedaov2(e.g., 7M-8M), and then query old height data(e.g., headerByHeight(6M)),  it is supposed to read from legacy filedao but it looks like reading from filedaov2 for now.

koseoyoung (approved):
We can either make API node to just use filedao_legacy or find another way to improve filedaoV2 headerByHeight as well. For the filedao_legacy improvement, LGTM

PR #2483, opened 2020-10-23
implement bloomfilterIndexer for getAllLogs API 

koseoyoung:
implemented basic idea first, I will send out 2nd PR to optimize the performance and need to do benchmark test in the future
- store both block-level BF and range-level BF in bloomfilter indexer.
- make use of range-level BF for "getAllLogs API"
- make use of block-level BF for "getLogsInBlock" (new BF will include Log.Address and Topic position-sensitive feature)


blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,174 @@
...
+// NewBloomfilterIndexer creates a new bloomfilterindexer struct by given kvstore and rangebloomfilter size
+func NewBloomfilterIndexer(kv db.KVStore, rangeSize uint64) (BloomFilterIndexer, error) {
+	if kv == nil {
+		return nil, errors.New("empty kvStore")
+	}
+	x := bloomfilterIndexer{
+		kvStore:   kv,
+		rangeSize: rangeSize,
+	}
+	return &x, nil
```
yutongp, line 56: nit
```
return &bloomfilterIndexer....
```

api/api.go
```diff
@@ -1485,6 +1488,52 @@ func (api *Server) getLogsInBlock(filter *LogFilter, start, count uint64) ([]*io
 	return logs, nil
 }
 
+// TODO: improve using goroutine
+func (api *Server) getLogsInRange(filter *LogFilter, start, end uint64) ([]*iotextypes.Log, error) {
```
yutongp, line 1492: I think we should still have constrain on the range of the query and the number of logs to return. it is not great for the node if user request to scan 1-7m block and got 1m logs back to return.

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,174 @@
...
+	// CurrentHeightKey indicates the key of current bf indexer height in underlying DB
+	CurrentHeightKey = "CurrentHeight"
+)
+
+type (
+	// BloomFilterIndexer is the interface for bloomfilter indexer
+	BloomFilterIndexer interface {
+		blockdao.BlockIndexer
+		RangeBloomFilterSize() uint64
+		RangeBloomFilter(uint64) (bloom.BloomFilter, error)
```
CoderZhi, line 35: Comments should be added to public struct/interface, so we need to move the comments up to here.

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,174 @@
...
+func (bfx *bloomfilterIndexer) Start(ctx context.Context) (err error) {
+	if err := bfx.kvStore.Start(ctx); err != nil {
+		return err
+	}
+	bfx.mutex.Lock()
+	defer bfx.mutex.Unlock()
+	bfx.curRangeBloomfilter, err = bloom.NewBloomFilter(2048, 3)
+	if err != nil {
+		return err
+	}
```
CoderZhi, line 69: on start, it should be loaded from db if it is in the middle of a range

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// RangeBloomFilter reads rangebloomfilter by given block number from underlying DB
+func (bfx *bloomfilterIndexer) RangeBloomFilter(blockNumber uint64) (bloom.BloomFilter, error) {
+	bfBytes, err := bfx.kvStore.Get(BloomFilterNamespace, byteutil.Uint64ToBytes(blockNumber))
+	if err != nil {
+		return nil, err
+	}
+	return bloom.BloomFilterFromBytes(bfBytes, 2048, 3)
+}
```
CoderZhi, line 146: as mentioned above, we should not expose this function.

blockindex/bloomfilterindexer.go
koseoyoung: Because it is an indexer, if it was in the middle of a range, when start(), it will check underlying DB tip Height which is smaller than dao.tipheight. And it will automatically catch up by putBlock() up to the blockdao's tip (it will regenerate the in-memory current bloomfilter).

api/api.go
koseoyoung: Yes I agree, we should have a limit in API for the number of return values.

api/logfilter.go
```diff
@@ -134,11 +135,84 @@ func (l *LogFilter) ExistInBloomFilter(filter bloom.BloomFilter) bool {
...
+	if len(l.pbFilter.Topics) > 0 {
+		for _, e := range l.pbFilter.Topics {
+			if e == nil || len(e.Topic) == 0 {
+				continue
+			}
+			for _, v := range e.Topic {
+				if bf.Exist(v) {
+					return true
+				}
+			}
```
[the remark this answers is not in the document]
koseoyoung: Can you more elaborate on position-sensitive topic filtering?

api/api.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/iotexproject/iotex-core/action/protocol/poll"
 	"github.com/iotexproject/iotex-core/action/protocol/rolldpos"
 	"github.com/iotexproject/iotex-core/actpool"
+	l "github.com/iotexproject/iotex-core/api/logfilter"
```
yutongp, line 42: l is more like a local var name instead of a package name, consider use the full `logfilter` instead

api/api.go
```diff
@@ -1469,19 +1473,75 @@ func (api *Server) getLogsInBlock(filter *LogFilter, start, count uint64) ([]*io
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	return filter.MatchLogs(receipts), nil
+}
+
+// TODO: improve using goroutine
+func (api *Server) getLogsInRange(filter *l.LogFilter, start, end uint64) ([]*iotextypes.Log, error) {
+	if start > 0 {
+		start = 0 // from the beginning
+	}
```
yutongp, line 1506: i am confused here. say a contract have more than `RangeQueryLimit` number of logs, so we want to query range by range. but with this check which force the start to be 0, we cannot query range by range then

api/api.go
koseoyoung: yes, I addressed it.

api/api.go
koseoyoung: I addressed your comment on the latest commit. Now RangeQuery does not force the start height to be 0.

api/logfilter.go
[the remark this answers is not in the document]
koseoyoung: Addressed on the latest comimit 025f7d675b776930b46536ebef07e50e037b1b4b

api/logfilter/logfilter.go
```diff
@@ -134,11 +140,90 @@ func (l *LogFilter) ExistInBloomFilter(filter bloom.BloomFilter) bool {
...
+		return true
+	}
+
+	// {} or nil matches any address or topic list
+	return true
+}
+
+// SelectBlocksFromRangeBloomFilter filters out RangeBloomFilter for selecting block numbers which have logFilter's topics/address
+// TODO[dorothy]: optimize using goroutine
+func (l *LogFilter) SelectBlocksFromRangeBloomFilter(bf bloom.BloomFilter, start, end uint64) []uint64 {
```
koseoyoung, line 191: This function is only height-sensitive. 

This is because for range level-bloomfilter and block level bloomfilter, we already check topic's position-sensitive filtering via ```ExistInBloomFilterv2```

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+		queryHeight += bfx.rangeSize
+	}
+
+	if end%bfx.rangeSize != 0 {
+		bigBloom, err := bfx.rangeBloomFilter(end)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get rangeBloomFilter from indexer by given height %d", end)
+		}
+		if l.ExistInBloomFilterv2(bigBloom) {
```
koseoyoung, line 179: Topics position-sensitive

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,253 @@
...
+			blocks := l.SelectBlocksFromRangeBloomFilter(bigBloom, end-(end%bfx.rangeSize)+1, end)
```
koseoyoung, line 180: Height-sensitive

api/logfilter/logfilter_test.go
```diff
@@ -69,88 +71,92 @@ var (
...
 		},
 		{
 			&action.Log{
 				Address: "topicN",
-				Topics:  []hash.Hash256{data1, dataB},
+				Topics:  []hash.Hash256{topic1, topicB},
 			},
 			[5]bool{true, false, true, false, false},
 			[5]bool{true, true, true, true, true},
+			[5]bool{true, false, true, false, false},
```
koseoyoung, line 132: bfv2 exist check is more accurate than bfv1 (since address/topic sensitive information is added on)

yutongp (commented):
overall LGTM

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,253 @@
...
+func (bfx *bloomfilterIndexer) BloomFilterByHeight(height uint64) (bloom.BloomFilter, error) {
+	return bfx.blockBloomFilter(height)
+}
+
+// FilterBlocksInRange returns the block numbers by given logFilter in range (start, end]
+func (bfx *bloomfilterIndexer) FilterBlocksInRange(l *filter.LogFilter, start, end uint64) ([]uint64, error) {
+	bfx.mutex.RLock()
+	defer bfx.mutex.RUnlock()
+	blockNumbers := make([]uint64, 0)
+	if start%bfx.rangeSize != 0 {
```
CoderZhi, line 158: It doesn't matter whether the mod result is 0 or not. we just need to pick the beginning height of the range which contains `start`

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil, errors.New("Invalid arguments")
+	}
+	queryHeight := start + bfx.rangeSize
+	for queryHeight <= end {
+		bigBloom, err := bfx.rangeBloomFilter(queryHeight)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get rangeBloomFilter from indexer by given height %d", queryHeight)
+		}
+		if l.ExistInBloomFilterv2(bigBloom) {
+			blocks := l.SelectBlocksFromRangeBloomFilter(bigBloom, queryHeight-bfx.rangeSize+1, queryHeight)
```
CoderZhi, line 168: if a block number is smaller than `start`, discard it. same for case larger than `end`

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,253 @@
...
+
+func (bfx *bloomfilterIndexer) addLogsToRangeBloomFilter(ctx context.Context, blockNumber uint64, receipts []*action.Receipt) {
+	Heightkey := append([]byte(filter.BlockHeightPrefix), byteutil.Uint64ToBytes(blockNumber)...)
+
+	for _, receipt := range receipts {
+		for _, l := range receipt.Logs() {
+			bfx.curRangeBloomfilter.Add([]byte(l.Address))
+			bfx.curRangeBloomfilter.Add(append(Heightkey, []byte(l.Address)...)) // concatenate with block number
+			for i, topic := range l.Topics {
+				bfx.curRangeBloomfilter.Add(append(byteutil.Uint64ToBytes(uint64(i)), topic[:]...)) //position-sensitive
```
CoderZhi: we may not need the position+topic at this level.

api/api.go
```diff
@@ -683,13 +687,21 @@ func (api *Server) GetLogs(
 		if req.Count > 1000 {
 			return nil, status.Error(codes.InvalidArgument, "maximum query range is 1000 blocks")
 		}
-		startBlock = req.FromBlock
-		count = req.Count
+		startBlock := req.FromBlock
+		count := req.Count
+		logs, err = api.getLogsInRange(logfilter.NewLogFilter(in.GetFilter(), nil, nil), startBlock, startBlock+count-1)
+	case in.GetByLongRange() != nil:
```
CoderZhi, line 693: if this is supported, why should we keep GetByRange which has a limit of count?

api/api.go
```diff
@@ -1469,19 +1481,79 @@ func (api *Server) getLogsInBlock(filter *LogFilter, start, count uint64) ([]*io
...
+	for _, i := range blockNumbers {
+		indexed, err := api.getLogsInBlock(filter, i)
 		if err != nil {
-			return logs, status.Error(codes.InvalidArgument, err.Error())
+			return nil, err
+		}
+		if indexed != nil {
+			logs = append(logs, indexed...)
+		}
+		if len(logs) > int(api.cfg.API.RangeQueryLimit) {
```
CoderZhi, line 1552: if this is the case, we should have a pagination param for query limit

blockindex/bloomfilterindexer.go
```diff
@@ -0,0 +1,253 @@
...
+	bfx.mutex.Lock()
+	defer bfx.mutex.Unlock()
+	bfx.curRangeBloomfilter, err = bloom.NewBloomFilter(2048, 3)
+	if err != nil {
+		return err
+	}
+
+	_, err = bfx.kvStore.Get(BlockBloomFilterNamespace, []byte(CurrentHeightKey))
+	switch errors.Cause(err) {
+	case nil:
```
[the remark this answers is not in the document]
koseoyoung: fixed on the latest commit, thx!

blockindex/bloomfilterindexer.go
koseoyoung: Addressed on the latest commit, thanks! 
I can use this way for ```start```

blockindex/bloomfilterindexer.go
koseoyoung: I discarded if it is smaller than start, but for ```end```, I think I still have to handle this.

api/api.go
koseoyoung: We could unify those two getLogsRequest, pls confirm this! https://github.com/iotexproject/iotex-proto/pull/102

blockindex/bloomfilterindexer.go
```diff
@@ -128,7 +131,10 @@ func (bfx *bloomfilterIndexer) DeleteTipBlock(blk *block.Block) (err error) {
 		return err
 	}
 	if (height-1)%bfx.rangeSize == 0 {
-		bfx.curRangeBloomfilter = nil
+		bfx.curRangeBloomfilter, err = bloom.NewBloomFilter(2048, 3)
+		if err != nil {
+			return err
+		}
```
CoderZhi, line 137: we need to delete the corresponding value in db.

CoderZhi (commented):
the others look fine

blockindex/bloomfilterindexer.go
koseoyoung: In line 130, at delete(), we are deleting the value at height in db, right?

blockindex/bloomfilterindexer.go
```diff
@@ -69,21 +69,23 @@ func (bfx *bloomfilterIndexer) Start(ctx context.Context) error {
...
 	defer bfx.mutex.Unlock()
-	tipHeight, err := bfx.kvStore.Get(BlockBloomFilterNamespace, []byte(CurrentHeightKey))
+	tipHeightData, err := bfx.kvStore.Get(RangeBloomFilterNamespace, []byte(CurrentHeightKey))
 	switch errors.Cause(err) {
 	case nil:
-		bfx.curRangeBloomfilter, err = bfx.rangeBloomFilter(byteutil.BytesToUint64(tipHeight))
-		if err != nil {
-			return err
+		tipHeight := byteutil.BytesToUint64(tipHeightData)
+		if tipHeight%bfx.rangeSize == 0 {
```
koseoyoung, line 76: when restarting in 4096, it initializes ```curRangeBloomfilter```

CoderZhi (approved):

Issue #2485, opened 2020-10-26
implement GetActPoolActions


yutongp:
1. protobuf is defined in this pr: https://github.com/iotexproject/iotex-proto/pull/101/files
2. actpool read api: https://github.com/iotexproject/iotex-core/blob/master/actpool/actpool.go#L48

add GetActPoolActions api handler in api package.
if actionhashes is empty, return all actions in actpool

raullenchai:
@koseoyoung can u pls make sure this has been done

koseoyoung:
I tried several times to query all actions in the action pool via the API, and the result was always empty. I think this is the expected result.

PR #2486, opened 2020-10-27
implement GetActPoolActions

LuGHuaaa:
#2485 
@yutongp


yutongp:
```
exported method Server.GetActPoolActions should have comment or be unexported
```
make sure you run make test before creating a pr

PR #2490, opened 2020-11-01
[ioctl] simplify action.Read()

dustinxie:


ioctl/cmd/action/action.go
```diff
@@ -374,8 +366,12 @@ func Read(contract address.Address, amount *big.Int, bytecode []byte) (string, e
...
 		ctx,
 		&iotexapi.ReadContractRequest{
-			Execution:     exec.Proto(),
-			CallerAddress: caller,
+			Execution: &iotextypes.Execution{
+				Amount:   amount,
+				Contract: contract.String(),
+				Data:     bytecode,
+			},
+			CallerAddress: address.ZeroAddress,
```
koseoyoung, line 374: can we set zero address here? When we read contract (e.g., ERC20 ```balannceOf``` method), we might need a caller address(msg.sender)?

koseoyoung (approved):

ioctl/cmd/action/action.go
dustinxie: discussed offline

PR #2498, opened 2020-11-07
introduce batch in bloomfilter indexer

koseoyoung:
introduce batch in bloomfilterIndexer to optimize DB write


CoderZhi:
problem resolved

Issue #2512, opened 2020-11-19
ioctl contract test's --with-arguments cannot handle param without name


raullenchai:
```
- constant: true
  inputs:
    - internalType: bytes32
      name: ''
      type: bytes32
  name: commitments
  outputs:
    - internalType: bool
      name: ''
      type: bool
  payable: false
  stateMutability: view
  type: function
```

`ioctl contract test function xxxx abc.abi commitments --with-arguments '{"":"b21b78b009d7ebcbc88f9d8986fa199a561b6d6a"}'` won't work.

Let's use `_` when the param name is empty, e.g., 
`ioctl contract test function xxxx abc.abi commitments --with-arguments '{"_":"b21b78b009d7ebcbc88f9d8986fa199a561b6d6a"}'` should work

PR #2515, opened 2020-11-19
ioctl contract handle param without name

Frankonly:
#2512


ioctl/cmd/contract/contract.go
```diff
@@ -141,7 +141,13 @@ func packArguments(targetAbi *abi.ABI, targetMethod string, rowInput string) ([]
 	for _, param := range method.Inputs {
```
dustinxie, line 141: i think this way would be more clear?
```go
if param.Name == "" {
   param.Name = "_"
}
```
before `rowArg, ok := rowArguments[param.Name]`

Frankonly: OK

dustinxie (approved):

PR #2517, opened 2020-11-19
catch unexpected status

CoderZhi:
if solc failed or some other unexpected happens, throw an error to prevent further damage.

### DIFF
--- a/ioctl/cmd/action/action.go
+++ b/ioctl/cmd/action/action.go
@@ -339,6 +339,9 @@ func SendAction(elp action.Envelope, signer string) error {
 
 // Execute sends signed execution transaction to blockchain
 func Execute(contract string, amount *big.Int, bytecode []byte) error {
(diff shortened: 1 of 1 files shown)


ioctl/cmd/action/action.go
```diff
@@ -339,6 +339,9 @@ func SendAction(elp action.Envelope, signer string) error {
 
 // Execute sends signed execution transaction to blockchain
 func Execute(contract string, amount *big.Int, bytecode []byte) error {
+	if len(contract) == 0 && len(bytecode) == 0 {
+		return output.NewError(0, "failed to deploy contract with empty bytecode", nil)
```
dustinxie, line 343: nit:
use output.NewError(output.InputError, xxx) instead of 0?

dustinxie (approved):

raullenchai (approved):

Issue #2516, opened 2020-11-19
Add an optional "amount" for deployment


CoderZhi:
**What would you like to be added**:
add an optional input argument

**Why is this needed**:
https://github.com/iotexproject/iotex-core/blob/6fc60f994391e63ca6f8fff74eacddb6ca568fd8/ioctl/cmd/contract/contractdeploysol.go#L113
amount shouldn't be hard coded to 0

CoderZhi:
same here: https://github.com/iotexproject/iotex-core/blob/6fc60f994391e63ca6f8fff74eacddb6ca568fd8/ioctl/cmd/contract/contractdeploybytecode.go#L64

Frankonly:
The `contract` commands adapt different number of arguments. If adding an option argument to those commands may cause confusion when ioctl identifies arguments. Therefore, I add a flag for initial amount of new deployed contract.

PR #2518, opened 2020-11-20
#2516

Frankonly:
#2516


CoderZhi (approved):

CoderZhi:
good to have, it would be better to have the format consistent with other commands.

Issue #1599, opened 2019-11-01
switch to go std errors wrap


yutongp:
plz refer to https://golang.org/pkg/errors/

raullenchai:
Let's start with one small package and work it out with @yutongp, then scale to other packages.

PR #2526, opened 2020-11-26
switch go std errors to github.com/pkg/errors #1599

millken:
#1599


api/blocklistener_test.go
```diff
@@ -7,10 +7,11 @@
 package api
 
 import (
-	"errors"
 	"testing"
 	"time"
 
+	"github.com/pkg/errors"
+
```
CoderZhi, line 14: delete the empty line
"github.com/pkg/errors" should be grouped with other third party packages
same for the rest

CoderZhi (approved):

Issue #2330, opened 2020-07-24
Start DB before p2p network


CoderZhi:
**What would you like to be added**:
Reorder blockchain and p2p.

**Why is this needed**:
If user starts with chain.db, it takes a long time for state db to catch up. As a result, the node will be kicked out of p2p network once the node is ready to sync.

raullenchai:
cc @Madongming

PR #2529, opened 2020-12-01
start chain services before p2p network #2330 

millken:
#2330


CoderZhi:
https://github.com/iotexproject/iotex-core/pull/2334

PR #2545, opened 2020-12-16
fix TestBlockEpochReward #1873

millken:


e2etest/rewarding_test.go
```diff
@@ -572,48 +589,44 @@ func updateExpectationWithPendingClaimList(
...
 	cfg.Genesis.Blockchain.NumDelegates = numNodes
 	cfg.Genesis.Blockchain.TimeBasedRotation = true
-	cfg.Genesis.Delegates = cfg.Genesis.Delegates[0:numNodes]
-
-	cfg.Genesis.BlockInterval = 500 * time.Millisecond
-	cfg.Genesis.EnableGravityChainVoting = true
-	cfg.Genesis.Rewarding.FoundationBonusLastEpoch = 2
+	cfg.Genesis.Delegates = cfg.Genesis.Delegates[3 : numNodes+3]
+	cfg.Genesis.EnableGravityChainVoting = false
+	cfg.Genesis.PollMode = "lifeLong"
```
dustinxie, line 630: can you explain what is the root-cause and fix?

saw a lot of params changed like 
1. adding "lifelong"
2. EnableGravityChainVoting changed to false
3. removing FoundationBonusLastEpoch = 2
4. block interval and TTL value change to seconds, but these are local tests, the original ~200ms value should allow the test to pass

millken: the config copy form other place.
if we don't "lifeLong"，got error "fail to create chain service: failed to generate poll protocol: unsupported poll mode"

e2etest/rewarding_test.go
```diff
@@ -572,26 +589,23 @@ func updateExpectationWithPendingClaimList(
...
-	indexDBPath string,
 	producerPriKey crypto.PrivateKey,
 	networkPort,
 	apiPort int,
-	numNodes uint64,
+	HTTPAdminPort int,
 ) config.Config {
 	cfg := config.Default
 
+	cfg.Plugins[config.GatewayPlugin] = true
```
dustinxie, line 599: if enable plugin, need to delete it in defer{}, otherwise it might cause problem for later tests.
search "// clear the gateway" you will see

dustinxie:
@millken  can you update the PR per our latest conversation?

e2etest/rewarding_test.go
```diff
@@ -159,6 +156,15 @@ func TestBlockEpochReward(t *testing.T) {
 
 	// Set mini-cluster configurations
 	rand.Seed(time.Now().UnixNano())
+	var delegates []genesis.Delegate
+	for i := 0; i < numNodes; i++ {
+		delegates = append(delegates, genesis.Delegate{
+			OperatorAddrStr: identityset.Address(i + numNodes).String(),
+			RewardAddrStr:   identityset.Address(i).String(),
```
dustinxie, line 163: 1. rewardAddr and operatorAddr got reversed
2. VotesStr = strconv.Itoa(1000 + rand.Intn(1000))

PR #2546, opened 2020-12-16
modify EstimateActionGasConsumption api to return execution revert error

koseoyoung:
After Hawaii height, EstimateActionGasConsumption API will return execution revert msg as well even before sending tx.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	"fmt"
 	"math"
 	"math/big"
 	"net"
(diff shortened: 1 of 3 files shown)


api/api.go
```diff
@@ -1548,7 +1549,10 @@ func (api *Server) estimateActionGasConsumptionForExecution(exec *iotextypes.Exe
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if receipt.Status != uint64(iotextypes.ReceiptStatus_Success) {
-		return nil, status.Error(codes.Internal, "execution simulation gets failure status")
+		if receipt.Status == uint64(iotextypes.ReceiptStatus_ErrExecutionReverted) && receipt.ExecutionRevertMsg() != "" {
```
dustinxie, line 1552: there are 4 other places also have  `if receipt.Status != uint64(iotextypes.ReceiptStatus_Success)`, do you care to modify those as well?

also i think `if receipt.ExecutionRevertMsg() != ""` is enough, since the revert msg is only set when status == 
 ReceiptStatus_ErrExecutionReverted

koseoyoung: we only need to modify bot/../grpcutil/grpc.go, others are used in create genesis (in consortium and staking protocol) 
, and addressed your comment!

dustinxie (approved):

PR #2550, opened 2020-12-22
return error if p2p agent not started

CoderZhi:
return error if p2pAgent.host is nil. This will trigger an error, instead of fatal


server/itx/server.go
```diff
@@ -90,14 +90,14 @@ func newServer(cfg config.Config, testing bool) (*Server, error) {
...
-	if err := s.p2pAgent.Start(cctx); err != nil {
-		return errors.Wrap(err, "error when starting P2P agent")
-	}
 	for _, cs := range s.chainservices {
 		if err := cs.Start(cctx); err != nil {
 			return errors.Wrap(err, "error when starting blockchain")
 		}
 	}
+	if err := s.p2pAgent.Start(cctx); err != nil {
+		return errors.Wrap(err, "error when starting P2P agent")
```
dustinxie, line 99: should we even start p2p after dispatcher? b/c dispatcher handles msg from p2p

dustinxie (approved):

Issue #2551, opened 2020-12-22
make indexer non-blocking


dustinxie:
**What would you like to be added**:
Currently, if indexer's height is not up to tip height, blockdao.Start() will block, waiting for indexer to catch up all the missing blocks. This creates extra delay when booting up the node.

For instance, yesterday i deleted the old file to test the bloomfilter index. When i restart the node, it keeps rebuilding the bloomfilter index for several hours. The user experience is not good.

**New proposal**:
Instead of let blockdao call indexer.PutBlock(), we can start a go-routine for each indexer, which keeps checking/pulling latest block committed to blockdao, and call PutBlock() to index it

By doing so, node can immediately finish Start() and begin to commit new blocks, even though some indexer are not yet synced to tip

This would also help reduce blockdao's commit time, by offloading indexer.PutBlock to other go-routine

raullenchai:
Closing as not-planned. `blockDAO.Start` still catches indexers up synchronously (`checkIndexers`), but this only bites when (re)building an indexer from scratch — a rare, operator-initiated event where blocking is arguably the safer behavior (the node shouldn't serve queries from a half-built index). Making it async adds coordination complexity for no routine benefit.

PR #2552, opened 2020-12-23
use larger bloom filter for indexer

dustinxie:


CoderZhi:
As I remember, we spent a large effort to make sf as a normal indexer. With our current implementation, we already have two ways of building indexers, async and sync. How could this PR be different?

dustinxie:
this is try to solve the issue of slow Start() in https://github.com/iotexproject/iotex-core/issues/2551

sf will be the only special indexer that is sync -- its PutBlock() needs to be called strictly after blockdao.PutBlock().

in next PR, all other indexers will be made async -- they will be added to subscriber, and PutBlock() called on a separate go-routine, similar to indexbuilder.go

finally, sf could possibly be made async as well -- considering our block time is 5 seconds, sf.PutBlock() will most likely finish before commit next block, but for safety it is better to keep sf (the only) sync indexer

koseoyoung (approved):

Issue #2521, opened 2020-11-21
go test -run TestProtocol_ClaimReward


dustinxie:
### What version of iotex-core are you using (`ioctl version`)?
latest master

### What operating system and processor architecture are you using?
macOS

### What did you do? If possible, provide a recipe for reproducing the error.
```go
cd action/protocol/rewarding
go test -run TestProtocol_ClaimReward
```

### What did you expect to see?
test should pass

### What did you see instead?
test fail
if you run `go test` it will pass
but `go test -run TestProtocol_ClaimReward` should pass by itself as well

PR #2555, opened 2020-12-25
fix go test -run TestProtocol_ClaimReward #2521

millken:


action/protocol/rewarding/reward_test.go
```diff
@@ -235,9 +235,11 @@ func TestProtocol_ClaimReward(t *testing.T) {
 		unclaimedBalance, _, err := p.UnclaimedBalance(ctx, sm, claimActionCtx.Caller)
 		require.NoError(t, err)
 		assert.Equal(t, big.NewInt(5), unclaimedBalance)
+		claimActionCtx.Caller = identityset.Address(28)
+
 		primAcc, err := accountutil.LoadAccount(sm, hash.BytesToHash160(claimActionCtx.Caller.Bytes()))
 		require.NoError(t, err)
-		assert.Equal(t, big.NewInt(1000005), primAcc.Balance)
+		assert.Equal(t, big.NewInt(980), primAcc.Balance)
```
dustinxie, line 242: i think we need to find out the root-cause?

as mentioned in #2521, this test can pass in `make test`, but only fails when running as a separate test. must be sth like testConfig() or a local variable used by earlier test but changes when running separately.

the proper fix is to understand why it happened that way and find out the exact root-cause

Liuhaai:
The issue has been solved by PR #2651

PR #2559, opened 2021-01-04
remove position-sensitive in range bf

koseoyoung:
It reduces the number of elements in range bf although it does not have a significant effect.


CoderZhi:
is this still essential after increase the bits of bf?

koseoyoung:
@CoderZhi After the increase the bits of bf, if it starts from 0 height, it takes 16 sec and in the worst case, it takes 45 sec (until max limit). And if it starts from 7M, it takes around 5 sec. If this performance is fine, this PR is not essential.

Issue #2560, opened 2021-01-17
Make iotexscan url clickable


raullenchai:
Using ioctl to initiate a tx, it will return the iotexscan URL like below. We shall make it clickable
```
....
Options: yes
Quit for anything else.
yes
Action has been sent to blockchain.
Wait for several seconds and query this action by hash:iotexscan.io/action/b49e2eca223d47cdbcf0cd9f10a9f6dcf12adff6f82a966ce6f009890aa6a0c0
```
<img width="1000" alt="Screen Shot 2021-01-16 at 7 10 03 PM" src="https://user-images.githubusercontent.com/989846/104829978-78998900-582e-11eb-9407-7b301b70a5e0.png">

PR #2561, opened 2021-01-17
Make iotexscan url clickable #2560

millken:
`ctrl/cmd`+`click`on a URL and it will open in the default browser, or right click on a URL.


millken:
`ctrl/cmd`+`click`on a URL and it will open in the default browser, or right click on a URL.

raullenchai (approved):

PR #2570, opened 2021-02-09
[p2p] re-connect in case of network disconnect

dustinxie:


p2p/agent.go
```diff
@@ -409,6 +370,69 @@ func (p *Agent) Neighbors(ctx context.Context) ([]peerstore.PeerInfo, error) {
...
+			break
+		}
+	}
+	return nil
+}
+
+func (p *Agent) reconnect() {
+	ctx := context.Background()
+	peers, err := p.Neighbors(ctx)
+	if err != nil || len(peers) == 0 {
```
dustinxie, line 430: reconnect if number of peers becomes 0

config/genesis_mainnet.yaml
```diff
@@ -0,0 +1,98 @@
+account:
+  initBalances:
```
raullenchai, line 2: Why do we need this for genesis_mainnet.yaml ?

dustinxie: this file is the exact config we used to bootstrap the mainnet, it is same as https://github.com/iotexproject/iotex-bootstrap/blob/master/genesis_mainnet.yaml

it is only used to create the genesis block at bootstrap time, does not affect later sync and chain operation afterwards

CoderZhi (approved):

PR #2572, opened 2021-02-23
[staking] fix change candidate bug

dustinxie:
When the new candidate is same as old one, staking handler are dealing
with 2 copies of the same candidate: vote of the first being deducted,
and vote of the second being added -- yet the second change overwrites
the first, b/c these 2 are the same candidate. This caused the total
vote to be incorrectly increased.

### DIFF
--- a/action/protocol/staking/handlers.go
+++ b/action/protocol/staking/handlers.go
@@ -313,6 +313,14 @@ func (p *Protocol) handleChangeCandidate(ctx context.Context, act *action.Change
 		}
 	}
 
+	if p.hu.IsPost(config.Hawaii, blkCtx.BlockHeight) && address.Equal(prevCandidate.Owner, candidate.Owner) {
+		// change to same candidate, do nothing
+		return log, &handleError{
+			err:           errors.New("change to same candidate"),
+			failureStatus: iotextypes.ReceiptStatus_ErrCandidateAlreadyExist,
+		}
+	}
+
 	// update bucket index
 	if err := delCandBucketIndex(csm, bucket.Candidate, act.BucketIndex()); err != nil {
 		return log, errors.Wrapf(err, "failed to delete candidate bucket index for candidate %s", bucket.Candidate.String())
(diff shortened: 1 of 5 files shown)


action/protocol/staking/vote_reviser_test.go
```diff
@@ -12,19 +12,14 @@ import (
...
 
 func TestVoteReviser(t *testing.T) {
 	r := require.New(t)
 
-	// make sure the prefix stays constant, they affect the key to store objects to DB
-	r.Equal(byte(0), _const)
-	r.Equal(byte(1), _bucket)
-	r.Equal(byte(2), _voterIndex)
-	r.Equal(byte(3), _candIndex)
-
```
dustinxie, line 27: duplicate test code

action/protocol/staking/handlers.go
```diff
@@ -313,6 +313,11 @@ func (p *Protocol) handleChangeCandidate(ctx context.Context, act *action.Change
 		}
 	}
 
+	if p.hu.IsPost(config.Hawaii, blkCtx.BlockHeight) && address.Equal(prevCandidate.Owner, candidate.Owner) {
```
CoderZhi, line 316: we can do this earlier by comparing bucket.Candidate and candidate.Owner, before loading "prevCandidate".

action/protocol/staking/handlers.go
```diff
@@ -382,6 +387,11 @@ func (p *Protocol) handleTransferStake(ctx context.Context, act *action.Transfer
 	}
 	log.AddTopics(byteutil.Uint64ToBytesBigEndian(bucket.Index), act.VoterAddress().Bytes(), bucket.Candidate.Bytes())
 
+	if p.hu.IsPost(config.Hawaii, blkCtx.BlockHeight) && address.Equal(newOwner, bucket.Owner) {
```
CoderZhi, line 390: what's the point?

chainservice/chainservice.go
```diff
@@ -254,7 +254,12 @@ func New(
 	)
 	// staking protocol need to be put in registry before poll protocol when enabling
 	if cfg.Chain.EnableStakingProtocol {
-		stakingProtocol, err = staking.NewProtocol(rewarding.DepositGas, cfg.Genesis.Staking, candBucketsIndexer, cfg.Genesis.GreenlandBlockHeight)
+		stakingProtocol, err = staking.NewProtocol(
+			rewarding.DepositGas,
+			cfg.Genesis.Staking, candBucketsIndexer,
+			cfg.Genesis.GreenlandBlockHeight,
+			cfg.Genesis.HawaiiBlockHeight,
```
dustinxie, line 261: revise at hawaii height

action/protocol/staking/handlers.go
dustinxie: line 309 above: if bucket is unstaked and after Greenland, it will return a diff error

this check needs to be after that check, otherwise block won't commit in case the tx meets all these conditions

action/protocol/staking/handlers.go
dustinxie: so similar to change candidate, if it's transferring to same owner, then don't need to modify the bucket, and can early exit

action/protocol/staking/handlers.go
dustinxie: actually block can commit, but it's inconsistent behavior for such block before Hawaii (return error in Greenland fix) vs. after Hawaii (return nil)

CoderZhi (approved):

PR #2576, opened 2021-03-01
fixed cant compile in mac m1(go1.16 darwin arm64)

millken:


pkg/log/redirectStderr_darwin_arm64.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2019 IoTeX Foundation
```
raullenchai, line 1: s/2019/20201

raullenchai (approved):

PR #2583, opened 2021-03-16
[log] integrate logger with ECS

dustinxie:


blocksync/buffer.go
```diff
@@ -78,15 +75,12 @@ func (b *blockBuffer) Flush(blk *block.Block) (bool, bCheckinResult) {
...
 		delete(b.blocks, heightToSync)
 		if err := commitBlock(b.bc, b.cs, blk); err != nil && errors.Cause(err) != blockchain.ErrInvalidTipHeight {
-			if errors.Cause(err) == poll.ErrProposedDelegatesLength || errors.Cause(err) == poll.ErrDelegatesNotAsExpected || errors.Cause(err) == db.ErrNotExist {
-				l.Debug("Failed to commit the block.", zap.Error(err), zap.Uint64("syncHeight", heightToSync))
-			} else {
-				l.Error("Failed to commit the block.", zap.Error(err), zap.Uint64("syncHeight", heightToSync))
-			}
+			l.With(
+				zap.Uint64("syncHeight", heightToSync),
+				zap.Strings("actionHash", blk.ActionHashs())).Error("Failed to commit the block.", zap.Error(err))
```
dustinxie, line 80: this if{} is added in https://github.com/iotexproject/iotex-core/commit/2bf61cdd85789412055cd529f49a8db997321e99
looks like to suppress a specific non-critical issue in the past

verified with latest code, this error no longer happens, so remove it and use l.Error() for any error returned by commitBlock()

action/protocol/staking/protocol.go
```diff
@@ -338,7 +339,9 @@ func (p *Protocol) handle(ctx context.Context, act action.Action, csm CandidateS
 	}
 
 	if receiptErr, ok := err.(ReceiptError); ok {
-		log.L().Debug("Non-critical error when processing staking action", zap.Error(err))
+		actionCtx := protocol.MustGetActionCtx(ctx)
```
raullenchai, line 342: this could panic

action/protocol/staking/vote_reviser.go
```diff
@@ -121,5 +121,11 @@ func (vr *VoteReviser) flush(height uint64, csm CandidateStateManager) error {
 			return err
 		}
 	}
+	log.L().Info("committed revise action",
+		zap.Uint64("height", height), zap.Int("number of cands", len(cands)))
+	for i := range cands {
```
raullenchai, line 126: why not log in the for-loop above?

CoderZhi (approved):

blockchain/block/block.go
```diff
@@ -152,3 +153,13 @@ func (b *Block) TransactionLog() *BlkTransactionLog {
 	}
 	return &blkLog
 }
+
+// ActionHashs returns action hashs in the block
+func (b *Block) ActionHashs() []string {
```
raullenchai, line 158: This doesn't below to `Block`

blocksync/buffer.go
raullenchai: Suggest do not touch this snippet of code if not necessary. You didn't see it perhaps b/c it is `Debug` log which is not enabled.

action/protocol/staking/protocol.go
dustinxie: ctx is passed from above with action context
MustGetActionCtx(ctx) is also called in downstream funcs like handleCreateStake(), it should not panic

action/protocol/staking/vote_reviser.go
dustinxie: fixed

PR #2586, opened 2021-03-26
[action] make Envelope an interface

dustinxie:
1. pls focus on 2nd commit, the 1st commit is mostly a clean-up
2. WIP to adjust/fix tests

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -41,9 +41,10 @@ type hasDestination interface {
 
 // Sign signs the action using sender's private key
 func Sign(act Envelope, sk crypto.PrivateKey) (SealedEnvelope, error) {
-	sealed := SealedEnvelope{Envelope: act}
-
-	sealed.srcPubkey = sk.PublicKey()
+	sealed := SealedEnvelope{
+		Envelope:  act,
+		srcPubkey: sk.PublicKey(),
+	}
 
 	hash := act.Hash()
 	sig, err := sk.Sign(hash[:])
@@ -51,7 +52,7 @@ func Sign(act Envelope, sk crypto.PrivateKey) (SealedEnvelope, error) {
 		return sealed, errors.Wrapf(ErrAction, "failed to sign action hash = %x", hash)
 	}
 	sealed.signature = sig
-	sealed.payload.SetEnvelopeContext(sealed)
+	act.SetSealedContext(sealed)
 	return sealed, nil
 }
 
@@ -62,7 +63,7 @@ func FakeSeal(act Envelope, pubk crypto.PublicKey) SealedEnvelope {
 		Envelope:  act,
 		srcPubkey: pubk,
 	}
-	sealed.payload.SetEnvelopeContext(sealed)
+	act.SetSealedContext(sealed)
 	return sealed
 }
 
@@ -74,7 +75,7 @@ func AssembleSealedEnvelope(act Envelope, pk crypto.PublicKey, sig []byte) Seale
 		srcPubkey: pk,
 		signature: sig,
 	}
-	sealed.payload.SetEnvelopeContext(sealed)
+	act.SetSealedContext(sealed)
 	return sealed
 }
 
--- a/action/envelope.go
+++ b/action/envelope.go
@@ -12,23 +12,41 @@ import (
 	"github.com/iotexproject/iotex-core/pkg/util/byteutil"
 )
 
-// Envelope defines an envelope wrapped on action with some envelope metadata.
-type Envelope struct {
-	version  uint32
-	nonce    uint64
-	gasLimit uint64
-	payload  actionPayload
-	gasPrice *big.Int
-}
+type (
+	// Envelope defines an envelope wrapped on action with some envelope metadata.
+	Envelope interface {
+		Version() uint32
+		Nonce() uint64
+		GasLimit() uint64
+		GasPrice() *big.Int
+		Destination() (string, bool)
+		Cost() (*big.Int, error)
+		IntrinsicGas() (uint64, error)
+		Action() Action
+		Hash() hash.Hash256
+		Proto() *iotextypes.ActionCore
+		LoadProto(pbAct *iotextypes.ActionCore) error
+		SetNonce(n uint64)
+		SetSealedContext(SealedEnvelope) error
+	}
+
+	envelope struct {
+		version  uint32
+		nonce    uint64
+		gasLimit uint64
+		gasPrice *big.Int
+		payload  actionPayload
+	}
+)
 
 // Version returns the version
-func (elp *Envelope) Version() uint32 { return elp.version }
+func (elp *envelope) Version() uint32 { return elp.version }
 
 // Nonce returns the nonce
-func (elp *Envelope) Nonce() uint64 { return elp.nonce }
+func (elp *envelope) Nonce() uint64 { return elp.nonce }
 
 // Destination returns the destination address
-func (elp *Envelope) Destination() (string, bool) {
+func (elp *envelope) Destination() (string, bool) {
 	r, ok := elp.payload.(hasDestination)
 	if !ok {
 		return "", false
@@ -38,10 +56,10 @@ func (elp *Envelope) Destination() (string, bool) {
 }
 
 // GasLimit returns the gas limit
-func (elp *Envelope) GasLimit() uint64 { return elp.gasLimit }
+func (elp *envelope) GasLimit() uint64 { return elp.gasLimit }
 
 // GasPrice returns the gas price
-func (elp *Envelope) GasPrice() *big.Int {
+func (elp *envelope) GasPrice() *big.Int {
 	p := &big.Int{}
 	if elp.gasPrice == nil {
 		return p
@@ -50,20 +68,20 @@ func (elp *Envelope) GasPrice() *big.Int {
 }
 
 // Cost returns cost of actions
-func (elp *Envelope) Cost() (*big.Int, error) {
+func (elp *envelope) Cost() (*big.Int, error) {
 	return elp.payload.Cost()
 }
 
 // IntrinsicGas returns intrinsic gas of action.
-func (elp *Envelope) IntrinsicGas() (uint64, error) {
+func (elp *envelope) IntrinsicGas() (uint64, error) {
 	return elp.payload.IntrinsicGas()
 }
 
 // Action returns the action payload.
-func (elp *Envelope) Action() Action { return elp.payload }
+func (elp *envelope) Action() Action { return elp.payload }
 
 // Proto convert Envelope to protobuf format.
-func (elp *Envelope) Proto() *iotextypes.ActionCore {
+func (elp *envelope) Proto() *iotextypes.ActionCore {
 	actCore := &iotextypes.ActionCore{
 		Version:  elp.version,
 		Nonce:    elp.nonce,
@@ -74,8 +92,7 @@ func (elp *Envelope) Proto() *iotextypes.ActionCore {
 	}
 
 	// TODO assert each action
-	act := elp.Action()
-	switch act := act.(type) {
+	switch act := elp.Action().(type) {
 	case *Transfer:
 		actCore.Action = &iotextypes.ActionCore_Transfer{Transfer: act.Proto()}
 	case *Execution:
@@ -113,14 +130,14 @@ func (elp *Envelope) Proto() *iotextypes.ActionCore {
 }
 
 // LoadProto loads fields from protobuf format.
-func (elp *Envelope) LoadProto(pbAct *iotextypes.ActionCore) error {
+func (elp *envelope) LoadProto(pbAct *iotextypes.ActionCore) error {
 	if pbAct == nil {
 		return errors.New("empty action proto to load")
 	}
 	if elp == nil {
 		return errors.New("nil action to load proto")
 	}
-	*elp = Envelope{}
+	*elp = envelope{}
 	elp.version = pbAct.GetVersion()
 	elp.nonce = pbAct.GetNonce()
 	elp.gasLimit = pbAct.GetGasLimit()
@@ -220,20 +237,25 @@ func (elp *Envelope) LoadProto(pbAct *iotextypes.ActionCore) error {
 		}
 		elp.payload = act
 	default:
-		return errors.Errorf("no applicable action to handle in action proto %+v", pbAct)
+		return errors.Errorf("no applicable action to handle proto type %T", pbAct.Action)
 	}
 	return nil
 }
 
-// Serialize returns encoded binary.
-func (elp *Envelope) Serialize() []byte {
+func (elp *envelope) serialize() []byte {
 	return byteutil.Must(proto.Marshal(elp.Proto()))
 }
 
-// Hash returns the hash value of SealedEnvelope.
-func (elp *Envelope) Hash() hash.Hash256 {
-	return hash.Hash256b(elp.Serialize())
+// Hash returns the Envelope's hash value (this hash is to be signed)
+func (elp *envelope) Hash() hash.Hash256 {
+	return hash.Hash256b(elp.serialize())
 }
 
 // SetNonce sets the nonce value
-func (elp *Envelope) SetNonce(n uint64) { elp.nonce = n }
+func (elp *envelope) SetNonce(n uint64) { elp.nonce = n }
+
+// SetSealedContext sets the SealedEnvelope context to envelope
+func (elp *envelope) SetSealedContext(selp SealedEnvelope) error {
+	elp.payload.SetEnvelopeContext(selp)
+	return nil
+}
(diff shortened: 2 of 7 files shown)


action/protocol/execution/evm/evm.go
```diff
@@ -189,7 +189,7 @@ func ExecuteContract(
 		blkCtx.BlockHeight,
 		hu.IsPre(config.Aleutian, blkCtx.BlockHeight),
 		hu.IsPost(config.Greenland, blkCtx.BlockHeight),
-		execution.Hash(),
+		actionCtx.ActionHash,
```
dustinxie, line 192: this is a wrong programming: should use action hash in context

that is why SetEnvelopeContext() need to do
```
act.hash = Selp.Hash()
```

action/protocol/poll/consortium.go
```diff
@@ -113,7 +114,7 @@ func (cc *consortiumCommittee) CreateGenesisStates(ctx context.Context, sm proto
 		return err
 	}
 	actionCtx.Nonce = consortiumCommitteeContractNonce
-	actionCtx.ActionHash = execution.Hash()
+	actionCtx.ActionHash = consortiumCommitteeContractHash
```
dustinxie, line 117: this is also improper programming:
execution.Hash() will be all zero, since it is a raw tx not signed

reason this is not an issue: this code is not run for mainnet/testnet

action/actctx.go
```diff
@@ -63,20 +71,15 @@ func (act *AbstractAction) BasicActionSize() uint32 {
...
 	}
-	ab := &Builder{}
-	*act = ab.SetVersion(selp.Version()).
-		SetNonce(selp.Nonce()).
-		SetSourcePublicKey(selp.SrcPubkey()).
-		SetGasLimit(selp.GasLimit()).
-		SetGasPrice(selp.GasPrice()).
-		Build()
-
-	// the reason to set hash here, after set act context, is because some actions use envelope information in their proto define. for example transfer use des addr as Receipt.
```
dustinxie, line 78: reason we need this is due to wrong programming in evm.go, see comments there

action/protocol/poll/staking_committee.go
```diff
@@ -124,7 +127,7 @@ func (sc *stakingCommittee) CreateGenesisStates(ctx context.Context, sm protocol
 		return err
 	}
 	actionCtx.Nonce = nativeStakingContractNonce
-	actionCtx.ActionHash = execution.Hash()
+	actionCtx.ActionHash = nativeStakingContractHash
```
dustinxie, line 130: same thing: all-zero hash

action/protocol/poll/consortium.go
```diff
@@ -25,6 +25,7 @@ import (
 var (
 	consortiumCommitteeContractCreator = address.ZeroAddress
 	consortiumCommitteeContractNonce   = uint64(0)
+	consortiumCommitteeContractHash, _ = hash.HexStringToHash256("00636f6e736f727469756d436f6d6d6974746565436f6e747261637448617368")
```
raullenchai, line 28: `00636f6e736f727469756d436f6d6d6974746565436f6e747261637448617368` - what is this magic string?

action/protocol/poll/staking_committee.go
```diff
@@ -34,8 +34,11 @@ import (
 	"github.com/iotexproject/iotex-core/state"
 )
 
-var nativeStakingContractCreator = address.ZeroAddress
-var nativeStakingContractNonce = uint64(0)
+var (
+	nativeStakingContractCreator = address.ZeroAddress
+	nativeStakingContractNonce   = uint64(0)
+	nativeStakingContractHash, _ = hash.HexStringToHash256("000000000000006e61746976655374616b696e67436f6e747261637448617368")
```
raullenchai, line 40: why do we need a magic string?

action/protocol/poll/consortium.go
dustinxie: this is the hex of "consortiumCommitteeContractHash"

action/protocol/poll/staking_committee.go
dustinxie: like i said, the current code gives an all-0 hash

action/actctx.go
```diff
@@ -40,6 +41,21 @@ func (act *AbstractAction) GasPrice() *big.Int {
 	return p.Set(act.gasPrice)
 }
 
+// IsExternal returns false for AbstractAction
+func (act *AbstractAction) IsExternal() bool {
```
raullenchai, line 45: isExternal is a parameter at the Envelope level, AbstractAction does not need to know about this concept.

Also suggest change to isRLP rather than isExternal

action/execution.go
```diff
@@ -142,3 +172,60 @@ func (ex *Execution) SanityCheck() error {
...
+	var ethAddr *common.Address
+	if ex.contract != EmptyAddress {
+		addr, err := address.FromString(ex.contract)
+		if err != nil {
+			return errors.Wrapf(err, "invalid recipient address %s", ex.contract)
+		}
+		eAddr := common.BytesToAddress(addr.Bytes())
+		ethAddr = &eAddr
+	}
+	items := []interface{}{
```
raullenchai, line 202: Dup code w/ https://github.com/iotexproject/iotex-core/pull/2586/files?file-filters%5B%5D=.go#diff-ede8b4d922fba992cf4cacfee119592fbac03586ade8fea4a8346866e4f2c07dR196 

Pls have one place to handle

action/actctx.go
raullenchai: Envelope is the only level that needs to know this parameter

action/transfer.go
```diff
@@ -100,6 +108,27 @@ func (tsf *Transfer) Proto() *iotextypes.Transfer {
...
+func (tsf *Transfer) RawHash() hash.Hash256 {
+	if !tsf.IsExternal() {
+		panic("should not call RawHash() on non-external transfer")
+	}
+	return tsf.rawHash
+}
+
+// Hash returns the signed hash
+func (tsf *Transfer) Hash() hash.Hash256 {
+	if !tsf.IsExternal() {
```
raullenchai, line 126: This is risky. Hash() becomes only available for rlp transfers. non rlp-tsf causes panic

action/actctx.go
dustinxie: yes, Envelope is the only level that needs to know this, but it is called upon Envelope's payload
```
func (elp *Envelope) Hash() hash.Hash256 {
	if elp.payload.IsExternal() {
		return elp.payload.RawHash()
	}
	return hash.Hash256b(elp.Serialize())
}
```
so this needs to be included in the `actionPayload` interface

if i remove this, the code won't compile

dustinxie: in my first version, i did use `isRLP`
but later you call out `extern chain ID` for the new field's name in our discussion, then i think `isExternal` is more comprehensive? so we could potentially use this for other type of serialization, not particularly RLP
leave me another comment if you do want to use isRLP, i am fine with it

action/execution.go
dustinxie: ok, will fix

action/transfer.go
dustinxie: here's the reason why i think panic() is better than return an error:

the ONLY legit source of RLP tx coming into our chain is from the wrapper proxy service, they come in as raw bytes, and gets processed by `sealEnvelope.LoadProto()`, where it computes the raw hash and hash, and saved in the RLP-transfer/execution

any normal tx (generated from ioctl, ioPay desktop, ioPay mobile, etc.) must be non-RLP, and RawHash(), Hash() must NOT be called on these normal tx. If it happens, means something is seriously wrong

since this func returns `hash.Hash256`, if don't panic, would need to return `hash.Hash256{}`, which is an all-0 hash, and no error is reported

so i would rather panic() to catch the unintended call, instead of let it silently pass

the same reason for this panic in `action.Sign`:
```
if act.payload.IsExternal() {
		panic("not allowed to sign external transaction")
	}
```

action/transfer_test.go
```diff
@@ -124,3 +136,125 @@ func TestTransfer(t *testing.T) {
...
+		tsf.payload = tx.Data()
+
+		// pack the action object
+		pb := &iotextypes.Action{
+			Core: &iotextypes.ActionCore{
+				Version:  1,
+				Nonce:    tsf.nonce,
+				GasLimit: tsf.gasLimit,
+				GasPrice: tsf.gasPrice.String(),
+				ChainID:  v.chain,
```
zjshen14, line 212: chain is not part of the content signed. This is a potential field being tampered.

zjshen14 (commented):
Per dong's request, I took a quick look, that said, may be wrong. Some suggestions:

1. Better not to use chain ID to gate the logic: 1) the field seems not to be part of being signed, 2) counter intuitive, the tx is still settled on the same chain. 
2. Instead of hacking is_external in side envelope, consider create another RlpEnvelope. Then, you basically have 2 envelope types: Rlp vs Proto, and receiving both on the wire. This will require more coding change to pipe through the Rlp msg from network layer bottom up, but will prevent the rlp logic from spilling into envelop.

action/transfer_test.go
dustinxie: chainID is 1 of 9 inputs to calculate the RLP message hash to be signed, so this should be OK

in other words, if you tamper with the chainID in proto bytes, the tx can still be deserialized and decoded, but signature verification will fail due to change of message hash

action/action.go
```diff
@@ -40,12 +40,16 @@ type hasDestination interface {
 }
 
 // Sign signs the action using sender's private key
-func Sign(act Envelope, sk crypto.PrivateKey) (SealedEnvelope, error) {
-	sealed := SealedEnvelope{Envelope: act}
-
-	sealed.srcPubkey = sk.PublicKey()
+func Sign(elp Envelope, sk crypto.PrivateKey) (SealedEnvelope, error) {
+	if elp.IsExternal() {
+		panic("not allowed to sign external tx")
```
yutongp, line 45: wonder if it is better to return error instead of panic? if it is possible for someone forms a bad request and make the node panic?

action/action.go
```diff
@@ -90,14 +94,14 @@ func Verify(sealed SealedEnvelope) error {
 	}
 
 	hash := sealed.Envelope.Hash()
-	if sealed.SrcPubkey().Verify(hash[:], sealed.Signature()) {
+	if sealed.SrcPubkey().Verify(hash[:], sealed.signature) {
```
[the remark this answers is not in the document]
yutongp: curious why change it to access the field directly?

action/envelope.go
```diff
@@ -230,10 +239,100 @@ func (elp *Envelope) Serialize() []byte {
...
+	return elp.chainID != 0 && elp.chainID != config.ExternChainID()
+}
+
+// IsExternal returns whether the action is external or not
+func (elp *Envelope) IsExternal() bool {
+	return elp.chainID == config.ExternChainID()
+}
+
+// ValidateExternal validates an external action
+func (elp *Envelope) ValidateExternal(sig []byte) error {
```
yutongp, line 264: this function has side effort rather than just validate the action

action/envelope.go
```diff
@@ -230,10 +239,100 @@ func (elp *Envelope) Serialize() []byte {
...
+type rlpTransaction interface {
+	Nonce() uint64
+	GasPrice() *big.Int
+	GasLimit() uint64
+	Recipient() string
+	Amount() *big.Int
+	Payload() []byte
+}
+
+func computeRlpHash(act rlpTransaction, chainID uint32, sig []byte) (hash.Hash256, hash.Hash256, error) {
```
yutongp, line 291: (hash.Hash256, hash.Hash256, error)
better to assign varriable name

config/config.go
```diff
@@ -86,6 +90,8 @@ const (
 	SigP256sm2 = "p256sm2"
 )
 
+var chainID sync.Once
```
yutongp, line 93: not sure if this is necessary since you only assign the value once

yutongp:
> Per dong's request, I took a quick look, that said, may be wrong. Some suggestions:
> 
> 1. Better not to use chain ID to gate the logic: 1) the field seems not to be part of being signed, 2) counter intuitive, the tx is still settled on the same chain.
> 2. Instead of hacking is_external in side envelope, consider create another RlpEnvelope. Then, you basically have 2 envelope types: Rlp vs Proto, and receiving both on the wire. This will require more coding change to pipe through the Rlp msg from network layer bottom up, but will prevent the rlp logic from spilling into envelop.
> 
> ==== dustin
> Thanks Zhijie for inputs!
> For 1, yes, it is kinda counter-intuitive to use a chainID, the primary purpose is to still use the current protobuf struct so we can avoid extra change to front-end. Regarding signing, the field is an input to RLP message hash, see comments below
> For 2, we'll evaluate if this could be a better way of processing the RLP tx

+1 for create a new RlpEnvelope

config/config.go
dustinxie: this is for safety, that multiple calls to SetExternChainID() will only get called once

action/action.go
dustinxie: Sign() only get calls by ioctl, so node won't panic

since RLP-tx would ONLY come from proxy, should never be generated from core or ioctl, it's trying to prevent such mis-use. Return an error might be ignored, panic make it clear that sth is seriously wrong

action/envelope.go
dustinxie: yes, it also computes the hash

action/envelope.go
dustinxie: ok

dustinxie:
> > Per dong's request, I took a quick look, that said, may be wrong. Some suggestions:
> > 
> > 1. Better not to use chain ID to gate the logic: 1) the field seems not to be part of being signed, 2) counter intuitive, the tx is still settled on the same chain.
> > 2. Instead of hacking is_external in side envelope, consider create another RlpEnvelope. Then, you basically have 2 envelope types: Rlp vs Proto, and receiving both on the wire. This will require more coding change to pipe through the Rlp msg from network layer bottom up, but will prevent the rlp logic from spilling into envelop.
> > 
> > ==== dustin
> > Thanks Zhijie for inputs!
> > For 1, yes, it is kinda counter-intuitive to use a chainID, the primary purpose is to still use the current protobuf struct so we can avoid extra change to front-end. Regarding signing, the field is an input to RLP message hash, see comments below
> > For 2, we'll evaluate if this could be a better way of processing the RLP tx
> 
> +1 for create a new RlpEnvelope

yes, i like the idea of a new RlpEnvelope, however due to current Envelope is a struct and embedded in SealedEnvelope, this entails turning Envelope to an interface (with current Envelope and RlpEnvelope implementing the interface). I am evaluating the amount of change

or is there a smarter way to plug a new `RlpEnvelope` into current code structure?

action/envelope.go
dustinxie: it actually decodes the tx, validation will be done later in Verify()
change name to DecodeExternal()

dustinxie:
closed due to too many msg, latest code push to new PR https://github.com/iotexproject/iotex-core/pull/2592

action/sealedenvelope.go
```diff
@@ -61,9 +61,12 @@ func (sealed *SealedEnvelope) LoadProto(pbAct *iotextypes.Action) error {
 	if err != nil {
 		return err
 	}
+
+	// clear 'sealed' and populate new value
+	*sealed = SealedEnvelope{}
```
CoderZhi, line 66: unnecessary.

action/action.go
```diff
@@ -41,17 +41,18 @@ type hasDestination interface {
...
+	}
 
 	hash := act.Hash()
 	sig, err := sk.Sign(hash[:])
 	if err != nil {
 		return sealed, errors.Wrapf(ErrAction, "failed to sign action hash = %x", hash)
 	}
 	sealed.signature = sig
-	sealed.payload.SetEnvelopeContext(sealed)
+	sealed.Action().SetEnvelopeContext(sealed)
```
CoderZhi, line 55: This is different from what I suggested in phone call. We may introduce a function "SetSealedContext" into Envelope, which is an interface after this PR. Then, we can call act.SetSealedContext(sealed).

Moreover, I am thinking about changing Envelope to be a SealedEnvelopeBuilder, say adding the following function to Envelope:
```Seal(sk crypto.PrivateKey) (*SealedEnvelope, error)```

dustinxie: yes, i  did
see https://github.com/iotexproject/iotex-core/pull/2586/commits/080d457a696f83775c902cdfe09a69317c249261

if you want, can have another PR for adding SealedEnvelopeBuilder and Seal()

dustinxie: updated: https://github.com/iotexproject/iotex-core/pull/2586/commits/631d6e9e8cad39d594cc2462d45aed4a7b632b11

CoderZhi (approved):

PR #2587, opened 2021-03-29
[action] improve envelope handling

dustinxie:
1. removed AbstractAction.hash. Hash of an action should be at the
level of Envelope and sealed envelope
2. removed unnecessary methods

### DIFF
--- a/action/actctx.go
+++ b/action/actctx.go
@@ -10,7 +10,6 @@ import (
 	"math/big"
 
 	"github.com/iotexproject/go-pkgs/crypto"
-	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/pkg/errors"
 )
 
@@ -18,10 +17,9 @@ import (
 type AbstractAction struct {
 	version   uint32
 	nonce     uint64
-	srcPubkey crypto.PublicKey
 	gasLimit  uint64
 	gasPrice  *big.Int
-	hash      hash.Hash256
+	srcPubkey crypto.PublicKey
 }
 
 // Version returns the version
@@ -29,9 +27,6 @@ func (act *AbstractAction) Version() uint32 { return act.version }
 
 // Nonce returns the nonce
 func (act *AbstractAction) Nonce() uint64 { return act.nonce }
-
-// SrcPubkey returns the source public key
-func (act *AbstractAction) SrcPubkey() crypto.PublicKey { return act.srcPubkey }
 
 // GasLimit returns the gas limit
 func (act *AbstractAction) GasLimit() uint64 { return act.gasLimit }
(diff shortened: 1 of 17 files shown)


action/protocol/execution/evm/evm.go
```diff
@@ -203,7 +203,7 @@ func ExecuteContract(
 	receipt := &action.Receipt{
 		GasConsumed:     ps.gas - remainingGas,
 		BlockHeight:     blkCtx.BlockHeight,
-		ActionHash:      execution.Hash(),
+		ActionHash:      actionCtx.ActionHash,
```
dustinxie, line 206: this is a wrong programming: should use action hash passed from context

and this is why SetEnvelopeContext() currently do:
```
act.hash = Selp.Hash()
```
which is not needed once this is corrected

action/protocol/poll/consortium.go
```diff
@@ -113,7 +114,7 @@ func (cc *consortiumCommittee) CreateGenesisStates(ctx context.Context, sm proto
 		return err
 	}
 	actionCtx.Nonce = consortiumCommitteeContractNonce
-	actionCtx.ActionHash = execution.Hash()
+	actionCtx.ActionHash = consortiumCommitteeContractHash
```
dustinxie, line 117: this is also improper programming:
execution.Hash() will be all zero, since it is a raw tx not signed

reason this is not an issue: this code is not run on mainnet/testnet now

if you start the chain from 0, this will result in an all-0 hash, colliding with the all-0 hash in staking_committee.go

action/protocol/poll/staking_committee.go
```diff
@@ -124,7 +127,7 @@ func (sc *stakingCommittee) CreateGenesisStates(ctx context.Context, sm protocol
 		return err
 	}
 	actionCtx.Nonce = nativeStakingContractNonce
-	actionCtx.ActionHash = execution.Hash()
+	actionCtx.ActionHash = nativeStakingContractHash
```
dustinxie, line 130: this is an all-0 hash, see comment above

action/actctx.go
```diff
@@ -67,16 +59,11 @@ func (act *AbstractAction) SetEnvelopeContext(selp SealedEnvelope) {
...
-	ab := &Builder{}
-	*act = ab.SetVersion(selp.Version()).
-		SetNonce(selp.Nonce()).
-		SetSourcePublicKey(selp.SrcPubkey()).
-		SetGasLimit(selp.GasLimit()).
-		SetGasPrice(selp.GasPrice()).
-		Build()
-
-	// the reason to set hash here, after set act context, is because some actions use envelope information in their proto define. for example transfer use des addr as Receipt.
-	act.hash = selp.Hash()
```
dustinxie, line 79: reason we need this is due to wrong programming in evm.go, see comments there

raullenchai (approved):

PR #1628, opened 2019-11-08
Modify EstimateActionGasConsumption API

lizhefeng:
Work on https://github.com/iotexproject/iotex-core/issues/1609

Issue #1609, opened 2019-11-05
EstimateGasForAction's result is not valid sometimes


Frankonly:
EstimateGasForAction returns gas consume of action execution simulation. However, when an action is being executed in evm, evm may request more gas than it needs actually and returns the rest later.

To given the valid gas limit for an execution action, EstimateGasForAction should return the highest gas request during evm running the action, rather than the gas consume evm returns after action executed.

PR #2588, opened 2021-03-31
[api] fix excessively large gas estimate

dustinxie:
Related to 
https://github.com/iotexproject/iotex-core/pull/1628
https://github.com/iotexproject/iotex-core/issues/1609

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1521,7 +1521,6 @@ func (api *Server) getLogsInRange(filter *logfilter.LogFilter, start, end, pagin
 	return logs, nil
 }
 
-// TODO: Since GasConsumed on the receipt may not be enough for the gas limit, we use binary search for the gas estimate. Need a better way to address it later.
 func (api *Server) estimateActionGasConsumptionForExecution(exec *iotextypes.Execution, sender string) (*iotexapi.EstimateActionGasConsumptionResponse, error) {
 	sc := &action.Execution{}
 	if err := sc.LoadProto(exec); err != nil {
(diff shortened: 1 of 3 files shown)


dustinxie:
here's the calculation of calling execution with gas limit = `g`

1. a = g - intrinsic
2. call EVM with gas amount = a
3. EVM returns remaining gas = r
4. refund = (g - r) / 2
5. b = r + refund = (g + r) / 2
6. gas consumed c = g - b = (g - r) / 2

since we input a to EVM and got r returned, `a-r` is enough to execute the tx
a - r = g - intrinsic - r = 2*c - intrinsic

to be safe, we can use 2*c

api/api.go
```diff
@@ -1562,28 +1561,10 @@ func (api *Server) estimateActionGasConsumptionForExecution(exec *iotextypes.Exe
...
-				estimatedGas = mid
-				break
-			}
-			low = mid + 1
-		}
-	}
 
+	// for our EVM, 2*GasConsumed will make it consume all the input gas
+	// add 25% as safety margin
+	estimatedGas := 2*receipt.GasConsumed + (receipt.GasConsumed >> 1)
```
raullenchai, line 1567: This is 2.5x, while the comment is 2.25x

raullenchai (approved):

api/api.go
dustinxie: exact amount = 2 * gasConsumed
add 25% = 2 * gasConsumed * 1.25 = 2.5 * gasConsumed

Issue #2589, opened 2021-04-01
e2etest fail


zsy-cn:
### What version of iotex-core are you using (`ioctl version`)?

latest master

### What operating system and processor architecture are you using?

macOS Big Sur 11.2.3

```env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.16.2/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.16.2/libexec/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.16.2"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/iotex-core/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
```

### What did you do? If possible, provide a recipe for reproducing the error.

```sh
git clone https://github.com/iotexproject/iotex-core.git
cd iotex-core
make
```

### What did you expect to see?

make success

### What did you see instead?

```sh
FAIL	github.com/iotexproject/iotex-core/e2etest	114.613s
```
```sh
2021-04-01T09:19:34.970+0800	ERROR	go-p2p@v0.2.12/p2p.go:449	Error when subscribing a broadcast message.	{"error": "subscription cancelled by calling sub.Cancel()"}
github.com/iotexproject/go-p2p.(*Host).AddBroadcastPubSub.func1
	/Users/apple/go/pkg/mod/github.com/iotexproject/go-p2p@v0.2.12/p2p.go:449
2021-04-01T09:19:38.383+0800	WARN	scheme/noop.go:45	Noop scheme could not calculate delegates by height
github.com/iotexproject/iotex-core/consensus/scheme.(*Noop).ValidateBlockFooter
	/Users/apple/Documents/iotex-core/consensus/scheme/noop.go:45
[...]

PR #2590, opened 2021-04-01
fix TestLocalTransfer failed #2589

millken:
#2589


e2etest/local_transfer_test.go
```diff
@@ -323,7 +323,7 @@ func TestLocalTransfer(t *testing.T) {
 	probeSvr := probe.New(7788)
 	require.NoError(probeSvr.Start(ctx))
 	defer func() {
-		require.NoError(probeSvr.Stop(ctx))
+		probeSvr.Stop(ctx)
```
dustinxie, line 326: think we better find out the root-cause and fix, instead of just removing `NoError`?

it erred b/c `context canceled`, so call this before context cancelled is the right fix

dustinxie (approved):

PR #2592, opened 2021-04-02
[action] add envelopeRLP to  handle RLP-encoded tx

dustinxie:


action/envelope_test.go
```diff
@@ -12,61 +11,35 @@ import (
 	"github.com/iotexproject/iotex-core/test/identityset"
 )
 
-func TestEnvelope_Version(t *testing.T) {
+func TestEnvelope_Basic(t *testing.T) {
```
dustinxie, line 14: combine multiple 1-line tests together

action/rlp_tx.go
```diff
@@ -0,0 +1,76 @@
...
+
+type rlpTransaction interface {
+	Nonce() uint64
+	GasPrice() *big.Int
+	GasLimit() uint64
+	Recipient() string
+	Amount() *big.Int
+	Payload() []byte
+}
+
```
dustinxie, line 30: we don't need a new struct{}
these are utility funcs

action/sealedenvelope_test.go
```diff
@@ -19,51 +19,40 @@ const (
 
 var signByte = []byte{1, 2, 3, 4, 5, 6, 7, 8, 9}
 
-func TestSealedEnvelope_Hash(t *testing.T) {
+func TestSealedEnvelope_Basic(t *testing.T) {
```
dustinxie, line 22: combine multiple 1-line tests together

action/rlp_tx.go
```diff
@@ -0,0 +1,76 @@
...
+func rlpRawHash(tx rlpTransaction, chainID uint32) hash.Hash256 {
+	rawTx, err := generateRlpTx(tx)
+	if err != nil {
+		return hash.ZeroHash256
```
CoderZhi, line 34: return error and leave the handling of the error to where it is called

action/rlp_tx.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	h := types.NewEIP155Signer(big.NewInt(int64(chainID))).Hash(rawTx)
+	return hash.BytesToHash256(h[:])
+}
+
+func rlpSignedHash(tx rlpTransaction, chainID uint32, sig []byte) hash.Hash256 {
```
CoderZhi, line 40: same for this function, return (hash.Hash256, error)

action/sealedenvelope.go
```diff
@@ -7,19 +7,39 @@ import (
...
 
+	"github.com/iotexproject/iotex-core/config"
 	"github.com/iotexproject/iotex-core/pkg/util/byteutil"
 )
 
 // SealedEnvelope is a signed action envelope.
 type SealedEnvelope struct {
 	Envelope
+	encoding      uint32
+	externChainID uint32
```
CoderZhi, line 18: externChainID is not part of the struct, because it can be told from the encoding for now.

action/sealedenvelope.go
```diff
@@ -7,19 +7,39 @@ import (
...
+	return sealed.Envelope.Hash()
 }
 
 // Hash returns the hash value of SealedEnvelope.
 func (sealed *SealedEnvelope) Hash() hash.Hash256 {
+	if IsRLP(iotextypes.Encoding(sealed.encoding)) {
+		if tx, err := actionToRLP(sealed.Action()); err == nil {
+			return rlpSignedHash(tx, sealed.externChainID, sealed.Signature())
+		}
+		return hash.ZeroHash256
```
CoderZhi, line 41: we should either panic or return the error

action/sealedenvelope.go
```diff
@@ -7,19 +7,39 @@ import (
...
+// RawHash returns the raw hash of embedded Envelope
+// this is the hash to be signed
+func (sealed *SealedEnvelope) RawHash() hash.Hash256 {
+	if IsRLP(iotextypes.Encoding(sealed.encoding)) {
+		if tx, err := actionToRLP(sealed.Action()); err == nil {
+			return rlpRawHash(tx, sealed.externChainID)
+		}
+		return hash.ZeroHash256
+	}
+	return sealed.Envelope.Hash()
```
CoderZhi, line 32: if we keep this function, Envelope.Hash, it will be confusing, because it is just the hash of the protobuf encoding format

action/sealedenvelope.go
```diff
@@ -7,19 +7,39 @@ import (
...
 	"github.com/pkg/errors"
 
+	"github.com/iotexproject/iotex-core/config"
 	"github.com/iotexproject/iotex-core/pkg/util/byteutil"
 )
 
 // SealedEnvelope is a signed action envelope.
 type SealedEnvelope struct {
 	Envelope
+	encoding      uint32
```
CoderZhi, line 17: encoding iotextypes.WireEncoding

action/sealedenvelope.go
```diff
@@ -51,10 +77,21 @@ func (sealed *SealedEnvelope) LoadProto(pbAct *iotextypes.Action) error {
...
 	if err := elp.LoadProto(pbAct.GetCore()); err != nil {
 		return err
 	}
+	if IsRLP(encoding) {
+		// verify action type can support RLP-encoding
+		if _, err := actionToRLP(elp.Action()); err != nil {
+			return err
+		}
+		sealed.externChainID = config.ExternChainID()
+	}
```
CoderZhi, line 94: if we need to do validation for rlp encoding, should we do it for protobuf as well?

config/config.go
```diff
@@ -246,6 +253,7 @@ type (
 		CandidateIndexDBPath   string           `yaml:"candidateIndexDBPath"`
 		StakingIndexDBPath     string           `yaml:"stakingIndexDBPath"`
 		ID                     uint32           `yaml:"id"`
+		ExternChainID          uint32           `yaml:"externChainID"`
```
CoderZhi, line 256: we may have a more precise name which can better explain itself from its name

config/config.go
```diff
@@ -467,6 +475,9 @@ func New(validates ...Validate) (Config, error) {
 		}
 	}
 
+	// populdate chain ID
+	SetExternChainID(cfg.Chain.ExternChainID)
```
CoderZhi, line 479: if the way of reading this parameter is so strict, what about _plugins and etc.? Thus, I don't see the benefit.

config/config.go
dustinxie: so do you have a suggestion?

config/config.go
dustinxie: can you give more details? _plugins?
what I did is to make sure the external chain ID is loaded exactly one

action/sealedenvelope.go
dustinxie: for protobuf, validation is done in Envelope.Load(), it will check all possible types.

here is to make sure for RLP case, the struct{} inside must be Transfer or Execution, i.e., to catch someone send in a StakeCreate with encoding = 1

action/sealedenvelope.go
dustinxie: we discussed last night, it cannot be told from the encoding (unless we send in all tx bytes, then you can decode and get it)

for now, we'll just set it according to your network config (4689/4690), and use it to calculate raw hash, and verify against the signature
if signature fail, means this tx is not intended for this network

config/config.go
dustinxie: do you mean like "-plugin=gateway"?

this is doable but I think putting in config.yaml is an easier way for delegates.

action/sealedenvelope.go
dustinxie: Hash() returns 32-byte array

i added a check to catch all-0 hash in Verify()

action/sealedenvelope.go
dustinxie: i think the logic is similar to what we discussed for signed hash?
```
if rlp()
   return RLP's raw hash
return sealed.Envelope.Hash(), which is native protobuf's raw hash
```

action/rlp_tx.go
dustinxie: since the caller is sealed.Hash(), it does not return err, so return all-0 hash to indicate error

added a check to catch all-0 hash in Sing() and Verify()

CoderZhi: even this is a private function, it is still a good practice to return the error and let the caller handle the error properly.

action/sealedenvelope.go
CoderZhi: If a sealed envelope could be assembled, we shouldn't expect any error here. Thus, we should panic, instead of having some hidden knowledge like ZeroHash

action/sealedenvelope.go
CoderZhi: ok

action/action.go
```diff
@@ -46,13 +40,16 @@ func Sign(act Envelope, sk crypto.PrivateKey) (SealedEnvelope, error) {
 		srcPubkey: sk.PublicKey(),
 	}
 
-	hash := act.Hash()
-	sig, err := sk.Sign(hash[:])
+	h := act.Hash()
+	if h == hash.ZeroHash256 {
```
CoderZhi, line 44: how could envelope.hash be zero hash?

action/sealedenvelope.go
CoderZhi: besides https://github.com/iotexproject/iotex-core/pull/2592/files#diff-fda8afdb7165ec30b8d2c40dfd1963def0ffeb117a3d9098167b8139c72527caR97, where else we set the externChainID? If we can use confg.ExternChainID() there for rlp encoding, why cannot we do it in the struct?

action/sealedenvelope.go
CoderZhi: Envelope.Hash is only used in Sign and Verify, in other word, Hash doesn't need to be an essential public API of Envelope.
Same for https://github.com/iotexproject/iotex-core/pull/2592/files#diff-fda8afdb7165ec30b8d2c40dfd1963def0ffeb117a3d9098167b8139c72527caR25, RawHash doesn't expose to external packages

config/config.go
CoderZhi: https://chainid.network/ EVM_NETWORK_ID

config/config.go
CoderZhi: think about all the scenarios how this function will be called (we don't need this function, but set the _evmNetworkID in the New function), these complicated handlings are redundant.

config/config.go
```diff
@@ -513,6 +524,18 @@ func NewSub(validates ...Validate) (Config, error) {
...
 
+// SetExternChainID sets the extern chain ID
+func SetExternChainID(id uint32) {
+	chainID.Do(func() {
+		_externChainID = id
+	})
+}
+
+// ExternChainID returns the extern chain ID
+func ExternChainID() uint32 {
```
CoderZhi, line 535: I can see why we have this function for now. But ideally, we should have a Signer and a Verifier which were created with cfg.EVMNetworkID. Global function is not a good choice, but we can leave it to future.

action/action.go
dustinxie: ok

action/sealedenvelope.go
dustinxie: that's the only place we set externChainID

what do you mean exactly "why cannot we do it in the struct?", as said above, that is the ONLY place where we need to set externChainID, since it only applies for IsRLP() case

action/sealedenvelope.go
dustinxie: ok, right, RawHash() need not expose

action/sealedenvelope.go
dustinxie: yes, ideally, if the signature is correct, rlpSignedHash() will not err.

i feel fallback to 0-hash is safer, b/c 0-hash will fail Verify(), and program will catch the error, and continue. Panic could bring down a node in case a tx somehow caused error in rlpSignedHash() -- though in theory it should not happen

config/config.go
dustinxie: i don't want to explicitly call out the name EVM, also this extern chain ID can in theory be used for another type of wire encoding format -- though that's not very likely in the near future

config/config.go
dustinxie: again my purpose is to make sure the external chain ID is loaded exactly one, i think it's better/safer to do that

action/sealedenvelope.go
dustinxie: will have another PR to remove Envelope.Hash

action/sealedenvelope.go
```diff
@@ -33,12 +64,18 @@ func (sealed *SealedEnvelope) Signature() []byte {
 	return sig
 }
 
+// Encoding returns the encoding
+func (sealed *SealedEnvelope) Encoding() uint32 {
```
CoderZhi, line 68: returntype iotextypes.WireEncoding

action/sealedenvelope.go
```diff
@@ -7,19 +7,50 @@ import (
...
+	}
+	return sealed.Envelope.Hash()
 }
 
 // Hash returns the hash value of SealedEnvelope.
+// an all-0 return value means the transaction is invalid
 func (sealed *SealedEnvelope) Hash() hash.Hash256 {
+	if IsRLP(sealed.encoding) {
+		tx, err := actionToRLP(sealed.Action())
+		if err != nil {
```
CoderZhi, line 45: we should panic here, but let's leave it to next PR

action/sealedenvelope.go
```diff
@@ -33,12 +64,18 @@ func (sealed *SealedEnvelope) Signature() []byte {
...
+	return uint32(sealed.encoding)
+}
+
 // Proto converts it to it's proto scheme.
 func (sealed *SealedEnvelope) Proto() *iotextypes.Action {
 	return &iotextypes.Action{
 		Core:         sealed.Envelope.Proto(),
 		SenderPubKey: sealed.srcPubkey.Bytes(),
 		Signature:    sealed.signature,
+		Encoding:     iotextypes.Encoding(sealed.encoding),
```
CoderZhi, line 78: redundant type casting?

config/config.go
CoderZhi: in what case this struct or function call could be thread unsafe? I don't see it

config/config.go
CoderZhi: it should be explicit and shouldn't be shared by other encoding type

action/sealedenvelope.go
```diff
@@ -7,19 +7,50 @@ import (
...
-	signature []byte
+// rawHash returns the raw hash of embedded Envelope (this is the hash to be signed)
+// an all-0 return value means the transaction is invalid
+func (sealed *SealedEnvelope) rawHash() hash.Hash256 {
+	if IsRLP(sealed.encoding) {
+		tx, err := actionToRLP(sealed.Action())
+		if err != nil {
+			return hash.ZeroHash256
+		}
+		h, err := rlpRawHash(tx, sealed.externChainID)
```
CoderZhi, line 31: this is the only place we need to call config.ExternChainID()

CoderZhi (approved):
Unblock this PR, but the problems in the comments should be fixed addressed in next PR.

action/sealedenvelope.go
dustinxie: also rlpSignedHash()

action/sealedenvelope.go
dustinxie: fixed

action/sealedenvelope.go
dustinxie: node receive tx thru API endpoint --> tx.LoadProto() --> actpool.Add(tx) --> call Verify(tx), if pass, tx is added to actpool, if fail, tx is rejected

if panic here, Verify() will panic, and node stop running
return all-0 hash will make Verify() fail gracefully, reject this tx, and continue running. i think this is better than panic

action/sealedenvelope.go
dustinxie: by saving externChainID in LoadProto() once, all later func calls like `rawHash(), Hash()` become self-contained so they don't need to call outside `config.ExternChainID()` again

PR #2594, opened 2021-04-03
add expectedDesc for local_transfer_test

maswalker:
original local_transfer_test.go did not checked the return error detail information.
it may cause confusion.
for example, a test case was expected to catch error when nonce field error.
but the case maybe got an error which was throw because balance not enough.
so I add "expectedDesc" field to verify error type.

### DIFF
--- a/e2etest/local_transfer_test.go
+++ b/e2etest/local_transfer_test.go
@@ -17,6 +17,9 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/status"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
 
 	"github.com/iotexproject/go-pkgs/crypto"
 	"github.com/iotexproject/iotex-address/address"
@@ -78,6 +81,7 @@ type simpleTransferTestCfg struct {
 	gasPrice        *big.Int
 	expectedResult  TransferState
 	message         string
+	expectedDesc    string
 }
 
 var (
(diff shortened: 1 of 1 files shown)


raullenchai (approved):

e2etest/local_transfer_test.go
```diff
@@ -178,40 +191,43 @@ var (
...
-			},
-		*/
 		{
 			AcntCreate, nil, big.NewInt(100000),
 			AcntCreate, nil, big.NewInt(100000),
 			1, big.NewInt(-100),
 			make([]byte, 4),
 			uint64(200000), big.NewInt(1),
 			TsfFail, "Transfer with negative amount",
+			"Invalid balance", // TODO: "Unknown",
```
dustinxie, line 230: remove TODO?

e2etest/local_transfer_test.go
```diff
@@ -178,40 +191,43 @@ var (
 			make([]byte, 100),             //payload
 			uint64(200000), big.NewInt(1), // gasLimit, gasPrice
 			TsfFail, "Normal transfer to a bad address",
+			"Unknown", // TODO: "Blacklisted address",
```
dustinxie, line 194: "Bad address", and remove TODO?

e2etest/local_transfer_test.go
```diff
@@ -78,6 +81,7 @@ type simpleTransferTestCfg struct {
 	gasPrice        *big.Int
 	expectedResult  TransferState
 	message         string
+	expectedDesc    string
```
dustinxie, line 84: feel put this after `expectedResult` and before `message` is better

dustinxie (approved):

Issue #2585, opened 2021-03-24
Add unit tests for https://github.com/iotexproject/iotex-core/pull/2580/


raullenchai:
**What would you like to be added**:
Unit tests

**Why is this needed**:
Lack of test coverage

PR #2597, opened 2021-04-06
add TestSortMap

millken:
#2585


dustinxie:
the fix in https://github.com/iotexproject/iotex-core/commit/495fc835841921d76aa3ecfb294c86e4e8c81905 took effect in Snapshot() and RevertToSnapshot()
so we'll need a test to trigger code in these 2 funcs, before the fix, the trie root hash will change in you run the test multiple (like 20) times. after the fix, the root hash keep fixed

dustinxie:
the fix in https://github.com/iotexproject/iotex-core/commit/495fc835841921d76aa3ecfb294c86e4e8c81905 took effect in Snapshot() and RevertToSnapshot()

so we need a piece of code that can trigger these 2 funcs to make multiple entries in the map, then run the code multiple (like 20) times. before the fix, the trie root hash will change. after the fix, the trie root hash will keep fixed

action/protocol/execution/evm/evmstatedbadapter_test.go
```diff
@@ -462,3 +464,43 @@ func TestPreimage(t *testing.T) {
...
+		testHash1 := func(cm contractMap) []hash.Hash160 {
+			addrs := make([]hash.Hash160, 0, len(cm))
+			for addr := range cm {
+				addrs = append(addrs, addr)
+			}
+			return addrs
+		}
+		t1 := testHash1(cMap)
+		for i := 0; i < size; i++ {
+			require.NotEqual(t1, testHash1(cMap))
```
CoderZhi, line 487: this is not always true

CoderZhi (changes requested):
This unit test should be against StateDBAdapter.Snapshot() and RevertToSnapshot() with sortCachedContracts as true or false. When it is set to true, the behavior is constant, if false, the behavior may be inconstant (you can run it for several times, and asset that is some inconsistency).

action/protocol/execution/evm/evmstatedbadapter_test.go
```diff
@@ -453,3 +454,69 @@ func TestPreimage(t *testing.T) {
...
+			stateDB.SetCode(addr, []byte("0123456789"))
+			stateDB.SetState(addr, k1, k2)
+		}
+		sn := stateDB.Snapshot()
+		caches := []string{}
+		for i := 0; i < size; i++ {
+			stateDB.RevertToSnapshot(sn)
+			s := ""
+			for _, addr := range stateDB.cachedContractAddrs() {
+				c := stateDB.cachedContract[addr]
```
dustinxie, line 515: why we calculate each contract's root? is there an entire root to check?

from your test result, it seems correct to do this. maybe confirm with Zhi as well

CoderZhi (approved):

dustinxie (approved):

PR #2598, opened 2021-04-07
[action] add Key() as unique key for DB storage

dustinxie:
1. Add Key() which is always hash(Marshall(proto)), consistent for all tx

2. Hash() == Key() for native tx, will be != Key() for upcoming RLP tx

3. Use Key() as key for DB storage, to maintain consistency

4. API query pass in Hash(), indexer will internally do a Hash() --> Key() lookup

### DIFF
--- a/action/sealedenvelope.go
+++ b/action/sealedenvelope.go
@@ -18,7 +18,12 @@ type SealedEnvelope struct {
 	signature []byte
 }
 
-// Hash returns the hash value of SealedEnvelope.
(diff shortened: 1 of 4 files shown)


CoderZhi (approved):
Let's use swap "Key" and "Hash", and then find a better name for "Key". Up to now, it is not a blocker, because we can change the naming later.

dustinxie:
close for now before we settle on https://github.com/iotexproject/iotex-core/pull/2592

PR #2601, opened 2021-04-12
Envelope hash

CoderZhi:
1. delete Envelope.Hash, which is only used inside of action package, move the logic into sealedenvelope
    the checks of hash in various actions are redundant, because signature verification already covers it
2. panic if sealedenvelope.Hash failed
    returning ZeroHash is hidden knowledge, and the failure is always unexpected.


action/sealedenvelope.go
```diff
@@ -14,44 +14,46 @@ import (
...
 		if err != nil {
-			return hash.ZeroHash256
+			panic(err)
 		}
 		return h
+	case iotextypes.Encoding_IOTEX_PROTOBUF:
+		return hash.Hash256b(byteutil.Must(proto.Marshal(sealed.Proto())))
 	}
-	return hash.Hash256b(byteutil.Must(proto.Marshal(sealed.Proto())))
+	panic("unknown encoding type")
```
dustinxie, line 56: if someone send in an action with valid ActionCore but encoding = 2, this would cause the node to panic and stop running

you can think 0-hash is a "hidden" knowledge, but IMO that a way to fail gracefully, and it is fine given the behavior is clearly documented and understood

action/sealedenvelope.go
```diff
@@ -14,44 +14,46 @@ import (
...
 // Hash returns the hash value of SealedEnvelope.
 // an all-0 return value means the transaction is invalid
 func (sealed *SealedEnvelope) Hash() hash.Hash256 {
-	if IsRLP(sealed.encoding) {
+	switch sealed.encoding {
+	case iotextypes.Encoding_ETHEREUM_RLP:
 		tx, err := actionToRLP(sealed.Action())
 		if err != nil {
-			return hash.ZeroHash256
+			panic(err)
```
dustinxie, line 46: same thing here, in case err != nil, the node stopped running

action/sealedenvelope.go
CoderZhi: I thought your code https://github.com/iotexproject/iotex-core/pull/2601/files#diff-fda8afdb7165ec30b8d2c40dfd1963def0ffeb117a3d9098167b8139c72527caL105 covers the test

go.sum
```diff
@@ -524,14 +524,17 @@ github.com/openzipkin/zipkin-go v0.1.6/go.mod h1:QgAqvLzwWbR/WpD4A3cGpPtJrZXNIiJ
 github.com/oschwald/maxminddb-golang v1.5.0 h1:rmyoIV6z2/s9TCJedUuDiKht2RN12LWJ1L7iRGtWY64=
 github.com/oschwald/maxminddb-golang v1.5.0/go.mod h1:3jhIUymTJ5VREKyIhWm66LJiQt04F0UCDdodShpjWsY=
 github.com/pascaldekloe/goe v0.0.0-20180627143212-57f6aae5913c/go.mod h1:lzWF7FIEvWOWxwDKqyGYQf6ZUaNfKdP144TG7ZOy1lc=
+github.com/pborman/uuid v1.2.0 h1:J7Q5mO4ysT1dv8hyrUGHb9+ooztCXu1D8MY8DZYsu3g=
```
dustinxie, line 527: try run `go mod tidy` see if these will clean-up

action/sealedenvelope.go
dustinxie: no, that is to check, if encoding = 1, the ActionCore inside MUST be either Transfer or Execution

someone can send in a valid Transfer/Execution with encoding = 2, causing a panic

action/transfer_test.go
```diff
@@ -35,8 +34,6 @@ func TestTransferSignVerify(t *testing.T) {
 	elp, ok := eb.(*envelope)
 	require.True(ok)
 
-	require.Equal("0801100118a08d0622023130522f0a0231301229696f3172633264326465377274757563616c7371763464396e673068323937743633773777766c7068", hex.EncodeToString(elp.serialize()))
-
```
dustinxie, line 39: suggest to keep `serialize()` and related tests, b/c they help verify protobuf backward compatibility

raullenchai (approved):

action/sealedenvelope.go
CoderZhi: how could that happen with https://github.com/iotexproject/iotex-core/pull/2601/files#diff-fda8afdb7165ec30b8d2c40dfd1963def0ffeb117a3d9098167b8139c72527caL96?

action/execution_test.go
```diff
@@ -36,8 +36,6 @@ func TestExecutionSignVerify(t *testing.T) {
 	elp, ok := eb.(*envelope)
 	require.True(ok)
 
-	require.Equal("0801100218a08d0622023130622f0a0231301229696f3172633264326465377274757563616c7371763464396e673068323937743633773777766c7068", hex.EncodeToString(elp.serialize()))
```
dustinxie, line 39: suggest to keep `serialize()` and related tests, b/c they help verify protobuf backward compatibility

action/sealedenvelope.go
```diff
@@ -14,44 +14,46 @@ import (
...
 		}
-		h, err := rlpSignedHash(tx, sealed.externChainID, sealed.Signature())
+		h, err := rlpSignedHash(tx, sealed.evmNetworkID, sealed.Signature())
 		if err != nil {
-			return hash.ZeroHash256
+			panic(err)
```
dustinxie, line 50: same thing here, in case err != nil, the node panics

action/sealedenvelope.go
dustinxie: ok, cannot get here
what about the `rlpSignedHash` above, in case err happens, it will panic
i think we better stay on safe side

dustinxie (commented):
couple of new comments

action/rlp_tx.go
```diff
@@ -34,15 +34,17 @@ func rlpSignedHash(tx rlpTransaction, chainID uint32, sig []byte) (hash.Hash256,
 	if len(sig) != 65 {
 		return hash.ZeroHash256, errors.Errorf("invalid signature length = %d, expecting 65", len(sig))
 	}
-	if sig[64] >= 27 {
-		sig[64] -= 27
+	sc := make([]byte, 65)
```
dustinxie, line 37: not necessary, input is `sealed.Signature()`, which already did make([]byte), copy

action/sealedenvelope.go
```diff
@@ -98,11 +98,20 @@ func (sealed *SealedEnvelope) LoadProto(pbAct *iotextypes.Action) error {
...
 	encoding := pbAct.GetEncoding()
 	switch encoding {
 	case iotextypes.Encoding_ETHEREUM_RLP:
 		// verify action type can support RLP-encoding
-		if _, err := actionToRLP(elp.Action()); err != nil {
+		tx, err := actionToRLP(elp.Action())
+		if err != nil {
+			return err
+		}
+		if _, err = rlpSignedHash(tx, config.EVMNetworkID(), pbAct.GetSignature()); err != nil {
```
dustinxie, line 114: IMO, this is doing extra work so that you can use `panic`, which is weird when you don't have to use `panic`

what if there are more func calls later in the SealEnvelope code? we need to call those all in advance, so that `panic` can be used

it defeats the purpose of `panic` where you have no choice, but here we can return 0-hash, which will fail `Verify` gracefully

raullenchai (approved):

PR #2603, opened 2021-04-14
[block] add GenesisBlock

dustinxie:


blockchain/block/genesis_block.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/iotexproject/iotex-core/config"
+	"github.com/iotexproject/iotex-core/pkg/version"
+)
+
+// constants
+const (
+	GenesisHashMainnet = "ab7d006c1f7a9345ad05eef1b4f062814a176c25c7558052e18896844ee71edb"
+	GenesisHashTestnet = "84b5d7158407b85c07ba1bcac0699dedc86d22bbcb29b6d77a5f696ac35047fd"
+)
```
CoderZhi, line 23: Where do we use them? Where should we put them?

config/config.go
```diff
@@ -474,8 +478,9 @@ func New(validates ...Validate) (Config, error) {
 		}
 	}
 
-	// populdate chain ID
+	// populdate chain ID and genesis timestamp
 	SetEVMNetworkID(cfg.Chain.EVMNetworkID)
+	SetGenesisTimestamp(cfg.Genesis.Timestamp)
```
CoderZhi, line 483: SetEVMNetworkID is not the right way to set network id. network id is part of the blockchain or chain service, but not a global variable of the codebase. Similarly, we shouldn't do set the genesis timestamp in this way.

dustinxie: sure, that is your opinion "not the right way", IMO these are just diff options, not necessarily "right" or "wrong", in main.go it goes like this:
```
cfg, err := config.New()

// create and start the node
svr, err := itx.NewServer(cfg)
```
since input `cfg` determines `NewServer`, which is the blockchain or chain service you mentioned, i think `config.New()` is the right place to set these system-wide variables, i don't see why we cannot do it this way

dustinxie: also, i would like to see more constructive/suggestive comments, maybe something like
```
consider set EVM network ID and genesis timestamp in chain service? since these are system-wide variables used by chain service
```
when I see `xxx is not the right way, we shouldn't do yyy in this way`, it does not give a clear message/clue as to what is the exact problem here? why the reviewer think this is wrong? what the reviewer suggest could be a better way? maybe i am too dumb to get it, but that is my honest feeling

blockchain/block/genesis_block.go
dustinxie: GenesisHashMainnet is used in test to make sure the calculated genesis hash (with mainnet genesis timestamp) is correct
can be moved, will fix

CoderZhi (approved):
approve to unblock. Will address the issue of config in following PR.

PR #2609, opened 2021-04-20
[log] remove networkAddr in log

dustinxie:


pkg/log/fields.go
```diff
@@ -22,3 +22,11 @@ import (
 func Hex(k string, d []byte) zap.Field {
 	return zap.String(k, hex.EncodeToString(d))
 }
+
+// DisableVerbose returns a zap field to log the error w/o adding "errorVerbose" filed
+func DisableVerbose(err error) zap.Field {
+	if err == nil {
+		return zap.Skip()
+	}
+	return zap.String("error", err.Error())
```
dustinxie, line 31: this conversion would make zap NOT add "errorVerbose" in the log

raullenchai (changes requested):
No need to disable zap

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -458,7 +458,9 @@ func (ctx *rollDPoSCtx) Commit(msg interface{}) (bool, error) {
 	if pendingBlock == nil {
 		return false, nil
 	}
-	ctx.logger().Info("consensus reached", zap.Uint64("blockHeight", ctx.round.Height()))
+	if ctx.round.Height()%100 == 0 {
+		ctx.logger().Info("consensus reached", zap.Uint64("blockHeight", ctx.round.Height()))
+	}
```
dustinxie, line 463: reduce log frequency

server/main.go
```diff
@@ -118,12 +121,9 @@ func main() {
 	<-livenessCtx.Done()
 }
 
-func initLogger(cfg config.Config) {
+func initLogger(cfg config.Config) error {
 	addr := cfg.ProducerAddress()
-	if err := log.InitLoggers(cfg.Log, cfg.SubLogs, zap.Fields(
+	return log.InitLoggers(cfg.Log, cfg.SubLogs, zap.Fields(
 		zap.String("ioAddr", addr.String()),
-		zap.String("networkAddr", fmt.Sprintf("%s:%d", cfg.Network.Host, cfg.Network.Port)),
```
dustinxie, line 125: normal log does not need "networkAddr", move it to heartbeat.Log()

blockchain/blockchain.go
```diff
@@ -472,7 +472,9 @@ func (bc *blockchain) commitBlock(blk *block.Block) error {
 		return err
 	}
 	blkHash := blk.HashBlock()
-	blk.HeaderLogger(log.L()).Info("Committed a block.", log.Hex("tipHash", blkHash[:]))
+	if blk.Height()%100 == 0 {
```
raullenchai, line 475: https://github.com/uber-go/zap/blob/master/config.go#L39

consensus/scheme/rolldpos/rolldposctx.go
raullenchai: https://github.com/uber-go/zap/blob/master/config.go#L39

dustinxie:
> No need to disable zap

this does not disable zap, but remove the "errorVerbose" part in message which is not very necessary

blockchain/blockchain.go
dustinxie: it is used to filter out excessive logs, say > 20 same entries per second

i don't think this fits our use-case? since we are reducing to a very small frequency, log 1 block out of 100. If we apply this sampling, means we would also only log 1 error out of 100.

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -458,7 +458,9 @@ func (ctx *rollDPoSCtx) Commit(msg interface{}) (bool, error) {
 	if pendingBlock == nil {
 		return false, nil
 	}
-	ctx.logger().Info("consensus reached", zap.Uint64("blockHeight", ctx.round.Height()))
+	if ctx.round.Height()%100 == 0 {
```
raullenchai, line 461: can we do DEBUG on this one w/o sampling?

server/main.go
```diff
@@ -59,7 +59,10 @@ func main() {
 	if err != nil {
 		glog.Fatalln("Failed to new config.", zap.Error(err))
 	}
-	initLogger(cfg)
+	if err = initLogger(cfg); err != nil {
+		glog.Println("Cannot config global logger, use default one: ", err)
```
raullenchai, line 63: glog.Fatalln

raullenchai (approved):

server/main.go
dustinxie: fixed

consensus/scheme/rolldpos/rolldposctx.go
dustinxie: i still want to keep this to indicate consensus OK, similar to `committed a block` above

it only adds 1 line every 100 blocks (8 minutes), so won't increase log size

PR #2616, opened 2021-04-27
call SetEVMNetworkID() and SetGenesisTimestamp() in main()

dustinxie:


config/config.go
```diff
@@ -478,10 +473,6 @@ func New(validates ...Validate) (Config, error) {
 		}
 	}
 
-	// populdate chain ID and genesis timestamp
-	SetEVMNetworkID(cfg.Chain.EVMNetworkID)
-	SetGenesisTimestamp(cfg.Genesis.Timestamp)
```
dustinxie, line 483: here cfg.Genesis is the default genesis, so it always use the same timestamp, resulting in same genesis block hash

server/main.go
```diff
@@ -54,6 +54,11 @@ func main() {
 	if err != nil {
 		glog.Fatalln("Failed to new genesis config.", zap.Error(err))
 	}
+	// set genesis timestamp
+	genesis.SetGenesisTimestamp(genesisCfg.Timestamp)
```
dustinxie, line 58: here 
```
genesisCfg, err := genesis.New()
```
takes timestamp input from config/yaml file, so should call SetGenesisTimestamp() here

CoderZhi (approved):

PR #2623, opened 2021-04-29
move evm network id and genesis block into context

CoderZhi:
1. Refine Genesis Block related code
2. Move EVM Network Chain ID into context

### DIFF
--- a/action/action_test.go
+++ b/action/action_test.go
@@ -7,6 +7,7 @@
 package action
 
 import (
+	"context"
 	"encoding/hex"
 	"math/big"
 	"strings"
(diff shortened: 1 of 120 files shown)


dustinxie:
IMHO this is not the preferred way to handle these 2:

we should not (and don't need to) add `context` to 90% of all APIs only b/c of these 2. For APIs like protocol handlers, a context makes sense, we use that to pass in per-action, per-block variables, like action hash, blockchain tip, block height, gas limit, etc. But for many APIs like `LoadProto`, `convertBlockPb`, and entire `blockdao.xxx`, it is not necessary to use a context. `blockdao.GetBlock(100)` is just to retrieve block height 100 from DB, i don't see why it needs a context.

PR #2648, opened 2021-05-05
[p2p] add QoS and better reconnect policy

dustinxie:
the first fix for reconnect is not robust, this PR changes reconnect policy to: reconnect if there's no active p2p message sent/received in past 5 min

also add the following QoS metrics into P2P agent, blocksync can use it to filter/select peer
```
// BroadcastSendTotal returns total amount of broadcast send
func (p *Agent) BroadcastSendTotal() uint64 {}

// BroadcastSendSuccessRate return the broadcast send success rate
func (p *Agent) BroadcastSendSuccessRate() float64 {}

// BroadcastRecvTotal returns total amount of broadcast recv
func (p *Agent) BroadcastRecvTotal() uint64 {}

// UnicastSendCount returns total amount of unicast send to peer
func (p *Agent) UnicastSendCount(name string) (uint64, bool) {}

// UnicastSendSuccessRate return the unicast send success rate
func (p *Agent) UnicastSendSuccessRate(name string) (float64, bool) {}

// UnicastRecvCount returns total amount of unicast recv from peer
func (p *Agent) UnicastRecvCount(name string) (uint64, bool) {}
```


raullenchai (changes requested):
We should revert the reconnect change

p2p/qos.go
```diff
@@ -0,0 +1,60 @@
...
+
+import "time"
+
+type qos struct {
+	broadcastSendCount    uint64
+	broadcastSendSuccess  uint64
+	broadcastRecvCount    uint64
+	unicastSendCount      uint64
+	unicastSendSuccess    uint64
+	unicastRecvCount      uint64
```
CoderZhi, line 17: not needed in this PR.

p2p/qos.go
```diff
@@ -0,0 +1,128 @@
...
+func (q *qos) updateSendUnicast(peername string, t time.Time, success bool) {
+	peer, exist := q.metrics[peername]
+	if !exist {
+		peer = new(transmitMetric)
+		q.metrics[peername] = peer
+	}
+	atomic.AddUint64(&peer.unicastSendCount, 1)
+	if success {
+		atomic.AddUint64(&peer.unicastSendSuccess, 1)
+		atomic.StoreInt64(&q.lastActiveUnicastTs, t.UnixNano())
```
CoderZhi, line 76: same here

p2p/qos.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (q *qos) updateRecvUnicast(peername string, t time.Time) {
+	peer, exist := q.metrics[peername]
+	if !exist {
+		peer = new(transmitMetric)
+		q.metrics[peername] = peer
+	}
```
CoderZhi, line 85: if we use atomic for the other metrics values, how about q.metrics?

p2p/qos.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (q *qos) lastUnicastTime() time.Time {
+	return time.Unix(0, atomic.LoadInt64(&q.lastActiveUnicastTs))
+}
+
+func (q *qos) updateSendBroadcast(t time.Time, success bool) {
+	atomic.AddUint64(&q.broadcastSendCount, 1)
+	if success {
+		atomic.AddUint64(&q.broadcastSendSuccess, 1)
+		atomic.StoreInt64(&q.lastActiveBroadcastTs, t.UnixNano())
```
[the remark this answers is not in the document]
dustinxie: it is both, either success send or receive means the p2p connection is good

p2p/qos.go
dustinxie: it is both

p2p/qos.go
dustinxie: yeah, in case of multiple thread, can become race condition here
so, seems cannot avoid a lock ... let me revise

CoderZhi (commented):
Two suggestions:
1. If we use lock in qosMetrics, standardize it to 
```
    lock.Lock
    defer lock.Unlock()
```
instead of mixture lock and atomic, because the efficiency impact is limited, but code is complicated
2. Expose qosMetrics, because we don't introduce anything else in the wrapper functions:
https://github.com/iotexproject/iotex-core/pull/2648/files#diff-0d3cb1ef41af0d930f0441c4b12ed50c607cc45d49b1a31a111225408833eaa5R414-R443

CoderZhi (approved):

raullenchai (approved):

PR #2652, opened 2021-05-07
get and set nonce in state db adapter

CoderZhi:
1. Add func "InitNonce" in StateDB
2. If usePendingNonce is enabled:
   InitNonce will be 1
   GetNonce return pending nonce
   SetNonce will receive next pending nonce

### DIFF
--- a/action/protocol/execution/evm/evmstatedbadapter.go
+++ b/action/protocol/execution/evm/evmstatedbadapter.go
@@ -65,6 +65,7 @@ type (
 		notFixTopicCopyBug  bool
 		asyncContractTrie   bool
 		sortCachedContracts bool
+		usePendingNonce     bool
 	}
 )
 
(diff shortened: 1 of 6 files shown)


raullenchai (approved):

guo (approved):

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -220,23 +229,37 @@ func (stateDB *StateDBAdapter) GetBalance(evmAddr common.Address) *big.Int {
...
 		log.L().Error("Failed to convert evm address.", zap.Error(err))
 		return 0
 	}
+	nonce := uint64(0)
 	state, err := stateDB.AccountState(addr.String())
 	if err != nil {
 		log.L().Error("Failed to get nonce.", zap.Error(err))
 		// stateDB.logError(err)
-		return 0
+	} else {
```
dustinxie, line 252: nit:
`return InitNonce()` in if{}? 
then can remove else{}

PR #2660, opened 2021-05-10
better block sync strategy

CoderZhi:
1. cache multiple blocks per height
2. query at least two DIFFERENT neighbors
3. update lastUpdateTime on flush

The above changes will reduce the hurts from attackers


blocksync/buffer.go
```diff
@@ -74,25 +74,35 @@ func (b *blockBuffer) AddBlock(tipHeight uint64, blk *block.Block) {
...
-
-		if _, ok := b.blocks[blkHeight]; !ok {
-			b.blocks[blkHeight] = blk
+		if _, ok := b.blockQueues[blkHeight]; !ok {
+			b.blockQueues[blkHeight] = &uniQueue{
+				blocks: []*block.Block{},
+				hashes: map[hash.Hash256]bool{},
+			}
 		}
+		b.blockQueues[blkHeight].enque(blk)
```
dustinxie, line 83: one thought from that wrong block (timestamp is 2 months back) you found:
we can save the last successfully committed block's time, and accept incoming blocks only if timestamp is later than that
```
if blk.Time.Before(b.tipTime) { continue }
```
b.tipTime is updated every time a block is successfully committed

PR #2685, opened 2021-05-26
better block sync strategy 2

maswalker:
1. add blocklist in blocksync
2. add limit rate in dispatcher

### DIFF
--- a/chainservice/chainservice.go
+++ b/chainservice/chainservice.go
@@ -378,19 +378,31 @@ func New(
 
 			return nil
 		},
-		func(ctx context.Context, start uint64, end uint64) error {
+		func(ctx context.Context, start uint64, end uint64, repeat int) {
 			peers, err := p2pAgent.Neighbors(ctx)
 			if err != nil {
-				return err
+				log.L().Error("failed to get neighbours", zap.Error(err))
+				return
 			}
 			if len(peers) == 0 {
-				return errors.New("no peers")
+				log.L().Error("no peers")
 			}
-			return p2pAgent.UnicastOutbound(
-				p2p.WitContext(ctx, p2p.Context{ChainID: chain.ChainID()}),
-				peers[rand.Intn(len(peers))],
-				&iotexrpc.BlockSync{Start: start, End: end},
-			)
+			if repeat < 2 {
+				repeat = 2
+			}
+			if repeat > len(peers) {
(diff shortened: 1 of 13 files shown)


dispatcher/dispatcher.go
```diff
@@ -210,14 +215,25 @@ func (d *IotxDispatcher) blockHandler() {
...
+
 // syncHandler handles incoming block sync requests
 func (d *IotxDispatcher) syncHandler() {
 	for {
 		select {
 		case m := <-d.syncChan:
-			d.handleBlockSyncMsg(m)
-			// TODO: better strategy of pacing
-			time.Sleep(5 * time.Second)
+			if (d.checkSyncPermission(m.peer.ID.Pretty())) {
```
dustinxie, line 234: if checkSyncPermission == false
the request is simply discarded
i don't think this is a good strategy to serve the sync request?
for example, request [0, 16], [17, 32], [33, 48], received by A, B, A
A first serves [0, 16], then discard [33, 48] due to syncInterval has not passed

chainservice/chainservice.go
```diff
@@ -378,19 +378,31 @@ func New(
...
-				return errors.New("no peers")
+				log.L().Error("no peers")
+			}
+			if repeat < 2 {
+				repeat = 2
+			}
+			if repeat > len(peers) {
+				repeat = len(peers)
+			}
+			for i := 0; i < repeat; i++ {
```
dustinxie, line 396: why we want to send same request to >= 2 peers?
do we have test/data to indicate there's a high probability of sync request not received/served by peers? my concern is that this will increase (unnecessary) data traffic on the p2p network

dispatcher/dispatcher.go
dustinxie: we definitely need a better way other than `time.Sleep(5 * time.Second)`, but IMO maybe not simply discarding the request, can we think of a better solution?

raullenchai (approved):

chainservice/chainservice.go
raullenchai: The ideal case would be `adaptive`, e.g., request from 1, if doesn't work, request from 2; then 4; 8, 16...

2 seems a good start. we can get this one onto testnet and see how it works.

raullenchai (approved):

PR #2689, opened 2021-05-29
Fix gas parameter passed to evm.Create

niklr:
Currently, the intrinsic gas of an execution is subtracted from the overall gas before passing it to evm.Create. In most cases this works just fine but for reasonably large contracts the deployment fails (not due to max. code size being exceeded!). This is most probably not intended because this calculation is carried out by the contract itself of go-ethereum. The function [UseGas](https://github.com/ethereum/go-ethereum/blob/4bb3c89d44e372e6a9ab85a8be0c9345265c763a/core/vm/contract.go#L123) is called inside [evm.Create](https://github.com/ethereum/go-ethereum/blob/4bb3c89d44e372e6a9ab85a8be0c9345265c763a/core/vm/evm.go#L346).

With this change all UMA protocol contracts can be easily deployed as requested [here](https://gitcoin.co/issue/iotexproject/halogrants/34/100025754)

### DIFF
--- a/action/protocol/execution/evm/evm.go
+++ b/action/protocol/execution/evm/evm.go
@@ -292,7 +292,7 @@ func executeInEVM(evmParams *Params, stateDB *StateDBAdapter, hu config.HeightUp
 	if evmParams.contract == nil {
 		// create contract
 		var evmContractAddress common.Address
-		_, evmContractAddress, remainingGas, evmErr = evm.Create(executor, evmParams.data, remainingGas, evmParams.amount)
+		_, evmContractAddress, remainingGas, evmErr = evm.Create(executor, evmParams.data, evmParams.gas, evmParams.amount)
 		log.L().Debug("evm Create.", log.Hex("addrHash", evmContractAddress[:]))
 		if evmErr == nil {
 			if contractAddress, err := address.FromBytes(evmContractAddress.Bytes()); err == nil {


action/protocol/execution/evm/evm.go
```diff
@@ -292,7 +292,7 @@ func executeInEVM(evmParams *Params, stateDB *StateDBAdapter, hu config.HeightUp
 	if evmParams.contract == nil {
 		// create contract
 		var evmContractAddress common.Address
-		_, evmContractAddress, remainingGas, evmErr = evm.Create(executor, evmParams.data, remainingGas, evmParams.amount)
+		_, evmContractAddress, remainingGas, evmErr = evm.Create(executor, evmParams.data, evmParams.gas, evmParams.amount)
```
dustinxie, line 295: it is not as simple as change `remainingGas` to `evmParams.gas`

intrinsicGas is first deducted as a base cost, so the evm is left to run with `remainingGas` amount of gas

dustinxie: as you mentioned, "reasonably large contracts the deployment fails", for such case the solution is to increase the gas limit, the transaction will only be charged the gas amount that was actually consumed by EVM, so increasing gas limit does no harm

dustinxie:
thx for your contribution! please see comment

niklr:
Not sure how the gas limit can be increased in this case. I'm using Hardhat in combination with [babel-api](https://github.com/iotexproject/babel-api) to deploy the contracts. Babel-api calls [estimateActionGasConsumptionForExecution](https://github.com/iotexproject/iotex-core/blob/27aba2d6fb3c513190dd470161849732696a9a27/api/api.go#L1507) and the gas limit is defined by `api.cfg.Genesis.BlockGasLimit`.

```
enough, receipt, err := api.isGasLimitEnough(callerAddr, sc, nonce, api.cfg.Genesis.BlockGasLimit)
```

By default the BlockGasLimit is 20000000 and increasing this value does not seem to work. For the deployed mainnet and testnet this value can't be changed anyways. What am I missing?

jrynkiew:
I think you should simply increase the gas limit of the originally deployed contract, no?

niklr:
Probably found the issue. The gas limit is reduced from 20000000 to 5000000 in [newParams](https://github.com/iotexproject/iotex-core/blob/4e5fb0a7403bcce5974eea0869a32a40f58fe66a/action/protocol/execution/evm/evm.go#L106) with an empty standalone testnet:

```
if blkCtx.BlockHeight > 0 && hu.IsPre(config.Aleutian, blkCtx.BlockHeight) && gasLimit > preAleutianActionGasLimit {
gasLimit = preAleutianActionGasLimit
}
```

PR #2709, opened 2021-06-16
Small code optimization in protocol.go

### LINKED ISSUE #2702: Use Mock in action/protocol/execution/protocol_test.go

**Why is this needed**:

Liuhaai:
Small code optimization when solving issue #2702.
I think we can close #2702, because it is hard to mock blockchain in the protocol_test.go


raullenchai (approved):

action/protocol/execution/protocol.go
```diff
@@ -75,15 +75,12 @@ func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.St
...
-	if !ok {
-		return nil
-	}
-	// Reject oversize execution
-	if exec.TotalSize() > ExecutionSizeLimit {
-		return errors.Wrap(action.ErrActPool, "oversized data")
+	if exec, ok := act.(*action.Execution); ok {
+		if exec.TotalSize() > ExecutionSizeLimit {
+			return errors.Wrap(action.ErrActPool, "oversized data")
+		}
```
dustinxie, line 83: let's use the current code `if !ok { return nil }`, this is an early exit style used very commonly in our code

think about if future we need to add another validation check, like gas amount, the early exit style is better than changing to a big if {} statement

action/protocol/execution/protocol.go
```diff
@@ -75,15 +75,12 @@ func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.St
 	return receipt, nil
 }
 
-// Validate validates an execution
+// Validate validates the size of an execution
```
dustinxie, line 78: revert this, current code only checks size, but `Validate` is meant to validate entire action

action/protocol/execution/protocol_test.go
```diff
@@ -521,14 +521,11 @@ func TestProtocol_Validate(t *testing.T) {
...
-	t.Run("Oversized data", func(t *testing.T) {
-		tmpPayload := [32769]byte{}
-		data := tmpPayload[:]
-		ex, err := action.NewExecution("2", uint64(1), big.NewInt(0), uint64(0), big.NewInt(0), data)
-		require.NoError(err)
-		require.Equal(action.ErrActPool, errors.Cause(p.Validate(context.Background(), ex, nil)))
-	})
+	ex, err := action.NewExecution("2", uint64(1), big.NewInt(0), uint64(0), big.NewInt(0), data)
+	require.NoError(err)
+	require.Equal(action.ErrActPool, errors.Cause(p.Validate(context.Background(), ex, nil)))
```
dustinxie, line 528: this is good clean-up

dustinxie (approved):

action/protocol/execution/protocol.go
Liuhaai: good suggestions

PR #2713, opened 2021-06-24
flush right after receiving block

CoderZhi:
flush blocks right after receiving a block

### DIFF
--- a/blocksync/blocksync.go
+++ b/blocksync/blocksync.go
@@ -54,7 +54,6 @@ type blockSyncer struct {
 	commitBlockHandler   CommitBlock
 	requestBlocksHandler RequestBlocks
 
-	flushTask     *routine.RecurringTask
 	syncTask      *routine.RecurringTask
 	syncStageTask *routine.RecurringTask
 
@@ -101,7 +100,6 @@ func NewBlockSyncer(
 	}
 	if bs.cfg.Interval != 0 {
 		bs.syncTask = routine.NewRecurringTask(bs.sync, bs.cfg.Interval)
-		bs.flushTask = routine.NewRecurringTask(bs.flush, bs.cfg.Interval)
 		bs.syncStageTask = routine.NewRecurringTask(bs.syncStageChecker, bs.cfg.Interval)
 	}
 	atomic.StoreUint64(&bs.syncBlockIncrease, 0)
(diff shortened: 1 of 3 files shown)


blocksync/blocksync.go
```diff
@@ -231,12 +198,28 @@ func (bs *blockSyncer) ProcessBlock(ctx context.Context, peer string, blk *block
...
+			break
+		}
+		syncedHeight++
+	}
+	bs.buf.Cleanup(syncedHeight)
+	log.L().Debug("flush blocks", zap.Uint64("start", tip), zap.Uint64("end", syncedHeight))
+	if syncedHeight > bs.lastTip {
+		bs.lastTip = syncedHeight
+		bs.lastTipUpdateTime = time.Now()
+	}
```
dustinxie, line 222: seems `lastTip` is set here, but never used, i think it should replace `targetHeight`?

for example, we received block 87, 88 first, then `ProcessBlock(block 86)`, as a result, 86, 87, 88 are committed, and got `bs.lastTip = 88` here
but `bs.targetHeight = blk.Height()`, value is 86, shouldn't it be 88 as well?

blocksync/blocksync.go
```diff
@@ -231,12 +198,28 @@ func (bs *blockSyncer) ProcessBlock(ctx context.Context, peer string, blk *block
...
-	bs.buf.AddBlock(bs.tipHeightHandler(), newPeerBlock(peer, blk))
+	tip := bs.tipHeightHandler()
+	if !bs.buf.AddBlock(tip, newPeerBlock(peer, blk)) {
+		return nil
+	}
 	bs.mu.Lock()
 	defer bs.mu.Unlock()
 	if blk.Height() > bs.targetHeight {
 		bs.targetHeight = blk.Height()
 	}
```
dustinxie, line 209: bs.targetHeight should be = bs.lastTip? see comment below

blocksync/blocksync.go
CoderZhi: lastTip is used in line 219.
targetHeight is only set if blk.height > targetHeight

dustinxie: bs.lastTip only used in:
```
if syncedHeight > bs.lastTip {
		bs.lastTip = syncedHeight
		bs.lastTipUpdateTime = time.Now()
	}
```
so value of `bs.lastTip` being updated, but not used anywhere else in the code

dustinxie: instead of 
```
if blk.Height() > bs.targetHeight {
		bs.targetHeight = blk.Height()
	}
```
should use bs.lastTip
```
if bs.lastTip > bs.targetHeight {
		bs.targetHeight = bs.lastTip
	}
```

dustinxie (approved):

blocksync/blocksync.go
CoderZhi: this is exactly what lasttip for

raullenchai (approved):

Issue #2567, opened 2021-02-01
Allow ioctl to show the list of actions for an account


simonerom:
**What would you like to be added**:
ioctl account actions ADDRESS|ALIAS [ SKIP [FIRST] ]

At least it returns (for each tx):
count + type + status + recipient + amount + hash
```
# lists all actions sorted from the most recent
$ ioctl account actions myAccount 

# lists all actions sorted from the most recent, skipping the latest 10
$ ioctl account actions myAccount 10 

# lists the most recent action
$ ioctl account actions myAccount 0 1 

# Example
12 transfer success io1jsljstd5xt62gzyde0usvqdjyf4vhh8peks9lf 12 IOTX ca01376611a11215589aaa664f71c8e466a5b5288438e50026f977e6928e2c48
```

**Why is this needed**:
When web explorers fail, we need a way to still be able to explore wallet transactions, then we use` ioctl account action hash `to get more details on each transaction

millken:
#2568

millken:
pr #2568 submitted

simonerom:
I tested the command and it works fine, but it seems to not include all actions: specifically if we take  `https://iotexscan.io/address/<address>` explorer view as a reference, the ioctl command seems to only include actions that are listed in the "Total Actions" tab, but all actions in the "Contract Actions" tab seem not to be included. 

Assuming that the XRC20 and NFT actions are already included, can we also include the Contract Actions in the output?

simonerom:
> I will check it later, can you give some example address?

Sure: let's take this address as an example: https://iotexscan.io/address/io1jzdxuv7etyfvs7th7jwyspswr6y660zjd7lykg

the command `ioctl account actions io1jzdxuv7etyfvs7th7jwyspswr6y660zjd7lykg --limit 138` returns all the 138 actions that you can see on the iotexscan account summary.
In fact, iotexscan shows 138 actions, and these 138 actions are listed in the **Total Actions** tab:

![image](https://user-images.githubusercontent.com/11096047/108717360-7c9b8380-751d-11eb-91b1-3615397d30ed.png)

[...]

PR #2724, opened 2021-07-09
Index txs in the action

Liuhaai:
Try to fix issue #2567


dustinxie:
this will capture IOTX token transfer by contract, but won't capture XRC20 token transfer to an address. For example: i this action https://iotexscan.io/action/fa45221706f521e3c18638610c52d45166ae5d91b8c23d89a5f4a49cba5e5b1c

68.404138 MLP:GFT-IOTX is transferred to `io1jzdxuv7etyfvs7th7jwyspswr6y660zjd7lykg`, but tx log won't have this info.

I think the solution is to get these transfers from iotex-analytics, which already had all these saved and can be queried

Issue #2699, opened 2021-06-08
Use Mock in blocksync/blocksync_test.go


CoderZhi:
**What would you like to be added**:

**Why is this needed**:

Liuhaai:
No easy implementation to mock commitBlock() action when BlockSyncer processes the block. 
Close this issue

PR #2726, opened 2021-07-11
fix bug in TestBlockSyncerPeerBlockList()

Liuhaai:
fix a test bug when trying to solve issue #2699

### DIFF
--- a/blocksync/blocksync_test.go
+++ b/blocksync/blocksync_test.go
@@ -528,7 +528,7 @@ func TestBlockSyncerPeerBlockList(t *testing.T) {
 	require.NoError(chain.Start(ctx))
 	cs := mock_consensus.NewMockConsensus(ctrl)
 	cs.EXPECT().ValidateBlockFooter(gomock.Any()).Return(nil).Times(3)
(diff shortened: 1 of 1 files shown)


blocksync/blocksync_test.go
```diff
@@ -555,8 +555,8 @@ func TestBlockSyncerPeerBlockList(t *testing.T) {
 		nil,
 	)
 
-	require.NoError(bs.ProcessBlock(ctx, "peer1", blk2))
-	require.NoError(bs.ProcessBlock(ctx, "peer2", blk1))
+	require.NoError(bs.ProcessBlock(ctx, "peer1", blk1))
+	require.NoError(bs.ProcessBlock(ctx, "peer2", blk2))
```
dustinxie, line 559: trying to understand why we need this PR?

the testing is passing on latest code, and I guess process blk2 before blk1 is intentionally?

Liuhaai: Yes, it's intentionally. I add some comments on original code for quick understanding

dustinxie: ok, so I think we don't need this PR any more?

or, your change tested more cases for this blocksync test?

PR #2760, opened 2021-08-09
[api] improve ReadContract() API

dustinxie:


CoderZhi (commented):
what if we change the min gas consumption to here: https://github.com/iotexproject/babel-api/blob/e7d1b318140a200a8947c106b83c450c2bf60998/src/service/api.service.ts#L154

dustinxie:
> what if we change the min gas consumption to here: https://github.com/iotexproject/babel-api/blob/e7d1b318140a200a8947c106b83c450c2bf60998/src/service/api.service.ts#L154

ok, makes sense, this 21000 is only for web3 side

CoderZhi (approved):

raullenchai (approved):

PR #2767, opened 2021-08-23
fix err return bug

Liuhaai:
Current code in agent.go returns nil when err exists


p2p/agent.go
```diff
@@ -170,14 +170,12 @@ func (p *Agent) Start(ctx context.Context) error {
 			p2pMsgLatency.WithLabelValues("broadcast", strconv.Itoa(int(broadcast.MsgType)), status).Observe(float64(latency))
 		}()
 		if err = proto.Unmarshal(data, &broadcast); err != nil {
-			err = errors.Wrap(err, "error when marshaling broadcast message")
-			return
+			return errors.Wrap(err, "error when marshaling broadcast message")
```
dustinxie, line 173: this is golang's named return, err is defined
so the current code has no problem

PR #2787, opened 2021-09-13
API Tracing

millken:
standalone test
1. start jaeger using docker
```
docker run --name jaeger \                          
  -e COLLECTOR_ZIPKIN_HOST_PORT=:9411 \
  -p 5775:5775/udp \
  -p 6831:6831/udp \
  -p 6832:6832/udp \
  -p 5778:5778 \
  -p 16686:16686 \
  -p 14268:14268 \
  -p 14250:14250 \
  -p 9411:9411 \
  jaegertracing/all-in-one:1.26
```
2. add config into `standalone-config.yaml`
```
tracer:
  endpoint: http://localhost:14268/api/traces
```
3. start standalone server

```
./bin/server -plugin=gateway -config-path=/pathto/standalone-config.yaml -genesis-path=/pathto/standalone-genesis.yaml
```
4. call test api
```
grpcurl --plaintext -d '{"address": "io197jzepp686snmmejfn359atttn28qk35sg05jv"}' 127.0.0.1:14014 iotexapi.APIService.GetAccount
```
5. open jaeger UI in browser [http://127.0.0.1:16686/](http://127.0.0.1:16686/)


server/main.go
```diff
@@ -108,6 +110,18 @@ func main() {
 		glog.Fatalln("EVM Network ID is not set, call config.New() first")
 	}
 
+	tp, err := tracer.NewProvider(
+		tracer.WithEndpoint(cfg.Tracer.EndPoint),
+		tracer.WithInstanceID(cfg.Tracer.InstanceID),
+	)
```
CoderZhi, line 116: if tracer config is null, we should skip the integration of tracer

millken: https://github.com/iotexproject/iotex-core/blob/baeec450ad4cad3c8992213dbec55f97a0d94c50/pkg/tracer/tracer.go#L54-L65
Yes, it has been considered. if tracer config is null, tracer does not  sampling.

raullenchai (approved):

config/config.go
```diff
@@ -353,6 +353,14 @@ type (
 		RangeBloomFilterNumHash uint64 `yaml:"rangeBloomFilterNumHash"`
 	}
 
+	// Tracer is the config for tracer
+	Tracer struct {
+		// EndPoint the jaeger endpoint
```
dustinxie, line 358: jaeger?

raullenchai: https://www.jaegertracing.io/

pkg/tracer/tracer.go
```diff
@@ -0,0 +1,83 @@
+package tracer
+
+import (
+	"github.com/iotexproject/iotex-core/pkg/version"
```
dustinxie, line 4: move `iotex-core/` to end of list

pkg/tracer/tracer.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/sdk/resource"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+)
+
+const (
+	service = "iotex-server"
```
dustinxie, line 14: `_service = iotex-tracer`?

server/main.go
maswalker: If cfg.Tracer.EndPoint not configured, tracer.NewProvider should return (nil, nil) so that subsequent tracer operation could be simply skipped.
Since the tracer will be enabled on api server only, Don't introduce additional code logic on non-api server sounds better.

config/config.go
```diff
@@ -353,6 +353,14 @@ type (
 		RangeBloomFilterNumHash uint64 `yaml:"rangeBloomFilterNumHash"`
 	}
 
+	// Tracer is the config for tracer
+	Tracer struct {
```
dustinxie, line 357: move this struct to /pkg/tracer

server/itx/server.go
```diff
@@ -170,6 +171,9 @@ func (s *Server) Dispatcher() dispatcher.Dispatcher {
 
 // StartServer starts a node server
 func StartServer(ctx context.Context, svr *Server, probeSvr *probe.Server, cfg config.Config) {
+	span := trace.SpanFromContext(ctx)
+	span.AddEvent("start server")
+	defer span.End()
```
dustinxie, line 176: move to after httpServer below

server/main.go
dustinxie: +1, can the start/stop of tracer be moved into `StartServer`? or even better if it can be all handled inside api server's `Start/Stop`

config/config.go
```diff
@@ -370,6 +378,7 @@ type (
 		Log                log.GlobalConfig            `yaml:"log"`
 		SubLogs            map[string]log.GlobalConfig `yaml:"subLogs"`
 		Genesis            genesis.Genesis             `yaml:"genesis"`
+		Tracer             Tracer                      `yaml:"tracer"`
```
dustinxie, line 381: so here it is `tracer.Tracer`, we are modifying config this way

config/config.go
dustinxie: also, we need to specify a default endpoint? like `iotex.status.me/tracing` or `tracing.iotex.io`

config/config.go
```diff
@@ -353,6 +353,14 @@ type (
 		RangeBloomFilterNumHash uint64 `yaml:"rangeBloomFilterNumHash"`
 	}
 
+	// Tracer is the config for tracer
+	Tracer struct {
+		// EndPoint the jaeger endpoint
+		EndPoint string `yaml:"endpoint"`
+		// InstanceID MUST be unique for each instance of the same
+		InstanceID string `yaml:"instanceID"`
```
dustinxie, line 361: does InstanceID need to be in config? since it is unique for each instance, it should be assigned to a random number in the `New()`?

server/main.go
dustinxie: also, considering only API used tracer, is it possible to just add the `Endpoint string` to `API struct`?

config/config.go
millken: The instanceID should be easy to quickly locate the container/machine, like `us-aws-node1` , and make new random number (restart server) will break tracing.

config/config.go
millken: tracer `endpoint` leave blank(default value) will skip the integration of tracer, tracer does not sampling. maybe tracer is only used by us.

PR #2792, opened 2021-09-15
change info to debug in staking protocol

CoderZhi:
the log level was not appropriate.

### DIFF
--- a/action/protocol/staking/protocol.go
+++ b/action/protocol/staking/protocol.go
@@ -339,7 +339,7 @@ func (p *Protocol) handle(ctx context.Context, act action.Action, csm CandidateS
 	if receiptErr, ok := err.(ReceiptError); ok {
 		actionCtx := protocol.MustGetActionCtx(ctx)
 		log.L().With(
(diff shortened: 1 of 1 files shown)


action/protocol/staking/protocol.go
```diff
@@ -339,7 +339,7 @@ func (p *Protocol) handle(ctx context.Context, act action.Action, csm CandidateS
 	if receiptErr, ok := err.(ReceiptError); ok {
 		actionCtx := protocol.MustGetActionCtx(ctx)
 		log.L().With(
-			zap.String("actionHash", hex.EncodeToString(actionCtx.ActionHash[:]))).Info("Failed to commit staking action", zap.Error(err))
+			zap.String("actionHash", hex.EncodeToString(actionCtx.ActionHash[:]))).Warn("Failed to commit staking action", zap.Error(err))
```
dustinxie, line 342: can you also change the "failed to get committed state" to DEBUG?

dustinxie (approved):

action/protocol/staking/protocol.go
dustinxie: also, WARN is actually higher level than info, if want to mute need to change to DEBUG?

raullenchai (approved):

PR #2796, opened 2021-09-16
retry on delta state mismatch before Hawaii height

CoderZhi:


blockchain/blockdao/blockdao.go
```diff
@@ -164,15 +164,23 @@ func (dao *blockDAO) checkIndexers(ctx context.Context) error {
...
-			if err := indexer.PutBlock(protocol.WithBlockCtx(
-				ctxWithTip,
-				protocol.BlockCtx{
-					BlockHeight:    i,
-					BlockTimeStamp: blk.Timestamp(),
-					Producer:       producer,
-					GasLimit:       g.BlockGasLimit,
-				},
-			), blk); err != nil {
+			for retry := 1; retry <= 10; retry++ {
```
dustinxie, line 167: we have this retry mechanism in `chainservice.go`, before the sortmap fix it will retry 4 times, can change to 10 times there

dustinxie: on a diff topic, this is why i don't prefer the method of passing/using function handler, it is likely to forget the logic has been moved to upper layer (`chainservice.go`), and need to make change up there

IMHO it is better to contain this function handler inside the `BlockSync` interface as a method `CommitBlock`, chainservice receive `BlockSync` as input

Any business/logic change will be made inside BlockSync implementation, no need to involve `chainservice.go`. Anyway just my 2 cents

dustinxie: sorry, misread the code, retry in `chainservice.go` is for sync block
this is in Start() and needed

dustinxie (approved):

raullenchai (approved):

PR #2807, opened 2021-09-21
set target height if block height is out of buffer size

CoderZhi:
target height was not set if the block height is larger than tip height + buffer size.
This PR set target height to tip height + buffer size, if blockchain is far behind.

### DIFF
--- a/blocksync/buffer.go
+++ b/blocksync/buffer.go
@@ -66,18 +66,21 @@ func (b *blockBuffer) Cleanup(height uint64) {
 }
 
 // AddBlock tries to put given block into buffer and flush buffer into blockchain.
-func (b *blockBuffer) AddBlock(tipHeight uint64, blk *peerBlock) bool {
+func (b *blockBuffer) AddBlock(tipHeight uint64, blk *peerBlock) (bool, uint64) {
 	b.mu.Lock()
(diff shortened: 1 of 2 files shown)


blocksync/buffer.go
```diff
@@ -66,18 +66,21 @@ func (b *blockBuffer) Cleanup(height uint64) {
...
-	if blkHeight > tipHeight && blkHeight <= tipHeight+b.bufferSize {
-		if _, ok := b.blockQueues[blkHeight]; !ok {
-			b.blockQueues[blkHeight] = newUniQueue()
-		}
-		b.blockQueues[blkHeight].enque(blk)
-		return true
+	if blkHeight <= tipHeight {
+		return false, 0
+	}
+	if blkHeight > tipHeight+b.bufferSize {
```
dustinxie, line 76: is it possible `blkHeight <= tipHeight`? how to handle that?

dustinxie (approved):

raullenchai (approved):

blocksync/buffer.go
CoderZhi: if blkHeight <= tipHeight, the block will be discarded.
this fix is effective for the case that the node is far behind the tip

Issue #2737, opened 2021-07-19
Call fallback function in transfer actions


CoderZhi:
https://github.com/iotexproject/iotex-core/blob/7c758ab75007a17606563303816229b3d252778f/action/protocol/account/transfer.go#L73

**What would you like to be added**:
Enable transfer to a smart contract and trigger fallback function if any
 
**Why is this needed**:
Be compatible with more dapps

maswalker:
I think that transfer to contract has already been supported by execution.
please refer to the commit,
https://github.com/iotexproject/pebble-contracts/commit/122498880907af062b7652ed8801b4f00315227d

raullenchai:
My testing example
```
pragma solidity 0.7.3;

contract Receiver {
    event Received(address, uint);
    uint256 cnt = 0;
    
    function getcount() public view returns (uint256) { 
        return cnt;
    }
    
    receive() external payable {
        cnt += 1;
        emit Received(msg.sender, msg.value);
    }
}
```

Verified that every transfer of IOTX to this contract increases count by 1, indicating it works.

raullenchai:
Pending on #1600

raullenchai:
Closing as by-design: plain transfer actions to a contract address intentionally return a Failure receipt (`action/protocol/account/transfer.go`); contract fallback functions are invoked via execution actions instead, as confirmed in this thread.

PR #2809, opened 2021-09-22
Call fallback function in transfer actions

maswalker:
https://github.com/iotexproject/iotex-core/issues/2737

https://docs.soliditylang.org/en/v0.6.0/contracts.html#receive-ether-function
https://docs.soliditylang.org/en/v0.6.0/contracts.html#fallback-function


dustinxie:
trying to understand, what is this change/feature for?
looks like if the tx is a `transfer` (but send to a contract address), then interpret/run it as a `contract execution`.

but in that case, shouldn't the user send the tx as contract in the first place?

maswalker:
https://docs.soliditylang.org/en/v0.6.0/contracts.html#receive-ether-function
https://docs.soliditylang.org/en/v0.6.0/contracts.html#fallback-function

action/protocol/execution/protocol.go
```diff
@@ -62,9 +63,36 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
+	featureCtx := protocol.MustGetFeatureCtx(ctx)
 	exec, ok := act.(*action.Execution)
 	if !ok {
-		return nil, nil
+		if !featureCtx.TransferToContractFallback {
+			return nil, nil
+		}
+
+		tsf, ok := act.(*action.Transfer)
+		if !ok {
```
dustinxie, line 74: `if !ok || len(tsf.Payload()) == 0`
receive/fallback is for a call with empty calldata

action/protocol/execution/protocol.go
```diff
@@ -62,9 +63,36 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
+			return nil, nil
+		}
+
+		recipientAddr, err := address.FromString(tsf.Recipient())
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to decode recipient address %s", tsf.Recipient())
+		}
+
+		recipientAcct, err := accountutil.LoadAccount(sm, recipientAddr)
+		if err == nil && recipientAcct.IsContract() {
```
dustinxie, line 84: we don't check `IsContract()` for regular calls, so don't need this?

action/protocol/execution/protocol.go
```diff
@@ -62,9 +63,36 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
+			exec, _ = action.NewExecution(
+				tsf.Recipient(),
+				tsf.Nonce(),
+				tsf.Amount(),
+				tsf.GasLimit(),
+				tsf.GasPrice(),
+				tsf.Payload(),
```
dustinxie, line 91: `nil`

action/protocol/account/transfer_test.go
```diff
@@ -94,7 +94,7 @@ func TestProtocol_HandleTransfer(t *testing.T) {
 		},
 		// not enough balance
 		{
-			alfa, 3, big.NewInt(30000), bravo.String(), 10000, big.NewInt(1), false, state.ErrNotEnoughBalance, uint64(iotextypes.ReceiptStatus_Failure), 1,
+			alfa, 3, big.NewInt(30000000000), bravo.String(), 10000, big.NewInt(1), false, state.ErrNotEnoughBalance, uint64(iotextypes.ReceiptStatus_Failure), 1,
```
dustinxie, line 97: why we need this change?

action/protocol/execution/protocol.go
dustinxie, line 84:
```
if err != nil {
    return nil, err
}
if recipientAcct.IsContract()
```

action/protocol/execution/protocol.go
CoderZhi: data could be nonempty

CoderZhi (approved):

action/protocol/account/transfer_test.go
maswalker: reverted

action/protocol/execution/protocol.go
dustinxie: from: https://docs.soliditylang.org/en/v0.6.0/contracts.html#receive-ether-function

This function cannot have arguments, cannot return anything and must have external visibility and payable state mutability. It is executed on a call to the contract with **empty calldata**.

dustinxie: if allow non-empty data, it could end up calling an existing valid `payable` function

for example, the contract has a function `receiveFromFool() payable`, with func signature = `0x1234abcd`

then a transfer with payload = `0x1234abcd` will be converted to a call to `receiveFromFool()`, we need to prevent that

dustinxie: or do we want to allow such interpretation/convertion?

action/protocol/execution/protocol.go
```diff
@@ -62,9 +63,38 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
 
 // Handle handles an execution
 func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.StateManager) (*action.Receipt, error) {
+	featureCtx := protocol.MustGetFeatureCtx(ctx)
 	exec, ok := act.(*action.Execution)
 	if !ok {
-		return nil, nil
+		if !featureCtx.ConvertTransferToExecution {
```
dustinxie, line 69: refactor these code into:
```
exec, ok := convertToExecution(ctx, act, sm)
if !ok {
   return nil, nil
}
```
all the new logic goes to `convertToExecution`

action/protocol/execution/protocol.go
```diff
@@ -60,11 +61,45 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
+		tsf.GasLimit(),
+		tsf.GasPrice(),
+		tsf.Payload(),
+	)
+	return exec, nil
+}
+
 // Handle handles an execution
 func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.StateManager) (*action.Receipt, error) {
 	exec, ok := act.(*action.Execution)
```
dustinxie, line 97: var err errror

action/protocol/execution/protocol.go
```diff
@@ -60,11 +61,45 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
 	if !ok {
-		return nil, nil
+		exec, err := p.convertToExecution(ctx, act, sm)
```
dustinxie, line 99: `exec, err =`

dustinxie (approved):

action/protocol/execution/protocol.go
```diff
@@ -60,11 +61,46 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
+	}
+	exec, _ := action.NewExecution(
+		tsf.Recipient(),
+		tsf.Nonce(),
+		tsf.Amount(),
+		tsf.GasLimit(),
+		tsf.GasPrice(),
+		tsf.Payload(),
+	)
+	return exec, nil
```
CoderZhi, line 92:
```
          return action.NewExecution(
		tsf.Recipient(),
		tsf.Nonce(),
		tsf.Amount(),
		tsf.GasLimit(),
		tsf.GasPrice(),
		tsf.Payload(),
	)
```

action/protocol/execution/protocol.go
```diff
@@ -60,11 +61,46 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
 // Handle handles an execution
 func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.StateManager) (*action.Receipt, error) {
+	var err error
 	exec, ok := act.(*action.Execution)
 	if !ok {
-		return nil, nil
+		exec, err = p.convertToExecution(ctx, act, sm)
+		if exec == nil {
+			return nil, err
+		}
```
CoderZhi, line 103:
```
        if err != nil {
              return nil, err
        }
```

action/protocol/execution/protocol.go
```diff
@@ -60,11 +61,46 @@ func FindProtocol(registry *protocol.Registry) *Protocol {
...
 
+func (p *Protocol) convertToExecution(ctx context.Context, act action.Action, sm protocol.StateManager) (*action.Execution, error) {
+	featureCtx := protocol.MustGetFeatureCtx(ctx)
+	if !featureCtx.ConvertTransferToExecution {
+		return nil, nil
+	}
+	tsf, ok := act.(*action.Transfer)
+	if !ok {
+		return nil, nil
+	}
```
CoderZhi, line 72: move to L100

CoderZhi (approved):

PR #2821, opened 2021-09-29
[actpool] speed up act with higher gas price

Liuhaai:


actpool/actqueue.go
```diff
@@ -103,17 +104,24 @@ func NewActQueue(ap *actPool, address string, ops ...ActQueueOption) ActQueue {
 	return aq
 }
 
-// Overlap returns whether the current queue contains the given nonce
+// Overlap returns whether the current queue contains the given act
 func (q *actQueue) Overlaps(act action.SealedEnvelope) bool {
-	_, exist := q.items[act.Nonce()]
-	return exist
+	if actInPool, exist := q.items[act.Nonce()]; exist && hex.EncodeToString(actInPool.Signature()) == hex.EncodeToString(act.Signature()) {
```
CoderZhi, line 109: how does this work?

Liuhaai: Same acts whose signatures and gas prices are same will be rejected in https://github.com/Liuhaai/iotex-core/blob/issue%232801/actpool/actpool.go#L374. The act with higher gas price will be put in  https://github.com/Liuhaai/iotex-core/blob/issue%232801/actpool/actqueue.go#L118.

dustinxie: also had the same question, why need to check signature is equal now?

in line 123 below, tx with higher gas will replace the existing one in the queue, so for each `nonce`, there is ONLY 1 tx in the queue, right? so no need to check signature is equal?

Liuhaai: > also had the same question, why need to check signature is equal now?
> 
> in line 123 below, tx with higher gas will replace the existing one in the queue, so for each `nonce`, there is ONLY 1 tx in the queue, right? so no need to check signature is equal?

Sig checking is bad idea, It will be removed

actpool/actqueue.go
```diff
@@ -103,17 +102,16 @@ func NewActQueue(ap *actPool, address string, ops ...ActQueueOption) ActQueue {
...
 func (q *actQueue) Put(act action.SealedEnvelope) error {
 	nonce := act.Nonce()
-	if _, exist := q.items[nonce]; exist {
-		return errors.Wrap(action.ErrNonce, "duplicate nonce")
+	if actInPool, exist := q.items[nonce]; exist {
+		// act of higher gas price cut in line
+		if act.GasPrice().Cmp(actInPool.GasPrice()) != 1 {
+			return errors.Wrap(action.ErrNonce, "gas price is smaller than the act of same nonce")
+		}
+		q.items[nonce] = act
```
CoderZhi, line 113: simply replace the action of the same nonce works fine in this case, just because https://github.com/iotexproject/iotex-core/blob/976c94abd634c2a7f2a998aacd776d94db8b5c83/actpool/actqueue.go#L32 uses nonce for comparison. Moreover, the ttl of the new action will be the same as the old one.

We could first delete the old action, removing it from the items and index, and then re-add it into the queue

CoderZhi (changes requested):

PR #2837, opened 2021-10-13
extend the time to receive proposal endorsement

CoderZhi:
delegates will have longer time to collect proposal endorsement to. if a delegate collects enough proposal endorsement, it will re-send this proposal in its round.


dustinxie:
is this the implementation of reusing already validated working set we talked about?
or this is a new idea? can you explain more details?

PR #2841, opened 2021-10-14
reducing block gas limit as round increasing

CoderZhi:
fullness, as uint8, is in range of [0, 255]. The number will decrease as round number increasing: fullness = 255 / (round + 1).

### DIFF
--- a/action/protocol/execution/protocol_test.go
+++ b/action/protocol/execution/protocol_test.go
@@ -314,7 +314,7 @@ func runExecutions(
 		}
 		hashes = append(hashes, selpHash)
 	}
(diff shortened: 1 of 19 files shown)


dustinxie:
as we discussed, commit https://github.com/iotexproject/iotex-core/commit/5cb71a640c48d2e121ccaf5eb2243472da2c020c has greatly improved the chance of completing consensus, so I feel this PR is not needed at this moment.

on the other hand, IMO we should focus on re-use cached minted block to guarantee block creation in most 2 rounds (draft https://github.com/iotexproject/iotex-core/pull/2868), which brings a definite better ROI

PR #2853, opened 2021-10-20
defer IIP-11

dustinxie:


raullenchai (approved):

blockchain/genesis/genesis.go
```diff
@@ -495,6 +500,11 @@ func (g *Blockchain) IsLordHowe(height uint64) bool {
 	return g.isPost(g.LordHoweBlockHeight, height)
 }
 
+// IsToBeEnabled checks whether height is equal to or larger than toBeEnabled height
```
Liuhaai, line 503: Could we add more details about what toBeEndabled feature do?

Liuhaai (approved):

blockchain/genesis/genesis.go
dustinxie: explained in `genesis.go`
new features will use `toBeEndabledHeight` so they are disabled by default, before release, replace it with actual height name

PR #2867, opened 2021-10-29
[api]Tracing sampling support

millken:
half of traces will be sampled
```
api:
  tracer:
    endpoint: http://localhost:14268/api/traces
    samplingRatio: .5
```

### DIFF
--- a/pkg/tracer/tracer.go
+++ b/pkg/tracer/tracer.go
@@ -65,7 +65,7 @@ func WithInstanceID(instanceID string) Option {
 	}
 }
 
-// WithSamplingRatio defines the sampling rate
+// WithSamplingRatio defines the sampling ratio
 func WithSamplingRatio(samplingRatio string) Option {
 	return func(ops *optionParams) error {
 		ops.samplingRatio = samplingRatio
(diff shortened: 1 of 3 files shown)


CoderZhi (approved):
the code of optionParams is hard to follow, but the function itself is working.

pkg/tracer/tracer.go
```diff
@@ -76,11 +91,19 @@ func NewProvider(opts ...Option) (*tracesdk.TracerProvider, error) {
 	} else {
 		//skipped tracing when endpoint no set
 		return nil, nil
-		//trackerTracerProviderOption = append(trackerTracerProviderOption, tracesdk.WithSampler(tracesdk.ParentBased(tracesdk.TraceIDRatioBased(-1))))
 	}
 	if ops.serviceName == "" {
 		ops.serviceName = _service
 	}
+	if ops.samplingRatio != "" {
+		ratio, err := strconv.ParseFloat(ops.samplingRatio, 64)
```
Liuhaai, line 99: Do we need to judge the range of samplingratio here? (ratio < 0, ratio > 1)

millken: ratio >= 1 will always sample (default),< 0 are treated as zero will no sample

Liuhaai (approved):

PR #2868, opened 2021-10-30
[consensus] re-use cached minted block to speed up block creation

dustinxie:


state/factory/util.go
```diff
@@ -71,8 +72,16 @@ func calculateLogsBloom(ctx context.Context, receipts []*action.Receipt) bloom.B
...
+func generateWorkingSetCacheKey(blkHeader block.Header, _ string) hash.Hash256 {
+	sum := byteutil.Uint64ToBytesBigEndian(blkHeader.Height())
+	h := blkHeader.PrevHash()
+	sum = append(sum, h[:]...)
+	h = blkHeader.TxRoot()
+	sum = append(sum, h[:]...)
+	h = blkHeader.DeltaStateDigest()
+	sum = append(sum, h[:]...)
+	h = blkHeader.ReceiptRoot()
+	sum = append(sum, h[:]...)
```
dustinxie, line 84: height + prev hash + 3 roots => the uniqueness and correctness of the proposed block

Liuhaai: why is blkHeader.SerializeCore() replaced by height + prev hash + 3 roots?

dustinxie: b/c header contains timestamp, which is changed when the next round proposer re-sign the block

PR #2898, opened 2021-11-12
[test] start and stop test for ServerV2

millken:


api/serverV2_test.go
```diff
@@ -0,0 +1,69 @@
...
+	require := require.New(t)
+	cfg := newConfig(t)
+	config.SetEVMNetworkID(1)
+	svr, bfIndexFile, err := createServerV2(cfg, false)
+	require.NoError(err)
+	defer func() {
+		testutil.CleanupPath(t, bfIndexFile)
+	}()
+	require.NoError(svr.Start())
+	time.Sleep(5 * time.Second) //let server have enough time to start.
```
Liuhaai, line 31: Is sleep time 5s too long?

raullenchai: Sleep in the test code will cause flakiness. Pls use wait-until pattern

api/serverV2_test.go
```diff
@@ -0,0 +1,71 @@
...
+	cfg := newConfig(t)
+	config.SetEVMNetworkID(1)
+	svr, bfIndexFile, err := createServerV2(cfg, false)
+	require.NoError(err)
+	defer func() {
+		testutil.CleanupPath(t, bfIndexFile)
+	}()
+	require.NoError(svr.Start())
+	require.NoError(testutil.WaitUntil(100*time.Millisecond, 2*time.Second, func() (bool, error) {
+		return true, svr.Stop()
```
Liuhaai, line 32: Is the condition always `true`?

api/serverV2_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestNewServerV2(t *testing.T) {
+	require := require.New(t)
+	cfg := newConfig(t)
+	config.SetEVMNetworkID(1)
+	svr, bfIndexFile, err := createServerV2(cfg, false)
+	require.NoError(err)
+	defer func() {
+		testutil.CleanupPath(t, bfIndexFile)
+	}()
+	require.NoError(svr.Start())
```
CoderZhi, line 30: after calling this line, what have not been started?

millken: go func() {
		if err := svr.grpcServer.Serve(lis); err != nil {
			log.L().Fatal("grpc failed to serve.", zap.Error(err))
		}
	}()

> after calling this line, what have not been started?

api/serverV2_test.go
millken: yes

PR #2927, opened 2021-11-26
[api] fix web3 empty return

Liuhaai:
fix the bug that the server returns `null` instead of `[]` when calling web3 api `eth_getLogs`,

### DIFF
--- a/api/web3server.go
+++ b/api/web3server.go
@@ -623,7 +623,7 @@ func (svr *Web3Server) getTransactionReceipt(in interface{}) (interface{}, error
 		return nil, err
 	}
 	tx, err := svr.getTransactionFromActionInfo(actInfo)
-	if err != nil || tx == nil {
+	if err != nil {
 		return nil, err
 	}
 
--- a/api/web3server.go
+++ b/api/web3server.go
@@ -239,11 +239,12 @@ func (svr *Web3Server) handlePOSTReq(req *http.Request) interface{} {
 			err := errors.Wrapf(errors.New("web3 method not found"), "method: %s\n", web3Req.Get("method"))
 			return packAPIResult(nil, err, 0)
 		}
-		if err != nil {
+		resp := packAPIResult(res, err, int(web3Req.Get("id").Int()))
+		if err != nil || !isResultValid(resp) {
 			// temporally used for monitor and debug
 			log.L().Error("web3 server err", zap.String("input", fmt.Sprintf("%+v", web3Req)), zap.Error(err))
 		}
-		web3Resps = append(web3Resps, packAPIResult(res, err, int(web3Req.Get("id").Int())))
+		web3Resps = append(web3Resps, resp)
 		web3ServerMtc.WithLabelValues(method.(string)).Inc()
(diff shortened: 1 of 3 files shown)


api/web3server.go
```diff
@@ -831,31 +830,33 @@ func (svr *Web3Server) getFilterChanges(in interface{}) (interface{}, error) {
 	filterID = removeHexPrefix(filterID)
 	filterObj, err := loadFilterFromCache(svr.cache, filterID)
 	if err != nil {
-		return []interface{}{}, err
+		return nil, err
```
dustinxie, line 833: return nil is OK?
`nil` may cause panic if trying to json-decode it, but `[]interface{}{}` will not

api/web3server.go
```diff
@@ -831,31 +830,33 @@ func (svr *Web3Server) getFilterChanges(in interface{}) (interface{}, error) {
...
-
 	switch filterObj.FilterType {
 	case "log":
+		if filterObj.LogHeight > tipHeight {
+			return []logsObject{}, nil
+		}
 		from, to, hasNewLogs, err := svr.getLogQueryRange(filterObj.FromBlock, filterObj.ToBlock, filterObj.LogHeight)
 		if err != nil {
 			return nil, err
 		} else if !hasNewLogs {
```
dustinxie, line 848: remove `else`

api/web3server_utils.go
```diff
@@ -284,7 +284,7 @@ func (svr *Web3Server) getTransactionFromActionInfo(actInfo *iotexapi.ActionInfo
 
 func (svr *Web3Server) getTransactionCreateFromActionInfo(actInfo *iotexapi.ActionInfo) (transactionObject, error) {
 	tx, err := svr.getTransactionFromActionInfo(actInfo)
-	if err != nil {
+	if err != nil || tx == nil {
```
dustinxie, line 287: can you check if there's other similar cases?

api/web3server.go
Liuhaai: no. `func packAPIResult()` will only encode err

api/web3server_test.go
```diff
@@ -642,3 +642,47 @@ func TestGetNetworkID(t *testing.T) {
...
 }
+
+func TestIsResultValid(t *testing.T) {
+	require := require.New(t)
+	testData := []struct {
+		data    web3Resp
+		isValid bool
+	}{
+		{
+			data: web3Resp{
```
dustinxie, line 653: can remove `data:` and `isValid:`

dustinxie (approved):

raullenchai (approved):

api/web3server.go
```diff
@@ -623,7 +623,7 @@ func (svr *Web3Server) getTransactionReceipt(in interface{}) (interface{}, error
 		return nil, err
 	}
 	tx, err := svr.getTransactionFromActionInfo(actInfo)
-	if err != nil {
+	if err != nil || tx == nil {
```
raullenchai, line 626: why tx can be nil?

raullenchai (approved):

PR #2931, opened 2021-12-01
[api] fix unallocated memory of array 

Liuhaai:
similar with fix #2927
the array in response should be [] instead of null. The null value is due to the unallocated memory of the array


dustinxie (approved):

raullenchai (approved):

raullenchai (approved):
test coverage needed

PR #2935, opened 2021-12-04
[injector] pass map object by reference

pocockn:
**What would you like to be added:**
Replace *map[string]*big.Int with map[string]*big.Int

**Why is this needed:**
https://github.com/iotexproject/iotex-core/blob/master/tools/util/injectorutil.go#L159
map is passed by reference in nature

### DIFF
--- a/action/envelope_test.go
+++ b/action/envelope_test.go
@@ -1,12 +1,14 @@
 package action
 
 import (
+	"math/big"
 	"testing"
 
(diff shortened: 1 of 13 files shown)


pocockn:
Fixes  github.com/iotexproject/iotex-core/issues/2926

dustinxie (dismissed):

raullenchai (approved):

dustinxie (approved):

PR #2942, opened 2021-12-07
[evm] check null pointer before dump calldata

dustinxie:
this might be the reason for api node restart

### DIFF
--- a/action/protocol/execution/evm/evm.go
+++ b/action/protocol/execution/evm/evm.go
@@ -371,8 +371,15 @@ func executeInEVM(evmParams *Params, stateDB *StateDBAdapter, g genesis.Blockcha
 	if evmErr != nil {
 		errCode = evmErrToErrStatusCode(evmErr, g, blockHeight)
 		if errCode == uint64(iotextypes.ReceiptStatus_ErrUnknown) {
+			var addr string
+			if evmParams.contract != nil {
+				ioAddr, _ := address.FromBytes((*evmParams.contract)[:])
+				addr = ioAddr.String()
+			} else {
+				addr = "contract creation"
+			}
 			log.L().Warn("evm internal error", zap.Error(evmErr),
-				log.Hex("contract", (*evmParams.contract)[:]),
+				zap.String("address", addr),
 				log.Hex("calldata", evmParams.data))
 		}
 	}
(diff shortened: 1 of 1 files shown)


Liuhaai (approved):

action/protocol/execution/evm/evm.go
```diff
@@ -439,7 +446,6 @@ func evmErrToErrStatusCode(evmErr error, g genesis.Blockchain, height uint64) (e
 			case "no compatible interpreter":
 				errStatusCode = uint64(iotextypes.ReceiptStatus_ErrNoCompatibleInterpreter)
 			default:
-				log.L().Error("evm internal error", zap.Error(evmErr))
```
Liuhaai, line 442: why is this log removed

action/protocol/execution/evm/evm.go
```diff
@@ -371,8 +372,14 @@ func executeInEVM(evmParams *Params, stateDB *StateDBAdapter, g genesis.Blockcha
 	if evmErr != nil {
 		errCode = evmErrToErrStatusCode(evmErr, g, blockHeight)
 		if errCode == uint64(iotextypes.ReceiptStatus_ErrUnknown) {
+			var address string
+			if evmParams.contract != nil {
```
Liuhaai, line 376: when contract address is empty (""), will the contract field be nil?

pkg/log/fields.go
```diff
@@ -20,5 +20,5 @@ import (
 
 // Hex creates a zap field which convert binary to hex.
 func Hex(k string, d []byte) zap.Field {
-	return zap.String(k, hex.EncodeToString(d))
+	return zap.String(k, "0x"+hex.EncodeToString(d))
```
CoderZhi, line 23: is this related to the bug? if not, revert it

action/protocol/execution/evm/evm.go
dustinxie: it becomes the dump data in the PR

CoderZhi (approved):

action/protocol/execution/evm/evm.go
dustinxie: yes, empty "" --> evmParams.contract == nil

PR #2943, opened 2021-12-09
[api] fix nonce bug in getTransactionCount

Liuhaai:
https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1552

When blk number is "pending", the pending nonce of the addr in the actpool should be returned.
Otherwise, the number of confirmed nonce should be returned.

### DIFF
--- a/api/web3server.go
+++ b/api/web3server.go
@@ -97,6 +97,10 @@ var (
 	errNotImplemented = errors.New("method not implemented")
 	errInvalidFiterID = errors.New("filter not found")
 	errInvalidBlock   = errors.New("invalid block")
+
+	pendingBlockNumber  = "pending"
+	latestBlockNumber   = "latest"
+	earliestBlockNumber = "earliest"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


api/web3server.go
```diff
@@ -97,6 +97,10 @@ var (
 	errNotImplemented = errors.New("method not implemented")
 	errInvalidFiterID = errors.New("filter not found")
 	errInvalidBlock   = errors.New("invalid block")
+
+	pendingBlockNumber  = "pending"
```
dustinxie, line 101: `_pendingBlock = "pending"`

dustinxie (approved):

api/web3server.go
```diff
@@ -372,29 +377,19 @@ func (svr *Web3Server) getTransactionCount(in interface{}) (interface{}, error)
 	if err != nil {
 		return nil, err
 	}
-	num, err := svr.parseBlockNumber(blkNum)
-	if err != nil {
-		return nil, err
+	if blkNum == pendingBlockNumber {
+		pendingNonce, err := svr.coreService.ap.GetPendingNonce(ioAddr.String())
```
dustinxie, line 381: does `GetPendingNonce` have test in /actpool?
if not, add a test here to verify it works as expected

Liuhaai: yes

CoderZhi (approved):

PR #2947, opened 2021-12-09
[db] indexer call PutBlock() multiple times before Hawaii height

dustinxie:
this is done in full-node sync code
need to do the same for indexer during index re-building phase


CoderZhi:
https://github.com/iotexproject/iotex-core/pull/2796#discussion_r710354311

dustinxie:
same as #2796 , so close this one

PR #2955, opened 2021-12-17
use nonce = 0 for very first tx

dustinxie:
There's a request to use nonce = 0 for a new account's first tx, see https://github.com/gnosis/safe-singleton-factory/issues/10

1. this should NOT be a hard-fork b/c all our existing tx start from nonce = 1, now allowing start from 0 would not break it
2. this is code change on a quick check, may not cover all cases. If we decide to do this, need more scrutiny and comprehensive testing

### DIFF
--- a/action/protocol/generic_validator.go
+++ b/action/protocol/generic_validator.go
@@ -51,8 +51,7 @@ func (v *GenericValidator) Validate(ctx context.Context, selp action.SealedEnvel
 		return errors.Wrapf(err, "invalid state of account %s", caller.String())
 	}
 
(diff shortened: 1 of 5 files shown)


huof6829:
Glad to receive your reply.Thanks very much！Best  regards~

CoderZhi (commented):
set nonce may have problem

dustinxie:
> set nonce may have problem

right, i realized then nonce = 1 will be missing, so this is NOT possible without a hard-fork (add a flag in Account to indicate starting from 0 or 1)

PR #2959, opened 2021-12-20
fix nested ctx

CoderZhi:


dustinxie (approved):

Liuhaai (approved):

state/factory/workingset.go
```diff
@@ -404,6 +404,7 @@ func (ws *workingSet) pickAndRunActions(
 
 	// initial action iterator
 	blkCtx := protocol.MustGetBlockCtx(ctx)
+	ctxWithBlockContext := ctx
```
dustinxie, line 407: nit: maybe call it `baseCtx`? the name is kind of confusing with `blkCtx` above

PR #2961, opened 2021-12-22
[api] fix addr and txHash of the log of tx receipt

Liuhaai:


uu-z (approved):

api/web3server.go
```diff
@@ -673,13 +673,13 @@ func (svr *Web3Server) getTransactionReceipt(in interface{}) (interface{}, error
...
-		if contractAddr != nil {
-			addr = *contractAddr
+		addr, err := ioAddrToEthAddr(v.Address)
+		if err != nil {
+			return nil, err
 		}
 		log := logsObject{
 			BlockHash:        "0x" + blkHash,
-			TransactionHash:  actHashStr,
+			TransactionHash:  "0x" + hex.EncodeToString(v.ActionHash[:]),
```
dustinxie, line 682: this is not necessary? see below

dustinxie (approved):

api/web3server.go
Liuhaai: tx hash without `0x` should be added prefix

Issue #2938, opened 2021-12-06
[ioctl] build `account nonce` command line into new ioctl 


Liuhaai:
The new ioctl (./ioctl/newcmd) is a refactored version of the old ioctl(./ioctl/cmd). It includes two major changes:
- The [Client interface](https://github.com/iotexproject/iotex-core/blob/master/ioctl/client.go) is passed when constructing the Cobra command.
- Unit tests are added for each command.

The code is a draft. The code of old ioctl can be used for reference.

**What would you like to be added**:
add `accountnonce.go` and `accountnonce_test.go` in `./ioctl/newcmd/account`. 
**Code for reference**:
https://github.com/iotexproject/iotex-core/blob/master/ioctl/cmd/account/accountnonce.go

Issue #2869, opened 2021-10-31
[api] integrate getActionByActionHash() into GetActionByActionHash()


Liuhaai:
**What would you like to be added**:

**Why is this needed**:
Both `getActionByActionHash()` and `ActionByActionHash()` do similar jobs, they can be integrated together to simplify the code.

Input type:  hash.Hash256
Output type: action.SealedEnvelope, hash.Hash256, uint64, uint32, error

PR #2963, opened 2021-12-27
add account nonce and delete GetActionByActionHash

huof6829:
add account nonce  for #2938
del  GetActionByActionHash  for #2869


huof6829 (commented):
modify comment of func NewAccountCmd for build_test_docker

PR #2973, opened 2022-01-04
[statedb] use read lock to access currentChainHeight in reader funcs

dustinxie:
currentChainHeight is only written in `ws.Commit(ctx)`, which is protected by a write lock: https://github.com/iotexproject/iotex-core/blob/master/state/factory/statedb.go#L424-L440 
for all other places, read lock should suffice

test coverage:
1. make test
2. verified on full-node it can successfully commit/sync blocks

### DIFF
--- a/state/factory/statedb.go
+++ b/state/factory/statedb.go
@@ -302,7 +302,7 @@ func (sdb *stateDB) Validate(ctx context.Context, blk *block.Block) error {
 		if err = ws.ValidateBlock(ctx, blk); err != nil {
 			return errors.Wrap(err, "failed to validate block with workingset in statedb")
 		}
-		sdb.putIntoWorkingSets(key, ws)
+		sdb.workingsets.Add(key, ws)
 	}
 	receipts, err := ws.Receipts()
 	if err != nil {
@@ -320,9 +320,9 @@ func (sdb *stateDB) NewBlockBuilder(
 ) (*block.Builder, error) {
 	ctx = protocol.WithRegistry(ctx, sdb.registry)
 	sdb.mutex.RLock()
-	h := sdb.currentChainHeight
+	currHeight := sdb.currentChainHeight
 	sdb.mutex.RUnlock()
-	ws, err := sdb.newWorkingSet(ctx, h+1)
+	ws, err := sdb.newWorkingSet(ctx, currHeight+1)
 	if err != nil {
 		return nil, err
 	}
@@ -349,7 +349,7 @@ func (sdb *stateDB) NewBlockBuilder(
 
 	blkCtx := protocol.MustGetBlockCtx(ctx)
 	key := generateWorkingSetCacheKey(blkBuilder.GetCurrentBlockHeader(), blkCtx.Producer.String())
-	sdb.putIntoWorkingSets(key, ws)
+	sdb.workingsets.Add(key, ws)
 	return blkBuilder, nil
 }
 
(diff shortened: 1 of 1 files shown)


state/factory/statedb.go
```diff
@@ -553,22 +556,20 @@ func (sdb *stateDB) createGenesisStates(ctx context.Context) error {
...
+	sdb.mutex.RLock()
+	h := sdb.currentChainHeight
+	sdb.mutex.RUnlock()
+	tx, err := sdb.newWorkingSet(ctx, h+1)
 	return tx, false, err
 }
 
 func (sdb *stateDB) putIntoWorkingSets(key hash.Hash256, ws *workingSet) {
-	sdb.mutex.Lock()
-	defer sdb.mutex.Unlock()
```
dustinxie, line 572: sdb.workingsets.Add() is thread-safe, no need to use lock here

state/factory/statedb.go
```diff
@@ -318,10 +318,11 @@ func (sdb *stateDB) NewBlockBuilder(
 	ap actpool.ActPool,
 	sign func(action.Envelope) (action.SealedEnvelope, error),
 ) (*block.Builder, error) {
-	sdb.mutex.Lock()
 	ctx = protocol.WithRegistry(ctx, sdb.registry)
-	ws, err := sdb.newWorkingSet(ctx, sdb.currentChainHeight+1)
-	sdb.mutex.Unlock()
+	sdb.mutex.RLock()
+	h := sdb.currentChainHeight
```
Liuhaai, line 323: Replacing single char var `h` with longer var like `chainHeight` will make code editor easy to locate the var

state/factory/statedb.go
```diff
@@ -553,22 +556,20 @@ func (sdb *stateDB) createGenesisStates(ctx context.Context) error {
...
+	sdb.mutex.RUnlock()
+	tx, err := sdb.newWorkingSet(ctx, h+1)
 	return tx, false, err
 }
 
 func (sdb *stateDB) putIntoWorkingSets(key hash.Hash256, ws *workingSet) {
-	sdb.mutex.Lock()
-	defer sdb.mutex.Unlock()
 	sdb.workingsets.Add(key, ws)
```
Liuhaai, line 574: This code is very straightforward. I think this func can be removed.

Liuhaai (approved):

state/factory/statedb.go
```diff
@@ -446,27 +449,27 @@ func (sdb *stateDB) DeleteTipBlock(_ *block.Block) error {
...
 	cfg, err := processOptions(opts...)
 	if err != nil {
 		return 0, err
 	}
-
-	return sdb.currentChainHeight, sdb.state(cfg.Namespace, cfg.Key, s)
+	sdb.mutex.RLock()
+	h := sdb.currentChainHeight
+	sdb.mutex.RUnlock()
+	return h, sdb.state(cfg.Namespace, cfg.Key, s)
```
CoderZhi, line 459: state will read data from dao and then deserialize it. If we don't lock the whole struct, will `h` be inconsistent with the result read from `sdb.state`?

state/factory/statedb.go
```diff
@@ -446,27 +449,27 @@ func (sdb *stateDB) DeleteTipBlock(_ *block.Block) error {
...
 func (sdb *stateDB) States(opts ...protocol.StateOption) (uint64, state.Iterator, error) {
-	sdb.mutex.RLock()
-	defer sdb.mutex.RUnlock()
 	cfg, err := processOptions(opts...)
 	if err != nil {
 		return 0, nil, err
 	}
+	sdb.mutex.RLock()
+	h := sdb.currentChainHeight
+	sdb.mutex.RUnlock()
```
CoderZhi, line 470: same question for `sdb.dao.Filter`

state/factory/statedb.go
dustinxie: fixed

state/factory/statedb.go
dustinxie: fixed

state/factory/statedb.go
dustinxie: fixed

state/factory/statedb.go
dustinxie: fixed

CoderZhi (approved):
The change won't have significant impact on the performance.

PR #2975, opened 2022-01-05
[actpool] replace struct's value with its pointer in the heap

Liuhaai:


actpool/actqueue.go
```diff
@@ -27,24 +27,23 @@ type nonceWithTTL struct {
...
+	if in, ok := x.(*nonceWithTTL); ok {
+		*h = append(*h, in)
 	}
-	*h = append(*h, in)
 }
 
 func (h *noncePriorityQueue) Pop() interface{} {
 	old := *h
 	n := len(old)
 	x := old[n-1]
```
millken, line 45: need check length

Liuhaai: good find

actpool/actqueue.go
```diff
@@ -27,24 +27,26 @@ type nonceWithTTL struct {
...
 }
 
 func (h *noncePriorityQueue) Pop() interface{} {
 	old := *h
 	n := len(old)
+	if n == 0 {
+		return nil
+	}
 	x := old[n-1]
+	old[n-1] = nil // avoid memory leak
```
CoderZhi, line 49: how could memory leak happen? why the memory gc cannot detect it?

Liuhaai: >As mentioned earlier, re-slicing a slice doesn’t make a copy of the underlying array. The full array will be kept in memory until it is no longer referenced. Occasionally this can cause the program to hold all the data in memory when only a small piece of it is needed.

https://go.dev/blog/slices-intro

CoderZhi: I must be dumb, could you point me to the exact place or line where explain the details?

CoderZhi: Got it.
How will this change (struct to pointer) help code efficiency?

Liuhaai: It should. The pointer implementation is referred from official pq (https://github.com/golang/go/blob/master/src/container/heap/example_pq_test.go).

CoderZhi: not an explanation.

CoderZhi (approved):

dustinxie (approved):

PR #2984, opened 2022-01-11
[Archive Mode] 1. move serializablebytes and kvstorefortrie into protocol

CoderZhi:
prepare for follow up fix for archive mode
1. move serializabletypes and kvstoretire into protocol namespace
2. new kvstore with state reader


action/protocol/kvstorefortrie.go
```diff
@@ -0,0 +1,100 @@
...
+func (kv *kvStoreForTrieWithStateReader) Start(context.Context) error {
+	return nil
+}
+
+func (kv *kvStoreForTrieWithStateReader) Stop(context.Context) error {
+	return nil
+}
+
+func (kv *kvStoreForTrieWithStateReader) Put(key []byte, value []byte) error {
+	return errors.New("not implemented")
```
dustinxie, line 85: is it possible to (and should we) use panic?

action/protocol/kvstorefortrie.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (kv *kvStoreForTrieWithStateReader) Delete(key []byte) error {
+	return errors.New("not implemented")
```
dustinxie, line 89: same here

dustinxie (approved):

Liuhaai (approved):

action/protocol/kvstorefortrie.go
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2019 IoTeX Foundation
+// This is an alpha (internal) release and is not suitable for production. This source code is provided 'as is' and no
+// warranties are given as to title or non-infringement, merchantability or fitness for purpose and, to the extent
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package protocol
```
dustinxie, line 7: also, why need to move this file here?

action/protocol/kvstorefortrie.go
CoderZhi: error handling should be handled by logic component

action/protocol/kvstorefortrie.go
CoderZhi: you may check the follow up PRs

PR #2987, opened 2022-01-11
[Archive Mode] 4. replace iteration with of multiple states with keysoption

CoderZhi:
Fix the way of reading states. The reason is that the key storing in db may not be the key on creation and the change is irreversible.

If "keys" is not set, iterate all values. If set, read values by keys.

### DIFF
--- a/action/protocol/managers.go
+++ b/action/protocol/managers.go
@@ -2,9 +2,9 @@ package protocol
 
 import (
 	"github.com/iotexproject/go-pkgs/hash"
-	"github.com/iotexproject/iotex-core/db"
-	"github.com/iotexproject/iotex-core/state"
 	"github.com/pkg/errors"
+
+	"github.com/iotexproject/iotex-core/state"
 )
 
 // NamespaceOption creates an option for given namesapce
@@ -24,23 +24,19 @@ func KeyOption(key []byte) StateOption {
 	}
 }
 
+// KeysOption sets the key for call
+func KeysOption(f func() ([][]byte, error)) StateOption {
+	return func(cfg *StateConfig) (err error) {
+		cfg.Keys, err = f()
+		return err
+	}
+}
+
 // LegacyKeyOption sets the key for call with legacy key
 func LegacyKeyOption(key hash.Hash160) StateOption {
 	return func(cfg *StateConfig) error {
 		cfg.Key = make([]byte, len(key[:]))
 		copy(cfg.Key, key[:])
-		return nil
-	}
-}
-
-// FilterOption sets the filter
-func FilterOption(cond db.Condition, minKey, maxKey []byte) StateOption {
-	return func(cfg *StateConfig) error {
-		cfg.Cond = cond
-		cfg.MinKey = make([]byte, len(minKey))
-		copy(cfg.MinKey, minKey)
-		cfg.MaxKey = make([]byte, len(maxKey))
-		copy(cfg.MaxKey, maxKey)
 		return nil
 	}
 }
@@ -61,9 +57,7 @@ type (
 	StateConfig struct {
 		Namespace string // namespace used by state's storage
 		Key       []byte
-		MinKey    []byte
-		MaxKey    []byte
-		Cond      db.Condition
+		Keys      [][]byte
 	}
 
 	// StateOption sets parameter for access state
--- a/action/protocol/poll/util.go
+++ b/action/protocol/poll/util.go
@@ -7,7 +7,6 @@
 package poll
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"math"
@@ -319,12 +318,15 @@ func setCurrentBlockMeta(
 
 // allBlockMetasFromDB returns all latest block meta structs
 func allBlockMetasFromDB(sr protocol.StateReader, blocksInEpoch uint64) ([]*BlockMeta, error) {
+	keys := [][]byte{blockMetaKey(math.MaxUint64, blocksInEpoch)}
+	for i := uint64(0); i < blocksInEpoch; i++ {
+		keys = append(keys, blockMetaKey(i, blocksInEpoch))
+	}
 	stateHeight, iter, err := sr.States(
 		protocol.NamespaceOption(protocol.SystemNamespace),
-		protocol.FilterOption(func(k, v []byte) bool {
-			prefix := candidatesutil.ConstructKey(blockMetaPrefix)
-			return bytes.HasPrefix(k, prefix[:])
-		}, blockMetaKey(0, blocksInEpoch), blockMetaKey(math.MaxUint64, blocksInEpoch)),
+		protocol.KeysOption(func() ([][]byte, error) {
+			return keys, nil
+		}),
 	)
 	if err != nil {
 		return nil, err
@@ -337,10 +339,13 @@ func allBlockMetasFromDB(sr protocol.StateReader, blocksInEpoch uint64) ([]*Bloc
 	blockmetas := make([]*BlockMeta, 0, iter.Size())
 	for i := 0; i < iter.Size(); i++ {
 		bm := &BlockMeta{}
-		if err := iter.Next(bm); err != nil {
+		switch err := iter.Next(bm); errors.Cause(err) {
+		case nil:
+			blockmetas = append(blockmetas, bm)
+		case state.ErrNilValue:
+		default:
 			return nil, errors.Wrapf(err, "failed to deserialize block meta")
 		}
-		blockmetas = append(blockmetas, bm)
 	}
 	return blockmetas, nil
 }
--- a/state/factory/factory.go
+++ b/state/factory/factory.go
@@ -327,12 +327,12 @@ func (sf *factory) newWorkingSet(ctx context.Context, height uint64) (*workingSe
 			return err
 		},
 		statesFunc: func(ns string, keys [][]byte) ([][]byte, error) {
+			values := [][]byte{}
 			if keys == nil {
 				iter, err := mptrie.NewLayerTwoLeafIterator(tlt, namespaceKey(ns), legacyKeyLen())
 				if err != nil {
 					return nil, err
 				}
-				values := [][]byte{}
 				for {
 					_, value, err := iter.Next()
 					if err == trie.ErrEndOfIterator {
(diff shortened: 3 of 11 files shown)


dustinxie:
can we improve the fix to be more simple, efficient and not touching statedb (non-archive mode)? we currently have total ~47k buckets, this solution will create a huge [][]byte, and a for loop of that size to read DB, my concern is that it's both memory-heavy and time-consuming (at least for non-archive mode)

there are 2 possible directions:

1. similar to the current code in PR, add an API `GetKeys()` to underlying KVStore/Dao interface
```
if underlyingKVStore.(GetKeys) {
   // archive-mode use this method
   code in this PR
} else {
   existing code
}
```

and IMO this option might be better in terms of memory and running time
2. the root-cause of this problem is b/c the buckets and index are stored into the same namespace. We can route the index to a different namespace in archive-mode's `WriteInfoTranslate` function upon writing to DB
```
if namespace == StakingNameSpace && (key[0] == _voterIndex || key[0] == _voterIndex) {
    namespace = "stakingIndex"
}
```
then the `staking` namespace only stores buckets, we can use the TrieIterator you already implemented to get all buckets.

dustinxie:
and just found, already has existing code similar to option 2:
```
func (sf *factory) flusherOptions(ctx context.Context, height uint64) []db.KVStoreFlusherOption {
	g := genesis.MustExtractGenesisContext(ctx)
	preEaster := !g.IsEaster(height)
	opts := []db.KVStoreFlusherOption{
		db.SerializeFilterOption(func(wi *batch.WriteInfo) bool {
			if wi.Namespace() == ArchiveTrieNamespace {
				return true
			}
			if wi.Namespace() != evm.CodeKVNameSpace {
				return false
			}
			return preEaster
		}),
		db.SerializeOption(func(wi *batch.WriteInfo) []byte {
			if preEaster {
				return wi.SerializeWithoutWriteType()
			}
			return wi.Serialize()
		}),
	}
```

CoderZhi:
47k is not a big number. Adding Keys to KVStore is a better way which I had considered before submitting this PR, but it will make the PR too complicated. I don't want to postpone this PR after investing so much time to fix the issue. Up to now, this implementation works well for both statedb and factory. If you have more opinion, we can schedule a meeting.

action/protocol/managers.go
```diff
@@ -24,6 +24,14 @@ func KeyOption(key []byte) StateOption {
 	}
 }
 
+// KeysOption sets the key for call
+func KeysOption(f func() ([][]byte, error)) StateOption {
```
dustinxie, line 28: can use `func KeysOption(keys [][]byte) StateOption`?

when `KeysOption` is called, the caller will provide/prepare the keys, if there's an error, caller can handle it on their stack

state/factory/workingset.go
```diff
@@ -53,7 +53,7 @@ type (
 		writeviewFunc func(name string, v interface{}) error
 		dbFunc        func() db.KVStore
 		delStateFunc  func(string, []byte) error
-		statesFunc    func(opts ...protocol.StateOption) (uint64, state.Iterator, error)
```
dustinxie, line 56: think no need to change this

state/factory/workingset.go
```diff
@@ -244,7 +244,18 @@ func (ws *workingSet) State(s interface{}, opts ...protocol.StateOption) (uint64
 }
 
 func (ws *workingSet) States(opts ...protocol.StateOption) (uint64, state.Iterator, error) {
-	return ws.statesFunc(opts...)
```
dustinxie, line 247: think no need to change this, let `statesFunc` handle options inside

state/factory/statedb.go
```diff
@@ -251,8 +251,31 @@ func (sdb *stateDB) newWorkingSet(ctx context.Context, height uint64) (*workingS
 			flusher.KVStoreWithBuffer().MustDelete(ns, key)
 			return nil
 		},
-		statesFunc: func(opts ...protocol.StateOption) (uint64, state.Iterator, error) {
-			return sdb.States(opts...)
```
dustinxie, line 255: think no need to change, let `States` process the options

state/factory/statedb.go
```diff
@@ -469,17 +495,27 @@ func (sdb *stateDB) States(opts ...protocol.StateOption) (uint64, state.Iterator
 	if cfg.Key != nil {
 		return sdb.currentChainHeight, nil, errors.Wrap(ErrNotSupported, "Read states with key option has not been implemented yet")
 	}
-	if cfg.Cond == nil {
-		cfg.Cond = func(k, v []byte) bool {
-			return true
+	var values [][]byte
+	if cfg.Keys == nil {
```
dustinxie, line 499: add a comment like
// no Keys set, default to get all key-values in the namespace

state/factory/factory.go
```diff
@@ -322,8 +326,37 @@ func (sf *factory) newWorkingSet(ctx context.Context, height uint64) (*workingSe
...
+						return nil, err
+					}
+					values = append(values, value)
+				}
+			} else {
+				for _, key := range keys {
+					value, err := readState(tlt, ns, key)
+					switch errors.Cause(err) {
+					case db.ErrNotExist, db.ErrBucketNotExist:
+						values = append(values, nil)
```
dustinxie, line 351: why need to append nil here?

dustinxie:
> 

had an offline discussion, accept PR to unblock archive-mode, with 2 notes:
1. this will have some memory/perf hit for non-archive mode
2. add a db.BatchRead(keys) API instead of a for loop to read all keys (another PR)

action/protocol/managers.go
dustinxie, line 28: can use `func KeysOption(keys [][]byte) StateOption`?

when `KeysOption` is called, the caller will provide/prepare the keys, if there's an error, caller can handle it on their stack

state/factory/workingset.go
dustinxie, line 56: basically, this PR replaces the current `FilterOption` with `KeysOption`, think no need to change here

state/factory/workingset.go
dustinxie, line 247: think no need to change here, still pass the opts and let downstack funcs process

state/factory/statedb.go
```diff
@@ -251,8 +251,31 @@ func (sdb *stateDB) newWorkingSet(ctx context.Context, height uint64) (*workingS
 			flusher.KVStoreWithBuffer().MustDelete(ns, key)
 			return nil
 		},
-		statesFunc: func(opts ...protocol.StateOption) (uint64, state.Iterator, error) {
-			return sdb.States(opts...)
+		statesFunc: func(ns string, keys [][]byte) ([][]byte, error) {
+			if keys == nil {
```
dustinxie, line 255: add a comment like
// no Keys set, default to get all key-values in the namespace

state/factory/statedb.go
```diff
@@ -455,6 +478,9 @@ func (sdb *stateDB) State(s interface{}, opts ...protocol.StateOption) (uint64,
 	}
 	sdb.mutex.RLock()
 	defer sdb.mutex.RUnlock()
+	if cfg.Keys != nil {
+		return 0, errors.Wrap(ErrNotSupported, "Read state with keys option has not been implemented yet")
```
dustinxie, line 482: add a comment like
// no Keys set, default to get all key-values in the namespace

state/factory/factory.go
```diff
@@ -638,17 +677,27 @@ func (sf *factory) States(opts ...protocol.StateOption) (uint64, state.Iterator,
...
-	_, values, err := sf.dao.Filter(cfg.Namespace, cfg.Cond, cfg.MinKey, cfg.MaxKey)
-	if err != nil {
-		if errors.Cause(err) == db.ErrNotExist || errors.Cause(err) == db.ErrBucketNotExist {
-			return sf.currentChainHeight, nil, errors.Wrapf(state.ErrStateNotExist, "failed to get states of ns = %x", cfg.Namespace)
+	} else {
+		for _, key := range cfg.Keys {
+			value, err := sf.dao.Get(cfg.Namespace, key)
+			switch errors.Cause(err) {
+			case db.ErrNotExist, db.ErrBucketNotExist:
+				values = append(values, nil)
```
dustinxie, line 694: why need to append nil here?

action/protocol/poll/util.go
```diff
@@ -337,10 +339,13 @@ func allBlockMetasFromDB(sr protocol.StateReader, blocksInEpoch uint64) ([]*Bloc
 	blockmetas := make([]*BlockMeta, 0, iter.Size())
 	for i := 0; i < iter.Size(); i++ {
 		bm := &BlockMeta{}
-		if err := iter.Next(bm); err != nil {
+		switch err := iter.Next(bm); errors.Cause(err) {
```
Liuhaai, line 342: why is the style changed here?

state/factory/statedb.go
```diff
@@ -469,17 +495,27 @@ func (sdb *stateDB) States(opts ...protocol.StateOption) (uint64, state.Iterator
 	if cfg.Key != nil {
 		return sdb.currentChainHeight, nil, errors.Wrap(ErrNotSupported, "Read states with key option has not been implemented yet")
 	}
-	if cfg.Cond == nil {
-		cfg.Cond = func(k, v []byte) bool {
-			return true
+	var values [][]byte
```
Liuhaai, line 498: Is the func `stateDB.States()` still necessary? It seems that the code have been split into `statesFunc` and `workingSet.States()`

Liuhaai (approved):

action/protocol/managers.go
```diff
@@ -33,18 +41,6 @@ func LegacyKeyOption(key hash.Hash160) StateOption {
...
-	return func(cfg *StateConfig) error {
-		cfg.Cond = cond
-		cfg.MinKey = make([]byte, len(minKey))
-		copy(cfg.MinKey, minKey)
-		cfg.MaxKey = make([]byte, len(maxKey))
-		copy(cfg.MaxKey, maxKey)
-		return nil
-	}
-}
-
```
dustinxie, line 47: let's still keep this option, in case we need better performance, it does not affect the correctness of this fix/PR

CoderZhi: you can add it back in the PR which it is needed.

action/protocol/managers.go
dustinxie: fix this?

dustinxie (approved):

PR #2990, opened 2022-01-12
[actpool] fix performance issue in UpdateQueue()(WIP)

Liuhaai:
### Problem

Do sort.Sort() whenever a new tx is put in the actpool is time-consuming.

### Workaround

[The original code](https://github.com/iotexproject/iotex-core/blob/58b8dd7490745175883ef93fa03972672cd8760c/actpool/actqueue.go#L185-L211) removes expensive txs whose nonces are larger than pending nonce when updating queue.

The workaround is removing the checking process when updating queue. However, balance checking is done when getting pending action from the pool.

Those expensive txs will be cleaned out by timeout mechanism

### DIFF
--- a/actpool/actqueue_test.go
+++ b/actpool/actqueue_test.go
@@ -117,9 +117,9 @@ func TestActQueueUpdateNonce(t *testing.T) {
 	require.NoError(q.Put(tsf4))
 	q.pendingBalance = big.NewInt(1000)
 	require.NoError(q.Put(tsf5))
-	removed := q.UpdateQueue(uint64(2))
+	_ = q.UpdateQueue()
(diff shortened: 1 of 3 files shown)


actpool/actqueue_test.go
```diff
@@ -166,38 +167,6 @@ func TestActQueueAllActs(t *testing.T) {
 	require.Equal([]action.SealedEnvelope{tsf1, tsf3}, actions)
 }
 
-func TestActQueueRemoveActs(t *testing.T) {
```
dustinxie, line 169: can you modify (or add) a test, to verify the actqueue's behavior is same before and after the change in this PR?

PR #2991, opened 2022-01-12
[db] replace 1d map with 2d map in kv_cache

Liuhaai:
Calculating the hash of namespace and key is time-consuming.
It is replaced by using namespace and key as the string key of 2d map 

```
// before
BenchmarkCachedBatch_Snapshot-12    	  255420	      4493 ns/op	    2388 B/op	      37 allocs/op

// after
BenchmarkCachedBatch_Snapshot-12    	  667080	      1976 ns/op	    2420 B/op	      35 allocs/op
```


dustinxie:
an additional map access would takes extra time, is it possible to still use 1d map with key = some type of concatenation as we discussed?

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+		}
 	}
 	return &clone
 }
+
+func (c *kvCache) cachePut(key *KvCacheKey, v []byte) {
+	if _, ok := c.cache[key.key1]; !ok {
+		c.cache[key.key1] = make(map[string][]byte)
+	}
+	c.cache[key.key1][key.key2] = v
```
dustinxie, line 113:
```
m, ok := c.cache[key.key1]
if !ok {}
m[key.key2] = v
```

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+	for k1 := range c.cache {
+		for k2 := range c.cache[k1] {
+			delete(c.cache[k1], k2)
+		}
+		delete(c.cache, k1)
+	}
+}
+
+func (c *kvCache) deletedPut(key *KvCacheKey) {
+	if _, ok := c.deleted[key.key1]; !ok {
```
dustinxie, line 144: same as above

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
-		Evict(hash.Hash160)
+		Evict(*KvCacheKey)
 		// Clear clear the cache
 		Clear()
 		// Clone clones the cache
 		Clone() KVStoreCache
 	}
 
+	// KvCacheKey is the key for 2D Map cache
+	KvCacheKey struct {
```
dustinxie, line 27: can be `kvCacheKey`, only use internally

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
 }
 
 // Clear clear the cache
 func (c *kvCache) Clear() {
-	c.cache = nil
-	c.deleted = nil
-	c.cache = make(map[hash.Hash160][]byte)
-	c.deleted = make(map[hash.Hash160]struct{})
+	c.cacheClear()
+	c.deletedClear()
```
dustinxie, line 82: can combine these 2 into a single `clear()`

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+	c.cache = make(map[string]map[string][]byte)
+	c.deleted = make(map[string]map[string]struct{})
```
dustinxie, line 84: also the way you implement `clear()`, is to iterate and delete all keys, instead of simply set `c.cache = nil`
so these 2 lines are not needed, the 2 maps goes back to empty map after `clear()`

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+	}
+	return nil
+}
+
+func (c *kvCache) cacheDelete(key *KvCacheKey) {
+	if v1, ok := c.cache[key.key1]; ok {
+		delete(v1, key.key2)
+		if len(v1) == 0 {
+			delete(c.cache, key.key1)
+		}
```
dustinxie, line 130: actually don't need to delete? if `v1` becomes an empty map, `cacheGet` continues to work correctly, while at the same time it avoids make() of a new map inside `cachePut`

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+	}
+	return false
+}
+
+func (c *kvCache) deletedDelete(key *KvCacheKey) {
+	if v1, ok := c.deleted[key.key1]; ok {
+		delete(v1, key.key2)
+		if len(v1) == 0 {
+			delete(c.deleted, key.key1)
+		}
```
dustinxie, line 164: same here

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+	if v1, ok := c.cache[key.key1]; ok {
+		delete(v1, key.key2)
+		if len(v1) == 0 {
+			delete(c.cache, key.key1)
+		}
+	}
+}
+
+func (c *kvCache) cacheClear() {
+	for k1 := range c.cache {
```
CoderZhi, line 135: why we need to delete it one by one? memory leak? can you elaborate it?

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+}
+
+func (c *kvCache) deletedPut(key *KvCacheKey) {
+	if _, ok := c.deleted[key.key1]; !ok {
+		c.deleted[key.key1] = make(map[string]struct{})
+	}
+	c.deleted[key.key1][key.key2] = struct{}{}
+}
+
+func (c *kvCache) deletedGet(key *KvCacheKey) bool {
```
CoderZhi, line 150: bad naming

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+	if v1, ok := c.deleted[key.key1]; ok {
+		if _, ok := v1[key.key2]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *kvCache) deletedDelete(key *KvCacheKey) {
```
CoderZhi, line 159: bad naming

db/batch/batch_impl.go
```diff
@@ -310,6 +309,6 @@ func (cb *cachedBatch) AddFillPercent(ns string, percent float64) {
 	cb.kvStoreBatch.AddFillPercent(ns, percent)
 }
 
-func (cb *cachedBatch) hash(namespace string, key []byte) hash.Hash160 {
-	return hash.Hash160b(append([]byte(namespace), key...))
+func (cb *cachedBatch) hash(namespace string, key []byte) *KvCacheKey {
+	return &KvCacheKey{namespace, string(key)}
```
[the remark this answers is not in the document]
dustinxie: think it's fine to use `string(key)` here. The resulting string may not be printable, but here it is only used as a key to the map

db/batch/kv_cache.go
Liuhaai: It should. Otherwise, it becomes zoombie in the map

db/batch/kv_cache.go
Liuhaai: delete reference to byte slice

db/batch/kv_cache.go
```diff
@@ -6,96 +6,170 @@
...
+	if _, ok := c.cache[key.key1]; !ok {
+		c.cache[key.key1] = make(map[string][]byte)
+	}
+	c.cache[key.key1][key.key2] = v
+}
+
+func (c *kvCache) cacheGet(key *KvCacheKey) []byte {
+	if v1, ok := c.cache[key.key1]; ok {
+		if v2, ok := v1[key.key2]; ok {
+			return v2
```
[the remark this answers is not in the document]
Liuhaai: when will v2 becomes `nil`?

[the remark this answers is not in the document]
CoderZhi: https://github.com/iotexproject/iotex-core/pull/2991/files/3cb9b260cf6e77c099665c9314e34ce395aa554c#diff-eaf9b1d985c02522083f33d22902e5966879874d7e60d06644add629559177bfR113 v could be nil

db/batch/kv_cache.go
CoderZhi: memory leak? I don't get it.

Liuhaai: After testing, the mem of the map will be removed automatically by GC after Clear(). So cacheClear() is unnecessary.
Here is the code for test
```
func TestKvCache(t *testing.T) {
	var out []byte
	c := NewKVCache()
	for i := 0; i < 2000; i++ {
		key1 := RandStringBytes(3)
		key2 := RandStringBytes(5)
		val := RandStringBytes(100)
		c.Write(&kvCacheKey{key1, key2}, []byte(val))
		if i == 0 {
			out, _ = c.Read(&kvCacheKey{key1, key2})
		}
	}
	fmt.Println(out)
	PrintMemUsage()
	c.Clear()
	runtime.GC()
	PrintMemUsage()
}
```

Liuhaai:
Due to poor clone performance, the pr of 2d map should be closed.
```
BenchmarkKvCache/clone-12        	      28	  43210051 ns/op	44098092 B/op	  186605 allocs/op
```

db/batch/batch_impl_test.go
```diff
@@ -254,24 +254,20 @@ func BenchmarkCachedBatch_Digest(b *testing.B) {
...
 	for i := 0; i < b.N; i++ {
-		k := hash.Hash256b([]byte(strconv.Itoa(i)))
-		var v [1024]byte
-		for i := range v {
-			v[i] = byte(rand.Intn(8))
-		}
 		cb.Put(bucket1, k[:], v[:], "")
-		value, err := cb.Get(bucket1, k[:])
-		require.NoError(b, err)
-		require.True(b, bytes.Equal(value, v[:]))
```
Liuhaai, line 267: `require` is bottleneck in benchmark test

db/batch/batch_impl_test.go
```diff
@@ -254,24 +254,20 @@ func BenchmarkCachedBatch_Digest(b *testing.B) {
...
+	var v [1024]byte
+	for i := range v {
+		v[i] = byte(rand.Intn(8))
+	}
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		k := hash.Hash256b([]byte(strconv.Itoa(i)))
-		var v [1024]byte
-		for i := range v {
-			v[i] = byte(rand.Intn(8))
```
Liuhaai, line 262: `rand.intn` is bottleneck in benchmark test

go.mod
```diff
@@ -186,4 +186,4 @@ require (
 
 replace github.com/ethereum/go-ethereum => github.com/iotexproject/go-ethereum v0.4.0
 
-replace golang.org/x/xerrors => golang.org/x/xerrors v0.0.0-20190212162355-a5947ffaace3
+replace golang.org/x/xerrors => golang.org/x/xerrors v0.0.0-20190212162355-a5947ffaace3
```
dustinxie, line 189: go mod tidy

dustinxie (approved):

PR #3007, opened 2022-01-19
[ioctl/newcmd]  use AskToConfirm in AccountDelete

huof6829:
modify as ioctl/cmd/accountDelete

### DIFF
--- a/ioctl/client.go
+++ b/ioctl/client.go
@@ -102,16 +102,9 @@ func (c *client) Config() config.Config {
 }
 
 func (c *client) AskToConfirm() bool {
-	msg, lang := c.SelectTranslation(confirmMessages)
-	fmt.Println(msg)
 	var confirm string
 	fmt.Scanf("%s", &confirm)
-	switch lang {
-	case config.Chinese:
-		return strings.EqualFold(confirm, "是")
-	default: // config.English
-		return strings.EqualFold(confirm, "yes")
-	}
+	return strings.EqualFold(confirm, "yes")
 }
 
 func (c *client) SelectTranslation(trls map[config.Language]string) (string, config.Language) {
--- a/ioctl/client.go
+++ b/ioctl/client.go
@@ -40,7 +40,7 @@ type (
 		// SelectTranslation select a translation based on UILanguage
 		SelectTranslation(map[config.Language]string) (string, config.Language)
 		// AskToConfirm asks user to confirm from terminal, true to continue
-		AskToConfirm() bool
+		AskToConfirm(string) bool
 		// ReadSecret reads password from terminal
 		ReadSecret() (string, error)
 		// Execute a bash command
@@ -101,7 +101,9 @@ func (c *client) Config() config.Config {
 	return c.cfg
 }
 
-func (c *client) AskToConfirm() bool {
+func (c *client) AskToConfirm(info string) bool {
+	message := output.ConfirmationMessage{Info: info, Options: []string{"yes"}}
+	fmt.Println(message.String())
 	var confirm string
 	fmt.Scanf("%s", &confirm)
 	return strings.EqualFold(confirm, "yes")
--- a/ioctl/newcmd/account/accountdelete.go
+++ b/ioctl/newcmd/account/accountdelete.go
@@ -106,7 +106,7 @@ func NewAccountDelete(c ioctl.Client) *cobra.Command {
 					}
 
 					if err := os.Remove(v.URL.Path); err != nil {
-						return output.NewError(output.ReadFileError, failToRemoveKeystoreFile, err)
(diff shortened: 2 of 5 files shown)


ioctl/newcmd/account/accountdelete.go
```diff
@@ -102,17 +100,15 @@ func NewAccountDelete(c ioctl.Client) *cobra.Command {
...
-					fmt.Scanf("%s", &confirm)
-					if !strings.EqualFold(confirm, "yes") {
+					if !c.AskToConfirm() {
 						output.PrintResult("quit")
 						return nil
 					}
 
 					if err := os.Remove(v.URL.Path); err != nil {
-						return output.NewError(output.WriteFileError, failToRemoveKeystoreFile, err)
+						return output.NewError(output.ReadFileError, failToRemoveKeystoreFile, err)
```
dustinxie, line 111: if cannot remove file, it should be `WriteFileError`

ioctl/newcmd/account/accountdelete.go
```diff
@@ -102,17 +100,15 @@ func NewAccountDelete(c ioctl.Client) *cobra.Command {
 			ks := c.NewKeyStore(config.ReadConfig.Wallet, keystore.StandardScryptN, keystore.StandardScryptP)
 			for _, v := range ks.Accounts() {
 				if bytes.Equal(account.Bytes(), v.Address.Bytes()) {
-					var confirm string
 					message := output.ConfirmationMessage{Info: infoWarn, Options: []string{"yes"}}
 					fmt.Println(message.String())
-					fmt.Scanf("%s", &confirm)
-					if !strings.EqualFold(confirm, "yes") {
+					if !c.AskToConfirm() {
```
dustinxie, line 105: if the command line has `-y`, that means user confirms ahead. In that case, don't need to call `AskToConfirm`, pls take that into consideration, things like:
1. can this check be done inside `AskToConfirm`? if yes, that's the best
2. if not, what are the solution? like add a `needConfirm` flag to command, and call it for those commands with the flag? or other alternatives?

Liuhaai (approved):

ioctl/newcmd/account/accountdelete.go
huof6829: ok

ioctl/newcmd/account/accountdelete.go
huof6829: i can add a new flag to check wether user write `-y`, and check this flag before `AskToConfirm`.  if has, delete account directly;  if not, ask user to confirm.  
in current version,  `-y` is not supported, i think this may be safer for deleting account cannot be recovered and avoid user write wrong account by accident. 
so can we need add `-y` ?

ioctl/newcmd/account/accountdelete_test.go
```diff
@@ -36,10 +36,16 @@ func TestNewAccountDelete(t *testing.T) {
...
+	client.EXPECT().Config().Return(config.ReadConfig).Times(2)
+	client.EXPECT().AskToConfirm().Return(false)
 	cmd := NewAccountDelete(client)
 	_, err := util.ExecuteCmd(cmd)
 	require.NoError(t, err)
+
+	client.EXPECT().AskToConfirm().Return(true)
+	cmd = NewAccountDelete(client)
+	_, err = util.ExecuteCmd(cmd)
+	require.NoError(t, err)
```
CoderZhi, line 50: what's the difference of execution result for the two cases?

ioctl/client.go
```diff
@@ -102,16 +102,9 @@ func (c *client) Config() config.Config {
 }
 
 func (c *client) AskToConfirm() bool {
-	msg, lang := c.SelectTranslation(confirmMessages)
-	fmt.Println(msg)
```
CoderZhi, line 106: if there is always a message before user input, we can keep the `msg` as an input parameter of this function.

huof6829: yes,  i  search  ioctl/cmd.  i can keep msg as the parameter

ioctl/newcmd/account/accountdelete_test.go
huof6829: if user input yes, print the result.  if not yes,  print quit and return at once.  i can see the print in cmd line.  after i move output to ioctl.Client,  we can check by require.Equal,

ioctl/newcmd/account/accountdelete.go
huof6829: not add, keep the same with ioctl/cmd

ioctl/newcmd/account/accountdelete.go
CoderZhi: address this comment before merging

CoderZhi (approved):

PR #3009, opened 2022-01-19
[ioutil/newcmd] add test for account

huof6829:
add unit test


ioctl/newcmd/account/account_test.go
```diff
@@ -0,0 +1,314 @@
...
+const (
+	testPath = "ksTest"
+)
+
+func TestNewAccountCmd(t *testing.T) {
+	require := require.New(t)
+	ctrl := gomock.NewController(t)
+	client := mock_ioctlclient.NewMockClient(ctrl)
+	client.EXPECT().SelectTranslation(gomock.Any()).Return("mockTranslationString", config.English).AnyTimes()
+	client.EXPECT().Config().Return(config.ReadConfig).AnyTimes()
```
CoderZhi, line 45: specify the number of times it will be called

ioctl/newcmd/account/account_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+const (
+	testPath = "ksTest"
+)
+
+func TestNewAccountCmd(t *testing.T) {
+	require := require.New(t)
+	ctrl := gomock.NewController(t)
+	client := mock_ioctlclient.NewMockClient(ctrl)
+	client.EXPECT().SelectTranslation(gomock.Any()).Return("mockTranslationString", config.English).AnyTimes()
```
CoderZhi, line 44: specify the number of times it will be called

ioctl/newcmd/account/account_test.go
```diff
@@ -0,0 +1,314 @@
...
+	require.NoError(os.MkdirAll(testWallet, os.ModePerm))
+	defer testutil.CleanupPath(t, testWallet)
+	config.ReadConfig.Wallet = testWallet
+
+	ks := keystore.NewKeyStore(config.ReadConfig.Wallet, keystore.StandardScryptN, keystore.StandardScryptP)
+	require.NotNil(ks)
+
+	// create accounts
+	nonce := strconv.FormatInt(rand.Int63(), 10)
+	passwd := "3dj,<>@@SF{}rj0ZF#" + nonce
```
CoderZhi, line 122: move into a test util function?

huof6829: yes, these codes generate a true account , and is used by other test functions.

ioctl/newcmd/account/account_test.go
huof6829: yes

ioctl/newcmd/account/account_test.go
huof6829: yes

ioctl/newcmd/account/account_test.go
huof6829: i suggest not modify.  because we do AddCommand( Create/Delete/List ... ) in NewAccountCmd,   the sub command calls  SelectTranslation() and Config().  if add one sub command, the numbers of time is added corresponding, otherwise it occurs error  in test.  this is trouble and forgot.  so use AnyTimes() for test ok when adding sub commands.

ioctl/newcmd/account/account_test.go
huof6829: the same above

CoderZhi (approved):

Liuhaai (approved):

PR #3010, opened 2022-01-19
[ioctl/newcmd] add hdwallet for new account calling

huof6829:
add function DeriveKey  for new account calling

### DIFF
--- a/ioctl/newcmd/account/account.go
+++ b/ioctl/newcmd/account/account.go
@@ -30,8 +30,8 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/iotexproject/iotex-core/ioctl"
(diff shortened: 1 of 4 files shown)


millken (approved):

CoderZhi (commented):
ioctl/newcmd/hdwallet/hdwalletderive.go could be moved out of 'newcmd'. what if we put it under `github.com/iotexproject/iotex-core/ioctl/hdwallet/` directory?

huof6829:
it  corresponds to the command:  ioctl hdwallet derive 
i think in newcmd may be suitable  considering the meaning

CoderZhi:
> it corresponds to the command: ioctl hdwallet derive i think in newcmd may be suitable considering the meaning

got it. the public function itself seems to be a util function.

CoderZhi (approved):

PR #3014, opened 2022-01-21
[evm] remove corresponding logs when tx is reverted

dustinxie:


action/protocol/execution/evm/evmstatedbadapter_test.go
```diff
@@ -463,6 +488,14 @@ func TestSnapshotRevertAndCommit(t *testing.T) {
 				require.Equal(e.suicide, stateDB.HasSuicided(e.addr))
 				require.Equal(e.exist, stateDB.Exist(e.addr))
 			}
+			// test logs
+			if revertLog {
+				require.Equal(test.logSize, len(stateDB.logs))
+				require.Equal(test.logAddr, stateDB.logs[test.logSize-1].Address)
+			} else {
+				require.Equal(6, len(stateDB.logs))
+				require.Equal("io1x3cv7c4w922k6wx5s8p6d8sjrcqlcfrxhkn5xe", stateDB.logs[5].Address)
```
dustinxie, line 497: if revertLog is not enabled, the number of logs stays at 6

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -892,10 +919,16 @@ func (stateDB *StateDBAdapter) clear() {
 	stateDB.suicideSnapshot = nil
 	stateDB.preimages = nil
 	stateDB.preimageSnapshot = nil
+	stateDB.logsSnapshot = nil
```
Liuhaai, line 922: Once the memory is unreachable, it will be freed.
No need to set it to `nil`

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -461,6 +472,18 @@ func (stateDB *StateDBAdapter) RevertToSnapshot(snapshot int) {
 		log.L().Error("Failed to get snapshot.", zap.Int("snapshot", snapshot))
 		return
 	}
+	// restore logs
+	if stateDB.revertLog {
+		stateDB.logs = stateDB.logs[:stateDB.logsSnapshot[snapshot]]
+		delete(stateDB.logsSnapshot, snapshot)
```
Liuhaai, line 478: why do we need `map` here? Is the distribution of sn sparse?
I think `slice` is enough

action/protocol/execution/evm/evm.go
```diff
@@ -243,12 +243,12 @@ func ExecuteContract(
 	if err := stateDB.CommitContracts(); err != nil {
 		return nil, nil, errors.Wrap(err, "failed to commit contracts to underlying db")
 	}
-	stateDB.clear()
 	receipt.AddLogs(stateDB.Logs()...).AddTransactionLogs(depositLog, burnLog)
 	if receipt.Status == uint64(iotextypes.ReceiptStatus_Success) ||
 		featureCtx.AddOutOfGasToTransactionLog && receipt.Status == uint64(iotextypes.ReceiptStatus_ErrCodeStoreOutOfGas) {
 		receipt.AddTransactionLogs(stateDB.TransactionLogs()...)
 	}
+	stateDB.clear()
```
CoderZhi, line 251: The position change of this line may be a hard fork.

action/protocol/execution/evm/evmstatedbadapter.go
dustinxie: yes, reason for this:
1. this is for code clarity (clearly tells reader the variable is not needed anymore), follow the existing code
2. in a sense help the runtime/GC to know immediately this memory is no longer needed and can be GC'ed, similar to https://github.com/iotexproject/iotex-core/commit/90badd6a83f9ff1b4e8b9addf58a83503f67e36e#diff-aa7bcf8c732d6ec1b1b2d3a3324071563c4c913f007f3298bf8f7e1f7c78aaa5R49

action/protocol/execution/evm/evm.go
dustinxie: from my understanding it's not a hard-fork, `clear()` does not clear Logs and TransactionLogs, which are used in line 246~249, now moving `clear()` to after line 249 and as a result, these 2 can be cleared as well

it is not a necessary change, just to make `clear()` a little better, that's why I put it in a 2nd commit.

action/protocol/execution/evm/evmstatedbadapter.go
dustinxie: updated in 3rd commit

action/protocol/execution/evm/evmstatedbadapter.go
Liuhaai: Unlike C++, Go is automatic memory management. The GC won't invoked immediately when the reference is removed.

dustinxie: right ... again the reasons explained above

action/protocol/execution/evm/evm.go
CoderZhi: then, it should be fine.

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -62,11 +62,13 @@ type (
 		suicideSnapshot     map[int]deleteAccount // snapshots of suicide accounts
 		preimages           preimageMap
 		preimageSnapshot    map[int]preimageMap
+		logsSnapshot        []int // logs is an array, save len(logs) at time of snapshot suffices
```
CoderZhi, line 65: shall we revert transaction logs as well? if so, it is better to have it fixed in this PR as well.

dustinxie: right, good catch

Liuhaai (approved):

CoderZhi (approved):

PR #3015, opened 2022-01-21
[ioctl/newcmd] add output in ioclt.Client

huof6829:
move print to ioctl.Client 
and make a test


ioctl/client.go
```diff
@@ -204,3 +210,15 @@ func (c *client) WriteConfig(cfg config.Config) error {
 	}
 	return nil
 }
+
+func (c *client) PrintError(err error) error {
+	return output.PrintError(err)
```
CoderZhi, line 215: According to https://github.com/huof6829/iotex-core/blob/2e51464f4170129c7f1c7d478d63584798cc68b1/ioctl/output/format.go#L213, if Format is not empty, no error will be returned.
Thus,
```
    if err := output.PrintError(err); err != nil {
        fmt.Println(err)
    }
```
no return is needed for this function.

ioctl/client.go
```diff
@@ -204,3 +210,15 @@ func (c *client) WriteConfig(cfg config.Config) error {
...
+}
+
+func (c *client) PrintResult(result string) {
+	output.PrintResult(result)
+}
+
+func (c *client) PrintQuery(query string) {
+	output.PrintQuery(query)
+}
```
CoderZhi, line 224: Unify to Print

ioctl/client.go
huof6829: yes, it always return nil

ioctl/client.go
huof6829: yes

ioctl/client.go
huof6829: in ioctl/cmd  the errors is printed by cobra   https://github.com/spf13/cobra/blob/v1.3.0/command.goL986  after execute RunE.   in newcmd  i add our own PrintError in ioctl.Client.   i think  after change to newcmd, we can move PintError to https://github.com/huof6829/iotex-core/blob/2e51464f4170129c7f1c7d478d63584798cc68b1/ioctl/output/format.go#L213 ,  this may be clearly

CoderZhi (approved):

Liuhaai:
update output in other files under `/newcmd`

ioctl/newcmd/account/accountcreate.go
```diff
@@ -57,15 +57,18 @@ func NewAccountCreate(c ioctl.Client) *cobra.Command {
 		Args:  cobra.ExactArgs(0),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cmd.SilenceUsage = true
+			cmd.SilenceErrors = true
 			newAccounts := make([]generatedAccount, 0)
 			for i := 0; i < int(numAccounts); i++ {
 				private, err := crypto.GenerateKey()
 				if err != nil {
-					return output.NewError(output.CryptoError, failToGenerateNewPrivateKey, err)
+					c.PrintError(output.NewError(output.CryptoError, failToGenerateNewPrivateKey, err))
```
dustinxie, line 65: can we directly use `output.PrintError(output.NewError(...))`?

ioctl/newcmd/account/accountcreate.go
```diff
@@ -76,14 +79,11 @@ func NewAccountCreate(c ioctl.Client) *cobra.Command {
 			}
 
 			message := createMessage{Accounts: newAccounts}
-
-			fmt.Println(message.String())
+			c.PrintInfo(message.String())
```
dustinxie, line 82: can we directly use `output.PrintResult(message.String())`?

ioctl/client.go
```diff
@@ -199,3 +203,13 @@ func (c *client) WriteConfig(cfg config.Config) error {
...
 }
+
+func (c *client) PrintError(err error) {
+	if err := output.PrintError(err); err != nil {
+		fmt.Println("Error:", err.Error())
+	}
+}
+
+func (c *client) PrintInfo(info string) {
+	output.PrintResult(info)
```
dustinxie, line 214: looks like these 2 funcs are simply calling `output.Printxxx`, then do we need to add these 2 new methods into the interface?

ioctl/newcmd/account/accountcreate.go
huof6829: use  client.PrintError , we can test the result wether is the wanting setting by gomock .  if use output.NewError,  we cannot set the wanting by gomock, then we cannot know whether it's ok.

ioctl/newcmd/account/accountcreate.go
huof6829: the same above

ioctl/client.go
huof6829: yes,  the aim is test the result of cmd by gomock.

huof6829:
> What's benefit if we print the err by ourself not by cobra

we can control the format of output.

Liuhaai:
Sorry, operate mistakenly

Liuhaai (approved):

PR #3020, opened 2022-01-23
[api] simplify BlockMetas() in coreservice

Liuhaai:


CoderZhi:
rename the title

CoderZhi (approved):

api/coreservice.go
```diff
@@ -1005,9 +1005,6 @@ func (core *coreService) ActionsByBlock(blkHash string, start uint64, count uint
 
 // BlockMetas returns blockmetas response within the height range
 func (core *coreService) BlockMetas(start uint64, count uint64) ([]*iotextypes.BlockMeta, error) {
-	if count == 0 {
-		return nil, status.Error(codes.InvalidArgument, "count must be greater than zero")
-	}
```
dustinxie, line 1010: this would cause behavior change of gRPC/web3 server, i.e., request with `count = 0` now returns empty result instead of an error,  would that be a concern?

Liuhaai: someone will never request with `count = 0`

dustinxie: `func (svr *GRPCServer) GetBlockMetas(ctx context.Context, in *iotexapi.GetBlockMetasRequest)` calls into here, user could possibly pass in `count = 0` in the RPC request

this is input validity/sanity test, either do it outside, or do it inside here, not sure why we want to specifically remove the case `count = 0`

Liuhaai: After discussion, removing `count==0` might cause panic if frontend doesn't access it properly

PR #3021, opened 2022-01-23
[blksync] simplify code in blocksync/buffer

Liuhaai:
-  dequeAll() can return a empty slice instead of `nil`
 - Rename `Delete` to `Pop`


blocksync/buffer.go
```diff
@@ -35,7 +35,7 @@ func newBlockBuffer(bufferSize, intervalSize uint64) *blockBuffer {
 	}
 }
 
-func (b *blockBuffer) Delete(height uint64) []*peerBlock {
+func (b *blockBuffer) Pop(height uint64) []*peerBlock {
```
CoderZhi, line 38: why do you rename this function?

CoderZhi (commented):
the title is too general. This PR doesn't "refactor" anything, but refine some of the corner cases.

blocksync/buffer.go
Liuhaai: `Delete` is to remove buffer of height x, whereas `Pop` is to remove buffer of height x and return the buffer to the caller.

blocksync/uniqueue.go
```diff
@@ -33,9 +33,6 @@ func (uq *uniQueue) enque(blk *peerBlock) {
 }
 
 func (uq *uniQueue) dequeAll() []*peerBlock {
-	if len(uq.blocks) == 0 {
-		return nil
-	}
```
dustinxie, line 38: i think it's OK to leave this?

CoderZhi (approved):

blocksync/uniqueue.go
Liuhaai: It is better to return an empty slice than nil when error type is unavilable in return values

dustinxie: ok, then use `return []*peerBlock{}`? 
line 40, 41 is not necessary for the case `if len(uq.blocks) == 0`

dustinxie (approved):

PR #3026, opened 2022-01-26
[config] enable 4 features at midway height

dustinxie:
1. allow correct and default ChainID
2. fix GetHashFunc in EVM
3. correct tx/log index for transaction receipt and EVM log
4. revert logs upon tx reversion in EVM


ioctl/newcmd/account/accountlist_test.go
```diff
@@ -35,7 +35,7 @@ func TestNewAccountList(t *testing.T) {
 		ks := keystore.NewKeyStore(testAccountFolder, keystore.StandardScryptN, keystore.StandardScryptP)
 		_, _ = ks.NewAccount("test")
 		_, _ = ks.NewAccount("test2")
-		
+
```
dustinxie, line 38: `make fmt` catch this

CoderZhi (approved):

PR #3027, opened 2022-01-26
change log level from error to debug

CoderZhi:
The txBytes could be very long, thus it is not suitable to print at the "Error" level.

### DIFF
--- a/api/coreservice.go
+++ b/api/coreservice.go
@@ -281,7 +281,7 @@ func (core *coreService) SendAction(ctx context.Context, in *iotextypes.Action) 
 		if serErr != nil {
 			l.Error("Data corruption", zap.Error(serErr))
 		} else {
-			l.With(zap.String("txBytes", hex.EncodeToString(txBytes))).Error("Failed to accept action", zap.Error(err))
+			l.With(zap.String("txBytes", hex.EncodeToString(txBytes))).Debug("Failed to accept action", zap.Error(err))
 		}
 		errMsg := core.cfg.ProducerAddress().String() + ": " + err.Error()
 		st := status.New(codes.Internal, errMsg)


api/coreservice.go
```diff
@@ -281,7 +281,7 @@ func (core *coreService) SendAction(ctx context.Context, in *iotextypes.Action)
 		if serErr != nil {
 			l.Error("Data corruption", zap.Error(serErr))
 		} else {
-			l.With(zap.String("txBytes", hex.EncodeToString(txBytes))).Error("Failed to accept action", zap.Error(err))
+			l.With(zap.String("txBytes", hex.EncodeToString(txBytes))).Debug("Failed to accept action", zap.Error(err))
```
dustinxie, line 284: i think it's worth to keep/print it? This is how we debug the web3 transfer hash not correct issue back when we were enabling babel. Also I did a search on Kibana, `Failed to accept action` did not occur for the past 90 days, so storage should not be a concern here IMO. In case such error happen, we want to be able to search on Kibana

dustinxie: BTW just notice the title says change to `warn` level, that is fine (Kibana can search warn level)

CoderZhi: what's the value of printing out bytecode? especially the case that the bytecode is very long.

dustinxie: back when we implement babel service, this is how we root-caused the issue that web3 transfer does not work (by checking the bytecode it is found that `value = nil` but should be `0x0`)
Also I did a search on Kibana, `"Failed to accept action"` does not occur for the past 90 days, so log file size/storage is not be a concern here. So IMO we should keep this for debug in case such error happens

CoderZhi: try "go test -v" by yourself

Liuhaai:
txdata is good to be muted. But we want to know whether there are errs from actpool or not.
```
l.Error("Failed to accept action", zap.Error(err))
l.With(zap.String("txBytes", hex.EncodeToString(txBytes))).Debug("Failed action's bytes")
```

api/coreservice.go
dustinxie: i see, it seems just this one special test `go test -run TestChainlinkErrTest`
are you OK we fix by below? i tested it can pass
Screen Shot 2022-01-26 at 12 46 00 PM

CoderZhi: even for.API node, printing the bytecode doesn't have any value

dustinxie: alright, I'm curious to know more on why it doesn't have any value? didn't it help us root-cause and fix the bug for babel case?

CoderZhi: if we suspect anything to do with bytecode, change logging level to "Debug" in command.

PR #3034, opened 2022-01-26
[statedb] clear snapshots after running each transaction

dustinxie:
when workingset is running all tx in the block, all snapshots made during one tx are meaningless for the next tx, so after running one tx, clear all snapshots to reduce memory consumption/growth of underlying `CachedBatch`

### DIFF
--- a/db/batch/batch_impl.go
+++ b/db/batch/batch_impl.go
@@ -57,7 +57,7 @@ func (b *baseKVStoreBatch) Unlock() {
 // ClearAndUnlock clears the write queue and unlocks the batch
 func (b *baseKVStoreBatch) ClearAndUnlock() {
 	defer b.mutex.Unlock()
-	b.writeQueue = nil
+	b.writeQueue = b.writeQueue[:0]
 
 	b.fillLock.Lock()
 	defer b.fillLock.Unlock()
@@ -116,7 +116,7 @@ func (b *baseKVStoreBatch) SerializeQueue(serialize WriteInfoSerialize, filter W
 func (b *baseKVStoreBatch) Clear() {
 	b.mutex.Lock()
 	defer b.mutex.Unlock()
-	b.writeQueue = nil
+	b.writeQueue = b.writeQueue[:0]
 
 	b.fillLock.Lock()
 	defer b.fillLock.Unlock()
@@ -228,10 +228,8 @@ func (cb *cachedBatch) ClearAndUnlock() {
 	cb.kvStoreBatch.Clear()
 	// clear all saved snapshots
 	cb.tag = 0
-	cb.batchShots = nil
-	cb.cacheShots = nil
-	cb.batchShots = make([]int, 0)
-	cb.cacheShots = make([]KVStoreCache, 0)
+	cb.batchShots = cb.batchShots[:0]
+	cb.cacheShots = cb.cacheShots[:0]
 }
 
 // Put inserts a <key, value> record
@@ -260,10 +258,8 @@ func (cb *cachedBatch) Clear() {
 	cb.kvStoreBatch.Clear()
 	// clear all saved snapshots
 	cb.tag = 0
-	cb.batchShots = nil
-	cb.cacheShots = nil
-	cb.batchShots = make([]int, 0)
-	cb.cacheShots = make([]KVStoreCache, 0)
+	cb.batchShots = cb.batchShots[:0]
+	cb.cacheShots = cb.cacheShots[:0]
 }
 
 // Get retrieves a record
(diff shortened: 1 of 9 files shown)


db/batch/batch_impl.go
```diff
@@ -228,10 +228,8 @@ func (cb *cachedBatch) ClearAndUnlock() {
 	cb.kvStoreBatch.Clear()
 	// clear all saved snapshots
 	cb.tag = 0
-	cb.batchShots = nil
-	cb.cacheShots = nil
-	cb.batchShots = make([]int, 0)
-	cb.cacheShots = make([]KVStoreCache, 0)
+	cb.batchShots = cb.batchShots[:0]
+	cb.cacheShots = cb.cacheShots[:0]
```
dustinxie, line 232: this will reuse the underlying memory, better than set to nil (discard current) and `make` a new slice

db/kvstorewithbuffer.go
```diff
@@ -115,9 +115,7 @@ func (f *flusher) Flush() error {
 		return err
 	}
 
-	f.kvb.buffer.Lock()
-	f.kvb.buffer.ClearAndUnlock()
-
+	f.kvb.buffer.Clear()
```
dustinxie, line 118: `Clear()` is equivalent to 2 calls on left side

state/factory/workingset.go
```diff
@@ -170,6 +171,7 @@ func (ws *workingSet) runAction(
 	}
 	for _, actionHandler := range reg.All() {
 		receipt, err := actionHandler.Handle(ctx, elp.Action(), ws)
+		ws.clearActFunc()
```
dustinxie, line 174: clear all snapshots made by current tx

Liuhaai (approved):

CoderZhi (commented):
why we should make this change if the current logic works fine and makes sense?

dustinxie:
> why we should make this change if the current logic works fine and makes sense?

1. this logic is more correct (snapshots from previous tx is useless for next tx) 
2. reduce memory usage, increase perf

dustinxie:
this change reduced memory usage while running actions, also does not hurt running time
before
```
BenchmarkMintAndCommitBlock-8   	       1	1050992664 ns/op
BenchmarkMintBlock-8            	                      2	 509487252 ns/op
BenchmarkValidateBlock-8        	                      1	1166334819 ns/op
```
after
```
BenchmarkMintAndCommitBlock-8   	       1	1031465595 ns/op
BenchmarkMintBlock-8            	                      2	 518841254 ns/op
BenchmarkValidateBlock-8        	                      1	1154082540 ns/op
```

dustinxie:
> why we should make this change if the current logic works fine and makes sense?

IMO this change removes unnecessary snapshots and reduces run-time memory, the logic still makes sense with this change

state/factory/workingset.go
CoderZhi:
```
 	elpHash, err := elp.Hash()
	if err != nil {
		return nil, errors.Wrapf(err, "Failed to get hash")
	}

	var receipt *action.Receipt
	var err error
	for _, actionHandler := range reg.All() {
		receipt, err = actionHandler.Handle(ctx, elp.Action(), ws)
		if err != nil {
			err = errors.Wrapf(
				err,
				"error when action %x mutates states",
				elpHash,
			)
			break
		}
		if receipt != nil {
			break
		}
	}
	ws.clearActFunc()
	// TODO (zhi): return error if both receipt and error are nil
	return receipt, err
```

Liuhaai: good improvement. this will save lock time

Liuhaai (approved):

state/factory/factory.go
```diff
@@ -281,7 +281,11 @@ func (sf *factory) newWorkingSet(ctx context.Context, height uint64) (*workingSe
 	span.AddEvent("factory.newWorkingSet")
 	defer span.End()
 
-	flusher, err := db.NewKVStoreFlusher(sf.dao, batch.NewCachedBatch(), sf.flusherOptions(ctx, height)...)
+	var opts []batch.CachedBatchOption
+	if genesis.MustExtractGenesisContext(ctx).IsMidway(height) {
+		opts = append(opts, batch.SafeRevertOption())
+	}
+	flusher, err := db.NewKVStoreFlusher(sf.dao, batch.NewCachedBatch(opts...), sf.flusherOptions(ctx, height)...)
```
dustinxie, line 288: after 1.7.0, we can verify sync and revert this

state/factory/statedb.go
```diff
@@ -213,7 +213,11 @@ func (sdb *stateDB) Height() (uint64, error) {
 }
 
 func (sdb *stateDB) newWorkingSet(ctx context.Context, height uint64) (*workingSet, error) {
-	flusher, err := db.NewKVStoreFlusher(sdb.dao, batch.NewCachedBatch(), sdb.flusherOptions(ctx, height)...)
+	var opts []batch.CachedBatchOption
+	if genesis.MustExtractGenesisContext(ctx).IsMidway(height) {
+		opts = append(opts, batch.SafeRevertOption())
+	}
+	flusher, err := db.NewKVStoreFlusher(sdb.dao, batch.NewCachedBatch(opts...), sdb.flusherOptions(ctx, height)...)
```
dustinxie, line 220: after 1.7.0, we can verify sync and revert this

Liuhaai: add TODO

Issue #2694, opened 2021-06-07
Move config definition into corresponding components


CoderZhi:
Right now, most of the configs of different components are defined in config/config.go, and config is used in different components. As a result, the dependency is very complex that loop dependency is a common problem.

**What would you like to be added**:
Move config definitions into corresponding components.

**Why is this needed**:
Independent component

CoderZhi:
Example 1: https://github.com/iotexproject/iotex-core/pull/2673
Example 2: https://github.com/iotexproject/iotex-core/pull/2695

CoderZhi:
https://iotex.larksuite.com/docs/docusUF0iZ9n3JF4CamxsQAfCLe#

dustinxie:
create PR for moving each struct, like Consensus struct {}, BlockSync struct {}

Liuhaai:
All unmoved config definitions in config.go directly or indirectly import action package. However, if the config structs in config.go are moved to their responding package, there will be a loop dependency. Because action package needs to read ChainID in config package.

Liuhaai:
Dependency would be decoupled after the pr https://github.com/iotexproject/iotex-core/pull/3035 is merged

millken:
has been split into #3739 #3736 #3735 #3721

PR #3035, opened 2022-01-27
[config] move EVMNetworkID to /pkg/chain

dustinxie:
this is to break the loop dependency so we can move forward on issue https://github.com/iotexproject/iotex-core/issues/2694

### DIFF
--- a/blockchain/genesis/genesis.go
+++ b/blockchain/genesis/genesis.go
@@ -212,7 +212,7 @@ type (
 		// 1. use correct chainID
 		// 2. check legacy address
 		// 3. enable web3 staking transaction
-		NewfoundlandBlockHeight uint64 `yaml:"newfoundlandBlockHeight"`
+		NewfoundlandBlockHeight uint64 `yaml:"newfoundlandHeight"`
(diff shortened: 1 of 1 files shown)


action/sealedenvelope.go
```diff
@@ -8,7 +14,7 @@ import (
 	"github.com/pkg/errors"
 	"google.golang.org/protobuf/proto"
 
-	"github.com/iotexproject/iotex-core/config"
```
dustinxie, line 11: this is the dependency causing a loop for issue #2694

Liuhaai (approved):

CoderZhi (changes requested):
evm network id is supposed to be in Context. Previously, we put it in config, which is temporary solution. This PR doesn't make it right.

dustinxie:
> evm network id is supposed to be in Context. Previously, we put it in config, which is temporary solution. This PR doesn't make it right.

If I understand correctly, you mean evm network id should be passed in Context in our programming. While the purpose of this PR is to solve the dependency loop so we can move forward on issue https://github.com/iotexproject/iotex-core/issues/2694, I think it's 2 different things

PR #3039, opened 2022-01-29
[action] fix contract addr in tsf

Liuhaai:
The contractAddr should be empty if the recipient is not contract


action/protocol/account/transfer.go
```diff
@@ -127,11 +127,10 @@ func (p *Protocol) handleTransfer(ctx context.Context, act action.Action, sm pro
 	}
 
 	receipt := &action.Receipt{
-		Status:          uint64(iotextypes.ReceiptStatus_Success),
-		BlockHeight:     blkCtx.BlockHeight,
-		ActionHash:      actionCtx.ActionHash,
-		GasConsumed:     actionCtx.IntrinsicGas,
-		ContractAddress: p.addr.String(),
```
millken, line 134: Why we removed it?

dustinxie: need this, otherwise the receipt root will change

action/protocol/rewarding/protocol.go
```diff
@@ -68,6 +68,16 @@ func NewProtocol(cfg genesis.Rewarding) *Protocol {
 	}
 }
 
+// ProtocolAddr returns the address generated from protocol id
+func ProtocolAddr() address.Address {
```
dustinxie, line 72: create a common func `hashStringToAddress()` under `/action/protocol`
and here
```
return hashStringToAddress(protocolID)
```

action/protocol/rolldpos/epoch.go
```diff
@@ -92,6 +95,16 @@ func NewProtocol(numCandidateDelegates, numDelegates, numSubEpochs uint64, opts
 	return p
 }
 
+// ProtocolAddr returns the address generated from protocol id
+func ProtocolAddr() address.Address {
```
dustinxie, line 99: same here

api/web3server_utils.go
```diff
@@ -239,7 +244,7 @@ func (svr *Web3Server) getTransactionFromActionInfo(actInfo *iotexapi.ActionInfo
 			if err != nil {
 				return transactionObject{}, err
 			}
-			addr, err := ioAddrToEthAddr(receipt.ContractAddress)
+			addr, err := getExecutionContractAddr(receipt.ContractAddress)
```
dustinxie, line 247: in `ioAddrToEthAddr`, for `input = ""`
```
return "0x0000000000000000000000000000000000000000", nil
``` 
now in `getExecutionContractAddr`, for `input = ""`
```
return "", nil
```
would this be an issue (behavior change for client)?

api/web3server_utils.go
```diff
@@ -372,6 +377,22 @@ func (svr *Web3Server) getLogsWithFilter(from uint64, to uint64, addrs []string,
 	return ret, nil
 }
 
+// contract addr is only returned when a contract was created, otherwise an empty string is returned
+func getExecutionContractAddr(addr string) (string, error) {
+	switch addr {
+	case "":
+		fallthrough
+	case account.ProtocolAddr().String(),
```
dustinxie, line 385: you can just combine `""` into here

action/protocol/rewarding/protocol.go
Liuhaai: I don't think exporting protocolID is better

api/web3server_utils.go
Liuhaai: Good finding. But I think there will never be a scenario where both  `to` and `contractAddr` are empty string

CoderZhi (approved):
the others look good to me

api/web3server_utils.go
```diff
@@ -374,6 +379,21 @@ func (svr *Web3Server) getLogsWithFilter(from uint64, to uint64, addrs []string,
...
 
+// contract addr is only returned when a contract was created, otherwise an empty string is returned
+func getExecutionContractAddr(addr string) (string, error) {
+	switch addr {
+	case "",
+		account.ProtocolAddr().String(),
+		poll.ProtocolAddr().String(),
+		rewarding.ProtocolAddr().String(),
+		rolldpos.ProtocolAddr().String(),
+		staking.ProtocolAddr().String():
```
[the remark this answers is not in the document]
Liuhaai: The addr `protocol.HashStringToAddress(protocolID)` is used in returned Receipt object

dustinxie (approved):

PR #3073, opened 2022-02-12
[api] Move the subscribers of coreService to serverV2

Liuhaai:
Attach subscriber to ServerV2, which receives the block from Pub/Sub manager and passes it to its subscriber.
After this pr, read_cache could be repalced by memKVStore()


api/serverV2.go
```diff
@@ -119,3 +120,9 @@ func (svr *ServerV2) Stop() error {
 	}
 	return nil
 }
+
+// ReceiveBlock receives the new block
+func (svr *ServerV2) ReceiveBlock(blk *block.Block) error {
+	svr.core.readCache.Clear()
+	return svr.core.chainListener.ReceiveBlock(blk)
```
CoderZhi, line 127: if both `readCache` & `chainListener` are variables of `core`, we may implement this function in core?

api/coreservice.go
```diff
@@ -1490,3 +1484,8 @@ func (core *coreService) ReadContractStorage(ctx context.Context, addr address.A
 	}
 	return core.sf.ReadContractStorage(ctx, addr, key)
 }
+
+func (core *coreService) receiveBlock(blk *block.Block) error {
```
CoderZhi, line 1488: now, CoreService is an interface. Maybe we can define ServerV2 as
```
type ServerV2 struct {
    CoreService
    ...
}

CoderZhi (approved):

dustinxie (approved):

Newman1515 (commented):
Review is ok know,..

PR #3090, opened 2022-02-20
[chainservice] Create createBlockSyner()

Liuhaai:


chainservice/chainservice.go
```diff
@@ -334,87 +334,21 @@ func New(
...
-						log.L().Error("failed to request blocks", zap.Error(err), zap.String("peer", peer.ID.Pretty()), zap.Uint64("start", start), zap.Uint64("end", end))
-					}
-				}
-			},
-		)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to create blockSyncer")
-		}
+	bs, err := createBlockSyner(
+		cfg.Consensus.Scheme == config.StandaloneScheme,
```
CoderZhi, line 342: I don't think a boolean parameter is not as readable as passing in the consensus scheme.

CoderZhi (approved):

chainservice/chainservice.go
dustinxie: yeah, i think pass in `cfg.Consensus.Scheme` itself might be better? now the logic is check standalone or not, so a boolean can work, in case need to check another mode in the future, boolean won't work

chainservice/chainservice.go
```diff
@@ -653,3 +587,85 @@ func (cs *ChainService) BlockSync() blocksync.BlockSync {
 
 // Registry returns a pointer to the registry
 func (cs *ChainService) Registry() *protocol.Registry { return cs.registry }
+
+func createBlockSyner(
```
dustinxie, line 591: Syncer

dustinxie (approved):

chainservice/chainservice.go
Liuhaai: There should be a const list (e.g. `DummyBlockSyncerMode`, `NormalBlockSyncerMode`, `LightBlockSyncerMode`) in `blocksyncer` package. It will be added in the future. Currently, using `bool` is a temp prototype

PR #3092, opened 2022-02-21
[action] Add abi encoding in stake create

Liuhaai:


action/stake_create.go
```diff
@@ -23,6 +25,49 @@ const (
...
+			"type": "function"
+		}
+	]`
+)
+
+var (
+	// _createStakeMethod is the interface of the abi encoding of stake action
+	_createStakeMethod abi.Method
+
+	errDecodeFailure = errors.New("failed to decode the data")
```
dustinxie, line 70: since the logic is exactly same, can add the change for other staking actions in this PR as well

CoderZhi (approved):

dustinxie (approved):

PR #3096, opened 2022-02-23
[action] remove crypto.Pubkey from AbstractAction

dustinxie:
`crypto.Pubkey` is from the signature, it is outside the scope of `AbstractAction` and should NOT reside in there

the reason it is there now, is b/c after we load protobuf into `SealEnvelope` struct, we need to set the length of pubkey for transfer and execution, and this is solved by adding a `hasPubKeySize` internal interface

2 additional benefits by fixing this:
1. `envelope` struct can directly contain an `AbstractAction`, more clear
2. we don't need pass in the entire sealedEnvelope struct to `SetEnvelopeContext()`, it is replaced by a `SetAbstractAction()`, which only needs the `AbstractAction` as input


action/actctx.go
```diff
@@ -62,27 +59,19 @@ func (act *AbstractAction) SetGasPrice(val *big.Int) {
...
 		size += len(act.gasPrice.Bytes())
 	}
 
 	return uint32(size)
 }
 
-// SetEnvelopeContext sets the SealedEnvelope context to action context.
-func (act *AbstractAction) SetEnvelopeContext(selp SealedEnvelope) {
+// SetAbstractAction sets the struct to input
+func (act *AbstractAction) SetAbstractAction(ab AbstractAction) {
```
dustinxie, line 70: now we don't need pass in the entire sealedEnvelope struct, only need AbstractAction

action/builder.go
```diff
@@ -31,12 +29,6 @@ func (b *Builder) SetNonce(n uint64) *Builder {
 	return b
 }
 
-// SetSourcePublicKey sets action's source's public key.
-func (b *Builder) SetSourcePublicKey(key crypto.PublicKey) *Builder {
-	b.act.srcPubkey = key
-	return b
-}
-
```
dustinxie, line 39: this is never used

action/execution.go
```diff
@@ -36,6 +39,7 @@ type Execution struct {
 	contract string
 	amount   *big.Int
 	data     []byte
+	sigSize  uint32
```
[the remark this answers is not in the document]
dustinxie: `TotalSize()` includes this length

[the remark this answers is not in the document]
Liuhaai: > `TotalSize()` need this length to compute

`TotalSize()` aims for what purpose?

CoderZhi:
https://github.com/iotexproject/iotex-core/search?q=TotalSize is only used in two places, thus, pubkey size could be calculated by modifying executionsizelimit and transfersizelimit values

dustinxie:
> https://github.com/iotexproject/iotex-core/search?q=TotalSize is only used in two places, thus, pubkey size could be calculated by modifying executionsizelimit and transfersizelimit values

our chain also can support sm2 signature scheme, whose pubkey size = 91 bytes. I feel it's better to read pubkey size from the protobuf data, rather than modifying the `ExecutionSizeLimit` and `TransferSizeLimit`

on the other hand, i do agree it's best not to have `pubkeySize` in transfer and execution struct, let me know if there's a better way of handing

Liuhaai:
> > https://github.com/iotexproject/iotex-core/search?q=TotalSize is only used in two places, thus, pubkey size could be calculated by modifying executionsizelimit and transfersizelimit values
> 
> our chain also can support sm2 signature scheme, whose pubkey size = 91 bytes. I feel it's better to read pubkey size from the protobuf data, rather than modifying the `ExecutionSizeLimit` and `TransferSizeLimit`
> 
> on the other hand, i do agree it's best not to have `pubkeySize` in transfer and execution struct, let me know if there's a better way of handing

could the key size be wrapped into in context?

action/transfer_test.go
```diff
@@ -40,12 +39,13 @@ func TestTransferSignVerify(t *testing.T) {
 	tsf2, ok := w.Envelope.Action().(*Transfer)
 	require.True(ok)
 	require.Equal(tsf, tsf2)
-	require.NotNil(tsf.srcPubkey)
 
 	// sign the transfer
 	selp, err := Sign(elp, senderKey)
 	require.NoError(err)
 	require.NotNil(selp)
+	require.EqualValues(21, tsf.BasicActionSize())
```
Liuhaai, line 47: ignored by last pr?

action/action.go
```diff
@@ -18,7 +18,7 @@ import (
 type (
 	// Action is the action can be Executed in protocols. The method is added to avoid mistakenly used empty interface as action.
 	Action interface {
-		SetEnvelopeContext(SealedEnvelope)
+		SetAbstractAction(AbstractAction)
```
Liuhaai, line 21: Why is `SetAbstractAction` needed in `Action` interface?

action/action.go
```diff
@@ -30,7 +30,7 @@ type (
 	actionPayload interface {
 		Cost() (*big.Int, error)
 		IntrinsicGas() (uint64, error)
-		SetEnvelopeContext(SealedEnvelope)
+		SetAbstractAction(AbstractAction)
```
Liuhaai, line 33: same here

action/envelope.go
```diff
@@ -28,12 +28,8 @@ type (
...
 
 	envelope struct {
-		version  uint32
-		chainID  uint32
-		nonce    uint64
-		gasLimit uint64
-		gasPrice *big.Int
-		payload  actionPayload
+		AbstractAction
+		payload actionPayload
```
Liuhaai, line 32: what is difference between ab in envelop and ab in payload?

action/action.go
dustinxie: good catch, it can be removed

action/action.go
dustinxie: in `func (elp *envelope) LoadProto()`, it calls: `elp.payload.SetAbstractAction(elp.AbstractAction)`, so `actionPayload ` needs this method

action/envelope.go
dustinxie: envelopeBuilder has its own copy of Ab, and in `build()`, use that to overwrites the payload's Ab
IMO this is actually one more thing we need to improve, we should remove `nonce, gas, gasPrice` etc from `NewTransfer()` b/c those values are useless, will be overwritten by `builder.SetNonce().SetGas()`

action/action.go
dustinxie: yes, removed

action/envelope.go
Liuhaai: It looks like they are duplicated. I think removing those params in envelope should be fine. 
```
envelop struct {
payload actionPayload
}
```

dustinxie: yes, it can be in future PR to improve

action/builder.go
```diff
@@ -137,5 +129,9 @@ func (b *EnvelopeBuilder) Build() Envelope {
 	if b.elp.version == 0 {
 		b.elp.version = version.ProtocolVersion
 	}
+	if b.elp.payload == nil {
+		panic("cannot build Envelope w/o a valid payload")
```
Liuhaai, line 133: why is panic added here?

action/action.go
Liuhaai: remove

action/builder.go
dustinxie: added to prevent such case -- `elp.payload` is nil --> no valid action is set --> meaningless to `build()`

action/transfer.go
```diff
@@ -79,8 +79,8 @@ func (tsf *Transfer) TotalSize() uint32 {
 	if tsf.amount != nil && len(tsf.amount.Bytes()) > 0 {
 		size += uint32(len(tsf.amount.Bytes()))
 	}
-
-	return size + uint32(len(tsf.payload))
+	// 65 is the pubkey size
+	return size + uint32(len(tsf.payload)) + 65
```
dustinxie, line 83: verified sm2 signature/tx won't be accepted (due to wrong length), so pubkey size = constant 65

Liuhaai (approved):

action/envelope.go
```diff
@@ -36,6 +36,9 @@ type (
 // Version returns the version
 func (elp *envelope) Version() uint32 { return elp.version }
 
+// ChainID return the chainID value
+func (elp *envelope) ChainID() uint32 { return elp.chainID }
+
```
dustinxie, line 41: move getter func together

action/envelope.go
```diff
@@ -28,18 +28,17 @@ type (
 	}
 
 	envelope struct {
-		version  uint32
-		chainID  uint32
-		nonce    uint64
-		gasLimit uint64
-		gasPrice *big.Int
-		payload  actionPayload
+		AbstractAction
```
CoderZhi, line 31: we should revert this change, which won't hurt this PR, and leave it to follow up PRs after discussion.

CoderZhi (approved):

PR #3049, opened 2022-02-03
[api] add `trace_call`, `trace_callMany`, `trace_rawTransaction` API

Liuhaai:
add `trace_call`, `trace_callMany`, `trace_rawTransaction` api from parity
https://openethereum.github.io/JSONRPC-trace-module

PR #3105, opened 2022-02-25
Fix mockgen and delete state factory

CoderZhi:
1. fix mockgen.sh
2. In current implementation of TraceTransactionStructLogs, `coreservice` was referred multiple times, for `Blockchain`, `BlockDAO`, `cfg`, `StateFactory`. Introduce `SimulateExecution` interface to `CoreServcie`, and delete `StateFactory` interface.

### DIFF
--- a/api/coreservice.go
+++ b/api/coreservice.go
@@ -129,11 +129,11 @@ type (
 		ChainID() uint32
 		// ReadContractStorage reads contract's storage
 		ReadContractStorage(ctx context.Context, addr address.Address, key []byte) ([]byte, error)
+		// SimulateExecution simulates execution
+		SimulateExecution(context.Context, address.Address, *iotextypes.Execution) ([]byte, *action.Receipt, error)
 
 		// BlockChain returns the member bc
 		BlockChain() blockchain.Blockchain
-		// StateFactory return the member sf
-		StateFactory() factory.Factory
 		// BlockDao return the member dao
 		BlockDao() blockdao.BlockDAO
 		// Indexer return the member indexer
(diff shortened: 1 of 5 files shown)


huof6829 (approved):

api/coreservice.go
```diff
@@ -1590,9 +1590,31 @@ func (core *coreService) BlockChain() blockchain.Blockchain {
 	return core.bc
 }
 
-// StateFactory return the member sf
-func (core *coreService) StateFactory() factory.Factory {
-	return core.sf
+func (core *coreService) SimulateExecution(ctx context.Context, addr address.Address, exec *iotextypes.Execution) ([]byte, *action.Receipt, error) {
```
Liuhaai, line 1593: Don't use the type from `iotextypes` as argument in new func in coreService

api/coreservice.go
```diff
@@ -1590,9 +1590,31 @@ func (core *coreService) BlockChain() blockchain.Blockchain {
...
+	}
+	ctx, err = core.bc.Context(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	sc, err := action.NewExecution(
+		exec.GetContract(),
+		state.Nonce+1,
+		amount,
+		core.cfg.Genesis.BlockGasLimit,
```
Liuhaai, line 1610: Current func used for `TraceTransactionStructLogs` is ok. 
But let's make this func usable in more scenario: an error should be returned if the the gaslimit or gasprice of exec is inadaquate. it is because we can't assume that the gaslimit and gasprice of input are reasonable.
See 
https://github.com/iotexproject/iotex-core/pull/3049/files#diff-a84e0d7fdb009bebf88eb3c2582dc4c2f5fce09f6d40a3a514b553c198675f42R383
https://github.com/iotexproject/iotex-core/pull/3049/files#diff-6dc4e8a6418cb4f7e5b7da626b745fb04ff21d7c37b77d883b8c8b9ab7aacfcbR985
https://github.com/iotexproject/iotex-core/pull/3049/files#diff-6dc4e8a6418cb4f7e5b7da626b745fb04ff21d7c37b77d883b8c8b9ab7aacfcbR1006

api/grpcserver_test.go
```diff
@@ -2379,10 +2379,13 @@ func TestGrpcServer_GetTransactionLogByActionHash(t *testing.T) {
 		require.Equal(log.Proto(), res.TransactionLog)
 	}
 
-	// check implicit transfer receiver balance
-	state, err := accountutil.LoadAccount(svr.core.StateFactory(), identityset.Address(31))
-	require.NoError(err)
-	require.Equal(big.NewInt(5), state.Balance)
+	// TODO (huof6829): Re-enable this test
```
Liuhaai, line 2382: Is this test failed?

api/coreservice.go
CoderZhi: Not the duty of this PR. I will leave this to you.

api/grpcserver_test.go
CoderZhi: The purpose of this PR is to delete `StateFactory` interface.

Liuhaai (approved):

api/grpcserver_test.go
dustinxie: since the PR is to delete SF interface, i think can just delete this code block, NO need to leave a TODO. We must have tests (in other places) for implicit transfer

dustinxie (approved):

PR #3108, opened 2022-02-25
[ioctl/newcmd] modify output points to os.Stdout 

huof6829:
in cobra, output is point to os.Stdout default


Liuhaai:
see https://github.com/iotexproject/iotex-core/issues/3109

huof6829:
> see #3109

if execute cmd successed, the `buf` is empty, the result info is printed in cmd line by client.PrintInfo
if execute cmd  failed,  assert err with expected err,  the err info is printed in cmd line by cobra inner function. 
so i think `buf` is useless.

Liuhaai:
> > see #3109
> 
> if execute cmd successed, the `buf` is empty, the result info is printed in cmd line by client.PrintInfo if execute cmd failed, assert err with expected err, the err info is printed in cmd line by cobra inner function. so i think `buf` is useless. and we should see the err info

buf is empty b/c no output redirection is used.

PR #3131, opened 2022-03-02
[ioctl/newcmd] move config.Language to client

huof6829:


ioctl/client.go
```diff
@@ -70,13 +70,16 @@ type (
...
+	Language int
+
 	client struct {
 		cfg       config.Config
 		conn      *grpc.ClientConn
 		cryptoSm2 bool
 
 		// TODO: merge into config
-		lang config.Language
+		lang Language
```
CoderZhi, line 82: I thought we need to move `lang` into `config.Config`.

huof6829: yes , i will create a new pr

PR #3132, opened 2022-03-02
[ioctl/newcmd] modify GetAddress to DefaultAddress 

huof6829:


ioctl/client.go
```diff
@@ -46,9 +50,9 @@ type (
 		ReadSecret() (string, error)
 		// Execute a bash command
 		Execute(string) error
-		// doing
-		GetAddress(in string) (string, error)
-		// doing
+		// DefaultAddress returns default address if input empty
+		DefaultAddress(in string) (string, error)
```
CoderZhi, line 54: the name is still not clear. how about `AddressWithDefaultIfNotExist`

ioctl/client.go
```diff
@@ -147,8 +151,34 @@ func (c *client) SelectTranslation(trls map[config.Language]string) (string, con
 }
 
 func (c *client) ReadSecret() (string, error) {
-	// TODO: delete util.ReadSecretFromStdin, and move code to here
-	return util.ReadSecretFromStdin()
+	signalListener := make(chan os.Signal, 1)
```
CoderZhi, line 154: revert this change. you can do it in a different PR or even leave as it is until we migrate to the new command

ioctl/client.go
huof6829: ok

ioctl/client.go
huof6829: ok

CoderZhi (approved):

Liuhaai (approved):

PR #3136, opened 2022-03-02
[test] add test for byteutil

### LINKED ISSUE #3114: Add unit test to https://github.com/iotexproject/iotex-core/tree/master/pkg/util/byteutil
Add unit test to enhance robustness

LuckyPigeon:
fix #3114

### DIFF
--- a/pkg/util/byteutil/byteutil_test.go
+++ b/pkg/util/byteutil/byteutil_test.go
@@ -1,0 +1,83 @@
+// Copyright (c) 2019 IoTeX Foundation
+// This is an alpha (internal) release and is not suitable for production. This source code is provided 'as is' and no
+// warranties are given as to title or non-infringement, merchantability or fitness for purpose and, to the extent
+// permitted by law, all liability for your use of the code is disclaimed. This source code is governed by Apache
+// License 2.0 that can be found in the LICENSE file.
+
+package byteutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestUint32ToBytes(t *testing.T) {
+	t.Run("convert uint32 to bytes", func(t *testing.T) {
+		expectedValue := []uint8([]byte{0x76, 0x5e, 0xdf, 0x1})
+
+		result := Uint32ToBytes(31415926)
+
+		require.Equal(t, expectedValue, result)
+	})
+}
+
+func TestUint64ToBytes(t *testing.T) {
+	t.Run("convert uint64 to bytes", func(t *testing.T) {
+		expectedValue := []uint8([]byte{0x99, 0x99, 0x99, 0x99, 0x99, 0x99, 0x99, 0x19})
+
+		result := Uint64ToBytes(1844674407370955161)
+
+		require.Equal(t, expectedValue, result)
+	})
+}
+
+func TestBytesToUint64(t *testing.T) {
+	t.Run("convert bytes to unit64", func(t *testing.T) {
+		expectedValue := uint64(1844674407370955161)
+
+		result := BytesToUint64([]byte{0x99, 0x99, 0x99, 0x99, 0x99, 0x99, 0x99, 0x19})
+
+		require.Equal(t, expectedValue, result)
+	})
+}
+
+func TestMust(t *testing.T) {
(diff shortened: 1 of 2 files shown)


huof6829:
please make fmt/lint before create pr

pkg/util/byteutil/byteutil_test.go
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2019 IoTeX Foundation
```
dustinxie, line 1: 2022

pkg/util/byteutil/byteutil_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+package byteutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestUint32(t *testing.T) {
```
dustinxie, line 15: no need to add this test
`Uint32ToBytes` and `Uint32ToBytesBigEndian` just calls standard golang library func

pkg/util/byteutil/byteutil_test.go
```diff
@@ -0,0 +1,79 @@
...
+	t.Run("converts a uint32 to 4 bytes in big-endian", func(t *testing.T) {
+		expectedValue := []uint8([]byte{0x1, 0xdf, 0x5e, 0x76})
+
+		result := Uint32ToBytesBigEndian(input)
+
+		require.Equal(t, expectedValue, result)
+	})
+}
+
+func TestUint64(t *testing.T) {
```
dustinxie, line 34: same, no need

pkg/util/byteutil/byteutil_test.go
```diff
@@ -0,0 +1,79 @@
...
+		expectedValue := uint64(11068046444225730841)
+
+		result := BytesToUint64BigEndian(byteInput)
+
+		require.Equal(t, expectedValue, result)
+	})
+}
+
+func TestMust(t *testing.T) {
+	t.Run("must return identical output when given nil error", func(t *testing.T) {
```
dustinxie, line 72: no need to use `t.Run`, just test 2 cases:
1. Must(a, nil) == a
2. Must(a, err) panics, search `require.Panics()` in the codebase you will see

pkg/util/byteutil/byteutil_test.go
dustinxie: also, pls run `make fmt` and `make lint` before pushing the changelist

dustinxie (approved):

pkg/util/byteutil/byteutil_test.go
CoderZhi: unit test is to test the component in our codebase, and it is supposed to cover all the possible cases.
Take
```
// Uint32ToBytesBigEndian converts a uint32 to 4 bytes in big-endian
func Uint32ToBytesBigEndian(value uint32) []byte {
	bytes := make([]byte, 4)
	binary.BigEndian.PutUint32(bytes, value)
	return bytes
}
```
for example, we may make mistake in `bytes := make([]byte, 4)`, so unit test is needed to ensure we don't pass wrong parameter to `binary.BigEndian.PutUnit32`.

Even for
```
// BytesToUint64BigEndian converts 8 bytes to uint64 in big-endian
func BytesToUint64BigEndian(value []byte) uint64 {
	return binary.BigEndian.Uint64(value)
}
```
in which we call the function `binary.BigEndian.Uint64` with the identical parameters, the unit test is not for `binary.BigEndian.Uint64`, but to validate the function `BytesToUint64BigEndian` is doing what it is claiming to do.

**Conclusion**: We do need unit tests for these functions.

pkg/util/byteutil/byteutil_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"testing"
+
+	"github.com/pkg/errors"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMust(t *testing.T) {
```
CoderZhi, line 17: This unit test is fine, because it is a unit test for `Must` function. If we only keep this function, the title of this PR should be changed to "add unit test for `byteutil.Must`"

CoderZhi (approved):

PR #3141, opened 2022-03-03
apply Readiness to RecurringTask

dustinxie:


pkg/lifecycle/ready.go
```diff
@@ -0,0 +1,54 @@
...
+	if atomic.CompareAndSwapInt32(&r.ready, _ready, _notReady) {
+		return nil
+	}
+	return ErrServiceNotReady
+}
+
+// IsNotReady returns whether the service is not ready (initial state)
+func (r *Readiness) IsNotReady() bool {
+	return atomic.LoadInt32(&r.ready) == _notReady
+}
```
CoderZhi, line 49: delete

pkg/lifecycle/ready.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+const (
+	_notReady = 0
+	_ready    = 1
+)
+
+// vars
+var (
+	ErrServiceNotReady = errors.New("service is not ready to turn on/off")
```
CoderZhi, line 22: the error definition is confusing

server/itx/server.go
```diff
@@ -105,12 +111,15 @@ func (s *Server) Start(ctx context.Context) error {
 	if err := s.dispatcher.Start(cctx); err != nil {
 		return errors.Wrap(err, "error when starting dispatcher")
 	}
-
```
CoderZhi, line 108: revert

server/itx/server.go
```diff
@@ -92,6 +94,10 @@ func newServer(cfg config.Config, testing bool) (*Server, error) {
 
 // Start starts the server
 func (s *Server) Start(ctx context.Context) error {
+	if err := s.TurnOn(); err != nil {
+		return errors.Wrapf(err, "failed to turn on itx server")
+	}
```
CoderZhi, line 99: if this function is called at beginning, what's the definition of `s.Ready()`?

pkg/lifecycle/ready.go
dustinxie: i also feel so, any good suggestion for a better/fit name?

server/itx/server.go
dustinxie: it is `s.Ready()`, feel `TurnOn/Off` is more clear (action verb instead of Ready/NotReady() can be interpreted as status)

pkg/lifecycle/ready.go
dustinxie: maybe ErrWrongState?

CoderZhi (approved):

server/itx/server.go
```diff
@@ -106,11 +108,15 @@ func (s *Server) Start(ctx context.Context) error {
 		return errors.Wrap(err, "error when starting dispatcher")
 	}
 
-	return nil
+	return s.TurnOn()
 }
 
 // Stop stops the server
 func (s *Server) Stop(ctx context.Context) error {
+	if err := s.TurnOff(); err != nil {
```
Liuhaai, line 116: What is the benefit if readiness is addded here?

pkg/routine/recurringtask.go
```diff
@@ -63,12 +64,14 @@ func (t *RecurringTask) Start(ctx context.Context) error {
 	}()
 
 	<-ready
-	return nil
+	return t.TurnOn()
 }
 
 // Stop stops the timer
 func (t *RecurringTask) Stop(_ context.Context) error {
-	// TODO: actually this happens when stop is called before init/start. We should prevent this from happening
```
Liuhaai, line 71: Could you find when will this scenario happen?

server/itx/server.go
dustinxie: Readiness serves as a common mechanism to prevent Stop() being called before Start() finishes
will file a series of tasks to add Readiness into other component like Dispatcher, p2p, etc.

dustinxie: after discussion, we'll determine what is the proper way to use Readiness

Liuhaai (approved):

PR #3157, opened 2022-03-06
[trie] Remove prometheus in trie

Liuhaai:
Prometheus metrics takes at least 50% time when doing trie.Upsert() benchmark. We can add metrics in `contract.go` if needed


CoderZhi (approved):
A better choice is to disable prometheus component for trie in product.

dustinxie (approved):

PR #3172, opened 2022-03-08
[trie] Add sortIfNeed() in sortedlist

Liuhaai:
Most trie operations are `getState()`, which load `branchnode` from `hashnode` and call `Search()`. Therefore, the trie performance improves by avoiding sorting work when loading `branchnode`


CoderZhi (commented):
what if a new created sortedlist is iterated?

CoderZhi (approved):

dustinxie (approved):

PR #3178, opened 2022-03-09
[db] use Readiness to check DB started

dustinxie:


Liuhaai (approved):

blockindex/indexbuilder_test.go
```diff
@@ -68,16 +68,14 @@ func TestIndexBuilder(t *testing.T) {
...
-			require.NoError(indexer.Stop(ctx))
-			require.NoError(dao.Stop(ctx))
-		}()
-
 		ib := &IndexBuilder{
 			dao:     dao,
 			indexer: indexer,
 		}
-		defer ib.Stop(context.Background())
+		defer func() {
```
CoderZhi, line 75: why is this change needed in this PR?

dustinxie: the original test is not properly written

ib is a wrapper of indexer, `ib.Stop()` calls `ib.indexer.Stop()` inside --> so `indexer.Stop()` is called twice

CoderZhi (approved):

PR #3180, opened 2022-03-10
[api] Improve log query with goroutine

### LINKED ISSUE #3047: [api] improve log query with goroutine
Doing logsInBlock() in each block could be done in parallel with goroutines.

**Why is this needed**:

Querying logs from mutiple blocks in `LogsInRange()` could be optimized with goroutine.
https://github.com/iotexproject/iotex-core/blob/da077c8a23a299cc4030368b8643f4382e92716f/api/coreservice.go#L1239

LuckyPigeon:
# Description

**summary of changes**
Prefix unexported top-level vars and consts with _ to make it clear when
they are used that they are global symbols.

Exception: Unexported error values, which should be prefixed with err.

**relevant motivation**
Top-level variables and constants have a package scope. Using a
generic name makes it easy to accidentally use the wrong value in a different
file.

Fixes #3047 

## Type of change

* Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

**[iotex-core@910c9f](https://github.com/iotexproject/iotex-core/tree/910c9fb0340505ad4ab93e2195eca3f615751a6e)**
* TestGrpcServer_GetLogs
* TestGetLogs

**This PR**
* TestLogsInRange
* BenchmarkLogsInRange

**Test Configuration**:

- Firmware version: Ubuntu 20.04
- Hardware:
- Toolchain: go 1.17.6
- SDK:

# Checklist:


api/coreservice.go
```diff
@@ -1379,25 +1383,35 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
 		paginationSize = 5000
 	}
+
+	c := make(chan []*iotextypes.Log)
 	logs := []*iotextypes.Log{}
+	if len(blockNumbers) == 0 {
+		return logs, nil
+	}
 	for _, i := range blockNumbers {
-		logsInBlock, err := core.LogsInBlock(filter, i)
```
CoderZhi, line 1391: it is not encouraged to call a public function of the same struct inside of another function of the struct.

api/coreservice.go
```diff
@@ -1340,8 +1340,12 @@ func (core *coreService) LogsInBlockByHash(filter *logfilter.LogFilter, blockHas
 }
 
 // LogsInBlock filter logs in the block x
-func (core *coreService) LogsInBlock(filter *logfilter.LogFilter, blockNumber uint64) ([]*iotextypes.Log, error) {
-	return core.logsInBlock(filter, blockNumber)
+func (core *coreService) LogsInBlock(filter *logfilter.LogFilter, blockNumber uint64, c chan []*iotextypes.Log) {
```
CoderZhi, line 1343: we may create a goroutine in `web3server` or `grpcserver`, instead of passing `c` for output purpose

Liuhaai, line 1343: It is unnecessary to modify Public method `LogsInBlock` for this issue. `logsInBlock()` could be directly used in `LogsInRange()` if you want

Liuhaai:
And twe requirements for the optimization:
 1.  logs should be returned in order
 2. All goroutines should be closed once `paginationSize` is satisfied or `blockNumbers` has been iterated

api/coreservice.go
LuckyPigeon: I'll use this approach

Liuhaai: > How about put `LogInBlock` in a goroutine function literal and try not to modify the original code structure

you could have a try

api/coreservice.go
LuckyPigeon: I can replace `core.LogsInBlock` to `core.logsInBlock`

api/coreservice.go
LuckyPigeon: I can use a goroutine function literal to solve this issue.

Liuhaai:
> > And twe requirements for the optimization:
> > 
> > 1. logs should be returned in order
> > 2. All goroutines should be closed once `paginationSize` is satisfied or `blockNumbers` has been iterated
> 
> All checked

Take a example for explaination: 
There are three blocks blk1\~3 to be quried. Each of them contains 3 logs in the block: log 1\~3 in blk1, log 4\~6 in blk2, log 7~9 in blk3. If no err happens, the returned result should be [log1, log2, log3, ..., log8, log9].

api/coreservice.go
```diff
@@ -1379,25 +1384,45 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+			ret := &_logsInBlock{
+				logsInBlock,
+				err,
+			}
+			c <- *ret
+		}(i)
+	}
+
+	i := 0
+	for logsInBlock := range c {
```
Liuhaai, line 1411: Msg from channel might not come in order

api/coreservice.go
```diff
@@ -1379,25 +1384,45 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
 	if paginationSize > 5000 {
 		paginationSize = 5000
 	}
+
+	c := make(chan _logsInBlock)
 	logs := []*iotextypes.Log{}
+	if len(blockNumbers) == 0 {
+		return logs, nil
+	}
 	for _, i := range blockNumbers {
```
CoderZhi, line 1399: this is not practical, because `len(blockNumbers)` could be as large as 5000.

api/coreservice.go
```diff
@@ -1379,25 +1384,45 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
-		logsInBlock, err := core.LogsInBlock(filter, i)
-		if err != nil {
+		go func(blockNumber uint64) {
+			logsInBlock, err := core.logsInBlock(filter, blockNumber)
```
CoderZhi, line 1401: if there is an error returned, we may early quit the for loop

api/coreservice.go
LuckyPigeon: But this is how we at the original

api/coreservice.go
LuckyPigeon: We have an error check for this, and early quit because of error is exactly what we need isn't it?

api/coreservice.go
LuckyPigeon: Still solving this, any suggestion?

api/coreservice.go
LuckyPigeon: How about set a max restrict `blockNumbers`, like 500 or something.

Liuhaai: > How about set a max restrict `blockNumbers`, like 500 or something.

you could write an uint test to test whether #goroutine will run out of cpu resources

api/coreservice.go
```diff
@@ -1379,20 +1386,51 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+
+		orderedBlocks = append(orderedBlocks, logsInBlock)
+
+		if i == len(blockNumbers)-1 {
+			close(c)
+		}
+		i++
+	}
+
+	sort.SliceStable(orderedBlocks, func(i, j int) bool {
```
Liuhaai, line 1428: Sorting is not encouraged. IMO, order can be achieved by `producer-consumer` model or by assigning the ans array `logs[i]` in the gorouine `i`, or a new design by yourself

api/coreservice.go
LuckyPigeon: @CoderZhi 
How about this producer-consumer model, this should prevent go routines running out of cpu resources.
https://github.com/iotexproject/iotex-core/blob/a25b77f1bc0747efd713e62a1da8cca9746ad11f/api/coreservice.go#L1401-L1439

api/coreservice.go
LuckyPigeon: The producer-consumer model should prevent `cpu` from consuming out of resources.

CoderZhi (commented):
https://gobyexample.com/worker-pools

LuckyPigeon:
> https://gobyexample.com/worker-pools

Thanks a lot!

api/coreservice.go
```diff
@@ -1392,56 +1392,44 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
-		}
-	}()
-
-	go func() {
-		logs := []*iotextypes.Log{}
-		for logsInBlock := range c {
-			if logsInBlock.err != nil {
-				done <- _logsInBlock{
-					nil,
+	for w := 1; w <= 5; w++ {
```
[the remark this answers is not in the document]
CoderZhi: 5 is fine for now. Or we can make it configurable.

api/coreservice.go
```diff
@@ -1379,26 +1384,51 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
 		paginationSize = 1000
 	}
 	if paginationSize > 5000 {
 		paginationSize = 5000
 	}
+
 	logs := []*iotextypes.Log{}
+	jobs := make(chan uint64, len(blockNumbers))
+	results := make(chan _logsInBlock, len(blockNumbers))
+	// defer close(done)
```
CoderZhi, line 1397: delete

api/coreservice.go
```diff
@@ -1379,26 +1384,51 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+				}
+			}
+		}(w, filter, jobs, results)
+	}
+
 	for _, i := range blockNumbers {
-		logsInBlock, err := core.LogsInBlock(filter, i)
-		if err != nil {
+		jobs <- i
+	}
```
CoderZhi, line 1415: prepare jobs before start workers

api/coreservice.go
```diff
@@ -1379,26 +1384,51 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+	close(jobs)
```
CoderZhi, line 1416: wait until all workers are done using waitgroup

api/coreservice.go
```diff
@@ -1379,26 +1384,51 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+	// defer close(done)
+	if len(blockNumbers) == 0 {
+		return logs, nil
+	}
+	for w := 1; w <= 5; w++ {
+		go func(w int, f *logfilter.LogFilter, jobs <-chan uint64, results chan<- _logsInBlock) {
+			for j := range jobs {
+				logsInBlock, err := core.logsInBlock(f, j)
+				results <- _logsInBlock{
+					logsInBlock,
```
CoderZhi, line 1406: early quit on error

Liuhaai:
Could you add unit test and benchmark test for the improvement?

LuckyPigeon:
Sure!

api/coreservice.go
```diff
@@ -1377,26 +1383,62 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+		go func() {
+			defer wg.Done()
+			for job := range jobs {
+				logsInBlock, err := core.logsInBlock(filter, job)
+				if err != nil {
+					results <- _logsInBlock{
+						nil,
+						err,
+					}
+					return
```
CoderZhi, line 1416: quit all workers by introducing a new parameter `quit := make(chan error)`
```
   for {
      select {
      case err := <-quit:
          quit <- err
          return
       default:
           job, ok := <-jobs
           if !ok {
             return
           }
           logsInBlock, err := core.logsInBlock(filter, job)
           if err != nil {
             quit <- err
             return
           }
           ...
        }
    }
    ...
    wg.Wait()
    close(jobs)
    close(quit)
    err := <-quit
    if err != nil {
        return err
    }
    ....
 ```
not confident it will work perfectly, but you get the idea

File: api/coreservice.go [hunk 13]
```diff
@@ -1377,26 +1383,62 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+					logsInBlock,
+					nil,
+				}
+			}
+		}()
+	}
+	wg.Wait()
+
+	for i := 0; i < len(blockNumbers); i++ {
+		logsInBlock := <-results
```
CoderZhi, line 1429: this is not in order. build a map[uint64]logs first, and then iterate by block number

api/coreservice.go
```diff
@@ -1381,26 +1382,65 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+	}
+	wg.Wait()
+	close(quit)
+	err = <-quit
+	if err != nil {
+		return nil, err
+	}
+
+	for _, i := range blockNumbers {
+		for _, log := range logsMap[i] {
```
CoderZhi, line 1436:
```
logs, ok := logsMap[i]
if !ok {
    // return some error
}
// process logs
```

api/coreservice.go
```diff
@@ -1381,26 +1382,65 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
 	}
 	if paginationSize > 5000 {
 		paginationSize = 5000
 	}
+
 	logs := []*iotextypes.Log{}
+	jobs := make(chan uint64, len(blockNumbers))
+	quit := make(chan error)
+	var wg sync.WaitGroup
+	logsMap := make(map[uint64][]*iotextypes.Log)
```
CoderZhi, line 1396: map is not thread safe. You may need to use `sync.Map` instead.

CoderZhi (approved):

api/coreservice.go
```diff
@@ -1381,26 +1382,70 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+		jobs <- i
+	}
+	close(jobs)
+	for w := 1; w <= 5; w++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				select {
+				case err := <-quit:
```
Liuhaai, line 1410: replace `quit` channel implementation with simple `errgroup.WithContext()`

api/coreservice.go
```diff
@@ -1381,26 +1382,70 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
-	// TODO: improve using goroutine
 	if paginationSize == 0 {
 		paginationSize = 1000
 	}
 	if paginationSize > 5000 {
 		paginationSize = 5000
 	}
+
 	logs := []*iotextypes.Log{}
+	jobs := make(chan uint64, len(blockNumbers))
```
Liuhaai, line 1393: Good use

api/coreservice.go
```diff
@@ -1381,26 +1382,70 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+	if len(blockNumbers) == 0 {
+		return logs, nil
+	}
 	for _, i := range blockNumbers {
-		logsInBlock, err := core.LogsInBlock(filter, i)
-		if err != nil {
+		jobs <- i
+	}
+	close(jobs)
+	for w := 1; w <= 5; w++ {
```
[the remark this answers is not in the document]
LuckyPigeon: Change `LogsInRange`'s parameters seems not a good idea, it might cause some unexpected result.
Any suggestion?

api/coreservice.go
LuckyPigeon: I can't make `errgroup` work functionally with `worker-pool`, needs some help, here's my current solution.
```go
logs := []*iotextypes.Log{}
jobs := make(chan uint64, len(blockNumbers))
logsMap := make(map[uint64][]*iotextypes.Log)
eg, ctx := errgroup.WithContext(context.Background())
if len(blockNumbers) == 0 {
    return logs, nil
}
for _, i := range blockNumbers {
    jobs <- i
}
close(jobs)
for w := 1; w <= 5; w++ {
    eg.Go(func() error {
        for {
		select {
		case <-ctx.Done():
			return nil
		default:
			job, ok := <-jobs
			if !ok {
				return errors.Errorf("invalid job number %d", job)
			}
			logsInBlock, err := core.logsInBlock(filter, job)
			if err != nil {
				return err
			}
			logsMap[job] = logsInBlock
		}
	}
    })
}
if err := eg.Wait(); err != nil {
    return nil, err
}

for _, i := range blockNumbers {
    logsInBlock := logsMap[i]
    for _, log := range logsInBlock {
	logs = append(logs, log)
	if len(logs) >= int(paginationSize) {
		return logs, nil
	}
    }
}
```

LuckyPigeon: And I haven't seen a working solution of `errgroup` + `worker-pool` by google search, is there any chance that these two are not suppose to use together?
The common solution is using `worker-pool` with `quit` channel, just like my previous commit.
Suggestion are welcome.

api/coreservice.go
```diff
@@ -1381,26 +1382,70 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+					job, ok := <-jobs
+					if !ok {
+						return
+					}
+					logsInBlock, err := core.logsInBlock(filter, job)
+					if err != nil {
+						quit <- err
+						return
+					}
+					logsMap.Store(job, logsInBlock)
```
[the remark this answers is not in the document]
LuckyPigeon: concurrent write might happen, check [this](https://app.circleci.com/pipelines/github/iotexproject/iotex-core/4623/workflows/f79b588b-115c-4531-bf6a-a1ee016d98e1/jobs/16620) test.

api/coreservice.go
Liuhaai: see https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext

Liuhaai: Here is my code for reference: 
```
var (
		jobs      = make(chan jobDesc, len(blockNumbers))
		logsInBlk = make([][]*iotextypes.Log, len(blockNumbers))
		eg, ctx   = errgroup.WithContext(context.Background())
	)
	for i, v := range blockNumbers {
		jobs <- jobDesc{i, v}
	}
	close(jobs)
	for w := 1; w <= 10; w++ {
		eg.Go(func() error {
			for {
				select {
				case <-ctx.Done():
					return ctx.Err()
				default:
					job, ok := <-jobs
					if !ok {
						return nil
					}
					logsInBlock, err := core.logsInBlock(filter, job.blkNum)
					if err != nil {
						return err
					}
					logsInBlk[job.idx] = logsInBlock
				}
			}
		})
	}
	if err := eg.Wait(); err != nil {
		return nil, err
	}
```

api/coreservice.go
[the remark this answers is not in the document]
Liuhaai: No need to change the arguments of `LogsInRange`. We can put the #worker in coreService struct

api/coreservice_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+			// LogsInRange will assign default value if from, to returns error
+			from, err := strconv.ParseInt(d.filter.FromBlock, 10, 64)
+			if i == 1 {
+				require.Error(err)
+			} else {
+				require.NoError(err)
+			}
+			to, err := strconv.ParseInt(d.filter.ToBlock, 10, 64)
+			if i == 1 {
```
Liuhaai, line 92: It looks like for loop is unnecessary here because of the expression `i == 1` and `i == 5`.  `t.Run()` would be better for each test

api/coreservice.go
LuckyPigeon: Thanks for the code and explanation.

api/coreservice.go
```diff
@@ -1377,26 +1387,59 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
-		}
-		for _, log := range logsInBlock {
+
+	var (
+		logs      = []*iotextypes.Log{}
+		logsInBlk = make([][]*iotextypes.Log, len(blockNumbers))
+		jobs      = make(chan jobDesc, len(blockNumbers))
+		eg, ctx   = errgroup.WithContext(context.Background())
+	)
+	if len(blockNumbers) == 0 {
```
Liuhaai, line 1403: This can be early returned below L1389

api/coreservice.go
```diff
@@ -1377,26 +1387,59 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
 		return nil, err
 	}
 
-	// TODO: improve using goroutine
 	if paginationSize == 0 {
```
Liuhaai, line 1390: L1390~L1395 can be put after L1432

api/coreservice.go
```diff
@@ -1377,26 +1387,59 @@ func (core *coreService) LogsInRange(filter *logfilter.LogFilter, start, end, pa
...
+		eg, ctx   = errgroup.WithContext(context.Background())
+	)
+	if len(blockNumbers) == 0 {
+		return logs, nil
+	}
+	for i, v := range blockNumbers {
+		jobs <- jobDesc{i, v}
+	}
+	close(jobs)
+	for w := 1; w <= core.workerNumbers; w++ {
```
Liuhaai, line 1410: for w := 0; w < core.workerNumbers; w++ {}

api/serverV2.go
```diff
@@ -76,9 +76,10 @@ func NewServerV2(
 	bfIndexer blockindex.BloomFilterIndexer,
 	actPool actpool.ActPool,
 	registry *protocol.Registry,
+	workerNumbers int,
```
Liuhaai, line 79: Let's wrap this params in `opts`
```
func WithWorkers() Option {}
```
The default number of workers in API config is 5. If the number needs to be changed,  WithWorkers() can be used then.

api/coreservice_test.go
```diff
@@ -0,0 +1,168 @@
...
+func TestLogsInRange(t *testing.T) {
+	require := require.New(t)
+	svr, _, _, _, cleanCallback := setupTestServer(t)
+	defer cleanCallback()
+
+	t.Run("blocks with four logs", func(t *testing.T) {
+		testData := &filterObject{FromBlock: "1", ToBlock: "4"}
+		filter, err := getTopicsAddress(testData.Address, testData.Topics)
+		require.NoError(err)
+		from, to, err := svr.web3Server.parseBlockRange(testData.FromBlock, testData.ToBlock)
```
Liuhaai, line 32: The method in web3Server is unnecessary in coreservice test.

api/coreservice_test.go
```diff
@@ -0,0 +1,168 @@
...
+	cfg.Chain.ChainDBPath = testDBPath
+	cfg.Chain.IndexDBPath = testIndexPath
+	cfg.System.SystemLogDBPath = testSystemLogPath
+	cfg.Chain.EnableAsyncIndexWrite = false
+	cfg.Genesis.EnableGravityChainVoting = true
+	cfg.ActPool.MinGasPriceStr = "0"
+	cfg.API.RangeQueryLimit = 100
+
+	config.SetEVMNetworkID(_evmNetworkID)
+	svr, _, _, _, _, _, bfIndexFile, _ := createServerV2(cfg, false)
```
Liuhaai, line 140: could the normal test pattern be used?
```
svr, _, _, _, cleanCallback := setupTestServer(t)
defer cleanCallback()
```

LuckyPigeon: No, `setupTestServer` can't accept b (`testing.B`) as argument.

api/serverV2.go
```diff
@@ -65,6 +66,14 @@ func WithActionIndex() Option {
 	}
 }
 
+// WithWorkers is the option to manipulate api.LogsInRange's go routine numbers
+func WithWorkers() Option {
```
Liuhaai, line 70: After discussion with other engineers in our dev team, the `worker pool` pattern might be used in other package. So `config.API` isn't the good place to store `workerNumbers` param. So let's hardcode the param in the `coreservice.go` at first. It will be moved in the future. Sorry about the trouble.

api/coreservice.go
```diff
@@ -159,6 +160,13 @@ type (
 		hasActionIndex    bool
 		electionCommittee committee.Committee
 		readCache         *ReadCache
+		workerNumbers     int
```
Liuhaai, line 163: var (
_workerNumbers int =5
)

api/coreservice_test.go
Liuhaai: thx for the feedback. you could remove the `t` argument by replacing `r.NoError(err)`  with `if err != nil {return panic(err)}` in `newConfig(t)`

api/coreservice_test.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func BenchmarkLogsInRange(b *testing.B) {
+	svr, _, _, _, cleanCallback := setupTestServer()
+	defer cleanCallback()
+
+	filter := &filterObject{FromBlock: "0x1"}
+	from, _ := strconv.ParseInt(filter.FromBlock, 10, 64)
+	to, _ := strconv.ParseInt(filter.ToBlock, 10, 64)
+	var filters iotexapi.LogsFilter
```
Liuhaai, line 130: `filters, err := getTopicsAddress(filter.Address, filter.Topics)`

Liuhaai (approved):

api/coreservice_test.go
```diff
@@ -0,0 +1,149 @@
...
+		for _, str := range tp {
+			b, _ := hexToBytes(str)
+			topic = append(topic, b)
+		}
+		filters.Topics = append(filters.Topics, &iotexapi.Topics{Topic: topic})
+	}
+
+	b.Run("five workers to extract logs", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			svr.web3Server.coreService.LogsInRange(logfilter.NewLogFilter(&filters, nil, nil), uint64(from), uint64(to), uint64(0))
```
Liuhaai, line 146: `FilterBlocksInRange` is the bottleneck of the entire benchmark. It would be better If it could be mocked here.

LuckyPigeon: But `mock api` needs create by these
```go
require := require.New(t)
ctrl := gomock.NewController(t)
defer ctrl.Finish()
core := mock_apicoreservice.NewMockCoreService(ctrl)
```

However, we can't use `t` inside `BenchmarkTest`, unless we don't use it.

Liuhaai: could you try 
```
	ctrl := gomock.NewController(b)
	defer ctrl.Finish()
	core := mock_apicoreservice.NewMockCoreService(ctrl)
```

api/coreservice_test.go
```diff
@@ -7,16 +7,139 @@
...
+	core := mock_apicoreservice.NewMockCoreService(ctrl)
+	
+	testData := &filterObject{FromBlock: "0x1"}
+	filter, _ := getTopicsAddress(testData.Address, testData.Topics)
+	from, _ := strconv.ParseInt(testData.FromBlock, 10, 64)
+	to, _ := strconv.ParseInt(testData.ToBlock, 10, 64)
+
+	b.Run("five workers to extract logs", func(b *testing.B) {
+		core.EXPECT().LogsInRange(logfilter.NewLogFilter(&filter, nil, nil), uint64(from), uint64(to), uint64(0)).DoAndReturn(func(filter *logfilter.LogFilter, start, end, paginationSize uint64) ([]*iotextypes.Log, error) {
+			logs, _ := svr.web3Server.coreService.LogsInRange(filter, start, end, paginationSize)
```
Liuhaai, line 114: This bottleneck is at
`blockNumbers, err := core.bfIndexer.FilterBlocksInRange(filter, start, end)`, which should be mocked.

LuckyPigeon: That means I need to generate a new mock `FilterBlocksInRange`?

test/mock_blockindex/mock_blockindex.go
```diff
@@ -0,0 +1,153 @@
+// Code generated by MockGen. DO NOT EDIT.
```
Liuhaai, line 1: Oh, I forgot `mock_blockindex` is missing in the ./test folder. Actually, this can be generated by adding cmds in `./misc/scripts/mockgen.sh` and `make mockgen` under root dir

Liuhaai:
thx for contribution

PR #3192, opened 2022-03-14
[trie] Fix duplicate flush and remove flush argument

Liuhaai:
`store()` reflush the children by `cn.hash(true)`

### DIFF
--- a/db/trie/mptrie/extensionnode.go
+++ b/db/trie/mptrie/extensionnode.go
@@ -36,7 +36,14 @@ func newExtensionNode(
 	e.cacheNode.serializable = e
 
 	if !mpt.async {
-		return e.store()
+		if err := e.Flush(); err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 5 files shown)


db/trie/mptrie/extensionnode.go
```diff
@@ -178,12 +176,15 @@ func (e *extensionNode) updatePath(path []byte, hashnode bool) (node, error) {
 	e.dirty = true
 
 	if !e.mpt.async {
-		hn, err := e.store()
-		if err != nil {
+		if err := e.Flush(); err != nil {
```
CoderZhi, line 179: we are trying to prevent calling Public functions inside of struct.

CoderZhi (commented):
How will this PR improve performance? If not significant, we may postpose it until we have all the perf PRs in first. There will be some structure level changes.

raullenchai:
Close b/c it has been inactive for 1+yr.

PR #3197, opened 2022-03-15
Move Verify to SealedEnvelope

RobertKwiatkowski:
Verify function is now SealedEnvelope method

### DIFF
--- a/action/sealedenvelope.go
+++ b/action/sealedenvelope.go
@@ -1,14 +1,18 @@
 package action
 
 import (
+	"encoding/hex"
+
 	"github.com/iotexproject/go-pkgs/crypto"
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
 	"github.com/iotexproject/iotex-proto/golang/iotextypes"
 	"github.com/pkg/errors"
+	"go.uber.org/zap"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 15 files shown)


RobertKwiatkowski:
I believe that the `Verify() `function test from `action_test.go` should also be moved to the `sealedenvelope_test.go` However there was no standalone test just for the Verify function thus I would like to receive some information/tips from your side.

Liuhaai (approved):

Liuhaai:
thx for contribution. It'd be better if you can write unit test for `Verify()`

dustinxie (approved):

action/sealedenvelope.go
```diff
@@ -179,3 +183,27 @@ func wrapStakingActionIntoExecution(ab AbstractAction, toAddr []byte, pb proto.M
 		data:           data,
 	}, nil
 }
+
+// Verify verifies the action using sender's public key
+func (sealed *SealedEnvelope) Verify() error {
```
CoderZhi, line 188: it is hard to say whether it is appropriate to have this function as a function of struct sealedenvelope for the following reasons:
1. It uses so many public functions of itself, which is not recommended in our codebase
2. The original `Verify` function is more like a util function, which groups some checks together. It is only used in `action/protocol/generic_validator.go`. The usage of this function in several test files are misuse. We can split it into several checks:
a. check public key
b. check gas limit
c. check signature
check a and b could be moved into `action/protocol/generic_validator.go`

PR #3206, opened 2022-03-17
[action] fix issue #3198

### LINKED ISSUE #3198: make sealedEnvelope.Verify() more clear
https://github.com/iotexproject/iotex-core/commit/887ebabb361bdb9ec5e803f46ce3bc7b1ef14134#diff-fda8afdb7165ec30b8d2c40dfd1963def0ffeb117a3d9098167b8139c72527caR192-R196
it actually only used/related to the `Envelope` inside sealedEnvelope, move these 4 lines to a new func 
```
Envelope interface {
    CheckGas() error
}
```
note some of the test for `Verify()` is actually testing `CheckGas()`, for example: https://github.com/iotexproject/iotex-core/commit/887ebabb361bdb9ec5e803f46ce3bc7b1ef14134#diff-1b86240cc6a71c0115a9fcf9282d5416f8391411d3b706a2e8bbfccd496ce308R69

**Why is this needed**:
better code structure/readability

Liuhaai:
fix #3198


action/sealedenvelope.go
```diff
@@ -184,17 +184,11 @@ func wrapStakingActionIntoExecution(ab AbstractAction, toAddr []byte, pb proto.M
 	}, nil
 }
 
-// Verify verifies the action using sender's public key
-func (sealed *SealedEnvelope) Verify() error {
+// VerifyPubKey verifies the action using sender's public key
+func (sealed *SealedEnvelope) VerifyPubKey() error {
```
dustinxie, line 188: `Verify()` is good
it is verifying the signature using pubkey, not verify pubkey

action/envelope.go
```diff
@@ -25,6 +25,7 @@ type (
 		LoadProto(pbAct *iotextypes.ActionCore) error
 		SetNonce(n uint64)
 		SetChainID(chainID uint32)
+		CheckGas() error
```
CoderZhi, line 28: as discussed offline, CheckGas is not the duty of envelope/action

action/sealedenvelope.go
CoderZhi, line 188: not only the public key, but also signature. Because public key is needed to validate signature, `VerifySignature` may be a better name.

dustinxie: IMO `Verify` is simple and clear enough

Liuhaai: I think `VerifySignature` is ok

CoderZhi (approved):

action/sealedenvelope.go
dustinxie: let's stick to `Verify`. IMO we should avoid keep changing func names, unless the name is very wrong? It may not be a big issue for each such change, but after 3~4 times refactor/change, the code can look quite differently, which IMO is not good for readability and grasp of the codebase

dustinxie (approved):

PR #3209, opened 2022-03-17
[api] Support staking action in web3

Liuhaai:
fix #2922 

 - Enable staking actions able to be boardcasted to network by `sendRawTransaction()`
 - Enable staking action queryable by its act hash
 - The gas estimation of staking actions has been implemented in #3174

PR #3212, opened 2022-03-18
[action] Replace actionPayload with Action in envelope

Liuhaai:
- Every `Action` could be signed into a sealed envelope, so it is reasonable to have `Action` as the payload of envelope. This will also make possible that `EnvelopeBuilder` construct an envelope with an `Action ` outside of `action` pkg.
 - `actionPayload` interface is used to assert those which have `Cost()` method ( If every `Action` should be of  `Cost()` method, it should be added into `Action` interface)


Liuhaai:
this pr is prerequisite for #3209

PR #3213, opened 2022-03-18
[context] add public key size into ActionCtx

dustinxie:
for #3096, it's not desired to add pubkey size into action struct, so add it to context (to be used at `Validate()` time)

### DIFF
--- a/action/protocol/execution/protocol.go
+++ b/action/protocol/execution/protocol.go
@@ -76,13 +76,13 @@ func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.St
 }
 
 // Validate validates an execution
-func (p *Protocol) Validate(_ context.Context, act action.Action, _ protocol.StateReader) error {
+func (p *Protocol) Validate(ctx context.Context, act action.Action, _ protocol.StateReader) error {
 	exec, ok := act.(*action.Execution)
 	if !ok {
 		return nil
 	}
 	// Reject oversize execution
-	if exec.TotalSize() > ExecutionSizeLimit {
+	if exec.TotalSize()+protocol.MustGetActionCtx(ctx).PubkeySize > ExecutionSizeLimit {
 		return action.ErrOversizedData
 	}
 	return nil
--- a/action/protocol/execution/protocol.go
+++ b/action/protocol/execution/protocol.go
@@ -82,7 +82,7 @@ func (p *Protocol) Validate(ctx context.Context, act action.Action, _ protocol.S
 		return nil
 	}
 	// Reject oversize execution
-	if exec.TotalSize()+protocol.MustGetActionCtx(ctx).PubkeySize > ExecutionSizeLimit {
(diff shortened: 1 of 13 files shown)


action/protocol/execution/protocol.go
```diff
@@ -76,13 +76,13 @@ func (p *Protocol) Handle(ctx context.Context, act action.Action, sm protocol.St
...
 // Validate validates an execution
-func (p *Protocol) Validate(_ context.Context, act action.Action, _ protocol.StateReader) error {
+func (p *Protocol) Validate(ctx context.Context, act action.Action, _ protocol.StateReader) error {
 	exec, ok := act.(*action.Execution)
 	if !ok {
 		return nil
 	}
 	// Reject oversize execution
-	if exec.TotalSize() > ExecutionSizeLimit {
+	if exec.TotalSize()+protocol.MustGetActionCtx(ctx).PubkeySize > ExecutionSizeLimit {
```
CoderZhi, line 85: `public key size` could be 65, or 64, or 128, which are all constant values determined by the crypto algorithm used by the blockchain/system. What we want to limit here is the size of an execution or a transfer, with ExecutionSizeLimit, a limit to execution, it makes sure that the size won't be too large unexpectedly. The main factor for the size limitation is `payload`. In other words, our purpose here is to limit payload size. In this case, the `public key size` doesn't matter, if we consider ExecutionSizeLimit is a limit to execution without considering the system specific parameter `public key size`.

Liuhaai (approved):

action/protocol/execution/protocol.go
dustinxie: yes i know, question is in the past, is there any corner case tx exist on our mainnet/testnet, say with
```
payload size = ExecutionSizeLimit - 60
```
such tx would be rejected in current code, but if we remove the `pubkeySize` out of the equation, it will be accepted, so we need to keep the `pubkey size` to make the behavior compatible/consistent

or we could also make this a HF (exclude `pubkeySize` after next height), and then later verify that removing it would sync all past blocks, then remove it completely. I am fine with this as well, let me know

CoderZhi: exclude pub key is not hf, it is backward compatible

dustinxie: i think it's similar to the chainID thing?
for the corner case example, 12 delegates running new code (exclude pubkey size) will accept the tx, 12 delegates running old code (counting pubkey size) will reject the tx. A block proposed by new code delegate will fail at old code delegate (though in reality i think the chance is very slim)

dustinxie: I also agree `pubkey size` should not matter for the purpose of limit payload size.
I would suggest make it a HF, and remove it later. let me know

CoderZhi: it is not a hard fork

CoderZhi (approved):
we should remove the new context parameter eventually.

dustinxie:
> we should remove the new context parameter eventually.

yes, agreed, will remove eventually

PR #3371, opened 2022-04-29
[p2p] fix bootnode connection

dustinxie:
# Description

Jeremi in Dev Champion team reporting his node cannot join mainnet. After I asked him to try removing bootnode-5/6 from config.yaml it works -- which led me to suspect that the node is **not** connecting to all bootnodes at start-up time. By checking the log it is confirmed such is the case, and this PR fixed the issue

## Type of change

# How Has This Been Tested?
- [x] make test
- [x] start fullnode

# Checklist:


p2p/agent.go
```diff
@@ -502,8 +502,9 @@ func (p *agent) connect(ctx context.Context) error {
 	connErrChan := make(chan error, len(p.cfg.BootstrapNodes))
 
 	// try to connect to all bootstrap node beside itself.
-	for _, bootAddr := range p.bootNodeAddr {
+	for i := range p.bootNodeAddr {
```
dustinxie, line 505:
```
for _, bootAddr := range p.bootNodeAddr {
```
`bootAddr` is a single variable defined before the loop, so its value is overwritten by the last element in the array. This is usually OK if the loop executes sequentially -- but here a goroutine is started for each loop, so all goroutines end up using the last element in the array

Liuhaai (approved):

Liuhaai (approved):

huof6829 (approved):

PR #3375, opened 2022-05-01
[chainservice] staking cand/bucket indexer should be a gateway component

dustinxie:
# Description
1. cand/bucket indexer should be a gateway component
2. make the logic same as before refactor: https://github.com/iotexproject/iotex-core/commit/1065ea8d7a4bc5119e86456fc5742d5372781ac9#diff-6ae88fef41b61972a03af3642ac7d1da3d3eaa753f5a31e080cef9d8376054ebL178-L209

## Type of change

# How Has This Been Tested?
- [x] make test

# Checklist:

### DIFF
--- a/chainservice/builder.go
+++ b/chainservice/builder.go
@@ -303,10 +303,9 @@ func (builder *Builder) createGateWayComponents(forTest bool) (
 		if err != nil {
 			return
 		}
-		if !builder.cfg.Chain.EnableStakingIndexer {
-			return
+		if builder.cfg.Chain.EnableStakingIndexer {
+			candBucketsIndexer, err = staking.NewStakingCandidatesBucketsIndexer(db.NewMemKVStore())
 		}
-		candBucketsIndexer, err = staking.NewStakingCandidatesBucketsIndexer(db.NewMemKVStore())
 		return
 	}
 	dbConfig := builder.cfg.DB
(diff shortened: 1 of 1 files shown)


chainservice/builder.go
```diff
@@ -240,13 +240,11 @@ func (builder *Builder) buildBlockDAO(forTest bool) error {
...
-			indexers = append(indexers, builder.cs.bfIndexer)
-		}
-		if builder.cs.indexer != nil {
-			indexers = append(indexers, builder.cs.indexer)
-		}
+	if !builder.cfg.Chain.EnableAsyncIndexWrite && builder.cs.indexer != nil {
+		indexers = append(indexers, builder.cs.indexer)
+	}
+	if builder.cs.bfIndexer != nil {
+		indexers = append(indexers, builder.cs.bfIndexer)
```
dustinxie, line 247: `bfIndexer` should always be added

chainservice/builder.go
```diff
@@ -320,7 +300,13 @@ func (builder *Builder) createGateWayComponents(forTest bool) (
 			return
 		}
 		candidateIndexer, err = poll.NewCandidateIndexer(db.NewMemKVStore())
-
+		if err != nil {
+			return
+		}
+		if !builder.cfg.Chain.EnableStakingIndexer {
```
Liuhaai, line 306: `if builder.cfg.Chain.EnableStakingIndexer {}` ?

Liuhaai, line 306: `if builder.cfg.Chain.EnableStakingIndexer {}` ?

Liuhaai (approved):

chainservice/builder.go
CoderZhi: This is definitely a bug, which won't work if `EnableAsyncIndexWrite` is true. After this PR, the logic is identical to previous implementation. But why `bfIndexer` should be updated synchronically, instead of something like https://github.com/iotexproject/iotex-core/blob/5237ffb5a3b93d9dde73595c4f28363df24c43a3/chainservice/builder.go#L354?

dustinxie: right, the PR is to make logic identical as before
checked the `bfIndexer` code, it is to save all logs in every block, so I think it can also be wrapped into a `indexBuilder` and handled the similar way by `AddSubscriber`, can be another PR

CoderZhi (approved):

PR #3376, opened 2022-05-02
[accountutil] LoadOrCreateAccount() use address.Address as input

dustinxie:
# Description
This PR servers 2 purposes:
1. by passing in `address.Address`, it saves 1 conversion from string to `address.Address`, and also consistent with other funcs in `util.go` using `address.Address` as input
2. a small patch for the legacy address PR: https://github.com/iotexproject/iotex-core/commit/5a1fe5d978d45b07446e3750a177f5fba6876e96#diff-9c655a6b1d90598fa3fa865bf5235dd04787e0005d400508212bad3503c893a9R68-R73, see comment in transfer.go

Fixes # (issue)

# How Has This Been Tested?
- [x] make test

# Checklist:

### DIFF
--- a/action/protocol/account/transfer.go
+++ b/action/protocol/account/transfer.go
@@ -32,7 +32,7 @@ func (p *Protocol) handleTransfer(ctx context.Context, act action.Action, sm pro
 		return nil, nil
 	}
 	// check sender
-	sender, err := accountutil.LoadOrCreateAccount(sm, actionCtx.Caller.String())
+	sender, err := accountutil.LoadOrCreateAccount(sm, actionCtx.Caller)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to load or create the account of sender %s", actionCtx.Caller.String())
 	}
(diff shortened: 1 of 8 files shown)


action/protocol/account/transfer.go
```diff
@@ -117,7 +117,7 @@ func (p *Protocol) handleTransfer(ctx context.Context, act action.Action, sm pro
 		return nil, errors.Wrap(err, "failed to update pending account changes to trie")
 	}
 	// check recipient
-	recipient, err := accountutil.LoadOrCreateAccount(sm, tsf.Recipient())
+	recipient, err := accountutil.LoadOrCreateAccount(sm, recipientAddr)
```
dustinxie, line 120: `tsf.Recipient()` is the string format, so we need to check similar as `address.FromLegacy()`
instead of doing that, directly use the `recipientAddr`

action/protocol/poll/util.go
```diff
@@ -169,7 +169,11 @@ func setCandidates(
 	}
 	loadCandidatesLegacy := featureCtx.LoadCandidatesLegacy(height)
 	for _, candidate := range candidates {
-		delegate, err := accountutil.LoadOrCreateAccount(sm, candidate.Address)
+		addr, err := address.FromString(candidate.Address)
```
dustinxie, line 172: poll is no longer used, so as long as it can pass fullsync test, it will be fine

Liuhaai (approved):

PR #3383, opened 2022-05-10
[filedao] avoid extra Deserialize

dustinxie:
# Description
as title

## Type of change

# How Has This Been Tested?
- [x] make test

# Checklist:

### DIFF
--- a/blockchain/filedao/filedao_v2.go
+++ b/blockchain/filedao/filedao_v2.go
@@ -51,6 +51,21 @@ type (
 	}
 )
 
+// createNewV2File creates a new v2 chain db file
+func createNewV2File(start uint64, cfg db.Config) error {
+	v2, err := newFileDAOv2(start, cfg)
+	if err != nil {
+		return err
+	}
+
+	// calling Start() will write the header
+	ctx := context.Background()
+	if err := v2.Start(ctx); err != nil {
+		return err
+	}
+	return v2.Stop(ctx)
+}
+
 // newFileDAOv2 creates a new v2 file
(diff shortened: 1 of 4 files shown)


blockchain/filedao/filedao_v2.go
```diff
@@ -51,6 +51,21 @@ type (
 	}
 )
 
+// createNewV2File creates a new v2 chain db file
+func createNewV2File(start uint64, cfg db.Config) error {
```
dustinxie, line 55: `v2.go` is a more proper place for this func

blockchain/filedao/staging_buffer.go
```diff
@@ -35,14 +35,10 @@ func (s *stagingBuffer) Get(pos uint64) (*block.Store, error) {
 	return s.buffer[pos], nil
 }
 
-func (s *stagingBuffer) Put(pos uint64, blkBytes []byte) (bool, error) {
+func (s *stagingBuffer) Put(pos uint64, blk *block.Store) (bool, error) {
```
dustinxie, line 38: directly pass in `*block.Store`, then we can save the extra de-serialization below

Liuhaai (approved):

blockchain/filedao/staging_buffer.go
Liuhaai: be careful about pointer sharing

dustinxie: after checking, using the incoming `blk` may not be a good idea, since caller may modify it.
and now i recall that's why it is not designed that way in the first place, so close this PR

PR #3436, opened 2022-06-06
[protocol] apply address format check only to transfer protocol

dustinxie:
# Description
at 180 HF, we apply the check to all protocols for safety. This PR removes unnecessary check for staking and execution protocol

## Type of change

# How Has This Been Tested?
- [x] make test
- [x] fullsync

# Checklist:


action/consignment_transfer.go
```diff
@@ -118,11 +118,7 @@ func processConsignmentEther(c ConsignJSON, tolerateLegacyAddress bool) (Consign
 	if con.signer == nil {
 		return nil, errors.New("failed to get address")
 	}
-	if tolerateLegacyAddress {
-		con.recipient, err = address.FromStringLegacy(msg.Recipient)
-	} else {
-		con.recipient, err = address.FromString(msg.Recipient)
-	}
+	con.recipient, err = address.FromString(msg.Recipient)
```
dustinxie, line 121: fullsync test shows we can directly use `FromString()`

Liuhaai (approved):

action/protocol/account/transfer.go
```diff
@@ -75,12 +75,10 @@ func (p *Protocol) handleTransfer(ctx context.Context, act action.Action, sm pro
 		return nil, errors.Wrapf(err, "failed to decode recipient address %s", tsf.Recipient())
 	}
 	recipientAcct, err := accountutil.LoadAccount(sm, recipientAddr)
-	if !fCtx.TolerateLegacyAddress {
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to load address %s", tsf.Recipient())
-		}
+	if err != nil {
```
CoderZhi, line 78: I will assume that no error was returned before the `toleratelegacyaddress` fork.

action/execution.go
```diff
@@ -198,6 +198,12 @@ func (ex *Execution) SanityCheck() error {
 	if ex.Amount().Sign() < 0 {
 		return errors.Wrap(ErrInvalidAmount, "negative value")
 	}
+	// check if contract's address is valid
+	if ex.Contract() != EmptyAddress {
+		if _, err := address.FromString(ex.Contract()); err != nil {
+			return errors.Wrapf(err, "error when validating contract's address %s", ex.Contract())
+		}
+	}
```
CoderZhi, line 206: for `transfer`, we don't check recipient, because of the fork. for `createstake`, we won't check `candidatename`. for `execution`, should we check `recipient` here? As we discussed offline last time, `recipient` is not import as long as the `sender` is legal. So, should we do this in `SanityCheck`?

action/protocol/account/transfer.go
dustinxie: right, fullsync test proves that

action/execution.go
dustinxie: this is to revert to the old code before the HF.

due to that wrong address we found in transfer, we need to call `FromStringLegacy` before the HF, and `FromString` after the HF, which requires to get the HF flag from `ctx`, so move the check code to `Validate(ctx`, and for safety, we apply this check to `staking` and `execution` as well

now fullsync test shows this safety is not necessary for `staking` and `execution`, we can use `FromString` even before the HF. Or put it another way, if there were a wrong address in any execution tx in the past, then the error would fire here, and fullsync test won't pass

dustinxie: thx for bringing up, this is a valid design issue

i think we should do in `SanityCheck`? because it is called by act pool, so a tx with wrong recipient could be detected and won't even enter actpool. If we move to protocol's `Validate` (like we have to for `transfer` protocol due to the HF), it will only be detected later in `pickAndRunActions`. We should kick out invalid tx as early as possible in the pipeline

dustinxie: 1. for `transfer`, we have to move to `Validate(` due to the HF
2. for all `staking` and `candidate` tx, the address is being checked inside `LoadProto` by calling `address.FromString`

CoderZhi (approved):

action/execution.go
dustinxie: as we chatted, move this into `Validate()` later

CoderZhi: we haven't reached agreement on that

PR #3606, opened 2022-08-22
limit account nonce

### LINKED ISSUE #3593: limit account nonce to 2^64-1
### What would you like to be added
limit account nonce to 2^64-1

### Why is this needed
see https://github.com/ethereum/go-ethereum/commit/f32feeb260054d7c90c1efd57dfec6b717717e18

### How important you think this is for IoTeX
- [] must have
- [x] should have
- [] nice to have

### Additional information
this is EIP-2681 requirements, at this point no account/address has nonce value even close to this upper limit. So this is not urgent, can be done at any time

huangzhiran:
# Description
limit account nonce to 2^64-1 (my implement limit to 2^64-2)

Fixes #3593

## Type of change

# How Has This Been Tested?
- [x] make test

**Test Configuration**:
- Firmware version: macOS Monterey arm64
- Hardware:
- Toolchain:
- SDK:

# Checklist:

### DIFF
--- a/state/account.go
+++ b/state/account.go
@@ -141,23 +141,21 @@ func (st *Account) AccountType() int32 {
 	return st.accountType
 }
 
-func (st *Account) isMaxNonce() bool {
-	// for legacy account, nonce will finally +2. so use +2 as overflow flag
-	return st.nonce+2 < st.nonce
-}
-
 // SetPendingNonce sets the pending nonce
 func (st *Account) SetPendingNonce(nonce uint64) error {
-	if st.isMaxNonce() {
-		return errors.Wrapf(ErrNonceOverflow, "current value %d", st.nonce)
-	}
 	switch st.accountType {
 	case 1:
+		if st.nonce+1 < st.nonce {
+			return errors.Wrapf(ErrNonceOverflow, "current value %d", st.nonce)
+		}
 		if nonce != st.nonce+1 {
 			return errors.Wrapf(ErrInvalidNonce, "actual value %d, %d expected", nonce, st.nonce+1)
 		}
 		st.nonce++
 	case 0:
+		if st.nonce+2 < st.nonce {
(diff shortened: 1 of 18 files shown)


state/account.go
```diff
@@ -139,8 +141,16 @@ func (st *Account) AccountType() int32 {
 	return st.accountType
 }
 
+func (st *Account) isNonceOverflow() bool {
+	// if st.accountType is 0, nonce will add 2. so use +2 as overflow flag
+	return st.nonce+2 < st.nonce
```
huof6829, line 146: Limit account nonce to be between 0 and 2^64-1

huangzhiran: for legacy account, nonce will finally +2 ( when SetPendingNonce() and PendingNonce() ), so maybe nonce in [0, 2^64-2] will be better

state/account.go
```diff
@@ -139,8 +141,16 @@ func (st *Account) AccountType() int32 {
...
 }
 
+func (st *Account) isMaxNonce() bool {
+	// for legacy account, nonce will finally +2. so use +2 as overflow flag
+	return st.nonce+2 < st.nonce
+}
+
 // SetPendingNonce sets the pending nonce
 func (st *Account) SetPendingNonce(nonce uint64) error {
+	if st.isMaxNonce() {
```
Liuhaai, line 151: the logic could be movd under switch

huangzhiran: if movd under switch. st.nonce will already ++ and maybe overflow already. pre-check maybe better

Liuhaai: Could you give an example?

Liuhaai (approved):

state/account.go
huangzhiran: changed already

millken (approved):

PR #3679, opened 2022-10-21
[staking] fix name and operator map alias

dustinxie:
# Description
Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [x] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:


action/protocol/staking/handlers.go
```diff
@@ -684,8 +684,9 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
 	if err := csm.Upsert(c); err != nil {
 		return log, nil, csmErrorToHandleError(owner.String(), err)
 	}
+	featureHeightCtx := protocol.MustGetFeatureWithHeightCtx(ctx)
 	height, _ := csm.SM().Height()
-	if p.needToWriteCandsMap(height) {
+	if p.needToWriteCandsMap(height) && !featureHeightCtx.FixCandCenterAlias(height) {
```
Liuhaai, line 689: if p.needToWriteCandsMap(ctx, height) {}

action/protocol/staking/handlers.go
```diff
@@ -767,8 +768,9 @@ func (p *Protocol) handleCandidateUpdate(ctx context.Context, act *action.Candid
 	if err := csm.Upsert(c); err != nil {
 		return log, csmErrorToHandleError(c.Owner.String(), err)
 	}
+	featureHeightCtx := protocol.MustGetFeatureWithHeightCtx(ctx)
 	height, _ := csm.SM().Height()
-	if p.needToWriteCandsMap(height) {
+	if p.needToWriteCandsMap(height) && !featureHeightCtx.FixCandCenterAlias(height) {
```
Liuhaai, line 773: same

action/protocol/staking/candidate_center.go
```diff
@@ -411,12 +426,13 @@ func (cc *candChange) delete(owner address.Address) {
 // candBase funcs
 //======================================
 
-func newCandBase() *candBase {
+func newCandBase(fixAlias bool) *candBase {
```
Liuhaai, line 429: This is a factory pattern apparently. No need to use `bool` to new a object.

action/protocol/staking/candidate_center.go
```diff
@@ -448,6 +464,11 @@ func (cb *candBase) commit(change *candChange) (int, error) {
 			return 0, err
 		}
 		d := v.Clone()
+		if curr, existing := cb.ownerMap[d.Owner.String()]; existing && cb.fixAlias {
+			// this is an existing candidate, remove it from name/operator map
```
Liuhaai, line 468: The code has explained what the comment says

action/protocol/staking/candidate_center_extra.go
```diff
@@ -6,10 +6,10 @@
 
 package staking
 
-func (cb *candBase) clone() *candBase {
+func (cb *candBase) clone(fixAlias bool) *candBase {
 	cb.lock.RLock()
 	defer cb.lock.RUnlock()
-	clone := newCandBase()
+	clone := newCandBase(fixAlias)
```
Liuhaai, line 12: use external `fixAlias`?

action/protocol/staking/candidate_center.go
dustinxie: can you be more clear? what is your suggested change?

Liuhaai: if fixAlias {
 c.base = newCandBaseLegacy()
} else {
 c.base = newCandBase()
}

action/protocol/staking/protocol.go
```diff
@@ -601,12 +602,14 @@ func (p *Protocol) settleAction(
 	return &r, nil
 }
 
-func (p *Protocol) needToReadCandsMap(height uint64) bool {
-	return height > p.config.PersistStakingPatchBlock
+func (p *Protocol) needToReadCandsMap(ctx context.Context, height uint64) bool {
```
Liuhaai, line 605: IMO, `needToReadCandsMap` and `needToWriteCandsMap` can be combined together due to the same code. But current implementation is fine.

Liuhaai (approved):

action/protocol/staking/protocol.go
dustinxie: resolved offline

action/protocol/staking/protocol.go
```diff
@@ -601,12 +602,14 @@ func (p *Protocol) settleAction(
...
+	fCtx := protocol.MustGetFeatureWithHeightCtx(ctx)
+	return height > p.config.PersistStakingPatchBlock && !fCtx.FixCandCenterAlias(height)
```
CoderZhi, line 607: why we have `PersistStakingPatchBlock` from p.config, but the other from `fCtx`?

dustinxie: `PersistStakingPatchBlock` is not a hard-fork height (so not in genesis), while `FixCandCenterAlias()` is according to the next hard-fork height in genesis

action/protocol/staking/protocol.go
```diff
@@ -458,7 +458,8 @@ func (p *Protocol) Validate(ctx context.Context, act action.Action, sr protocol.
 
 // ActiveCandidates returns all active candidates in candidate center
 func (p *Protocol) ActiveCandidates(ctx context.Context, sr protocol.StateReader, height uint64) (state.CandidateList, error) {
-	c, err := GetStakingStateReader(sr)
+	featureCtx := protocol.MustGetFeatureWithHeightCtx(ctx)
+	c, err := GetStakingStateReader(sr, featureCtx.FixCandCenterAlias(height))
```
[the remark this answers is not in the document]
dustinxie: it was finally called in https://github.com/iotexproject/iotex-core/blob/master/action/protocol/poll/util.go#L88
```
ds, err := p.CalculateCandidatesByHeight(ctx, sr, blkCtx.BlockHeight)
```
so the caller pass in the `height`

of course,  here`blkCtx.BlockHeight` is effectively equal to `sr.Height()`, so we could actually remove `height` from the input parameters, that can be an improvement/refactor later

action/protocol/staking/candidate_center.go
```diff
@@ -512,3 +554,34 @@ func (cb *candBase) delete(owner address.Address) {
...
+	for _, v := range change.dirty {
+		if err := v.Validate(); err != nil {
+			return 0, err
+		}
+		d := v.Clone()
+		if curr, existing := cb.ownerMap[d.Owner.String()]; existing {
+			// this is an existing candidate, remove it from name/operator map
+			delete(cb.nameMap, curr.Name)
+			delete(cb.operatorMap, curr.Operator.String())
+		}
```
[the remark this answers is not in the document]
dustinxie: discussed in weekly sync, both ways have pros and cons

action/protocol/staking/protocol.go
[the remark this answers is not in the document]
CoderZhi: https://github.com/iotexproject/iotex-core/blob/ba21de20289f770af8c2a33fb9f6cd3b3e04c003/action/protocol/poll/util.go#L132

action/protocol/staking/candidate_center.go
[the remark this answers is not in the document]
CoderZhi: over-engineering

action/protocol/staking/protocol.go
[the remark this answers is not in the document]
dustinxie: i still think we should use `height` here? to give an example, assuming:
```
HF height = 10000
sr.Height() = 10100
epochHeight = 9900
```
in this case, the func tries to read candidates at height 9900, which is before the HF. so we should not enable flags for HF=10000, is that right? 
or maybe I still miss sth, let me know if I'm wrong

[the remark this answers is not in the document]
Liuhaai: agree with changing from `height` to `sr.height`, because state from `GetStakingStateReader(sr)` is the state at current chain height.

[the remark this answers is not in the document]
dustinxie: @CoderZhi @Liuhaai  how do you think of the example I give?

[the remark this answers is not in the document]
dustinxie: i'm changing to set the flag = `true`, since it won't impact the results

but I still not convinced whether to use `sr.Height()` or `height`, so write down the below for record

say HF=10000, at this height, the bug will be fixed, with `newCandidateCenter(fixAlias = true)`. Hence for any read request with height < 10000, it should have`fixAlias = false`, and for any read request with height >= 10000, it should have `fixAlias = true`

in other words, for calls like:
https://github.com/iotexproject/iotex-core/blob/ba21de20289f770af8c2a33fb9f6cd3b3e04c003/action/protocol/poll/util.go#L132

input `epochHeight` is useless, it is overriden by `sr.Height()`, which defeats the intention of this function to calculate candidates at `epochHeight`. of course, this limitation is (in another way) due to we don't support archive mode for now, so `sr.Height()` is always the current height, but cannot be set to an earlier height

action/protocol/staking/protocol.go
```diff
@@ -458,7 +458,7 @@ func (p *Protocol) Validate(ctx context.Context, act action.Action, sr protocol.
 
 // ActiveCandidates returns all active candidates in candidate center
 func (p *Protocol) ActiveCandidates(ctx context.Context, sr protocol.StateReader, height uint64) (state.CandidateList, error) {
-	c, err := GetStakingStateReader(sr)
+	c, err := GetStakingStateReader(sr, true)
```
CoderZhi, line 461: For `StakingStateReader`, which sounds like a readonly component, why we need the extra parameter?

action/protocol/staking/candidate_center.go
```diff
@@ -52,16 +87,23 @@ func listToCandChange(l CandidateList) (*candChange, error) {
 }
 
 // NewCandidateCenter creates an instance of CandidateCenter
-func NewCandidateCenter(all CandidateList) (*CandidateCenter, error) {
+func NewCandidateCenter(all CandidateList, opts ...CandidateCenterOption) (*CandidateCenter, error) {
```
CoderZhi, line 90: for candidate center, which may be constructed from https://github.com/iotexproject/iotex-core/blob/master/action/protocol/staking/candidate_statemanager.go#L84, how could the current code return a CandidateCenter with the right parameter?

dustinxie: this will be done in PR3688

action/protocol/staking/protocol.go
dustinxie: in case the view does not exist, it will need to call `CreateBaseView`, which will call `NewCandidateCenter`, which requires the flag

dustinxie: `GetStakingStateReader` calls `CreateBaseView`, which requires this flag

dustinxie: `GetStakingStateReader ` has been removed in latest master

PR #3846, opened 2023-04-17
add private key detecting in pre-commit hook

edwardTianY:
@dustinxie Could u pls help to review the config for the precommit hook for private-key detecting, thx.


edwardTianY:
Change the two commiters to one.

PR #3906, opened 2023-07-26
[blockindex] introduce indexergroup

envestcc:
# Description
Currently, when the node starts up, each `block indexer` is checked separately, meaning that one indexer will only be checked when it reaches the tip height before another indexer is checked. However, the `contract staking indexer` and `factory indexer` are not completely independent, and checking them separately may cause block validation failures. Therefore, these two indexers need to be checked synchronously.

To address this, we introduce the `IndexerGroup`, which is a special indexer that can contain other indexers. Indexers added to the IndexerGroup can support synchronous checking.

Fixes #(issue)
 
## Type of change

# How Has This Been Tested?
- [x] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:


Liuhaai:
> Currently, when the node starts up, each block indexer is checked separately, meaning that one indexer will only be checked when it reaches the tip height before another indexer is checked.

Should the issue be fixed here instead?
https://github.com/iotexproject/iotex-core/blob/v1.11.0/blockchain/blockdao/blockdao.go#L95-L115

blockchain/blockdao/blockindexer.go
```diff
@@ -32,7 +32,7 @@ type (
 	BlockIndexerWithStart interface {
 		BlockIndexer
 		// StartHeight returns the start height of the indexer
-		StartHeight() uint64
```
dustinxie, line 35: we don't need this?

blockindex/indexer_group.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"context"
+
+	"github.com/iotexproject/iotex-core/blockchain/block"
+	"github.com/iotexproject/iotex-core/blockchain/blockdao"
+)
+
+// IndexerGroup is a special index that includes multiple indexes,
+// which stay in sync when blocks are added.
+type IndexerGroup struct {
```
dustinxie, line 17: name it `SynchronousIndexers`?

blockindex/indexer_group.go
```diff
@@ -0,0 +1,123 @@
...
+	indexers []blockdao.BlockIndexer
```
dustinxie, line 18: add a `startHeights []uint64`

blockindex/indexer_group.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+	return nil
+}
+
+// PutBlock puts a block into the indexers in the group
+func (ig *IndexerGroup) PutBlock(ctx context.Context, blk *block.Block) error {
+	for _, indexer := range ig.indexers {
+		// check if the indexer is a BlockIndexerWithStart
+		if indexerWithStart, ok := indexer.(blockdao.BlockIndexerWithStart); ok {
```
dustinxie, line 50: use `startHeight[i]`, so no need to do the type insertion every time

blockindex/indexer_group.go
```diff
@@ -0,0 +1,123 @@
...
+// StartHeight returns the minimum start height of the indexers in the group
+func (ig *IndexerGroup) StartHeight() (uint64, error) {
+	var result uint64
+	for i, indexer := range ig.indexers {
+		tipHeight, err := indexer.Height()
+		if err != nil {
+			return 0, err
+		}
+		indexStartHeight := tipHeight + 1
+		if indexerWithStart, ok := indexer.(blockdao.BlockIndexerWithStart); ok {
```
dustinxie, line 94: use `startHeight[i]`, see above comment

blockindex/indexer_group.go
```diff
@@ -0,0 +1,123 @@
...
+			if blk.Height() < startHeight {
+				continue
+			}
+		}
+		// check if the block is higher than the indexer's height
+		height, err := indexer.Height()
+		if err != nil {
+			return err
+		}
+		if blk.Height() <= height {
```
dustinxie, line 64: should be `if blk.Height() != height+1`?

blockindex/indexer_group.go
```diff
@@ -0,0 +1,123 @@
...
+			}
+		}
+		if i == 0 || indexStartHeight < result {
+			result = indexStartHeight
+		}
+	}
+	return result, nil
+}
+
+// Height returns the minimum height of the indexers in the group
```
dustinxie, line 110: it seems we should return minimum
pls double-check the logic in `func (bic *BlockIndexerChecker) CheckIndexer()` carefully, see if minimum is logically correct
sgdIndexer will also be added into the group, consider the case of 3 indexers in the group

chainservice/builder.go
```diff
@@ -257,7 +257,11 @@ func (builder *Builder) buildBlockDAO(forTest bool) error {
 	}
 
 	var indexers []blockdao.BlockIndexer
-	indexers = append(indexers, builder.cs.factory)
+	if builder.cs.contractStakingIndexer != nil {
+		indexers = append(indexers, blockindex.NewIndexerGroup(builder.cs.factory, builder.cs.contractStakingIndexer))
```
dustinxie, line 261: also need to add `sgdIndexer`

envestcc:
> Should the issue be fixed here instead? https://github.com/iotexproject/iotex-core/blob/v1.11.0/blockchain/blockdao/blockdao.go#L95-L115 loop at each height not each indexer.

Yeah, it's another workable solution. However, using `IndexerGroup` is a more simple way to reduce modifications to existing code.

blockindex/indexer_group.go
envestcc: Here is benchmark for type assert:
```go
func BenchmarkTypeAssert(b *testing.B) {
	ctrl := gomock.NewController(b)
	var indexer interface{}
	indexer = mock_blockdao.NewMockBlockIndexer(ctrl)
	b.ResetTimer()

	b.Run("type assert", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			_, _ = indexer.(blockdao.BlockIndexerWithStart)
		}
	})
}
```

And the result:
```
❯ go test -benchmem -run=^$ -bench ^BenchmarkTypeAssert$ github.com/iotexproject/iotex-core/blockindex
goos: darwin
goarch: arm64
pkg: github.com/iotexproject/iotex-core/blockindex
BenchmarkTypeAssert/type_assert-10              344110524                3.417 ns/op           0 B/op          0 allocs/op
PASS
ok      github.com/iotexproject/iotex-core/blockindex   1.870s
```

I think the `3.4ns` is acceptable for function calls, so there is no need to introduce more complex logic to optimize at this point

blockindex/indexer_group.go
envestcc: it's not work for `indexerWithStartHeight`

blockindex/indexer_group.go
dustinxie: 1. the benchmark does not have `if`, also `_, _` means the result is not used, go compiler may optimize to not call the right side at all (but I am not sure about that for bench-marking code)
2. it's not all about performance. When a group is created, the start heights are known at creation time. If something can be done just one-time at creation, why keep doing it in a `for` loop for thousands of times ... To me the current way is complex, b/c there is a much simpler/clean way of just doing it one-time

blockindex/indexer_group.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// IndexerGroup is a special index that includes multiple indexes,
+// which stay in sync when blocks are added.
+type IndexerGroup struct {
+	indexers []blockdao.BlockIndexer
+}
+
+// NewIndexerGroup creates a new indexer group
+func NewIndexerGroup(indexers ...blockdao.BlockIndexer) *IndexerGroup {
```
dustinxie, line 22: `NewIndexerGroup(factory blockdao.BlockIndexer, indexers ...blockdao.BlockIndexer)`

blockindex/indexer_group.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}
+	return nil
+}
+
+// PutBlock puts a block into the indexers in the group
+func (ig *IndexerGroup) PutBlock(ctx context.Context, blk *block.Block) error {
+	for i, indexer := range ig.indexers {
+		// check if the block is higher than the indexer's start height
+		if blk.Height() < ig.startHeights[i] {
```
dustinxie, line 52: should be `<=`

chainservice/builder.go
```diff
@@ -257,19 +257,25 @@ func (builder *Builder) buildBlockDAO(forTest bool) error {
 	}
 
 	var indexers []blockdao.BlockIndexer
-	indexers = append(indexers, builder.cs.factory)
+	// indexers in synchronizedIndexers will be putblock one by one for every block
```
dustinxie, line 260: will need to run `PutBlock()` one by one

blockindex/indexer_group.go
dustinxie: name it `SyncIndexGroup`?

Liuhaai:
> > Should the issue be fixed here instead? https://github.com/iotexproject/iotex-core/blob/v1.11.0/blockchain/blockdao/blockdao.go#L95-L115 loop at each height not each indexer.
> 
> Yeah, it's another workable solution. However, using `IndexerGroup` is a more simple way to reduce modifications to existing code.

https://github.com/iotexproject/iotex-core/pull/3912/commits/d6331c6c6a8a3e5e571157f70b6ef9fbafbe0898

dustinxie (approved):

envestcc:
> Around 30 lines of change should have done the work [d6331c6](https://github.com/iotexproject/iotex-core/commit/d6331c6c6a8a3e5e571157f70b6ef9fbafbe0898)

是的，单纯从修改的代码量来说，这样更少。但是这个改法有几个不足：

millken (approved):

PR #3965, opened 2023-11-09
[action] support pre-EIP155 unprotected transaction

dustinxie:
# Description
Currently, our chain only supports Ether-compatible tx with EIP-155 signature. This PR adds support for old unprotected tx (before EIP-155)

For more details, see https://iotex.larksuite.com/wiki/QN4RwY6fOilXrWk08VHuj9uGs5d

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [x] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/iotexproject/iotex-address v0.2.8
 	github.com/iotexproject/iotex-antenna-go/v2 v2.5.1
 	github.com/iotexproject/iotex-election v0.3.5-0.20210611041425-20ddf674363d
-	github.com/iotexproject/iotex-proto v0.5.14
+	github.com/iotexproject/iotex-proto v0.5.15-0.20231108184801-bc89358f2344
```
dustinxie, line 26: will change to tag later

action/sealedenvelope.go
```diff
@@ -39,14 +41,31 @@ func (sealed *SealedEnvelope) envelopeHash() (hash.Hash256, error) {
...
 
+func EthSigner(txType iotextypes.Encoding, chainID uint32) (types.Signer, error) {
+	switch txType {
+	case iotextypes.Encoding_ETHEREUM_EIP155:
+		return types.NewEIP155Signer(big.NewInt(int64(chainID))), nil
+	case iotextypes.Encoding_ETHEREUM_UNPROTECTED:
+		return types.HomesteadSigner{}, nil
+	case iotextypes.Encoding_ETHEREUM_ACCESSLIST:
+		return types.NewEIP2930Signer(big.NewInt(int64(chainID))), nil
+	default:
```
dustinxie, line 64: use proper signer according to Eth-compatible tx type

action/rlp_tx.go
```diff
@@ -83,3 +82,68 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
...
+		encoding iotextypes.Encoding
+		sig      []byte
+		pubkey   crypto.PublicKey
+	)
+	//remove Hex prefix and decode string to byte
+	rawData = strings.Replace(rawData, "0x", "", -1)
+	rawData = strings.Replace(rawData, "0X", "", -1)
+	rawTxBytes, err := hex.DecodeString(rawData)
+	if err != nil {
+		return &tx, encoding, sig, pubkey, err
```
CoderZhi, line 99: `return nil, encoding, sig, pubkey, err`

action/rlp_tx.go
```diff
@@ -83,3 +82,68 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
...
+	}
+
+	// decode raw data into rlp tx
+	if err = tx.UnmarshalBinary(rawTxBytes); err != nil {
+		return &tx, encoding, sig, pubkey, err
```
CoderZhi, line 104: same here

action/rlp_tx.go
```diff
@@ -83,3 +82,68 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
...
+	}
+
+	// extract correct V value
+	var (
+		signer  types.Signer
+		V, R, S = tx.RawSignatureValues()
```
CoderZhi, line 110: v, r, s

action/rlp_tx.go
```diff
@@ -83,3 +82,68 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
...
+			signer = types.HomesteadSigner{}
+		}
+	case types.AccessListTxType:
+		// AL txs are defined to use 0 and 1 as their recovery
+		// id, add 27 to become equivalent to unprotected Homestead signatures.
+		V = new(big.Int).Add(V, big.NewInt(27))
+		encoding = iotextypes.Encoding_ETHEREUM_ACCESSLIST
+		signer = types.NewEIP2930Signer(tx.ChainId())
+	default:
+		return &tx, encoding, sig, pubkey, ErrNotSupported
```
CoderZhi, line 132: nil

action/rlp_tx.go
```diff
@@ -83,3 +82,68 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
...
+	}
+
+	// construct signature
+	if V.BitLen() > 8 {
+		return &tx, encoding, sig, pubkey, ErrNotSupported
```
CoderZhi, line 137: nil

api/web3server.go
```diff
@@ -433,7 +435,24 @@ func (svr *web3Handler) sendRawTransaction(in *gjson.Result) (interface{}, error
...
+		tx       *types.Transaction
+		sig      []byte
+		pubkey   crypto.PublicKey
+		err      error
+		encoding iotextypes.Encoding
+		cs       = svr.coreService
+	)
+	if g := cs.Genesis(); g.IsSumatra(cs.TipHeight()) {
+		tx, encoding, sig, pubkey, err = action.DecodeEtherTx(dataStr.String())
+		if !tx.Protected() && tx.ChainId().Uint64() != uint64(cs.EVMNetworkID()) {
```
CoderZhi, line 448: If `err != nil`, what will happen here? tx may not be initialized properly, and we shouldn't have any hidden assumption of the default value of tx

action/rlp_tx.go
```diff
@@ -83,3 +82,68 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
 	pubkey, err = crypto.RecoverPubkey(rawHash[:], sig)
 	return
 }
+
+// DecodeEtherTx decodes raw data string into eth tx
+func DecodeEtherTx(rawData string) (*types.Transaction, iotextypes.Encoding, []byte, crypto.PublicKey, error) {
```
CoderZhi, line 87: we may decouple the generation of `tx` and `encoding, sig, publickey` into `func DecodeEthTx(rawData string) (*types.Transaction, error)` and `func ExtractXXX(tx) (iotextypes.Encoding, []byte, crypto.PublicKey, error)`
```

File: action/rlp_tx_test.go [hunk 10]
```diff
@@ -390,18 +395,36 @@ func TestIssue3944(t *testing.T) {
...
 	r.Equal(sig[:32], q.Bytes())
 	r.Equal(sig[32:64], s.Bytes())
 	r.Equal("9415", v.String()) // this is the correct V value corresponding to chainID = 4690
 	r.Equal(hash, tx1.Hash().Hex())
 }
+
+func TestDetermine(t *testing.T) {
+	r := require.New(t)
+
+	tx, encoding, sig, pubkey, err := DecodeEtherTx("0xf8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222")
```
CoderZhi, line 422: unit test of `DecodeEtherTx` never covers the case of `err != nil`

api/web3server.go
```diff
@@ -433,7 +435,24 @@ func (svr *web3Handler) sendRawTransaction(in *gjson.Result) (interface{}, error
...
+	var (
+		tx       *types.Transaction
+		sig      []byte
+		pubkey   crypto.PublicKey
+		err      error
+		encoding iotextypes.Encoding
+		cs       = svr.coreService
+	)
+	if g := cs.Genesis(); g.IsSumatra(cs.TipHeight()) {
+		tx, encoding, sig, pubkey, err = action.DecodeEtherTx(dataStr.String())
```
CoderZhi, line 447: Don't see the value of wrapping the logic in a public function in package `action`, which is only called here.

action/sealedenvelope.go
```diff
@@ -39,14 +41,31 @@ func (sealed *SealedEnvelope) envelopeHash() (hash.Hash256, error) {
...
+		}
+		return rlpRawHash(tx, signer)
 	case iotextypes.Encoding_IOTEX_PROTOBUF:
 		return hash.Hash256b(byteutil.Must(proto.Marshal(sealed.Envelope.Proto()))), nil
 	default:
 		return hash.ZeroHash256, errors.Errorf("unknown encoding type %v", sealed.encoding)
 	}
 }
 
+func EthSigner(txType iotextypes.Encoding, chainID uint32) (types.Signer, error) {
```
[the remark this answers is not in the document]
dustinxie: fixed

action/rlp_tx.go
```diff
@@ -83,3 +82,68 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
...
+	var (
+		tx       types.Transaction
+		encoding iotextypes.Encoding
+		sig      []byte
+		pubkey   crypto.PublicKey
+	)
+	//remove Hex prefix and decode string to byte
+	rawData = strings.Replace(rawData, "0x", "", -1)
+	rawData = strings.Replace(rawData, "0X", "", -1)
```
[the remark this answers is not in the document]
dustinxie: need to remove `0x` at front, changed to simple code

action/rlp_tx.go
dustinxie: why want to return `nil` for tx specifically? we are returning the default value of these types here, and there's `err` to be checked

action/rlp_tx.go
dustinxie: this is internal var so should be OK? and below we have
```
r, s := R.Bytes(), S.Bytes()
copy(sig[32-len(r):32], r)
copy(sig[64-len(s):64], s)
```

api/web3server.go
dustinxie: good catch, fixed

action/rlp_tx.go
dustinxie: we could, but just as you commented below, it is only called 1 time in `api` package. If splitting to 2 parts can make them re-use more than 1 times, then it's definitely worth it

api/web3server.go
dustinxie: 2 reasons:
1. the current code is `DecodeRawTx` (which only works for EIP-155), so put this func together with current code
2. this func is to deal with Eth-compatible tx decoding/conversion, so it makes more sense to put it (and all related codes) in the `action` package. I think it's better and clear to have all these codes in 1 place

api/web3server.go
dustinxie: common `err` is checked at L456 below, but you are right, logically should check `err` first

action/rlp_tx_test.go
dustinxie: right, I am working to add more tests

action/rlp_tx.go
CoderZhi: If there was an error, the return value of non-nil `tx` just brings up some confusion to the caller side

api/web3server_utils.go
```diff
@@ -117,9 +118,28 @@ func (svr *web3Handler) getTransactionFromActionInfo(blkHash hash.Hash256, selp
...
+			return nil, err
+		}
+	} else {
+		if sig[64] >= 27 {
+			sig[64] -= 27
+		}
+		tx, err = ethTx.WithSignature(types.HomesteadSigner{}, sig)
+		if err != nil {
+			return nil, err
+		}
```
CoderZhi, line 142:
```
    if encoding != iotextypes.Encoding_IOTEX_PROTOBUF {
        signer, err = ...
    } else {
        signer = types.Homestead...
    }
    tx, err := action.RawTxToSignedTx(ethTx, signer, sig)
    if err != nil {
        return nil, err
    }
```
logic like `sig[64] -= 27` is something we want to hide in some unified util function. util function is help us to keep our logic clean.

action/rlp_tx.go
CoderZhi: `DecodeEtherTx` is not clearly describe what it is doing. Legacy problem cannot be used as excuse to latest implementation especially that the legacy function is not used.

Liuhaai:
one question: does this pr depend on  https://github.com/iotexproject/iotex-core/pull/3967 ?

api/web3server.go
CoderZhi: your modification here answers your own question:
```just curious, why want to return nil for tx specifically? we are returning the default value of these types here, and there's err to be checked```
so, let's return `nil` inside of that function, and also split it into two functions pls

File: action/rlp_tx.go (hunk 3 above)
Comment (dustinxie) [reply action/rlp_tx.go:114]: fixed

File: action/rlp_tx.go (hunk 7 above)
Comment (dustinxie) [reply action/rlp_tx.go:152]: fixed

File: action/rlp_tx.go (hunk 6 above)
Comment (dustinxie) [reply action/rlp_tx.go:147]: fixed

File: action/rlp_tx.go (hunk 9 above)
Comment (dustinxie) [reply action/rlp_tx.go:101]: updated in latest commit

File: action/rlp_tx.go [hunk 15]
```diff
@@ -87,14 +85,80 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
...
+	// decode raw data into eth tx
+	tx := types.Transaction{}
+	if err = tx.UnmarshalBinary(rawTxBytes); err != nil {
+		return nil, err
+	}
+	return &tx, nil
+}
+
+// ExtractTypeSigPubkey extracts tx type, signature, and pubkey
+func ExtractTypeSigPubkey(tx *types.Transaction) (iotextypes.Encoding, []byte, crypto.PublicKey, error) {
```
dustinxie, line 121: split to 2 funcs for clarity

action/rlp_tx_test.go
```diff
@@ -393,19 +714,33 @@ func TestIssue3944(t *testing.T) {
...
+	r.True(tx1.Protected())
+	r.EqualValues(4690, tx1.ChainId().Uint64())
 	v, q, s = tx1.RawSignatureValues()
 	r.Equal(sig[:32], q.Bytes())
 	r.Equal(sig[32:64], s.Bytes())
 	r.Equal("9415", v.String()) // this is the correct V value corresponding to chainID = 4690
 	r.Equal(hash, tx1.Hash().Hex())
 }
+
+func TestBackwardComp(t *testing.T) {
```
dustinxie, line 738: test to verify new unprotected tx will cause error using current code

action/rlp_tx_test.go
```diff
@@ -393,19 +714,33 @@ func TestIssue3944(t *testing.T) {
...
+	r := require.New(t)
+	// example unprotected tx at https://etherscan.io/tx/0xeddf9e61fb9d8f5111840daef55e5fde0041f5702856532cdbb5a02998033d26
+	unprotectedTx := "0xf8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222"
+	for _, chainID := range []uint32{_evmNetworkID, _evmNetworkID + 1, 0} {
+		_, _, _, err := DecodeRawTx(unprotectedTx, chainID)
+		r.Equal(crypto.ErrInvalidKey, err)
```
dustinxie, line 744: test to verify new unprotected tx will cause error using current code

api/web3server_utils.go
dustinxie: fixed

dustinxie:
> one question: does this pr depend on #3967 ?

yes, this PR has now rebased to 3967, pls check again

CoderZhi (approved):

api/web3server.go
```diff
@@ -433,9 +435,33 @@ func (svr *web3Handler) sendRawTransaction(in *gjson.Result) (interface{}, error
...
+	var (
+		cs       = svr.coreService
+		tx       *types.Transaction
+		encoding iotextypes.Encoding
+		sig      []byte
+		pubkey   crypto.PublicKey
+		err      error
+	)
+	if g := cs.Genesis(); g.IsSumatra(cs.TipHeight()) {
+		tx, err = action.DecodeEtherTx(dataStr.String())
```
dustinxie, line 447: use new code path after hard-fork

action/rlp_tx.go
```diff
@@ -87,14 +85,80 @@ func DecodeRawTx(rawData string, chainID uint32) (tx *types.Transaction, sig []b
 
 // NewEthSigner returns the proper signer for Eth-compatible tx
 func NewEthSigner(txType iotextypes.Encoding, chainID uint32) (types.Signer, error) {
-	// TODO: use proper signer according to tx type
 	switch txType {
 	case iotextypes.Encoding_IOTEX_PROTOBUF:
 		// native tx use same signature format as that of Homestead
 		return types.HomesteadSigner{}, nil
-	case iotextypes.Encoding_ETHEREUM_RLP:
+	case iotextypes.Encoding_ETHEREUM_EIP155:
```
Liuhaai, line 92: iotextypes.Encoding_ETHEREUM_POST_EIP155
iotextypes.Encoding_ETHEREUM_PRE_EIP155?

action/rlp_tx_test.go
```diff
@@ -344,14 +348,331 @@ func TestRlpDecodeVerify(t *testing.T) {
...
+			"0",
+			"0xA576C141e5659137ddDa4223d209d4744b2106BE",
+			_evmNetworkID,
+			iotextypes.Encoding_ETHEREUM_EIP155,
+			100,
+			"842fea9a292c0bc7a1e05a14a8255ed8dc945fdae7c01a6b70f5213ebe35583b",
+			"04830579b50e01602c2015c24e72fbc48bca1cca1e601b119ca73abe2e0b5bd61fcb7874567e091030d6b644f927445d80e00b3f9ca0c566c21c30615e94c343da",
+			"8d38efe45794d7fceea10b2262c23c12245959db",
+		},
+		{
```
Liuhaai, line 579: what error if integrated with `TestRlpDecodeVerify``

api/coreservice.go
```diff
@@ -1589,6 +1591,11 @@ func (core *coreService) ActionsInActPool(actHashes []string) ([]action.SealedEn
 	return ret, nil
 }
 
+// Genesis returns the genesis of the chain
+func (core *coreService) Genesis() genesis.Genesis {
+	return core.bc.Genesis()
```
Liuhaai, line 1596: coreService is the adapter for other services, where Genesis() shouldn't be exposed and tx decoding should be done

action/rlp_tx.go
dustinxie: `Encoding_ETHEREUM_UNPROTECTED` is a better name/term, Ethereum call it unprotected tx, and there is `Protected() bool` func

dustinxie: there won't be backwards compatibility issue, _EIP155 is an alias of _RLP
```
Encoding_ETHEREUM_EIP155 == Encoding_ETHEREUM_RLP == 1
```

action/rlp_tx_test.go
dustinxie: see `func TestBackwardComp(t *testing.T)` below, it verifies error will occur running the new tx with current code

Liuhaai: ok, could testdata be shared in `Var` ?

api/coreservice.go
Liuhaai: there are some grpc-only interfaces in coreservice as well

action/rlp_tx_test.go
dustinxie: updated in latest commit

api/coreservice.go
dustinxie: ok, that is not the reason to add another one ... here are my 3 thoughts when adding this to coreService:
1. there are existing similar funcs like `ChainID()`, `EvmNetworkID()`, `TipHeight()`, these actually all belong to the `Blockchain()` interface, so I don't see why cannot add a new `Genesis()`. We should have another PR to add a  `Blockchain()` funcs to deprecate these
2. like you said, we can add a `DecodeTx()` funcs instead of `Genesis()`, but given the facts in 1, doing so does not make our programming better IMO. The `DecodeTx()` is just a wrapper of some calls, it's not simpler or better readable than `Genesis()`
3. finally, after the hard-fork, we will switch to use the new code path, this `Genesis()` func won't be needed anymore

Liuhaai (approved):

PR #4041, opened 2024-01-04
[db] add KvVersioned interface

dustinxie:
# Description
Currently the staking index DB saves all delegates and candidates at each epoch, the DB size grows very rapidly due to large size of data stored.

This PR introduces a new DB which can save an item at different version (blockchain's height) only when the value changes.  In a quick POC testing, this has significantly reduced the DB size from ~80GB to ~4GB

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:


db/db_versioned.go
```diff
@@ -0,0 +1,151 @@
...
+func (b *BoltDBVersioned) Version(ns string, key []byte) (uint64, error) {
+	if !b.IsReady() {
+		return 0, ErrDBNotStarted
+	}
+	// TODO: implement Version
+	return 0, nil
+}
+
+// SetVersion sets the version which a Get()/Put() wants to access
+func (b *BoltDBVersioned) SetVersion(v uint64) KvVersioned {
```
millken, line 139: the v means `blockHeight` in production?

db/db_versioned.go
```diff
@@ -0,0 +1,151 @@
...
+	b.version = v
+	return b
+}
+
+// CreateVersionedNamespace creates a namespace to store versioned keys
+func (b *BoltDBVersioned) CreateVersionedNamespace(ns string, n uint32) error {
```
millken, line 145: `n` what means?

db/db_versioned.go
```diff
@@ -0,0 +1,151 @@
...
+
+		// CreateVersionedNamespace creates a namespace to store versioned keys
+		CreateVersionedNamespace(string, uint32) error
+	}
+
+	// BoltDBVersioned is KvVersioned implementation based on bolt DB
+	BoltDBVersioned struct {
+		*BoltDB
+		version      uint64          // version for Get/Put()
+		versioned    map[string]int  // buckets for versioned keys
```
millken, line 56: If there are many keys, will these two maps be very large?

db/db_versioned.go
```diff
@@ -0,0 +1,151 @@
...
+	// 2. the version when the key is lastly written
+	// 3. the version when the key is deleted
+	// 4. hash of the key's last written value
+	// If the location does not store a value, the key has never been written.
+	//
+	// Here's an example of a versioned DB which has 3 buckets:
+	// 1. "mta" -- regular bucket storing metadata, key is not versioned
+	// 2. "act" -- versioned namespace, key length = 20
+	// 3. "stk" -- versioned namespace, key length = 8
+	KvVersioned interface {
```
envestcc, line 39: I prefer the following interface:
```go
KvVersioned interface{
    Version(ns string, key []byte) (ver uint64, err error)
    WithVersion(ver uint64) KVStore
    CreateNamespace(ns string, keySize uint32) error
}
```
Simplify the definition of `KvVersioned`, it does not support operations such as put/get itself, and must specify the version before proceeding. If need an interface that supports both versioned and non-versioned operations, can define another one.

envestcc: If this is an interface that supports both versioned and non-versioned operations at the same time, a method `IsVersionedNamespaces(ns string)` should also be added, so that when the interface is used, it is known whether to call `Put()` or `SetVersion().Put()`.

db/db_versioned.go
envestcc: i think the map key should be namespace instead of key, is it right?

dustinxie: yes, the map key is the namespace

db/db_versioned.go
dustinxie: yes, that's correct

db/db_versioned.go
dustinxie: n is the size of key in this namespace

db/db_versioned.go
dustinxie: good suggestions, updated PR
yes, `Put()` is a common interface for both versioned and non-versioned, the code will check inside and process accordingly

db/db_versioned.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Delete deletes a record,if key is nil,this will delete the whole bucket
+func (b *BoltDBVersioned) Delete(ns string, key []byte) error {
+	if !b.IsReady() {
+		return ErrDBNotStarted
+	}
+	// TODO: implement Delete
+	return nil
+}
```
envestcc, line 88: Do we still need `Put`, `Get`, `Delete` the three methods?

dustinxie: yes, that is the core function of this `KvVersioned` concept, and will be implemented in later PRs

db/db_versioned.go
```diff
@@ -0,0 +1,158 @@
...
+func (b *BoltDBVersioned) Get(ns string, version uint64, key []byte) ([]byte, error) {
+	if !b.db.IsReady() {
+		return nil, ErrDBNotStarted
+	}
+	// TODO: implement Get
+	return nil, nil
+}
+
+// Delete deletes a record,if key is nil,this will delete the whole bucket
+func (b *BoltDBVersioned) Delete(ns string, key []byte) error {
```
envestcc, line 113: version param missing

db/db_versioned.go
```diff
@@ -0,0 +1,158 @@
...
+	//
+	// kv := db.KvStore() // creates a KvStore instance
+	// value, err := kv.SetVersion(5).Get("ns", key) // read 'key' at version 5
+	// err = kv.SetVersion(8).Put("ns", key, value) // write 'key' at version 8
+
+	DbVersioned interface {
+		// Version returns the key's most recent version
+		Version(string, []byte) (uint64, error)
+		// KvStore returns a handler to perform versioned read/write
+		KvStore() KvVersioned
```
envestcc, line 43: KvStore(version uint64) KvVersioned

db/db_versioned.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	KvVersioned interface {
```
envestcc, line 46: it's useless

Liuhaai:
pls add more details as context in the description

envestcc (approved):

db/versionpb/version.proto
```diff
@@ -0,0 +1,23 @@
...
+syntax = "proto3";
+package versionpb;
+option go_package = "github.com/iotexproject/iotex-core/db/versionpb";
+
+message VersionedNamespace {
+    string name = 1;
+    uint32 keyLen = 2;
+}
+
+message KeyMeta {
```
Liuhaai, line 18: how is this used

Liuhaai:
How to retrieve value at certain height?

db/db_versioned.go
```diff
@@ -0,0 +1,156 @@
...
+func (b *BoltDBVersioned) Stop(ctx context.Context) error {
+	return b.db.Stop(ctx)
+}
+
+// Put writes a <key, value> record
+func (b *BoltDBVersioned) Put(ns string, version uint64, key, value []byte) error {
+	if !b.db.IsReady() {
+		return ErrDBNotStarted
+	}
+	// TODO: implement Put
```
Liuhaai, line 88: add implementation in this pr if no other dependency?

Liuhaai:
From my understand, the idea of pr is to build a mapping for deduplicated data. Changing the implementation of KVStore is more reasonable than changing boltdb itself

db/versionpb/version.proto
dustinxie: will be used in future PR

dustinxie:
> From my understand, the idea of pr is to build a mapping for deduplicated data. Changing the implementation of KVStore is more reasonable than changing boltdb itself, where does simple kv storage jobs.

it's not changing boltdb, rather it's using boltdb's underlying API to construct a KVStore that can return multiple versions of value for the same key

millken (approved):

PR #4152, opened 2024-02-28
[db] Reset only if Snapshots exceed size of 10

envestcc:
# Description

## Backgroud
`ResetSnapshots` occupies almost half of the time of `MintBlock` under high TPS, so we need to optimize it. 

## What to do
Currently, `ResetSnapshot` is called after each action is processed, but in most cases, the number of snapshots is only 1. In this case, performing `ResetSnapshot` is not very meaningful, so we have changed it to only reset when the snapshot count exceeds 10. 

Why choose 10? This threshold will lead to too many calls of `ResetSnapshot` if it's too small, resulting in waste; a threshold that is too large may cause a significant impact on the kv read/write performance due to too many snapshots. When the threshold is set to 10, the time of `ResetSnapshots` decreases from 1s to 100ms, and the MintBlock time decreases from 2.2s to 1.3s, which is acceptable

## Type of change

# How Has This Been Tested?
- [x] make test

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:

### DIFF
--- a/db/batch/batch_impl.go
+++ b/db/batch/batch_impl.go
@@ -11,6 +11,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// resetSnapshotIgnoreThreshold is the threshold of snapshot number to ignore reset snapshots
+	resetSnapshotIgnoreThreshold = 10
+)
+
 type (
 	// baseKVStoreBatch is the base implementation of KVStoreBatch
 	baseKVStoreBatch struct {
@@ -359,6 +364,10 @@ func (cb *cachedBatch) ResetSnapshots() {
 	cb.lock.Lock()
 	defer cb.lock.Unlock()
 
+	// ignore reset snapshots if the snapshot number is less than threshold
+	if cb.tag <= resetSnapshotIgnoreThreshold {
+		return
+	}
 	cb.tag = 0
 	cb.batchShots = nil
 	cb.batchShots = make([]int, 0)


CoderZhi (approved):

db/batch/batch_impl.go
```diff
@@ -11,6 +11,11 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// resetSnapshotIgnoreThreshold is the threshold of snapshot number to ignore reset snapshots
+	resetSnapshotIgnoreThreshold = 10
```
dustinxie, line 16: `_resetSnapshotIgnoreThreshold`

dustinxie (approved):

db/batch/batch_impl.go
envestcc: In go, the convention should be to start the name of a private variable with a lowercase letter.

millken (approved):

Liuhaai (changes requested):

db/batch/batch_impl.go
```diff
@@ -359,6 +364,10 @@ func (cb *cachedBatch) ResetSnapshots() {
 	cb.lock.Lock()
 	defer cb.lock.Unlock()
 
+	// ignore reset snapshots if the snapshot number is less than threshold
+	if cb.tag <= resetSnapshotIgnoreThreshold {
+		return
```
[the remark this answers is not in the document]
CoderZhi: Could you elaborate it?
My understanding: snapshot id is just an id. Inside of evm, it will be called as follows:
```
  id := snapshot()
  some other code
  id2 := snapshot()
  some other code
  if need to revert id
      revertSnapshot(id)
  
```
Thus, the real value of the `id` doesn't matter.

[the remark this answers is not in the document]
Liuhaai: I have checked that the SerializeQueue() function wouldn't be impacted by the improvement, so no hard fork is needed. However, it would be more straightforward if there were tests to ensure that the improvement does not impact the function.

Liuhaai (approved):

db/batch/batch_impl.go
dustinxie: in our codebase, lots of private constants name start with _

PR #4171, opened 2024-03-07
[evm] add transaction log for Suicide()

dustinxie:
# Description
When `Suicide()` is called upon destruction, the contract's balance will be transferred to a beneficiary. This needs to be captured in a transaction log  

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [x] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -118,6 +118,7 @@ type (
 		SharedGasWithDapp                       bool
 		CandidateRegisterMustWithStake          bool
 		DisableDelegateEndorsement              bool
+		SuicideTxLogMismatchPanic               bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
@@ -261,6 +262,7 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			SharedGasWithDapp:                       g.IsToBeEnabled(height),
 			CandidateRegisterMustWithStake:          !g.IsToBeEnabled(height),
 			DisableDelegateEndorsement:              !g.IsToBeEnabled(height),
+			SuicideTxLogMismatchPanic:               g.IsToBeEnabled(height),
 		},
 	)
 }
--- a/action/protocol/execution/evm/evmstatedbadapter.go
+++ b/action/protocol/execution/evm/evmstatedbadapter.go
@@ -70,6 +70,7 @@ type (
 		fixSnapshotOrder           bool
 		revertLog                  bool
 		manualCorrectGasRefund     bool
+		suicideTxLogMismatchPanic  bool
 	}
 )
 
@@ -139,6 +140,14 @@ func ManualCorrectGasRefundOption() StateDBAdapterOption {
 		// has caused gas refund to change, which needs to be manually adjusted after
 		// the tx is reverted. After Okhotsk height, it is fixed inside RevertToSnapshot()
 		adapter.manualCorrectGasRefund = true
+		return nil
+	}
+}
+
+// SuicideTxLogMismatchPanicOption set suicideTxLogMismatchPanic as true
+func SuicideTxLogMismatchPanicOption() StateDBAdapterOption {
+	return func(adapter *StateDBAdapter) error {
+		adapter.suicideTxLogMismatchPanic = true
 		return nil
 	}
 }
@@ -406,6 +415,7 @@ func (stateDB *StateDBAdapter) Suicide(evmAddr common.Address) bool {
 		return false
 	}
 	// clears the account balance
+	actBalance := new(big.Int).Set(s.Balance)
 	if err := s.SubBalance(s.Balance); err != nil {
 		log.L().Debug("failed to clear balance", zap.Error(err), zap.String("address", evmAddr.Hex()))
 		return false
@@ -418,14 +428,22 @@ func (stateDB *StateDBAdapter) Suicide(evmAddr common.Address) bool {
 	}
 	// before calling Suicide, EVM will transfer the contract's balance to beneficiary
 	// need to create a transaction log on successful suicide
-	if len(stateDB.lastAddBalanceAmount.Bytes()) > 0 {
-		from, _ := address.FromBytes(evmAddr[:])
-		stateDB.addTransactionLogs(&action.TransactionLog{
-			Type:      iotextypes.TransactionLogType_IN_CONTRACT_TRANSFER,
-			Sender:    from.String(),
-			Recipient: stateDB.lastAddBalanceAddr,
-			Amount:    stateDB.lastAddBalanceAmount,
-		})
+	if stateDB.lastAddBalanceAmount.Cmp(actBalance) == 0 {
+		if len(stateDB.lastAddBalanceAmount.Bytes()) > 0 {
+			from, _ := address.FromBytes(evmAddr[:])
+			stateDB.addTransactionLogs(&action.TransactionLog{
+				Type:      iotextypes.TransactionLogType_IN_CONTRACT_TRANSFER,
+				Sender:    from.String(),
+				Recipient: stateDB.lastAddBalanceAddr,
+				Amount:    stateDB.lastAddBalanceAmount,
+			})
+		}
+	} else {
+		if stateDB.suicideTxLogMismatchPanic {
+			log.L().Panic("suicide contract's balance does not match",
+				zap.String("suicide", actBalance.String()),
+				zap.String("beneficiary", stateDB.lastAddBalanceAmount.String()))
+		}
 	}
 	// mark it as deleted
 	stateDB.suicided[addrHash] = struct{}{}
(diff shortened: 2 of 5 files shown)


action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -406,6 +412,17 @@ func (stateDB *StateDBAdapter) Suicide(evmAddr common.Address) bool {
 		stateDB.logError(err)
 		return false
 	}
+	// before calling Suicide, EVM will transfer the contract's balance to beneficiary
+	// need to create a transaction log on successful suicide
+	if len(stateDB.lastAddBalanceAmount.Bytes()) > 0 {
```
millken, line 417: should be check `if stateDB.lastAddBalanceAmount != nil && stateDB.lastAddBalanceAmount.Cmp(big.NewInt(0))>0` to prevent nil or negative amount

dustinxie: in `AddBalance()`, nil or negative amount will be caught and return error, in other words, `lastAddBalanceAmount` must be valid positive number here

add initialization of the value `lastAddBalanceAmount` in `new` to prevent being nil

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -242,6 +245,7 @@ func (stateDB *StateDBAdapter) SubBalance(evmAddr common.Address, amount *big.In
 
 // AddBalance adds balance to account
 func (stateDB *StateDBAdapter) AddBalance(evmAddr common.Address, amount *big.Int) {
+	stateDB.lastAddBalanceAmount.SetUint64(0)
```
millken, line 248: the change is only in AddBalance, does `SubBalance` also need it

CoderZhi, line 248: what's the purpose?

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -406,6 +416,17 @@ func (stateDB *StateDBAdapter) Suicide(evmAddr common.Address) bool {
 		stateDB.logError(err)
 		return false
 	}
+	// before calling Suicide, EVM will transfer the contract's balance to beneficiary
+	// need to create a transaction log on successful suicide
+	if len(stateDB.lastAddBalanceAmount.Bytes()) > 0 {
+		from, _ := address.FromBytes(evmAddr[:])
+		stateDB.addTransactionLogs(&action.TransactionLog{
```
CoderZhi, line 423: if lastAddBalanceAmount is not set or the amount is different from balance, panic?

envestcc (commented):
I think it is not a safe enough way to handle EVM upgrade in the future. If EVM has unexpected calls to `AddBalance` and `SubBalance` in the future, it would be difficult for us to detect.

Here is my proposal:
1. Basically, operations on `AddBalance` and `SubBalance` not triggered by `MakeTransfer` are unexpected behaviors in terms of recording transaction logs, and these operations should be append to a `UnexpectedOpList`.

2. Based on the current logic of the EVM code, when `Suicide` is called, we can remove unexpected `AddBalance` records with the same address and amount from the list, thereby generating a transaction log.

3. After the action is executed, we should check the `UnexpectedOpList`. If it is not empty, we should return an error or panic.

action/protocol/execution/evm/evmstatedbadapter.go
dustinxie: fixed in latest commit

dustinxie:
> I think it is not a safe enough way to handle EVM upgrade in the future. If EVM has unexpected calls to `AddBalance` and `SubBalance` in the future, it would be difficult for us to detect.
> 
> Here is my proposal:
> 
> 1. Basically, operations on `AddBalance` and `SubBalance` not triggered by `MakeTransfer` are unexpected behaviors in terms of recording transaction logs, and these operations should be append to a `UnexpectedOpList`.
> 2. Based on the current logic of the EVM code, when `Suicide` is called, we can remove unexpected `AddBalance` records with the same address and amount from the list, thereby generating a transaction log.
> 3. After the action is executed, we should check the `UnexpectedOpList`. If it is not empty, we should return an error or panic.

good suggestion, will take care in further PRs

action/protocol/execution/evm/evmstatedbadapter.go
dustinxie: to clear the amount for each AddBalance(), consider the following:
1. AddBalance(4000)
2. AddBalance(0)
3. Inside Suicide, it calls AddBalance(4000) to beneficiary but failed, if the amount is not cleared, it will mistakenly think the amount match and issue a tx log with amount=4000, but the 4000 is in step 1 and is mistaken

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -406,6 +426,25 @@ func (stateDB *StateDBAdapter) Suicide(evmAddr common.Address) bool {
...
+			from, _ := address.FromBytes(evmAddr[:])
+			stateDB.addTransactionLogs(&action.TransactionLog{
+				Type:      iotextypes.TransactionLogType_IN_CONTRACT_TRANSFER,
+				Sender:    from.String(),
+				Recipient: stateDB.lastAddBalanceAddr,
+				Amount:    stateDB.lastAddBalanceAmount,
+			})
+		}
+	} else {
+		if stateDB.suicideTxLogMismatchPanic {
```
dustinxie, line 442: for safety, we gate the panic with a flag
after hard-fork, we can run a fullsync w/o the flag, and remove this flag if fullsync pass

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -406,6 +426,25 @@ func (stateDB *StateDBAdapter) Suicide(evmAddr common.Address) bool {
 		stateDB.logError(err)
 		return false
 	}
+	// before calling Suicide, EVM will transfer the contract's balance to beneficiary
+	// need to create a transaction log on successful suicide
+	if stateDB.lastAddBalanceAmount.Cmp(actBalance) == 0 {
+		if len(stateDB.lastAddBalanceAmount.Bytes()) > 0 {
+			from, _ := address.FromBytes(evmAddr[:])
+			stateDB.addTransactionLogs(&action.TransactionLog{
```
envestcc, line 434: is not a hard-fork?

CoderZhi: We believe it is not, because transaction logs are not included in delta. @dustinxie may confirm this.

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -406,6 +426,25 @@ func (stateDB *StateDBAdapter) Suicide(evmAddr common.Address) bool {
 		stateDB.logError(err)
 		return false
 	}
+	// before calling Suicide, EVM will transfer the contract's balance to beneficiary
+	// need to create a transaction log on successful suicide
+	if stateDB.lastAddBalanceAmount.Cmp(actBalance) == 0 {
+		if len(stateDB.lastAddBalanceAmount.Bytes()) > 0 {
```
CoderZhi, line 432: ignore efficiency and use `stateDB.lastAddBalanceAmount.Cmp(big.NewInt(0)) > 0`

action/protocol/execution/evm/evmstatedbadapter.go
CoderZhi: https://github.com/iotexproject/iotex-core/pull/4171/files#r1519122441 if we panic here, what's the purpose of resetting the mount?

dustinxie: I gave an example above, if we don't reset the amount, it is possible to generate a mistaken txLog for an earlier `AddBalance()` whose amount happens to equal to suicide address

action/protocol/context.go
```diff
@@ -261,6 +262,7 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			SharedGasWithDapp:                       g.IsToBeEnabled(height),
 			CandidateRegisterMustWithStake:          !g.IsToBeEnabled(height),
 			DisableDelegateEndorsement:              !g.IsToBeEnabled(height),
+			SuicideTxLogMismatchPanic:               g.IsToBeEnabled(height),
```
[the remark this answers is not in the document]
dustinxie: is this a must? to me using a positive (enable after height) flag name will help improve code readability and reduce chance of programming mistake

action/protocol/execution/evm/evmstatedbadapter.go
dustinxie: updated

action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -48,6 +48,8 @@ type (
 		err                        error
 		blockHeight                uint64
 		executionHash              hash.Hash256
+		lastAddBalanceAddr         string
```
Liuhaai, line 51: getStateObject(addr common.Address) in geth(https://github.com/ethereum/go-ethereum/blob/ebf9e11af2ff701d0961623e817d37b421b96802/core/state/statedb.go#L540) seems more reasonable than two new variables under StateDBAdapter for me

dustinxie: can you elaborate? we cannot get the beneficiary's address in StateDBAdapter, so capture it in `AddBalance` every time. also we need the transfer amount, not the beneficiary's entire balance

CoderZhi (approved):

envestcc (approved):

PR #4256, opened 2024-04-26
[db] implement Get/Put/Delete/Version() for BoltDBVersioned

dustinxie:
# Description
as title, implement 4 methos for versioned DB.

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [x] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:


db/db_versioned_util.go
```diff
@@ -46,3 +43,72 @@ func deserializeVersionedNamespace(buf []byte) (*versionedNamespace, error) {
...
+	if km == nil {
+		// key not yet written
+		return &keyMeta{
+			lastWrite:    value,
+			firstVersion: version,
+			lastVersion:  version,
+		}, false
+	}
+	if version < km.lastVersion || version < km.deleteVersion {
+		// writing to an earlier version complicates things, for now it is not allowed
```
envestcc, line 107: why not disable to updates for historical versions

db/db_versioned_util.go
```diff
@@ -46,3 +43,72 @@ func deserializeVersionedNamespace(buf []byte) (*versionedNamespace, error) {
...
+// deserializeKeyMeta deserializes byte-stream to key meta
+func deserializeKeyMeta(buf []byte) (*keyMeta, error) {
+	var km versionpb.KeyMeta
+	if err := proto.Unmarshal(buf, &km); err != nil {
+		return nil, err
+	}
+	return fromProtoKM(&km), nil
+}
+
+func (km *keyMeta) updateRead(version uint64) (bool, error) {
```
envestcc, line 87: Misleading naming, it does not update any attr of the `keyMeta`

db/db_versioned.go
```diff
@@ -80,50 +95,195 @@ func (b *BoltDBVersioned) Stop(ctx context.Context) error {
...
+	km, err := b.checkNamespaceAndKey(ns, key)
+	if err != nil {
+		if cause := errors.Cause(err); cause != ErrNotExist && cause != ErrInvalid {
+			return err
+		}
+	}
+	if km == nil || version < km.lastVersion || version <= km.deleteVersion {
+		return nil
+	}
+	km.deleteVersion = version
```
envestcc, line 205: Only saving the latest delete will cause getting wrong value if multiple deletes. 

for example, there are opt sequnces as follows:
- put k1 v1 at 1
- del k1 at 5
- put k1 v2 at 10
- del k1 at 15

then get k1 at 7 should return NotExist, but actually returns v1

db/db_versioned_util.go
```diff
@@ -103,12 +105,54 @@ func (km *keyMeta) updateWrite(version uint64, value []byte) (*keyMeta, bool) {
...
+}
+
+func (km *keyMeta) lastDelete() uint64 {
+	if numDelete := len(km.deleteVersion); numDelete > 0 {
+		return km.deleteVersion[numDelete-1]
+	}
+	return 0
+}
+
+func (km *keyMeta) deleteInBetween(write, read uint64) (bool, error) {
```
envestcc, line 133: naming func to `notDeleteInBetween()`?

db/db_versioned_util.go
```diff
@@ -46,3 +45,114 @@ func deserializeVersionedNamespace(buf []byte) (*versionedNamespace, error) {
...
+	}
+	if !hasDelete {
+		return true, nil
+	}
+	if write < nextDelete {
+		// there's a delete after last write
+		return false, ErrDeleted
+	}
+	// delete and write fall on the same version, need to check further
+	// if it's write-after-delete or delete-after-write
```
envestcc, line 156: For each singel version, only the last record (delete/put) can be retained, which can help avoid this.

db/db_versioned.go
```diff
@@ -80,50 +95,229 @@ func (b *BoltDBVersioned) Stop(ctx context.Context) error {
...
+		return err
+	}
+	if km == nil {
+		return errors.Wrapf(ErrNotExist, "key = %x doesn't exist", key)
+	}
+	if err = km.updateDelete(version); err != nil {
+		return err
+	}
+	if version == km.lastVersion {
+		// write <key, nil> to indicate this is a delete-after-write
```
envestcc, line 225: what if the two updates on same version:
```
del k
put k nil
```

db/db_versioned_util.go
dustinxie: updated name

db/db_versioned.go
dustinxie: yes, but this will be a very rare corner case
1. usually a key's value is some valid data
2. if the user intends to use `nil` as a special case, then he can use a special []byte value to avoid the corner case

db/db_versioned.go
```diff
@@ -80,50 +96,192 @@ func (b *BoltDBVersioned) Stop(ctx context.Context) error {
...
 }
 
+// Base returns the underlying KVStore
+func (b *BoltDBVersioned) Base() KVStore {
+	return b.db
+}
+
 // Put writes a <key, value> record
-func (b *BoltDBVersioned) Put(ns string, version uint64, key, value []byte) error {
+func (b *BoltDBVersioned) Put(version uint64, ns string, key, value []byte) error {
```
envestcc, line 105: Do we really need to implement these single function`Put/Delete`? I prefer only keep `WriteBatch`

db/db_versioned.go
```diff
@@ -80,50 +96,192 @@ func (b *BoltDBVersioned) Stop(ctx context.Context) error {
...
+	} else {
+		if len(key) != int(vn.keyLen) {
+			return errors.Wrapf(ErrInvalid, "invalid key length, expecting %d, got %d", vn.keyLen, len(key))
+		}
+		last, _, err := b.get(math.MaxUint64, ns, key)
+		if !isNotExist(err) && version < last {
+			// not allowed to perform write on an earlier version
+			return ErrInvalid
+		}
+		buf.Delete(ns, keyForDelete(key, version), fmt.Sprintf("failed to delete key %x", key))
```
envestcc, line 129: why delete?

db/db_versioned.go
```diff
@@ -80,50 +96,192 @@ func (b *BoltDBVersioned) Stop(ctx context.Context) error {
...
+	if isNotExist(err) {
+		return err
+	}
+	if version < last {
+		// not allowed to perform delete on an earlier version
+		return ErrInvalid
+	}
+	buf := batch.NewBatch()
+	buf.Put(ns, keyForDelete(key, version), nil, fmt.Sprintf("failed to delete key %x", key))
+	buf.Delete(ns, keyForWrite(key, version), fmt.Sprintf("failed to delete key %x", key))
```
envestcc, line 203: why delete write key?

dustinxie: discussed offline
1. it is to keep only the last operation on the key
2. i recall in the original POC, we have cases where the same key is deleted, then written in the batch at around height 400k. So it is a valid request (to delete then write on the same height). We'll allow this behavior (and also write then delete on the same height)

db/db_versioned.go
dustinxie: discussed offline, it does not hurt to keep it

db/db_versioned.go
```diff
@@ -7,47 +7,42 @@
...
-	// kv := db.SetVersion(5)
-	// value, err := kv.Get("ns", key) // read 'key' at version 5
-	// kv = db.SetVersion(8)
-	// err := kv.Put("ns", key, value) // write 'key' at version 8
-
-	KvVersioned interface {
+	VersionedDB interface {
 		lifecycle.StartStopper
 
+		// Put insert or update a record identified by (namespace, key)
```
envestcc, line 32: Why to modify the interface this way? I would recommend that the version should be for all keys rather than individual keys. Do we have a use case that requires this? I think  the original interface design is easier for integration, or will there be a wrapper to encapsulate it?

dustinxie: yes, `KvVersioned` is the new wrapper to encapsulate it

envestcc (approved):

huangzhiran (approved):

PR #4276, opened 2024-05-23
[staking] ReadState API support Candidate ID

envestcc:
# Description

Due to the introduction of a new `ID` as a unique identifier by the Candidate, the following modifications need to be made to the API: 
- Upgrade the `candidateByAddress`, `candidateByName`, and `candidates` APIs to v3, adding 'id' to the returned results. 
- Add a new API called `candidateByID`; the original meaning of the `candidateByAddress` is to query based on the owner's address.

It's based on https://github.com/iotexproject/iotex-proto/pull/151

## Type of change

# How Has This Been Tested?
- [] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:

### DIFF
--- a/action/protocol/staking/candidate_statereader.go
+++ b/action/protocol/staking/candidate_statereader.go
@@ -478,11 +478,20 @@ func (c *candSR) readStateCandidateByName(ctx context.Context, req *iotexapi.Rea
 }
 
 func (c *candSR) readStateCandidateByAddress(ctx context.Context, req *iotexapi.ReadStakingDataRequest_CandidateByAddress) (*iotextypes.CandidateV2, uint64, error) {
-	owner, err := address.FromString(req.GetOwnerAddr())
-	if err != nil {
-		return nil, 0, err
+	var cand *Candidate
+	if len(req.GetOwnerAddr()) > 0 {
+		owner, err := address.FromString(req.GetOwnerAddr())
+		if err != nil {
+			return nil, 0, err
+		}
+		cand = c.GetCandidateByOwner(owner)
+	} else {
+		id, err := address.FromString(req.GetId())
+		if err != nil {
+			return nil, 0, err
+		}
+		cand = c.GetByIdentifier(id)
 	}
-	cand := c.GetCandidateByOwner(owner)
 	if cand == nil {
 		return &iotextypes.CandidateV2{}, c.Height(), nil
 	}
--- a/action/protocol/staking/candidate_statereader.go
+++ b/action/protocol/staking/candidate_statereader.go
@@ -478,19 +478,27 @@ func (c *candSR) readStateCandidateByName(ctx context.Context, req *iotexapi.Rea
 }
 
 func (c *candSR) readStateCandidateByAddress(ctx context.Context, req *iotexapi.ReadStakingDataRequest_CandidateByAddress) (*iotextypes.CandidateV2, uint64, error) {
-	var cand *Candidate
+	var cand, candByOwner, candByID *Candidate
 	if len(req.GetOwnerAddr()) > 0 {
 		owner, err := address.FromString(req.GetOwnerAddr())
 		if err != nil {
 			return nil, 0, err
 		}
-		cand = c.GetCandidateByOwner(owner)
-	} else {
+		candByOwner = c.GetCandidateByOwner(owner)
+	}
+	if len(req.GetId()) > 0 {
 		id, err := address.FromString(req.GetId())
 		if err != nil {
 			return nil, 0, err
 		}
-		cand = c.GetByIdentifier(id)
+		candByID = c.GetByIdentifier(id)
+	}
+	if candByOwner != nil && candByID != nil && address.Equal(candByID.GetIdentifier(), candByOwner.GetIdentifier()) {
+		cand = candByID
+	} else if candByID != nil {
+		cand = candByID
+	} else if candByOwner != nil {
+		cand = candByOwner
 	}
 	if cand == nil {
 		return &iotextypes.CandidateV2{}, c.Height(), nil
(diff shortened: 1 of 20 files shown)


dustinxie:
can you rebase this PR? i think lots of code has already checked in

action/protocol/staking/candidate_statereader.go
```diff
@@ -478,11 +478,20 @@ func (c *candSR) readStateCandidateByName(ctx context.Context, req *iotexapi.Rea
...
+		if err != nil {
+			return nil, 0, err
+		}
+		cand = c.GetCandidateByOwner(owner)
+	} else {
+		id, err := address.FromString(req.GetId())
+		if err != nil {
+			return nil, 0, err
+		}
+		cand = c.GetByIdentifier(id)
```
dustinxie, line 493: is it possible `Id` and `OwnerAddress` both exist? if yes, which one takes precedence?

action/protocol/staking/ethabi/common/candidatebyaddress.go
```diff
@@ -76,7 +76,7 @@ func (r *CandidateByAddressStateContext) EncodeToEth(resp *iotexapi.ReadStateRes
 
 	data, err := r.Method.Outputs.Pack(cand)
 	if err != nil {
-		return "", nil
+		return "", err
```
dustinxie, line 79: looks like this is a typo/bug in current code? we don't have a test for this?

CoderZhi:
As discussed offline, let's change the logic of creating delegate to:
```
if delegateByOwner(creator) != nil {
    return duplicate error
}
id := creator
while delegateByIdentifier(id) != nil { // up to N = 1000 times
    id = hash(id)
}
create delegate with id
```
such that an account who doesn't own a delegate could always create a delegate.

Make sure the scenario works in this PR.

dustinxie:
> Member



> As discussed offline, let's change the logic of creating delegate to:
> 
> ```
> if delegateByOwner(creator) != nil {
>     return duplicate error
> }
> id := creator
> while delegateByIdentifier(id) != nil { // up to N = 1000 times
>     id = hash(id)
> }
> create delegate with id
> ```
> 
> such that an account who doesn't own a delegate could always create a delegate.
> 
> Make sure the scenario works in this PR.

agreed, discussed offline we can use logic similar to `id = hash(address, blockheight)`

action/protocol/staking/candidate_statereader.go
envestcc: updated, now if both conditions are set, it means that they need to be fulfilled at mean while.

action/protocol/staking/ethabi/common/candidatebyaddress.go
envestcc: yeah, existed typo, but it's not easy to cause error for abi pack

action/protocol/staking/handlers.go
```diff
@@ -892,6 +906,26 @@ func (p *Protocol) fetchBucketAndValidate(
 	return bucket, nil
 }
 
+func (p *Protocol) generateCandidateID(owner address.Address, height uint64, csm CandidateStateManager) (address.Address, error) {
+	isValidID := func(id address.Address) bool {
+		return csm.GetByIdentifier(id) == nil && csm.GetByOwner(id) == nil
+	}
+	if isValidID(owner) {
+		return owner, nil
+	}
```
dustinxie, line 915: can have 
```
h = append(owner.Bytes(), append(byteutil.Uint64ToBytesBigEndian(height))
```
before the for loop, it is common for all iterations

action/protocol/staking/handlers.go
```diff
@@ -892,6 +906,26 @@ func (p *Protocol) fetchBucketAndValidate(
...
+	for i := range make([]struct{}, 1000) {
```
dustinxie, line 916: why not `for i := 0; i < 1000; i++`?

action/protocol/staking/handlers.go
```diff
@@ -892,6 +906,27 @@ func (p *Protocol) fetchBucketAndValidate(
...
 
+func (p *Protocol) generateCandidateID(owner address.Address, height uint64, csm CandidateStateManager) (address.Address, error) {
+	isValidID := func(id address.Address) bool {
+		return csm.GetByIdentifier(id) == nil && csm.GetByOwner(id) == nil
+	}
+	if isValidID(owner) {
+		return owner, nil
+	}
+	h := append(owner.Bytes(), byteutil.Uint64ToBytesBigEndian(height)...)
+	for i := 0; i < 1000; i++ {
```
dustinxie, line 917: can you quickly test how much time this 1000 loop will take? I'm thinking 128 should be enough to generate a diff ID

envestcc:
```
goos: darwin
goarch: arm64
pkg: github.com/iotexproject/iotex-core/action/protocol/staking
BenchmarkLoop-10            2366            489271 ns/op           33397 B/op       1024 allocs/op
```
around 0.5ms, which exclude query from csm. I think it's acceptable

action/protocol/staking/candidate_statereader.go
```diff
@@ -478,11 +478,28 @@ func (c *candSR) readStateCandidateByName(ctx context.Context, req *iotexapi.Rea
 }
 
 func (c *candSR) readStateCandidateByAddress(ctx context.Context, req *iotexapi.ReadStakingDataRequest_CandidateByAddress) (*iotextypes.CandidateV2, uint64, error) {
```
CoderZhi, line 480: deprecate this api

CoderZhi (approved):

dustinxie (approved):

PR #4319, opened 2024-07-01
[staking] make transfering ownership back valid

envestcc:
# Description

1. ensure that there is no conflict between the owner and identifier during candidateRegister and candidateTransferOwnership.
2. can transfer the candidate back to the previous owner.

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:

### DIFF
--- a/action/protocol/staking/handlers.go
+++ b/action/protocol/staking/handlers.go
@@ -691,6 +691,13 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
 	// cannot collide with existing identifier
 	candID := owner
 	if !featureCtx.CandidateIdentifiedByOwner {
+		// cannot collide with existing identifier
+		if csm.GetByIdentifier(owner) != nil {
+			return log, nil, &handleError{
+				err:           ErrInvalidOwner,
+				failureStatus: iotextypes.ReceiptStatus_ErrCandidateAlreadyExist,
+			}
+		}
 		id, err := p.generateCandidateID(owner, blkCtx.BlockHeight, csm)
 		if err != nil {
 			return log, nil, &handleError{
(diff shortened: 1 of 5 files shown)


action/protocol/staking/handlers.go
```diff
@@ -691,20 +691,21 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
 	// cannot collide with existing identifier
 	candID := owner
 	if !featureCtx.CandidateIdentifiedByOwner {
-		id, err := p.generateCandidateID(owner, blkCtx.BlockHeight, csm)
-		if err != nil {
+		// cannot collide with existing identifier
+		if csm.GetByIdentifier(owner) != nil {
```
envestcc, line 695: This is to ensure that the owner is not used as an ID for any existed candidate.

action/protocol/staking/handlers.go
```diff
@@ -691,20 +691,21 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
...
 			return log, nil, &handleError{
-				err:           errors.Wrap(err, "failed to generate candidate ID"),
+				err:           ErrInvalidOwner,
 				failureStatus: iotextypes.ReceiptStatus_ErrCandidateAlreadyExist,
 			}
 		}
-		candID = id
-		if csm.GetByIdentifier(candID) != nil {
```
envestcc, line 702: The reason for deletion here is that there is no need to check for ID conflicts anymore, as this is already ensured in generateCandidateID.

action/protocol/staking/handlers.go
```diff
@@ -691,20 +691,21 @@ func (p *Protocol) handleCandidateRegister(ctx context.Context, act *action.Cand
...
+		id, err := p.generateCandidateID(owner, blkCtx.BlockHeight, csm)
```
dustinxie, line 701: by moving the conflict check ahead, `generateCandidateID` will always return the original owner address, the for loop will never enter. this would cause 1 address can register only once

as discussed last evening, should remove the 3 lines, and let for loop to generate a random identifier. Then, the same address (after transfer ownership) can register again

CoderZhi: Yes, the first point is correct that only owner address will be used as identifier.
You might miss some part of the discussion. After the hard fork, preventing the same address to generate a new candidate is on purpose, to prevent confusion: CandidateByOwner and CandidateByIdentifier returns two different candidates given the same address. The reason I didn't ask @envestcc to delete the logic in `generateCandidateID` in this PR is that it really doesn't matter to keep it.

dustinxie (approved):

PR #4362, opened 2024-08-13
[actpool] impl a store to cache actions on disk

envestcc:
# Description
blobstore is a customized, high-performance datastore for transient actions, which is backed by [billy](https://github.com/holiman/billy) (a super simple datastore, also the blobpool storage option chosen by Ethereum). It will be used for cache actions so that can load from local when restart, avoiding large load on the network.

## Type of change

# How Has This Been Tested?
- [] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:


actpool/blobstore_test.go
```diff
@@ -0,0 +1,131 @@
...
+	acts := []*action.SealedEnvelope{}
+	r.NoError(store.Open(func(selp *action.SealedEnvelope) {
+		acts = append(acts, selp)
+	}))
+	acts[0].Hash()
+	r.Len(acts, 1)
+	r.Equal(act, acts[0])
+}
+
+func BenchmarkDatabase(b *testing.B) {
```
envestcc, line 67: benchmark: around 100x faster than kv
```
goos: darwin
goarch: arm64
pkg: github.com/iotexproject/iotex-core/actpool
BenchmarkDatabase/billy-put-10             10000            138064 ns/op          140654 B/op          1 allocs/op
BenchmarkDatabase/pebbledb-put-10            158           6843763 ns/op           24024 B/op         27 allocs/op
BenchmarkDatabase/boltdb-put-10              120           8863447 ns/op          541060 B/op         64 allocs/op
PASS
ok      github.com/iotexproject/iotex-core/actpool      5.885s
```

CoderZhi (approved):

actpool/blobstore.go
```diff
@@ -0,0 +1,245 @@
...
+		config: cfg,
+		lookup: make(map[hash.Hash256]uint64),
+		encode: encode,
+		decode: decode,
+	}, nil
+}
+
+func (s *blobStore) Open(onData onAction) error {
+	dir := s.config.Datadir
+	if err := os.MkdirAll(dir, 0700); err != nil {
```
dustinxie, line 87: default should be 0600 since it is not executable? or 0644 if we want to allow others to read the file

actpool/blobstore.go
```diff
@@ -0,0 +1,245 @@
...
+			return
+		}
+		id, ok := s.lookup[h]
+		if !ok {
+			log.L().Error("worst action not found in lookup", zap.String("hash", hex.EncodeToString(h[:])))
+			continue
+		}
+		if err := s.store.Delete(id); err != nil {
+			log.L().Error("failed to delete worst action", zap.Error(err))
+		}
```
dustinxie, line 213: delete(s.lookup, h)?

actpool/blobstore.go
```diff
@@ -0,0 +1,245 @@
...
+func (s *blobStore) drop() {
+	for {
+		h, ok := s.evict()
+		if !ok {
+			log.L().Error("no worst action found")
+			return
+		}
+		id, ok := s.lookup[h]
+		if !ok {
+			log.L().Error("worst action not found in lookup", zap.String("hash", hex.EncodeToString(h[:])))
```
dustinxie, line 208: this should not happen according to `evict()` below?

actpool/blobstore.go
```diff
@@ -0,0 +1,245 @@
...
+
+	for h := range s.lookup {
+		if !fn(h) {
+			return
+		}
+	}
+}
+
+func (s *blobStore) drop() {
+	for {
```
dustinxie, line 200: just lock the `s.lookup` map, and delete all items?

actpool/blobstore.go
```diff
@@ -0,0 +1,245 @@
...
+		stored uint64         // Useful data size of all transactions on disk
+
+		lookup map[hash.Hash256]uint64 // Lookup table mapping hashes to tx billy entries
+		lock   sync.RWMutex            // Mutex protecting the store
+
+		encode encodeAction // Encoder for the tx
+		decode decodeAction // Decoder for the tx
+	}
+	blobStoreConfig struct {
+		Datadir string `yaml:"datadir"` // Data directory containing the currently executable blobs
```
dustinxie, line 55: so billy has multiple files in a dir, not a single file right?

actpool/blobstore.go
```diff
@@ -0,0 +1,245 @@
...
+	decodeAction func([]byte) (*action.SealedEnvelope, error)
+)
+
+var (
+	errBlobNotFound = fmt.Errorf("blob not found")
+)
+
+var defaultBlobStoreConfig = blobStoreConfig{
+	Datadir: "blobpool",
+	Datacap: 10 * 1024 * 1024 * 1024,
```
dustinxie, line 70: evmNetworkID should be one of the config value?

actpool/blobstore.go
```diff
@@ -0,0 +1,245 @@
...
+	// Index all transactions on disk and delete anything inprocessable
+	var fails []uint64
+	index := func(id uint64, size uint32, blob []byte) {
+		act, err := s.decode(blob)
+		if err != nil {
+			fails = append(fails, id)
+			log.L().Warn("Failed to decode action", zap.Error(err))
+			return
+		}
+		if err = onData(act); err != nil {
```
dustinxie, line 99: why need to call this `onData`? In our blobpool, what would this function be?

actpool/blobstore.go
envestcc: yes, and one file per shelve, which includes bunch of fixed size slots

actpool/blobstore.go
envestcc: not required, it's inner part of decode

actpool/blobstore.go
envestcc: executable permission for a folder should mean it can enter the directory and subdirectories.

actpool/blobstore.go
envestcc: actpool will use it to load actions when startup

actpool/blobstore.go
envestcc: lock/unlock should be called only in public methods

actpool/blobstore.go
envestcc: yes in this impl of evict, but it does't mater?

actpool/blobstore.go
envestcc: good catch

actpool/blobstore.go
dustinxie: then define a public method? the current implementation is actually not correct, access to `s.lookup` should be protected with a lock (just like in other funcs) right?

actpool/blobstore.go
dustinxie: i see

actpool/blobstore.go
dustinxie: yes it's part of decode, think about we start the pool on testnet vs. mainnet, there needs to be a way to pass in `evmNetworkID`?

actpool/blobstore.go
envestcc: drop() is called by Put(), should not be called by outside

actpool/blobstore.go
envestcc: can ref to here for the usage: https://github.com/iotexproject/iotex-core/pull/4366/files#diff-c65dd5a657c73eb89458c69a82a25fc0993655650b3ad477136955c31e50e2bbR249

actpool/blobstore.go
dustinxie: OK I see, add a TODO or file an issue for a better drop policy, for now it is simply deleting all items, the newly added item could also be deleted

actpool/blobstore.go
```diff
@@ -0,0 +1,260 @@
...
+		if !ok {
+			log.L().Error("no worst action found")
+			return
+		}
+		id, ok := s.lookup[h]
+		if !ok {
+			log.L().Error("worst action not found in lookup", zap.String("hash", hex.EncodeToString(h[:])))
+			continue
+		}
+		if err := s.store.Delete(id); err != nil {
```
dustinxie, line 225: also it will deadlock here, since Put() --> drop() already lock the mutex, need a private delete() func
```
func (s *blobStore) Delete(hash hash.Hash256) error {
	s.lock.Lock()
	defer s.lock.Unlock()
	return s.delete()
}

func (s *blobStore) delete()
```

actpool/blobstore.go
dustinxie: I see, so pass in the `decode` func which takes care of the evmNetworkID

actpool/blobstore.go
```diff
@@ -0,0 +1,260 @@
...
+func (s *blobStore) Close() error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	return s.store.Close()
+}
+
+func (s *blobStore) Get(hash hash.Hash256) (*action.SealedEnvelope, error) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	if s.store == nil {
```
dustinxie, line 139: add a `Readiness` to s, and use `s.IsReady()`

actpool/blobstore.go
dustinxie: as discuss, no need `for`, just delete the first item returned from `evict`

dustinxie (approved):

actpool/blobstore.go
```diff
@@ -0,0 +1,270 @@
...
+	if err := s.store.Delete(id); err != nil {
+		log.L().Error("failed to delete worst action", zap.Error(err))
+	}
+	delete(s.lookup, h)
+	return
+}
+
+// TODO: implement a proper eviction policy
+func (s *blobStore) evict() (hash.Hash256, bool) {
+	for h := range s.lookup {
```
dustinxie, line 245: add a TODO, to find the worst action according to some policy

PR #4427, opened 2024-10-14
[actpool] check blob count before verifying proof

envestcc:
# Description

Move blob count per account checking before proof verification in actpool, because verification is time-consuming calling.

Implement it in two steps:
- add subscriber in actpool to knowing actions for other components
- make a new validator to wrap the checking, and put it front to generic_validator

## Type of change

# How Has This Been Tested?
- [] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:

### DIFF
--- a/actpool/actqueue.go
+++ b/actpool/actqueue.go
@@ -60,8 +60,6 @@ type actQueue struct {
 	clock          clock.Clock
 	ttl            time.Duration
 	mu             sync.RWMutex
-	// count of blob txs in the queue
-	blobCount int
 }
 
 // NewActQueue create a new action queue
@@ -139,12 +137,8 @@ func (q *actQueue) Put(act *action.SealedEnvelope) error {
 			}
 		}
 		q.updateFromNonce(nonce)
+		q.ap.removeInvalidActs([]*action.SealedEnvelope{actInPool})
 		return nil
-	}
-	// check max number of blob txs per account
-	isBlobTx := len(act.BlobHashes()) > 0
-	if isBlobTx && q.blobCount >= int(q.ap.cfg.MaxNumBlobsPerAcct) {
-		return errors.Wrap(action.ErrNonceTooHigh, "too many blob txs in the queue")
 	}
 	nttl := &nonceWithTTL{nonce: nonce, deadline: q.clock.Now().Add(q.ttl)}
 	heap.Push(&q.ascQueue, nttl)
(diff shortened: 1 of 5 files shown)


actpool/actqueue.go
```diff
@@ -139,23 +137,16 @@ func (q *actQueue) Put(act *action.SealedEnvelope) error {
 			}
 		}
 		q.updateFromNonce(nonce)
+		q.ap.removeInvalidActs([]*action.SealedEnvelope{actInPool})
```
CoderZhi, line 140: what's the purpose?

envestcc: it's to remove replaced action by same nonce from actpool

CoderZhi (approved):

actpool/validator.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (v *blobValidator) OnRemoved(act *action.SealedEnvelope) {
+	if len(act.BlobHashes()) == 0 {
+		return
+	}
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+	sender := act.SenderAddress().String()
+	v.blobCntPerAcc[sender]--
```
dustinxie, line 56: add `if v.blobCntPerAcc[sender] > 0` for safety?

dustinxie, line 56: add `if v.blobCntPerAcc[sender] > 0` for safety?

actpool/actqueue.go
dustinxie: just curious, if this is a bug, (invalid act not removed now), would some test fail/pass due to this change?

envestcc: It break the test at actpool_test.go:379 after the introducing subscriber and blob validation, because the missing call of onRemoved, the blobTx was wrongly rejected

actpool/validator.go
envestcc: I think it's not required, if something wrong in subscriber mechnisam, the blobCntPerAcc is inaccurate, regardless of whether this check is included or not

actpool/actqueue.go
dustinxie: so should fix the test by correctly adding `onRemoved()`, but not this change?

actpool/validator.go
dustinxie: if something wrong and an underflow happens, blobCntPerAcc will become `uint64.MAX`, then it will reject any blob tx forever (at line 33)

dustinxie (approved):

PR #4432, opened 2024-10-15
[action] AntiqueTx to handle out-of-band version value

dustinxie:
# Description
at block height = 18879571, tx d03127451c120caed0c4e2ed91c58872231192e027daad3760c475e5aa2feffd used the value `ActionCore.Version = 4689` 

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -24,6 +24,11 @@ const (
 	BlobTxType       = 4
 )
 
+var (
+	// block height = 18879571, tx d03127451c120caed0c4e2ed91c58872231192e027daad3760c475e5aa2feffd used this value
+	_outOfBandTxType18879571 uint32 = 4689
+)
+
 type (
 	// Action is the action can be Executed in protocols. The method is added to avoid mistakenly used empty interface as action.
 	Action interface {
--- a/action/action.go
+++ b/action/action.go
@@ -24,11 +24,6 @@ const (
 	BlobTxType       = 4
 )
 
-var (
-	// block height = 18879571, tx d03127451c120caed0c4e2ed91c58872231192e027daad3760c475e5aa2feffd used this value
-	_outOfBandTxType18879571 uint32 = 4689
-)
-
 type (
 	// Action is the action can be Executed in protocols. The method is added to avoid mistakenly used empty interface as action.
 	Action interface {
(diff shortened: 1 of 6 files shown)


action/action.go
```diff
@@ -24,6 +24,11 @@ const (
 	BlobTxType       = 4
 )
 
+var (
+	// block height = 18879571, tx d03127451c120caed0c4e2ed91c58872231192e027daad3760c475e5aa2feffd used this value
+	_outOfBandTxType18879571 uint32 = 4689
```
CoderZhi, line 29: obviously, this is the evmNetworkID.

CoderZhi (commented):
please add some description to the PR explaining the root cause of the misusing version value. Is it possible to reproduce an action with the unexpected version? if no, explain why. If yes, what value could it be and does it make sense to hardcode it?

action/tx_antique.go
```diff
@@ -16,26 +16,36 @@ var _ TxCommonInternal = (*AntiqueTx)(nil)
...
 			gasLimit: gasLimit,
 			gasPrice: gasPrice,
 		},
+		version: version,
 	}
 }
 
 func (tx *AntiqueTx) Version() uint32 {
-	return AntiqueTxType
+	return tx.version
```
CoderZhi, line 36: shall we hide the real version value or expose it to outside caller?

action/action.go
dustinxie: yes, but it is a co-incidence. someone just send in a tx at height 18879571 by setting `version = 4689`, that's why name it that way

action/tx_antique.go
dustinxie: i think we should expose, b/c this is the actual value of this tx's version?

dustinxie:
actually this solution has a potential risk, if another tx with arbitrary value, say `version = 355` was broadcast to network before hard-fork, then old/current code will accept the tx and mint it into block, while new code will reject it.
A safer solution would be to handle in `default` case, will update the PR later

PR #4581, opened 2025-03-14
Delete dock and fix potential corruption state issue

CoderZhi:
1. make a copy of view in every workingset
2. support creating workingset based on an existing workingset
3. every protocol is in charge of maintaining its own view

Fixes #(issue)

## Type of change

# How Has This Been Tested?
- [] make test
- [] fullsync

**Test Configuration**:
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

# Checklist:


action/protocol/staking/protocol.go
```diff
@@ -412,16 +411,25 @@ func (p *Protocol) Commit(ctx context.Context, sm protocol.StateManager) error {
...
 	if err != nil {
 		return nil, err
 	}
-	csm, err := NewCandidateStateManager(sm, featureWithHeightCtx.ReadStateFromDB(height))
+	view, err := sm.ReadView(_protocolID)
 	if err != nil {
 		return nil, err
 	}
-	return p.handle(ctx, elp, csm)
+	snapshot := view.(*ViewData).Snapshot()
```
dustinxie, line 422: can directly do `view.Snapshot()`

action/protocol/staking/protocol.go
```diff
@@ -412,16 +411,25 @@ func (p *Protocol) Commit(ctx context.Context, sm protocol.StateManager) error {
...
+	receipt, err := p.handle(ctx, elp, csm)
+	if err != nil {
+		if err := view.(*ViewData).Revert(snapshot); err != nil {
```
dustinxie, line 425: can directly do `view.Revert(snapshot)`

action/protocol/staking/protocol.go
```diff
@@ -280,22 +280,26 @@ func (p *Protocol) CreateGenesisStates(
...
 		csr, err := ConstructBaseView(sm)
 		if err != nil {
 			return err
 		}
+		view := csr.BaseView()
 		if _, err = sm.PutState(csr.BaseView().bucketPool.total, protocol.NamespaceOption(_stakingNameSpace), protocol.KeyOption(_bucketPoolAddrKey)); err != nil {
 			return err
 		}
+		view.bucketPool.EnableSMStorage()
+		if err := sm.WriteView(_protocolID, view); err != nil {
```
dustinxie, line 297: wondering is this necessarily needed? it's now base view constructed/read from `csr`, there's no changes to view/sm yet

consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -120,7 +120,10 @@ func (cm *chainManager) BlockCommitTime(height uint64) (time.Time, error) {
 }
 
 // MintNewBlock creates a new block with given actions
-func (cm *chainManager) MintNewBlock(timestamp time.Time) (*block.Block, error) {
+func (cm *chainManager) MintNewBlock(height uint64, timestamp time.Time) (*block.Block, error) {
+	if cm.bc.TipHeight() != height-1 {
+		// TODO: mint new block for given height
```
dustinxie, line 125: need to finalize this `TODO`?

state/factory/factory.go
```diff
@@ -251,21 +251,29 @@ func (sf *factory) Height() (uint64, error) {
 	return byteutil.BytesToUint64(height), nil
 }
 
+func (sf *factory) newReadOnlyWorkingSet(ctx context.Context, height uint64) (*workingSet, error) {
```
dustinxie, line 254: would be cleaner to put this into another new PR

state/factory/readonlykv.go
```diff
@@ -0,0 +1,22 @@
...
+	return db.ErrNotSupported
+}
+
+func (r *readOnlyKV) Delete(string, []byte) error {
+	return db.ErrNotSupported
+}
+
+func (r *readOnlyKV) WriteBatch(batch.KVStoreBatch) error {
+	return db.ErrNotSupported
+}
```
dustinxie, line 22: would be cleaner to put this into another new PR

state/factory/util.go
```diff
@@ -100,7 +100,7 @@ func protocolPreCommit(ctx context.Context, sr protocol.StateManager) error {
 	if reg, ok := protocol.GetRegistry(ctx); ok {
 		for _, p := range reg.All() {
 			post, ok := p.(protocol.PreCommitter)
-			if ok && sr.ProtocolDirty(p.Name()) {
+			if ok {
```
dustinxie, line 103: better to double-check, AFAIK `PreCommit()` is meant to be called only when the protocol is dirty, i.e., had some pending change. change to calling it always may change behavior of past blocks

state/factory/util.go
```diff
@@ -114,7 +114,7 @@ func protocolCommit(ctx context.Context, sr protocol.StateManager) error {
 	if reg, ok := protocol.GetRegistry(ctx); ok {
 		for _, p := range reg.All() {
 			post, ok := p.(protocol.Committer)
-			if ok && sr.ProtocolDirty(p.Name()) {
+			if ok {
```
dustinxie, line 117: same here

state/factory/statedb.go
```diff
@@ -232,11 +250,28 @@ func (sdb *stateDB) NewBlockBuilder(
...
+		ws  *workingSet
+		err error
+	)
 	sdb.mutex.RLock()
 	currHeight := sdb.currentChainHeight
 	sdb.mutex.RUnlock()
-	ws, err := sdb.newWorkingSet(ctx, currHeight+1)
+	switch {
+	case currHeight+1 < expectedBlockHeight:
+		parent, ok := sdb.workingsets.Get(bcCtx.Tip.Hash)
```
envestcc, line 265: the key in `workingsets` is not blockHash, we may need cache ws by hash meanwhile

state/factory/workingsetstore.go
```diff
@@ -8,35 +8,32 @@ package factory
...
-func (store *workingSetStoreCommon) ReadView(name string) (interface{}, error) {
-	return store.view.Read(name)
+func (store *workingSetStoreCommon) Filter(ns string, cond db.Condition, start, limit []byte) ([][]byte, [][]byte, error) {
+	return store.flusher.KVStoreWithBuffer().Filter(ns, cond, start, limit)
 }
 
-func (store *workingSetStoreCommon) WriteView(name string, value interface{}) error {
-	return store.view.Write(name, value)
+func (store *workingSetStoreCommon) WriteBatch(bat batch.KVStoreBatch) error {
+	return store.flusher.KVStoreWithBuffer().WriteBatch(bat)
```
envestcc, line 36: calling `store.flusher.Flush()` is also required

action/protocol/staking/protocol.go
```diff
@@ -412,16 +411,25 @@ func (p *Protocol) Commit(ctx context.Context, sm protocol.StateManager) error {
...
-	return p.handle(ctx, elp, csm)
+	snapshot := view.(*ViewData).Snapshot()
+	receipt, err := p.handle(ctx, elp, csm)
+	if err != nil {
+		if err := view.(*ViewData).Revert(snapshot); err != nil {
+			return nil, errors.Wrap(err, "failed to revert view")
+		}
+		return receipt, err
+	}
+	sm.WriteView(_protocolID, csm.DirtyView())
```
envestcc, line 430: the `view` should already updated when handling the ation, thus we do not need to `WriteView` here

action/protocol/staking/candidate_center_extra.go
```diff
@@ -21,6 +21,9 @@ func (cb *candBase) clone() *candBase {
 	for bucket, cand := range cb.selfStkBucketMap {
 		clone.selfStkBucketMap[bucket] = cand.Clone()
 	}
+	for identity, cand := range cb.identifierMap {
```
envestcc, line 24: Is it a bug? we need to take a closer look at whether it's a hard fork.

state/factory/workingsetstore.go
CoderZhi: https://github.com/iotexproject/iotex-core/pull/4581/files#diff-1e8edee1ee86092b9b6a7e60c38d852dc068c1105b6d3d70e1d87dba244fa959R40
for `Put` and `Delete`, `Flush()` is not called

action/protocol/staking/candidate_center_extra.go
dustinxie: this is used by the 1.8.4 bug fix
we can maybe define a new `clone2`, or run a fullsync test

state/factory/workingsetstore.go
```diff
@@ -33,24 +36,49 @@ func (store *workingSetStoreCommon) Filter(ns string, cond db.Condition, start,
 }
 
 func (store *workingSetStoreCommon) WriteBatch(bat batch.KVStoreBatch) error {
-	return store.flusher.KVStoreWithBuffer().WriteBatch(bat)
+	if err := store.flusher.KVStoreWithBuffer().WriteBatch(bat); err != nil {
```
envestcc, line 39: It should could work well. This way maybe more clear:
```
if store.commited {
    return store.flusher.BaseKVStore().WriteBatch(bat)
}
return store.flusher.KVStoreWithBuffer().WriteBatch(bat)
```

state/factory/workingsetstore.go
envestcc, line 1: suggestion: maybe we can return a new db.KVStore object using `workingSetStore.KVStore()` to pass it to the child, instead of implementing this interface in workingSetStore. This may make things clearer.

```
type workingSetKVStore struct {
    // read methods based on buffer
    Get(), Filter()
    // writes directly for underlying db
    Put(), Delete(), WriteBatch()
}
```

state/factory/statedb.go
```diff
@@ -224,6 +238,7 @@ func (sdb *stateDB) Validate(ctx context.Context, blk *block.Block) error {
 			return errors.Wrap(err, "failed to validate block with workingset in statedb")
 		}
 		sdb.workingsets.Add(key, ws)
+		sdb.workingsets.Add(blk.HashBlock(), ws)
```
envestcc, line 241: this will miss the minted block, we should cache hash even if existed

state/factory/workingsetstore.go
CoderZhi: will it cause a problem what the flusher.buffer is not considered?

state/factory/workingsetstore.go
CoderZhi: we may need an extra lock here

envestcc (approved):

state/factory/workingsetstore.go
```diff
@@ -33,24 +36,49 @@ func (store *workingSetStoreCommon) Filter(ns string, cond db.Condition, start,
...
 
 func (store *workingSetStoreCommon) Put(ns string, key []byte, value []byte) error {
-	return store.flusher.KVStoreWithBuffer().Put(ns, key, value)
+	if err := store.flusher.KVStoreWithBuffer().Put(ns, key, value); err != nil {
+		return errors.Wrap(err, "failed to put value")
+	}
+	if !store.committed {
+		return nil
+	}
+	return store.flusher.Flush()
```
dustinxie, line 62: when it gets here, the workingsetStore is already committed and `flusher.Flush()` called at L77. so in what use-case `Put()` will be called after committed, and why in such case need to call `Flush()` again?

dustinxie: also `func (store *factoryWorkingSetStore) Put()` call this and do extra tlt handling
```
func (store *factoryWorkingSetStore) Put(ns string, key []byte, value []byte) error {
	store.workingSetStoreCommon.Put(ns, key, value)
	nsHash := hash.Hash160b([]byte(ns))

	return store.tlt.Upsert(nsHash[:], toLegacyKey(key), value)
}
```
does the tlt require special handling?

and same for `Delete()`

dustinxie: as we discussed, after being committed, this `workingSetStoreCommon` functionally becomes a `KVStore` (for the next block N+1's workingset), so Put should write to underlying DB, that's correct
On the other hand, it is also a workingset's store, after being committed, it should not write to DB anymore, that's why the behavior here is somewhat confusing.
A better approach is to wrap it into another struct, similar to the recently `readOnlyKV` check-in, small change and clearly shows the intention, maybe a later PR to improve

dustinxie (approved):

CoderZhi:
after #4599, all `state.factory` related implementation will be deleted. So, we may ignore the problem in `state.factory` and its workingset store

PR #4792, opened 2026-03-09
fix: correct typo 'unsufficient' to 'insufficient'

envestcc:
Fix typo in error message in bucket_validation.go: unsufficient to insufficient.


envestcc (commented):
the test TestValidateBucket/validate_bucket_min_amount breaks

PR #4837, opened 2026-05-20
feat(ioctl): support sending EIP-7702 SetCode tx

envestcc:
## Summary

Adds first-class support for sending EIP-7702 SetCode transactions through ioctl.

- **`ioctl action sign-auth DELEGATE_ADDR -n NONCE -s AUTHORITY`** — signs an EIP-7702 authorization with a keystore key and prints it as 0x-hex RLP. Offline, no broadcast.
- **`--auth 0xHEX` flag** (repeatable) on `ioctl action transfer` and `ioctl contract invoke`. When present, the action is built as a SetCodeTx instead of a legacy tx. Without it, behavior is unchanged.
- Non-legacy tx types are dispatched with `TX_CONTAINER` encoding — the chain's generic validator rejects `IOTEX_PROTOBUF`-encoded modern tx types (`action/protocol/generic_validator.go`). The shared helper builds + eth-signs the tx, wraps it via `EthRawToContainer`, and sends it through the existing `SendAction` path.
- `--auth` is intentionally **not** registered on `contract deploy`: a SetCodeTx is always a CALL and the chain rejects a nil `to`.
- The existing `--gas-price` flag is reused for both `gasTipCap` and `gasFeeCap` on the SetCodeTx, so no new gas flags are introduced.

## Test plan

Verified end-to-end on a nightly node (chainID 3 / evmNetworkID 4691, Yap hardfork active):

- [x] `action transfer <addr> 0 --auth $REVOKE` → txType=4, TX_CONTAINER encoding, status 1, delegation cleared (`eth_getCode` → `0x`)
- [x] `action transfer <addr> <amount>` without `--auth` → txType=0, IOTEX_PROTOBUF encoding, status 1 (legacy path unchanged)
- [x] `contract invoke bytecode <addr> 0x --auth $REDELEGATE` → txType=4, status 1, delegation re-established
- [x] `contract deploy bytecode ... --auth ...` → rejected with `unknown flag: --auth`
- [x] `sign-auth` output RLP-decodes to the expected `SetCodeAuthorization`; recovered authority matches the signer
- [x] `make ioctl` + `go vet ./ioctl/...` clean


ioctl/cmd/action/actioneth.go
```diff
@@ -0,0 +1,243 @@
...
+	case 1:
+		return 4689, nil
+	case 2:
+		return 4690, nil
+	case 3:
+		return 4691, nil
+	default:
+		return 0, output.NewError(output.InputError,
+			fmt.Sprintf("unknown iotex chain id %d, set --evm-chain-id explicitly", iotexChainID), nil)
+	}
```
envestcc, line 101: can we get evm chainid from chain rpc, instead of hard code?

envestcc: Good point. I looked into it: the gRPC `APIService` doesn't currently expose `evmNetworkID` anywhere — `GetChainMeta` only returns the iotex `chainID`, and `GetServerMeta` is version strings. The only RPC that exposes the EVM chain id today is the Web3 `eth_chainId`, which ioctl isn't configured to reach.

The clean fix is to add `evmNetworkID` to the `ChainMeta` proto and populate it server-side (the node already knows it via `bc.EvmNetworkID()`), which is a cross-repo change (iotex-proto + core). I've filed #4838 to track that as a follow-up.

For this PR I've kept the hardcoded mapping with the `--evm-chain-id` override as an escape hatch; once #4838 lands, ioctl can query the value and fall back to the mapping only for older nodes.

ioctl/cmd/action/actioneth.go
```diff
@@ -0,0 +1,247 @@
...
+	core, err := action.EthRawToContainer(iotexChainID, hex.EncodeToString(rawBytes))
+	if err != nil {
+		return output.NewError(output.SerializationError, "failed to build tx container", err)
+	}
+	req := &iotextypes.Action{
+		Core:         core,
+		SenderPubKey: pubkey.Bytes(),
+		Signature:    sig,
+		Encoding:     iotextypes.Encoding_TX_CONTAINER,
```
envestcc, line 197: Encoding_TX_CONTAINER hasn't previously been used to send gRPC transactions directly. Can you confirm whether this is viable, and whether it introduces any risks or vulnerabilities?

envestcc: Confirmed viable — it's not actually a new server-side code path.

Both ingresses converge on the same internal method `coreService.SendAction(ctx, *iotextypes.Action)`:
- gRPC: `gRPCHandler.SendAction` (grpcserver.go:350) is a thin wrapper — it just forwards `in.GetAction()` to `coreService.SendAction`.
- Web3: `sendRawTransaction` (web3server.go:762) calls the same `cs.SendAction(ctx, req)`.

There is no ingress-specific branching in `coreService.SendAction` (deserialize → validateChainID → replay-deployer check → actpool.Add).

More importantly, the Web3 path **already produces `TX_CONTAINER`-encoded actions in production**: since Vanuatu, `decodeEthTx` (web3server.go:724-733) wraps every non-staking/rewarding eth tx with `action.EthRawToContainer(...)` + `Encoding: TX_CONTAINER` before handing it to `coreService.SendAction`. The `iotextypes.Action` this PR constructs in `actioneth.go` is built the same way. So the server-side handling of a `TX_CONTAINER` action arriving via the gRPC `SendAction` RPC is identical to one arriving via `eth_sendRawTransaction`.

On risk/vulnerability:
- **No new attack surface.** The gRPC `SendAction` RPC has always accepted an arbitrary `Encoding` value; `loadProto` dispatches on it and rejects unknown encodings via a `default` branch. A client could already set `Encoding: TX_CONTAINER` — this PR adds an ioctl *client* capability, it does not widen what the server accepts.
- **Signature verification is unchanged.** `TX_CONTAINER` is verified via `envelopeHash → rlpRawHash` (the ethereum RLP hash) and `VerifySignature` at actpool admission — same as the Web3 path.
- **Chain-id / replay.** `coreService.SendAction` deserializes with the node's own `EVMNetworkID()`. If a client signs for the wrong EVM chain id, server-side verification fails and the action is rejected — no fund risk.

Live-verified on a nightly node: 6+ SetCodeTx sent through this exact ioctl path, all mined with status=1 and the expected EIP-7702 delegation state changes on chain.

CoderZhi (dismissed):

ioctl/cmd/action/actionsignauth.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// Multi-language support
+var (
+	_signAuthCmdShorts = map[config.Language]string{
+		config.English: "Sign an EIP-7702 SetCode authorization (offline, no broadcast)",
+		config.Chinese: "签名一个 EIP-7702 SetCode 授权 (离线, 不广播)",
+	}
+	_signAuthCmdUses = map[config.Language]string{
+		config.English: "sign-auth DELEGATE_ADDR_HEX -n NONCE -s AUTHORITY [-P PASSWORD] [--evm-chain-id N]",
```
envestcc: `-s AUTHORITY` is not needed, should be derived from signer

envestcc: Good catch — the code already does that (`Signer()` returns the `-s` value or the config default). The `-s AUTHORITY` in the usage string was just confusing naming that made it read like a separate concept. Fixed in b48a4dc20: usage now reads `[-s SIGNER]` matching the rest of ioctl, and the Short clarifies the signer's key is what becomes the authority.

CoderZhi (approved):

PR #4839, opened 2026-05-26
disable premint on api nodes

CoderZhi:
## Summary
- API nodes (`cfg.Chain.HistoryIndexPath != ""`) wire a `workingSetStoreWithSecondary` whose `KVStore()` returns nil. The pre-mint path in `prepareNextProposal` forks a previous workingset via `workingSet.NewWorkingSet`, which needs `KVStore()` and errors with `"KVStore() not supported in *workingSetStoreWithSecondary"` — blocking the api node from minting.
- Add `DisablePremint(bool)` to the `RDPoSCtx` interface and gate `prepareNextProposal` on it. `Builder.Build()` turns it on when `HistoryIndexPath` is configured (same condition that creates the secondary erigon store). With pre-mint off, mint goes through the default `newWorkingSet` path, which works correctly for api nodes.
- No new config knob for per-block work limits: existing `MintTimeout` already drives the deadline cutoff in `workingset.go:821-826`. Operators set a tighter `MintTimeout` (or larger `AllowedBlockGasResidue`) on api nodes via existing config.

## Test plan
- [x] `go test ./consensus/scheme/rolldpos/ -count=1` — 48 pass (was 45; +3 new)
- [x] `TestDisablePremint` — default state, setter toggles, `prepareNextProposal` short-circuits to nil without calling `chain.Fork`, and toggling off restores original error behavior
- [x] `TestNewRollDPoS/api-node-disables-premint` — Builder sets the flag when `HistoryIndexPath` is non-empty
- [x] `TestNewRollDPoS/non-api-node-keeps-premint` — flag stays off otherwise
- [ ] Manual: deploy on an api node with a secondary erigon store and confirm it can mint a block (out-of-band)

### DIFF
--- a/consensus/scheme/rolldpos/rolldpos.go
+++ b/consensus/scheme/rolldpos/rolldpos.go
@@ -347,6 +347,14 @@ func (b *Builder) Build() (*RollDPoS, error) {
 		b.clock = clock.New()
 	}
 	b.cfg.DB.DbPath = b.cfg.Consensus.ConsensusDBPath
+	var opts []Option
+	// Premint forks a previous workingset to mint a block ahead of the committed chain tip.
+	// That path needs ws.store.KVStore(), which is nil when a secondary erigon store is wired
+	// (api nodes set HistoryIndexPath). Disable premint there so mint falls back to building
+	// a workingset from scratch.
+	if len(b.cfg.Chain.HistoryIndexPath) > 0 {
+		opts = append(opts, WithPremintDisabled())
+	}
 	ctx, err := NewRollDPoSCtx(
 		consensusfsm.NewConsensusConfig(b.cfg.Consensus.FSM, b.cfg.DardanellesUpgrade, b.cfg.WakeUpgrade, b.cfg.Genesis, b.cfg.Consensus.Delay),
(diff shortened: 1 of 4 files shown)


consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -366,6 +366,13 @@ func (b *Builder) Build() (*RollDPoS, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "error when constructing consensus context")
 	}
+	// Premint forks a previous workingset to mint a block ahead of the committed chain tip.
+	// That path needs ws.store.KVStore(), which is nil when a secondary erigon store is wired
+	// (api nodes set HistoryIndexPath). Disable premint there so mint falls back to building
+	// a workingset from scratch.
+	if len(b.cfg.Chain.HistoryIndexPath) > 0 {
+		ctx.DisablePremint(true)
```
envestcc: It's better to add param in `NewRollDPoSCtx` instead, b/c it's should only be set when init

envestcc (dismissed):

envestcc (approved):

PR #4841, opened 2026-05-26
feat(consensus): plumb BLS producer keys + aggregation feature gate (IIP-52)

envestcc:
## Summary

Scaffolding for the BLS signature aggregation work tracked in [IIP-52](https://github.com/iotexproject/iips/blob/master/iip-52.md). **No on-chain behavior change yet** — `EnableBLSAggregation` is gated on `IsToBeEnabled`, and the BLS keys plumbed into `rollDPoSCtx` are not yet used to sign or verify endorsements. Subsequent PRs will add the COMMIT-stage BLS signing path and the BlockFooter aggregation/verification.

- `blockchain/config.go`: add `Chain.BLSProducerPrivKey` (comma-separated hex) and `BLSProducerPrivateKeys()`. Empty value falls back to deriving each BLS key from the corresponding ECDSA producer key via `crypto.GenerateBLS12381PrivateKey` (consistent with the existing `ioctl account blskey` derivation).
- `consensus/scheme/rolldpos`: `Builder.SetBLSPriKey`; `NewRollDPoSCtx` accepts `[]*crypto.BLS12381PrivateKey` aligned 1:1 with the producer ECDSA keys; `rollDPoSCtx.blsPriKeys` is stored for the upcoming signing path.
- `consensus/consensus.go`: wire `SetBLSPriKey(cfg.Chain.BLSProducerPrivateKeys()...)`.
- `action/protocol/context.go`: `FeatureCtx.EnableBLSAggregation` gated on `g.IsToBeEnabled(height)`. Will be flipped to a named hardfork height in a later PR once the rest of Phase 2 lands.
- `go.mod`: bump iotex-proto to envestcc/iotex-proto `bls-aggregate` (52e72a6) for the new `BlockFooter.aggregated_signature` / `signer_bitmap` fields and the `BLSEndorsement` message. Tracking PR: iotexproject/iotex-proto#169.

## Why a separate BLS key field on Chain config

Per IIP-52 there are two equally valid modes for operators: maintain a standalone BLS keystore, or derive deterministically from the producer ECDSA key. The config accepts both — `BLSProducerPrivKey` is the explicit override, empty means derive. Length must match the active producer key range; otherwise the node panics on startup.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] Targeted tests: `TestNewRollDPoSCtx`, `TestCheckVoteEndorser`, `TestCheckBlockProposer` and related (15 tests pass)
[...]

PR #169, opened 2018-10-12
IOTEX-88 Unify action size calculation and byte stream generation

CoderZhi:
Use byteutil in the two functions. This is based on 168, will rebase onto it once its accepted

PR #4842, opened 2026-05-27
feat(consensus): BLS endorsement sender + wire format (IIP-52)

envestcc:
**Depends on #4841** — stacked on top of the BLS key plumbing PR. Once #4841 merges, this PR's diff will collapse to just the new changes.

Sender + wire-format half of the BLS signature aggregation work tracked in [IIP-52](https://github.com/iotexproject/iips/blob/master/iip-52.md). Receiver verification + endorsement-manager quorum integration land in a follow-up PR; with the feature gate parked at \`IsToBeEnabled\` this PR is dead code in production.

## What's in (new vs. #4841)

- \`endorsement.BLSEndorsement\` — new Go type holding (timestamp, endorser iotex address, BLS signature). Endorser is the delegate's iotex address, not the BLS pubkey; receivers look up the registered pubkey from candidate state, mirroring the BlockFooter aggregate verification model.
- \`EndorsedConsensusMessage\` — polymorphic over \`*Endorsement\` and \`*BLSEndorsement\`, round-trips through \`ConsensusMessage.endorsement\` and \`ConsensusMessage.bls_endorsement\`. At most one is set per message.
- \`ProofOfLock\` — internal struct that carries either \`[]*Endorsement\` or \`[]*BLSEndorsement\` so the proof-of-lock cascade doesn't sprout if/else branches at every callsite. \`blockProposal\`, \`roundCtx.proofOfLock\`, and the \`BlockProposal\` proto Serialize / LoadProto all flow through it.
- \`ConsensusConfig.BLSAggregationEnabled(height)\` — feature gate wired off \`Genesis.ToBeEnabledBlockHeight\`. Will be re-pointed at a named hardfork height once the full Phase-2 stack lands.
- Sender side:
  - \`rollDPoSCtx.newEndorsement\` now signs PROPOSAL, LOCK and COMMIT votes with BLS post-fork (skipping delegates without a configured BLS key).
  - \`rollDPoSCtx.endorseBlockProposal\` signs the proposer's wrapping endorsement with BLS post-fork. Block header signing remains on the ECDSA producer key — that signature ties the block to a chain identity and is unrelated to the consensus vote layer.
  - The proposer's \`producerKey\` (ECDSA + BLS + address) is threaded through \`Proposal / mintNewBlock / endorseBlockProposal\` so the branch can pick the right key without a separate lookup.
- iotex-proto bump to \`envestcc/iotex-proto@03a99cc\` ([PR #169](https://github.com/iotexproject/iotex-proto/pull/169)): \`BlockProposal.bls_endorsements\` added so post-fork proof-of-lock can ride the wire.

## What's deferred to the follow-up PR

- Endorsement manager BLS collection + DB persistence
- \`roundCtx.AddBLSVoteEndorsement\`
- \`RollDPoS.HandleConsensusMsg\` BLS dispatch + signature verification (\`CheckBlockProposer\` / \`CheckVoteEndorser\` BLS path)
- BLS pubkey lookup mechanism (from candidate state at round start)
- Proof-of-lock BLS replay in \`CheckBlockProposer\` (a TODO marker is in place where BLS endorsements are decoded but not yet counted toward majority)
- End-to-end tests for the BLS COMMIT / relock path

## Why all three topics, not just COMMIT

The original IIP-52 motivation (compress the on-chain footer) only requires aggregating COMMIT votes. Implementation-wise though, keeping PROPOSAL/LOCK on ECDSA while COMMIT runs on BLS creates a mixed-mode coupling at the lock check: a delegate that emitted COMMIT (BLS) but not PROPOSAL still counts toward locking, but the proof-of-lock would then need to carry mixed-type entries on the wire. Switching all three topics to BLS post-fork keeps the lock check / proof-of-lock paths uniform and leaves room for future LOCK aggregation in BlockProposal relay (\`bls_endorsements\` field would then aggregate). The CPU cost (extra ~36 ms verify per 5-second round) is negligible.

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] Targeted tests pass: \`TestNewRollDPoSCtx\`, \`TestCheckVoteEndorser\`, \`TestCheckBlockProposer\`, etc. (13 tests)
- [ ] Full BLS COMMIT path tests land with the receiver PR

### DIFF
--- a/consensus/scheme/rolldpos/rolldposctx.go
+++ b/consensus/scheme/rolldpos/rolldposctx.go
@@ -403,29 +403,29 @@ func (ctx *rollDPoSCtx) HasDelegate() bool {
 func (ctx *rollDPoSCtx) Proposal() (interface{}, error) {
 	ctx.mutex.RLock()
 	defer ctx.mutex.RUnlock()
-	var privateKey crypto.PrivateKey = nil
+	var key *producerKey
 	proposer := ctx.round.Proposer()
 	// TODO: this is to pass unit tests, remove it after the unit tests are fixed
 	if proposer == "" {
-		privateKey = ctx.producerKeys[0].ecdsa
+		key = &ctx.producerKeys[0]
 	} else {
-		for _, pk := range ctx.producerKeys {
-			if pk.address == proposer {
-				privateKey = pk.ecdsa
+		for i := range ctx.producerKeys {
+			if ctx.producerKeys[i].address == proposer {
+				key = &ctx.producerKeys[i]
 				break
 			}
 		}
 	}
-	if privateKey == nil {
+	if key == nil {
 		return nil, nil
 	}
 	if ctx.round.IsLocked() {
 		return ctx.endorseBlockProposal(newBlockProposal(
 			ctx.round.Block(ctx.round.HashOfBlockInLock()),
 			ctx.round.ProofOfLock(),
-		), privateKey)
+		), key)
 	}
-	return ctx.mintNewBlock(privateKey)
+	return ctx.mintNewBlock(key)
 }
 
 func (ctx *rollDPoSCtx) prepareNextProposal(prevHeight uint64, prevHash hash.Hash256) error {
@@ -733,12 +733,15 @@ func (ctx *rollDPoSCtx) Active() bool {
 // private functions
 ///////////////////////////////////////////
 
-func (ctx *rollDPoSCtx) mintNewBlock(privateKey crypto.PrivateKey) (*EndorsedConsensusMessage, error) {
+func (ctx *rollDPoSCtx) mintNewBlock(key *producerKey) (*EndorsedConsensusMessage, error) {
 	var err error
 	blk := ctx.round.CachedMintedBlock()
 	if blk == nil {
 		// in case that there is no cached block in eManagerDB, it mints a new block.
-		blk, err = ctx.chain.MintNewBlock(ctx.round.StartTime(), privateKey, ctx.round.PrevHash())
+		// Block header signing stays on the ECDSA producer key regardless of
+		// the BLS-aggregation feature flag; the BLS key is only used for the
+		// consensus endorsement wrapping this proposal.
+		blk, err = ctx.chain.MintNewBlock(ctx.round.StartTime(), key.ecdsa, ctx.round.PrevHash())
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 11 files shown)


endorsement/bls_endorsement.go
envestcc, line 1: I think there is no need to add a new `BLSEndorsement`; we can simply continue using `Endorsement`. The length of the signature should be sufficient to distinguish whether it is a BLS or an ECDSA signature. This way, many parts of the code would not need to be modified.

envestcc:
Refactored per the review suggestion — force-pushed `73e2d57`.

`Endorsement` is now reused as-is for BLS post-fork; signature length (65B secp256k1 vs 96B BLS) is the discriminator. The endorser field keeps its existing semantics (secp256k1 producer pubkey), so `Endorser().Address()` still resolves the iotex address; receivers will look up the registered BLS verifying key from candidate state by that address.

Dropped:
- `endorsement.BLSEndorsement` Go type
- `EndorsedConsensusMessage` polymorphism (back to single endorsement field)
- `ProofOfLock` internal struct (back to `[]*Endorsement`)
- `blockProposal` proto serialize/deserialize branching

Also amended **iotex-proto PR #169** to drop the standalone `BLSEndorsement` message and the `bls_endorsement` / `bls_endorsements` fields on `ConsensusMessage` / `BlockProposal`. The proto changes there are now just the `BlockFooter` aggregate fields plus a comment clarifying `Endorsement.signature` semantics. New SHA: `e4439ef`.

Diff size: 12 files / +396 → 6 files / +147. Build and 15 targeted tests still green.

(I tried to update the PR body to reflect this too, but GitHub's GraphQL keeps 5xx'ing on edits — sorry for the stale description until that clears.)

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -881,8 +902,37 @@ func (ctx *rollDPoSCtx) newEndorsement(
...
 	msgs := make([]*EndorsedConsensusMessage, 0, len(ens))
 	for _, en := range ens {
-		msgs = append(msgs, NewEndorsedConsensusMessage(ctx.round.Height(), vote, en))
+		msgs = append(msgs, NewEndorsedConsensusMessage(height, vote, en))
 	}
 
 	return msgs, nil
 }
+
+// newBLSEndorsements signs the consensus vote with each producer's BLS key
```
envestcc: Is it necessary to separate a new `BLSEndorsements` method, or can we directly modify the existing `newEndorsement` method?

envestcc:
Done in `ef3235d` — merged `newBLSEndorsements` back into `newEndorsement` with a single `signVote` helper that branches internally on `useBLS`. Also tightened `endorseBlockProposal` to use the same helper, dropping its duplicated ECDSA/BLS branch. The `rolldposctx.go` portion of the diff drops from +80 to +51 lines.

PR #4843, opened 2026-05-28
feat(consensus): BLS endorsement receiver path (IIP-52)

envestcc:
**Draft — depends on #4842 (which depends on #4841).** Stacks on top of the BLS sender PR. Once #4841 and #4842 merge in order, this PR's diff will collapse to just the receiver-path additions.

Receiver half of the BLS signature aggregation work for [IIP-52](https://github.com/iotexproject/iips/blob/master/iip-52.md). With #4842 a delegate post-fork emits BLS-signed endorsements; this PR makes peer nodes accept, verify, and count them toward quorum. The feature gate is still parked at `IsToBeEnabled`, so the BLS path is exercised only by tests in production builds.

## What's in (new vs. #4842)

- `BLSPubKeysByEpochFunc` callback type. `consensus.go` provides the implementation: reads the epoch's delegate list from candidate state and extracts each candidate's registered `BLSPubKey`, returning a map keyed by operator iotex address.
- `roundCalculator` caches the BLS pubkey index per round (decoded as `*crypto.BLS12381PublicKey`). `UpdateRound` carries it across height transitions inside an epoch and re-fetches on epoch boundaries.
- `roundCtx.BLSPubKey(addr)` accessor; `roundCtx.verifyEndorsement(doc, en)` dispatches on signature length (65 B secp256k1 vs 96 B BLS).
- `rollDPoSCtx.VerifyEndorsement(height, doc, en)` is the public entry point. Same length-aware dispatch plus pre/post-fork gating — pre-fork rejects 96 B sigs, post-fork rejects 65 B sigs.
- `HandleConsensusMsg` replaces the unconditional ECDSA verify (`endorsement.VerifyEndorsedDocument`) with `r.ctx.VerifyEndorsement(...)`. The `CheckBlockProposer` proof-of-lock replay path flows through `round.AddVoteEndorsement`, which now dispatches on signature length internally, so BLS endorsements in proof-of-lock are verified transparently.
- 11 new unit tests covering: BLS endorse/verify round-trip (positive + 4 negative cases), the two-layer dispatch (`roundCtx.verifyEndorsement` + `rollDPoSCtx.VerifyEndorsement`), length gating in both directions, and the sender confirming `newEndorsement` emits the expected signature scheme for each fork state.

## Notes on design choices

- **Pubkey lookup once per round, not per message.** The delegate set is round-stable, so the BLS pubkey map is built at `roundCalculator.newRound`/`UpdateRound` time and held on `roundCtx`. Verify paths just do `ctx.round.BLSPubKey(addr)` — no state read per incoming endorsement.
- **Endorser stays on the secp256k1 producer pubkey.** Per the discussion on #4842, `Endorsement.Endorser()` remains the ECDSA pubkey across fork. `Endorser().Address()` still works for delegate-set membership checks; receivers look up the BLS verifying key from candidate state by that address.
- **`endorsement.VerifyEndorsedDocument` is no longer the entry point.** It still works for pre-fork ECDSA-only verify, but the active call site (`HandleConsensusMsg`) now routes through the round-aware dispatcher so BLS pubkey access is available.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] Targeted tests: 51 in `endorsement/` + `consensus/scheme/rolldpos/`, including the 11 new BLS tests
- [ ] Integration test of the BLS aggregation footer path lands with the next PR (Phase 2 deliverable)

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -9,6 +9,7 @@ import (
 	"context"
 
 	"github.com/facebookgo/clock"
+	"github.com/iotexproject/go-pkgs/crypto"
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-proto/golang/iotextypes"
 	"github.com/pkg/errors"
@@ -113,7 +114,7 @@ func NewConsensus(
 			return nil, errors.New("block builder factory is not set")
 		}
 		chainMgr := rolldpos.NewChainManager(bc, sf, ops.bbf)
-		delegatesByEpochFunc := func(epochNum uint64, prevHash []byte) ([]string, error) {
+		delegatesByEpochFunc := func(epochNum uint64, prevHash []byte) ([]*rolldpos.Delegate, error) {
 			fork, err := chainMgr.Fork(hash.Hash256(prevHash))
 			if err != nil {
 				return nil, err
@@ -145,53 +146,21 @@ func NewConsensus(
 			if err != nil {
 				return nil, err
 			}
-			addrs := []string{}
+			delegates := make([]*rolldpos.Delegate, 0, len(candidatesList))
 			for _, candidate := range candidatesList {
-				addrs = append(addrs, candidate.Address)
+				d := &rolldpos.Delegate{Address: candidate.Address}
+				if len(candidate.BLSPubKey) > 0 {
+					blsPubKey, err := crypto.BLS12381PublicKeyFromBytes(candidate.BLSPubKey)
+					if err != nil {
+						return nil, errors.Wrapf(err, "invalid BLS pubkey for delegate %s", candidate.Address)
+					}
+					d.BLSPubKey = blsPubKey
+				}
+				delegates = append(delegates, d)
 			}
-			return addrs, nil
+			return delegates, nil
 		}
 		proposersByEpochFunc := delegatesByEpochFunc
-		blsPubKeysByEpochFunc := func(epochNum uint64, prevHash []byte) (map[string][]byte, error) {
-			fork, err := chainMgr.Fork(hash.Hash256(prevHash))
-			if err != nil {
-				return nil, err
-			}
-			forkSF, err := fork.StateReader()
-			if err != nil {
-				return nil, err
-			}
-			re := protocol.NewRegistry()
-			if err := ops.rp.Register(re); err != nil {
-				return nil, err
-			}
-			ctx := genesis.WithGenesisContext(
-				protocol.WithRegistry(context.Background(), re),
-				cfg.Genesis,
-			)
-			ctx = protocol.WithFeatureWithHeightCtx(ctx)
-			tipHeight := fork.TipHeight()
-			tipEpochNum := ops.rp.GetEpochNum(tipHeight)
-			var candidatesList state.CandidateList
-			switch epochNum {
-			case tipEpochNum:
-				candidatesList, err = ops.pp.Delegates(ctx, forkSF)
-			case tipEpochNum + 1:
-				candidatesList, err = ops.pp.NextDelegates(ctx, forkSF)
-			default:
-				err = errors.Errorf("invalid epoch number %d compared to tip epoch number %d", epochNum, tipEpochNum)
-			}
-			if err != nil {
-				return nil, err
-			}
-			out := make(map[string][]byte, len(candidatesList))
-			for _, candidate := range candidatesList {
-				if len(candidate.BLSPubKey) > 0 {
-					out[candidate.Address] = candidate.BLSPubKey
-				}
-			}
-			return out, nil
-		}
 		bd := rolldpos.NewRollDPoSBuilder().
 			SetPriKey(cfg.Chain.ProducerPrivateKeys()...).
 			SetBLSPriKey(cfg.Chain.BLSProducerPrivateKeys()...).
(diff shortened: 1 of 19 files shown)


consensus/scheme/rolldpos/roundctx.go
```diff
@@ -50,6 +51,12 @@ type roundCtx struct {
 	proofOfLock []*endorsement.Endorsement
 	status      status
 	eManager    *endorsementManager
+
+	// blsPubKeys maps the epoch delegate's operator iotex address to their
+	// registered BLS12-381 public key. Populated at round construction (via
+	// the BLSPubKeysByEpochFunc callback) so verify paths can resolve the
+	// pubkey by address without hitting state per message.
+	blsPubKeys map[string]*crypto.BLS12381PublicKey
```
envestcc: I think it makes more sense for `blsPubKey` to be part of `proposers`/`delegates`, with both types of public keys forming a new struct.

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -84,6 +90,12 @@ type (
 		Clock() clock.Clock
 		CheckBlockProposer(uint64, *blockProposal, *endorsement.Endorsement) error
 		CheckVoteEndorser(uint64, *ConsensusVote, *endorsement.Endorsement) error
+		// VerifyEndorsement verifies an endorsement's signature against the
+		// given document, dispatching on signature length: secp256k1 (65 B)
+		// pre-fork or BLS12-381 (96 B) post-fork. Post-fork BLS verify
+		// resolves the verifying public key from the current round's
+		// candidate-state-derived index by endorser iotex address.
+		VerifyEndorsement(uint64, endorsement.Document, *endorsement.Endorsement) error
```
envestcc: It looks like changing the function parameter to a single `EndorsedConsensusMessage` would be more concise.

consensus/scheme/rolldpos/rolldposctx.go
```diff
@@ -243,6 +257,55 @@ func (ctx *rollDPoSCtx) Clock() clock.Clock {
...
+// or a post-fork message carrying a 65 B signature is rejected. For BLS,
+// the verifying pubkey is resolved from the current round's BLS pubkey
+// index by the endorser's iotex address.
+func (ctx *rollDPoSCtx) VerifyEndorsement(
+	height uint64,
+	doc endorsement.Document,
+	en *endorsement.Endorsement,
+) error {
+	ctx.mutex.RLock()
+	defer ctx.mutex.RUnlock()
```
envestcc: Why is a lock needed? If a lock is necessary, release it as soon as possible.

envestcc:
Pushed `24b8ac8` addressing all three review comments:

1. **`blsPubKey` part of `delegates` struct** — `roundCtx.delegates` is now `[]delegate` where `delegate` carries both `address` and `blsPubKey`. The parallel `blsPubKeys map` is gone. `roundCalc.delegatesAt` merges the address callback and the BLS pubkey callback into one slice at round construction, so the alignment is structural.
2. **`VerifyEndorsement` takes `*EndorsedConsensusMessage`** — single param, the message already carries height/doc/endorsement.
3. **Lock scope** — `VerifyEndorsement` now grabs the RLock only long enough to snapshot the round pointer and the BLS aggregation feature flag, then releases before the (potentially slow) signature verification. Safe because `*roundCtx` is replaced atomically, never mutated in place; once a caller holds the pointer they're working with a stable snapshot.

Build + vet clean. Targeted tests still all pass (`TestRound*`, `TestRollDPoSCtx_VerifyEndorsement_*`, `TestEndorse*`, plus the existing roundctx/check suites). The only failing test in the rolldpos package is `TestRollDPoSConsensus/1-epoch`, which is flaky on master independent of this PR.

consensus/consensus.go
```diff
@@ -152,6 +152,46 @@ func NewConsensus(
 			return addrs, nil
 		}
 		proposersByEpochFunc := delegatesByEpochFunc
+		blsPubKeysByEpochFunc := func(epochNum uint64, prevHash []byte) (map[string][]byte, error) {
```
envestcc: You can consider removing the separate `blsPubKeysByEpochFunc` and simply return both types of addresses within `delegatesByEpochFunc`, or wrap them together in a struct.

envestcc:
Pushed `3e323cf` for the follow-up comment — removed the separate `blsPubKeysByEpochFunc`.

`NodesSelectionByEpochFunc` now returns `[]*Delegate` (exported, `{Address, BLSPubKey}`). `consensus.go`'s `delegatesByEpochFunc` builds them in one pass over candidate state, decoding each registered BLS pubkey once per epoch; `proposersByEpochFunc` reuses it. `roundCalculator` stores `[]*Delegate` directly — no more parallel lookup/merge (`delegatesAt`/`blsPubKeysFor` are gone), and the `roundCtx.blsPubKeys` map introduced earlier is fully subsumed by the delegate slice.

Net -68 lines vs the previous revision. Build + vet clean, targeted tests pass.

consensus/scheme/rolldpos/roundcalculator.go
```diff
@@ -32,7 +32,7 @@ type roundCalculator struct {
 func (c *roundCalculator) UpdateRound(round *roundCtx, height uint64, blockInterval time.Duration, now time.Time, toleratedOvertime time.Duration) (*roundCtx, error) {
 	epochNum := round.EpochNum()
 	epochStartHeight := round.EpochStartHeight()
-	delegates := round.Delegates()
+	delegates := round.delegates
```
envestcc: Why not keep using `round.Delegates()`?

envestcc:
Good catch — pushed `1a2f24e`. `UpdateRound` was reaching into `round.delegates` directly only because `Delegates()` returned `[]string` while the field is `[]*Delegate`. Made the accessor return `[]*Delegate` so `UpdateRound` uses `round.Delegates()` again. The single genuine `[]string` consumer (`ConsensusMetrics.LatestDelegates`, external metrics) extracts addresses at the call site; `endorsementManager.Log`'s delegates param (currently unused) was retyped to match.

PR #4846, opened 2026-06-01
feat(rewarding): implement IIP-62 productive inflation (curve + 80/20 split)

CoderZhi:
## Summary

Implements IIP-62 productive inflation: post-activation, a per-block protocol mint replaces the fixed block/epoch reward, split 80/20 between the staker reward pool and the Machina DAO along the 5‑20‑Half disinflation curve. The curve and split follow the proposal; the notes below cover where this implementation makes decisions the spec does not pin down.

## Implementation decisions (beyond IIP-62)

- **State is derived, not stored.** `InflationState` persists only the three year‑boundary fields (`outstandingSupplyAtYearStart`, `currentInflationBps`, `currentYearIndex`). The running supply, cumulative mint, per‑year remainder, and epoch surplus are recomputed on demand (deterministic from genesis + height), so the record is written **once per year** instead of every block — archive growth for the `"inf"` key stays bounded. `OutstandingSupply` / `PostActivationMinted` derive their answer at the query height.
- **Split rounding.** `SplitMint` is stateless: `staker = ⌊mTotal·bps/denom⌋`, `Machina = mTotal − staker`. Conserves `mTotal` exactly every block; the sub‑Rau truncation biases toward Machina (well under a micro‑IOTX over the chain's lifetime), so no per‑block dust accumulator is carried.
- **Epoch reward.** Funded by the per‑block staker surplus over the (clamped) block reward, summed in closed form over each epoch's block range (`EpochInflationSurplus`) rather than a stored accumulator. Assumes the block reward is constant within an epoch — holds post‑Wake, where IIP-62 activates.
- **Block‑reward clamp.** `effective_block_reward = min(blockReward, mStaker)` so grants track down past the Y12+ floor and the fund never underflows.
- **Activation** reuses the existing `ToBeEnabledBlockHeight` gate (no new hardfork field). `OutstandingSupplyAtActivation` is a genesis‑pinned snapshot of circulating supply at `activation‑1` (produced by a forthcoming snapshot subcommand).
- **Machina credit** is a pure balance credit (`AddBalance`) — no EVM call into the recipient, which must tolerate passive accrual (a multisig / Safe works natively; a contract relying on `receive()` for internal accounting would not see it).
- **Mint‑only supply.** IoTeX deposits the EIP‑1559 base fee into the rewarding pool (not burned), so `OutstandingSupply` only grows; there is no burn‑side debit.

## Tests

`go test ./action/protocol/rewarding/...` — curve spec values (§1.1 / §1.3) and 20‑year table walk; the per‑block and epoch derivations cross‑checked against brute‑force per‑block sums (single‑year, year‑final, straddle, pre‑activation); fund‑invariant and reorg‑safety walks; and end‑to‑end `OutstandingSupply` / `PostActivationMinted` getter coverage.

## Follow-ups (not in this PR)

Snapshot subcommand for `OutstandingSupplyAtActivation`; e2e activation + replay test across a compressed Y1→Y2 boundary; Machina DAO multisig (externally generated — this PR only credits the balance).

## Dependency

The paired `iotex-proto` PR adding the `INFLATION_MINT_{STAKER,MACHINA}` transaction‑log types must merge first; `go.mod` pins to it.


action/protocol/rewarding/inflation_state.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+// Encode satisfies the systemcontracts.GenericValueContainer interface so the
+// state can ride the erigon storage path used by Fund / Admin.
+func (s *inflationState) Encode() (systemcontracts.GenericValue, error) {
+	d, err := proto.Marshal(s.toProto())
+	if err != nil {
+		return systemcontracts.GenericValue{}, err
+	}
+	return systemcontracts.GenericValue{PrimaryData: d}, nil
```
CoderZhi: split into generic value and secondary value

action/protocol/rewarding/inflation.go
```diff
@@ -0,0 +1,131 @@
...
+	if height < activation || blocksPerYear == 0 {
+		return 0
+	}
+	return (height-activation)/blocksPerYear + 1
+}
+
+// IsYearBoundary reports whether height is the first block of a new Year ≥ 2.
+// The activation block itself is the start of Year 1 and returns false here — its
+// initialization is the activation hook, not a year transition.
+func IsYearBoundary(activation, blocksPerYear, height uint64) bool {
```
envestcc, line 31: seems not used

action/protocol/rewarding/rewardingpb/rewarding.proto
```diff
@@ -52,3 +52,25 @@ message RewardLog {
...
+// InflationState holds only the IIP-62 fields that change at year boundaries. All other
+// per-block values — the running supply (field 1, outstandingSupply), cumulative mint
+// (field 3, postActivationMinted), per-year remainder (field 8, yearMintRemainder), epoch
+// surplus (field 9, epochRemainderAccumulator), and the dust accumulators (fields 6,7) —
+// were deterministic and are now derived on read (CumulativeMinted / EpochInflationSurplus
+// / PerBlockMint in the rewarding package), so this record is written once per year
+// instead of every block. IIP-62 is mint-only (the EIP-1559 base fee is deposited to the
+// rewarding pool, not burned), so there is no burn-side debit.
+message InflationState {
+    reserved 1, 3, 6, 7, 8, 9;
```
envestcc, line 68: nothing need to be reserved

PR #4847, opened 2026-06-02
feat(consensus): BlockFooter BLS aggregation (IIP-52)

envestcc:
**Draft — depends on #4843 → #4842 → #4841.** Stacks on the BLS endorsement receiver PR. Once those merge in order, this PR's diff will collapse to just the footer-aggregation changes.

Phase 2 deliverable for [IIP-52](https://github.com/iotexproject/iips/blob/master/iip-52.md). Proposer aggregates the per-block COMMIT BLS signatures into a single 96-byte signature plus a signer bitmap, stored in `BlockFooter.aggregated_signature` / `BlockFooter.signer_bitmap`. Verifiers reconstruct the signer set from the bitmap, look up each BLS pubkey from the round's delegate index, and `FastAggregateVerify` against the shared COMMIT-vote hash.

With the feature gate parked at `IsToBeEnabled`, this PR is dead code in production until the gate flips.

## Wire layout (per IIP-52, unchanged from PR #169)

```
BlockFooter {
  repeated Endorsement endorsements = 1;   // legacy, empty post-fork
  google.protobuf.Timestamp timestamp = 2;
  bytes aggregated_signature = 3;          // 96B BLS aggregate, populated post-fork
  bytes signer_bitmap         = 4;          // bit i = delegates[i] of the epoch, LSB-first
}
```

## What's in

- `blockchain/block/footer.go` — new fields, proto round-trip, `IsAggregated` / `AggregatedSignature` / `SignerBitmap` accessors. Pre-fork blocks serialize empty bytes in fields 3/4 — append-only safe.
- `blockchain/block/block.go` — new `Block.FinalizeWithAggregate(aggSig, bitmap, ts)` mirroring the existing `Finalize` contract. One-shot enforcement moved off `len(endorsements) != 0` onto a `commitTime` witness so both paths error consistently on a second call.
- `consensus/scheme/rolldpos/aggregate.go` — `aggregateCommitEndorsements` builds the aggregate sig + bitmap from a slice of BLS COMMIT endorsements, indexed against the round's `[]*Delegate`. `bitmapSigners` is the inverse for the verifier (with an out-of-range check).
- `consensus/scheme/rolldpos/rolldposctx.go` — at commit time, branch on `BLSAggregationEnabled(height)` and call `FinalizeWithAggregate` post-fork.
- `consensus/scheme/rolldpos/rolldpos.go` — `ValidateBlockFooter` routes aggregated footers through `validateAggregatedFooter`: bitmap → delegates → BLS pubkeys → `BLSAggregateSignature.Verify`, with a separate 2/3 majority check.
- `action/protocol/staking/protocol.go` — `ActiveCandidates` filters out candidates without a registered BLS pubkey post-fork so the aggregate signer set is well-defined.
- `endorsement/endorsement.go` — expose `SigningHash` so the verifier can reconstruct the COMMIT-vote hash from `blk.CommitTime()` outside an `Endorsement` struct.

## Why the common signing hash works

`FastAggregateVerify` requires every signer to have signed the **same** message. The COMMIT endorsement timestamp is a deterministic function of `round.StartTime() + AcceptBlockTTL + AcceptProposalEndorsementTTL + AcceptLockEndorsementTTL` — every delegate computes it identically. The proposer stores that timestamp on `Block.commitTime` at finalize, and the verifier feeds it back into `endorsement.SigningHash(vote, blk.CommitTime())` to recompute the hash. No per-signer message variation.

## Tests

- `TestAggregateCommitEndorsements_RoundTrip` — 4-of-4 set, aggregate verifies via `BLSAggregateSignature.Verify`
- `TestAggregateCommitEndorsements_PartialSet` — 3-of-5 set, bitmap correctly identifies signers
- `TestAggregateCommitEndorsements_RejectsNonBLSSig` — secp256k1 endorsement in the input rejected
- `TestAggregateCommitEndorsements_RejectsUnknownEndorser` — endorser not in delegate set rejected
- `TestBitmapSigners_BitBeyondDelegateCount` — bitmap claiming non-existent delegate rejected
- `TestBitmapSigners_EmptyBitmap` — empty input yields empty signer list
- `TestFooter_AggregateSerDes` — footer proto round-trip preserves agg sig + bitmap + IsAggregated

All 25 BLS-related tests across `endorsement/`, `blockchain/block/`, `consensus/scheme/rolldpos/` pass.

## Not yet in (future work)

- Hardfork height naming: still `IsToBeEnabled`. Will be flipped to a named fork once the full stack is reviewed.
- Tooling: ioctl `bc block` decode of the new footer fields; explorer / API updates. Separate small PR.
- Pre-fork → post-fork transition stress tests on a testnet with the gate active.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] All BLS-affected packages green
- [ ] Testnet activation rehearsal

### DIFF
--- a/consensus/scheme/rolldpos/rolldpos.go
+++ b/consensus/scheme/rolldpos/rolldpos.go
@@ -208,10 +208,15 @@ func (r *RollDPoS) Metrics() (scheme.ConsensusMetrics, error) {
 		return metrics, errors.Wrap(err, "error when calculating round")
 	}
 
+	delegates := round.Delegates()
+	delegateAddrs := make([]string, len(delegates))
+	for i, d := range delegates {
+		delegateAddrs[i] = d.Address
+	}
 	return scheme.ConsensusMetrics{
 		LatestEpoch:         round.EpochNum(),
 		LatestHeight:        height,
-		LatestDelegates:     round.Delegates(),
+		LatestDelegates:     delegateAddrs,
 		LatestBlockProducer: round.proposer,
 	}, nil
 }
--- a/consensus/scheme/rolldpos/rolldpos.go
+++ b/consensus/scheme/rolldpos/rolldpos.go
@@ -261,6 +261,7 @@ type (
 		// TODO: we should use keystore in the future
 		encodedAddr       string
 		priKey            []crypto.PrivateKey
+		blsPriKey         []*crypto.BLS12381PrivateKey
 		chain             ChainManager
 		blockDeserializer *block.Deserializer
 		broadcastHandler  scheme.Broadcast
@@ -286,6 +287,13 @@ func (b *Builder) SetConfig(cfg BuilderConfig) *Builder {
 // SetPriKey sets the private key
 func (b *Builder) SetPriKey(priKeys ...crypto.PrivateKey) *Builder {
 	b.priKey = priKeys
+	return b
+}
+
+// SetBLSPriKey sets the BLS12-381 producer private keys, aligned 1:1 with
+// the ECDSA keys set via SetPriKey.
+func (b *Builder) SetBLSPriKey(blsPriKeys ...*crypto.BLS12381PrivateKey) *Builder {
+	b.blsPriKey = blsPriKeys
 	return b
 }
 
@@ -360,6 +368,7 @@ func (b *Builder) Build() (*RollDPoS, error) {
 		b.delegatesByEpochFunc,
 		b.proposersByEpochFunc,
 		b.priKey,
+		b.blsPriKey,
 		b.clock,
 		b.cfg.Genesis.BeringBlockHeight,
 	)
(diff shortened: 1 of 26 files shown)


consensus/scheme/rolldpos/rolldpos.go
```diff
@@ -184,6 +185,9 @@ func (r *RollDPoS) ValidateBlockFooter(blk *block.Block) error {
 		return err
 	}
 	blkHash := blk.HashBlock()
+	if blk.Footer.IsAggregated() {
```
envestcc, line 188: When verifying whether there are enough endorsements, for the two handling approaches before and after the hardfork, I think it would be best to use the same piece of verification code for cases where votes exceed 2/3. Currently, it is implemented separately in two places, which is not good for maintainability.

envestcc:
Done in `ff589bf`. Extracted the 2/3 quorum check into a single primitive on `roundCtx` — `isMajorityCount(count int) bool`. Both paths now share the same threshold rule:

- Pre-fork `isMajority([]Endorsement)` becomes a thin wrapper over `isMajorityCount(len(endorsements))`.
- Post-fork `validateAggregatedFooter` calls `round.isMajorityCount(len(signers))` directly, dropping the inlined `3*signers > 2*numDelegates` arithmetic.

Quorum-rule changes (if we ever revisit 2/3) stay in one place.

PR #4850, opened 2026-06-05
feat(protocol): fixed system signer key for system actions (BLS Producer Identity)

envestcc:
> Stacks on #4847 (and the IIP-52 chain #4841 → #4842 → #4843 → #4847). The diff includes those PRs' changes until they merge; review only the last two commits (`feat(protocol): protocol-fixed system signer key + actpool filter` and `feat(protocol): route system action signing through the fixed system signer`).

## Summary

Introduces a protocol-fixed ECDSA private key for sealing system actions (`GrantReward`, `PutPollResult`, `ScheduleCandidateDeactivation`), decoupling system-action issuance from the block producer's node-resident ECDSA key. First step toward the BLS Producer Identity follow-up to IIP-52 — a delegate node can drop its ECDSA producer key entirely once header signing is also migrated to BLS.

The key is derived deterministically from `keccak256("iotex.system.signer.v1")` and is intentionally public; the resulting `SealedEnvelope` is fully wire-compatible with the existing tx-hash pipeline, so `eth_getTransactionByHash`, explorer/wallet UIs, and indexers see post-fork system actions as ordinary signed transactions (with `from = SystemSenderAddress`).

The public-key risk is mitigated by two layers of defense:
1. `actpool` rejects any externally submitted envelope whose sender equals `SystemSenderAddress` (catches non-system-action types signed with the public key, complementing the existing `IsSystemAction` type filter).
2. Per-handler content rules (reward amount, poll-result content, deactivation eligibility) enforce action validity independently of signer identity — analogous to Bitcoin's coinbase tx.

Determinism of the signature is a hard prerequisite: validators must produce identical signature bytes for the same envelope or they would diverge on the tx hash. iotex's secp256k1 path uses libsecp256k1 (RFC 6979 deterministic-k); a regression test signs the same message 1000 times and asserts identical bytes.

## Changes

**Sys1 — protocol-fixed system signer key + actpool filter**
- `action/protocol/system_signer.go`: `systemSignerPrivKey` (private), `SystemSenderAddress` (exported), `SignAsSystem(envelope)` API
- `action/protocol/system_signer_determinism_test.go`: RFC 6979 determinism (1000 iterations), stable identity lock-in, `SignAsSystem` round-trip
- `actpool/actpool.go`: sender-based filter on `SystemSenderAddress`
- `actpool/actpool_test.go`: `TestActPool_RejectSystemSender` — re-derives the system key from the documented seed and verifies a `Transfer` sent from `SystemSenderAddress` is rejected

**Sys2 — route system action signing through the fixed key**
- `state/factory/statedb.go`: `Mint` sign closure dispatches on `EnableBLSAggregation` — post-fork calls `protocol.SignAsSystem`, pre-fork unchanged. Block header signing keeps using the producer ECDSA key until that path migrates to BLS (PR-Y2)
- `action/protocol/rewarding/protocol.go`: `GrantReward.Validate` accepts `Caller == SystemSenderAddress` post-fork, `Caller == blkCtx.Producer` pre-fork
- `action/protocol/poll/util.go`: `PutPollResult.validate` same gating
- `action/protocol/rewarding/protocol_test.go`: pinned `ToBeEnabledBlockHeight = ^uint64(0)` in the existing test so it exercises the pre-fork branch (it was relying on a zero-value `Genesis` which now activates `EnableBLSAggregation` at height 0); added `TestProtocol_Validate_PostFork` covering the post-fork branch (reject `Caller==Producer`, accept `Caller==SystemSenderAddress`)

## Activation gate

Shares `EnableBLSAggregation` (= `IsToBeEnabled(height)`) with IIP-52. BLS Producer Identity activates with IIP-52's fork height: BLS-signed headers presume IIP-52's `ActiveCandidates` BLS filter, and shipping them together removes any intermediate state.

## What's not in this PR

- Block header BLS signing — separate PR (Y2)
- Config-schema changes to drop `producerPrivKey` — separate PR (Y5); the field is still loaded pre-fork
- `handleCandidateUpdateByOperator` post-fork gate — separate PR (Y7)
- End-to-end fork-transition tests — final PR once the rest of the stack lands

## Test plan
- [ ] CI green on the IIP-52 stack base (#4847)
- [x] `go test ./action/protocol/ ./actpool/ ./state/factory/ ./action/protocol/rewarding/ ./action/protocol/poll/` — all passing locally
- [x] `go vet` on the touched packages — clean
- [ ] Sync a fresh node against current testnet and confirm pre-fork block validation is unaffected
- [ ] On a local fork that flips `ToBeEnabledBlockHeight` mid-chain, confirm system actions post-fork show `from = SystemSenderAddress` via `eth_getTransactionByHash`

### DIFF
--- a/action/protocol/poll/util.go
+++ b/action/protocol/poll/util.go
@@ -77,8 +77,18 @@ func validate(ctx context.Context, sr protocol.StateReader, p Protocol, act acti
 	actionCtx := protocol.MustGetActionCtx(ctx)
 	blkCtx := protocol.MustGetBlockCtx(ctx)
 
-	if blkCtx.Producer.String() != actionCtx.Caller.String() {
-		return errors.New("Only producer could create this protocol")
+	// Post-fork (BLS Producer Identity IIP), PutPollResult is sealed by the
+	// protocol-fixed system signer, so the caller is SystemSenderAddress
+	// regardless of which delegate produced the block. Pre-fork the caller
+	// is the block producer.
+	if protocol.MustGetFeatureCtx(ctx).EnableBLSAggregation {
+		if actionCtx.Caller.String() != protocol.SystemSenderAddress.String() {
+			return errors.New("only the system signer could create this protocol")
+		}
+	} else {
+		if blkCtx.Producer.String() != actionCtx.Caller.String() {
+			return errors.New("Only producer could create this protocol")
+		}
 	}
 	proposedDelegates := ppr.Candidates()
 	if err := validateDelegates(proposedDelegates); err != nil {
--- a/action/protocol/poll/util.go
+++ b/action/protocol/poll/util.go
@@ -81,7 +81,7 @@ func validate(ctx context.Context, sr protocol.StateReader, p Protocol, act acti
 	// protocol-fixed system signer, so the caller is SystemSenderAddress
 	// regardless of which delegate produced the block. Pre-fork the caller
 	// is the block producer.
-	if protocol.MustGetFeatureCtx(ctx).EnableBLSAggregation {
+	if protocol.MustGetFeatureCtx(ctx).UseSystemSigner {
 		if actionCtx.Caller.String() != protocol.SystemSenderAddress.String() {
 			return errors.New("only the system signer could create this protocol")
 		}
--- a/actpool/actpool.go
+++ b/actpool/actpool.go
@@ -296,6 +296,17 @@ func (ap *actPool) add(ctx context.Context, act *action.SealedEnvelope) error {
 	if action.IsSystemAction(act) {
 		return action.ErrInvalidAct
 	}
+	// The protocol-fixed system signer key (BLS Producer Identity IIP) is
+	// public, so anyone holding it can produce a syntactically valid
+	// SealedEnvelope with SystemSenderAddress as the sender — including for
+	// non-system-action types that the IsSystemAction check above would not
+	// catch. Such actions are only legitimate when produced internally by the
+	// state factory at block construction; any externally received envelope
+	// from this address is by definition invalid.
+	if act.SenderAddress() != nil && act.SenderAddress().String() == protocol.SystemSenderAddress.String() {
+		_actpoolMtc.WithLabelValues("systemSender").Inc()
+		return action.ErrInvalidAct
+	}
 
 	if err := checkSelpData(act); err != nil {
 		return err
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -173,6 +173,15 @@ type (
 		AlwaysWriteCachedContract bool
 		NoCandidateExitQueue      bool
 		EnableBLSAggregation      bool
+		// UseSystemSigner gates the BLS Producer Identity change where
+		// system actions (GrantReward, PutPollResult,
+		// ScheduleCandidateDeactivation) are sealed by the protocol-fixed
+		// system signer key rather than by the block producer's ECDSA key.
+		// Shares the IsToBeEnabled height with EnableBLSAggregation today
+		// but is a semantically distinct switch — keep them separate so
+		// call sites read self-descriptively and so the two can diverge if
+		// ever required by a hotfix or fork-height adjustment.
+		UseSystemSigner bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
@@ -348,6 +357,7 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			AlwaysWriteCachedContract:               !g.IsYap(height),
 			NoCandidateExitQueue:                    !g.IsYap(height),
 			EnableBLSAggregation:                    g.IsToBeEnabled(height),
+			UseSystemSigner:                         g.IsToBeEnabled(height),
 		},
 	)
 }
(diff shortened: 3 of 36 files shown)


action/protocol/poll/util.go
```diff
@@ -77,8 +77,18 @@ func validate(ctx context.Context, sr protocol.StateReader, p Protocol, act acti
 	actionCtx := protocol.MustGetActionCtx(ctx)
 	blkCtx := protocol.MustGetBlockCtx(ctx)
 
-	if blkCtx.Producer.String() != actionCtx.Caller.String() {
-		return errors.New("Only producer could create this protocol")
+	// Post-fork (BLS Producer Identity IIP), PutPollResult is sealed by the
+	// protocol-fixed system signer, so the caller is SystemSenderAddress
+	// regardless of which delegate produced the block. Pre-fork the caller
+	// is the block producer.
+	if protocol.MustGetFeatureCtx(ctx).EnableBLSAggregation {
```
envestcc: You can use a standalone feature instead of sharing EnableBLSAggregation.

actpool/actpool.go
```diff
@@ -296,6 +296,17 @@ func (ap *actPool) add(ctx context.Context, act *action.SealedEnvelope) error {
...
 		return action.ErrInvalidAct
 	}
+	// The protocol-fixed system signer key (BLS Producer Identity IIP) is
+	// public, so anyone holding it can produce a syntactically valid
+	// SealedEnvelope with SystemSenderAddress as the sender — including for
+	// non-system-action types that the IsSystemAction check above would not
+	// catch. Such actions are only legitimate when produced internally by the
+	// state factory at block construction; any externally received envelope
+	// from this address is by definition invalid.
+	if act.SenderAddress() != nil && act.SenderAddress().String() == protocol.SystemSenderAddress.String() {
```
envestcc, line 306: Do we really need to prohibit this system sender from sending transactions externally? That should depend on how we view the system sender. If we strictly regard it as a system-level account, then we should not accept externally signed transactions from it. However, we could also treat it as a regular account, except that the system uses it to sign system actions. In that case, we could allow it to sign external transactions as well. Could you explain your thoughts?

envestcc: Good question — I lean toward keeping the strict (a) interpretation (system-level account, no external txs). Three reasons:

**1. Preserve "from SystemSenderAddress = protocol-generated" as an architectural invariant.** This is the strongest argument. EIP-4788 uses its SystemAddress exactly this way: anything sent from `0xfff...fffe` is by definition consensus-issued, never user-issued. If we ever add a protocol or contract that trusts `msg.sender == SystemSenderAddress` as a signal (e.g., a future precompile, a permissioned upgrade hook, a "system-only" callable on a builtin contract), allowing external txs would silently break that trust the moment any user submits one. The invariant costs us one line of mempool filtering today and a defensive test; reclaiming it later is a much harder fork-time conversation.

**2. The address has no real owner.** The private key is documented in the IIP and committed to source; anyone can sign as SystemSenderAddress. Under model (b), the explorer / wallet / indexer view would show this single address as the apparent sender of arbitrary, unrelated transactions from anonymous third parties. That's confusing UX with no real benefit — there's no "user" of this account whose transactions are meaningful.

**3. Nonce model stays trivially deterministic.** Today, system-action nonces tick up purely as a function of how many system actions the protocol has emitted, which is fully deterministic across validators. Under (b) we'd still be deterministic in principle (mempool txs only count once they're in a block), but we'd be relying on workingset.go's ordering invariant ("system actions run after user actions") to keep nonces sane. That's a fragile constraint to hand to future maintainers — anyone refactoring action ordering could quietly break system-action issuance.

The mempool filter is essentially zero cost — one branch + a metric label — and it locks in (1) as a property the rest of the codebase can rely on without further checks.

I'd suggest we keep the filter, and I'll fold the rationale into a comment at the filter site + the IIP draft (Section S) so future readers don't have to re-derive it. Happy to revisit if you'd prefer (b); concretely that's just removing the `SystemSenderAddress` check in actpool.add and the corresponding test.

envestcc: In that case, not only do we need to filter out transactions sent by it in the actpool, but we also need to verify that the sender of user transactions cannot be it when validating the block.

envestcc: Agreed — good catch. The actpool filter only stops the mempool entrance; a validator could embed a `from=SystemSenderAddress` Transfer (or Execution) directly in its proposed block and current validation would let it through.

Pushed `d696e2f61` adding the symmetric check on the block-validation path: `GenericValidator.ValidateWithState` now rejects any non-system-action `SealedEnvelope` whose sender is `SystemSenderAddress`. The check sits after the existing `IsSystemAction` early-return so legitimate `GrantReward` / `PutPollResult` / `ScheduleCandidateDeactivation` envelopes signed by the same key are unaffected. Feature-gated by `UseSystemSigner` to preserve pre-fork behavior.

`ValidateWithState` runs from `workingset` on both the block-processing path and the `Mint` action-picking path, so the "from SystemSenderAddress = protocol-generated" invariant now holds end-to-end (mempool entrance, our own block production, peer-block validation). Added a focused test covering the post-fork reject, pre-fork allow, and legitimate-system-action pass-through.

PR #4851, opened 2026-06-05
feat(block): length-based dispatch for Header signature scheme (BLS Producer Identity)

envestcc:
> Stacks on #4847 (and the IIP-52 chain #4841 → #4842 → #4843 → #4847). The diff includes those PRs' changes until they merge; review only the last commit (`feat(block): length-based dispatch for Header signature scheme`).

## Summary

Foundational decoupling step for the BLS Producer Identity follow-up to IIP-52. Removes `crypto.PublicKey` (secp256k1-shaped) from `Header` storage so the existing accessors transparently handle BLS-signed headers post-fork. No behavior change for pre-fork blocks.

The signature scheme is implied by `len(blockSig)`:
- 65 bytes → secp256k1 (pre-fork)
- 96 bytes → BLS12-381 G2 compressed (post-fork)

## Changes

- `blockchain/block/header.go`:
  - `Header.pubkey crypto.PublicKey` → `Header.producerPubkey []byte` (raw storage)
  - `Header.PublicKey()` returns nil for non-secp256k1 producer pubkeys
  - New `Header.ProducerPubKey() []byte` exposes raw bytes regardless of scheme
  - `Header.VerifySignature()` dispatches on `len(blockSig)`; BLS path uses `crypto.BLS12381PublicKeyFromBytes + Verify` against `HashHeaderCore`
  - `Header.ProducerAddress()` dispatches on `len(blockSig)`; BLS path returns `hex.EncodeToString(producerPubkey)` (BLS pubkeys have no account semantics in iotex and are intentionally not derived into a 20-byte iotex address)

- `blockchain/block/builder.go`, `blockchain/block/testing.go`: write the raw pubkey bytes directly into `producerPubkey` (previously a typed `crypto.PublicKey` assignment).

- `blockchain/block/header_bls_test.go` (new): pre-fork secp256k1 path, post-fork BLS path, wrong-scheme combinations (BLS-length sig with secp256k1 pubkey and vice versa), empty pubkey rejection, proto round-trip.

## Why dispatch on signature length rather than on a scheme tag

The proto schema is unchanged (no new field). `len(blockSig)` is unambiguous (65 vs 96 bytes), survives proto round-trip without coordination, and keeps pre-fork verification on its existing fast path. A scheme tag would require a proto bump and additional validation.

## Why not extend `iotex-address` to handle BLS pubkey strings

BLS public keys carry no account semantics in iotex (no balance, no tx sender role). The `iotex-address` interface contract is "20-byte public-key hash"; adding a 48-byte BLS variant would silently truncate through `hash.BytesToHash160`, creating phantom account state. The string returned by `ProducerAddress()` is opaque to all current consumers (map keys, `==` comparison), so length-based dispatch is sufficient without an interface change.

## What's not in this PR

- Block-header signing path (`Builder.SignAndBuild` / `Mint`) still uses secp256k1; the BLS signer integration is a separate PR (Y2)
- Consumer fork-branching (reward attribution, `ValidateBlockFooter`) — separate PR (Y4)
- System-action signing migration — handled by #4850 (Sys1+Sys2)
- Config-schema changes to drop `producerPrivKey` — separate PR (Y5)

## Test plan

- [x] `go test ./blockchain/block/` — all passing locally (32 tests, including the new header_bls_test.go)
- [x] `go build ./...` — clean
- [ ] CI green on the IIP-52 stack base (#4847)
- [ ] Pre-fork header serialization is byte-identical across the change (round-trip test in `TestSerDesHeader`, unchanged)

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -172,6 +172,7 @@ type (
 		// contracts are committed and written back
 		AlwaysWriteCachedContract bool
 		NoCandidateExitQueue      bool
+		EnableBLSAggregation      bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
@@ -346,6 +347,7 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			PrePectraEVM:                            !g.IsYap(height),
 			AlwaysWriteCachedContract:               !g.IsYap(height),
 			NoCandidateExitQueue:                    !g.IsYap(height),
+			EnableBLSAggregation:                    g.IsToBeEnabled(height),
 		},
 	)
 }
--- a/action/protocol/staking/protocol.go
+++ b/action/protocol/staking/protocol.go
@@ -972,9 +972,17 @@ func (p *Protocol) ActiveCandidates(ctx context.Context, sr protocol.StateReader
 		if err != nil {
 			return nil, err
 		}
-		if active {
-			cand = append(cand, list[i])
+		if !active {
+			continue
 		}
+		// Post-fork (IIP-52): drop candidates without a registered BLS public
+		// key so the per-block aggregate signature has a well-defined signer
+		// set. Pre-fork the BLSPubKey field is empty for every candidate and
+		// this filter is a no-op.
+		if fCtx.EnableBLSAggregation && len(list[i].BLSPubKey) == 0 {
+			continue
+		}
+		cand = append(cand, list[i])
 	}
 	return cand.toStateCandidateList(protocol.MustGetFeatureWithHeightCtx(ctx).CandidateWithoutIdentityStorage(height))
 }
--- a/blockchain/block/block.go
+++ b/blockchain/block/block.go
@@ -74,15 +74,47 @@ func (b *Block) RunnableActions() RunnableActions {
 	return RunnableActions{actions: b.Actions, txHash: b.txRoot}
 }
 
-// Finalize creates a footer for the block
+// Finalize creates a footer for the block using the per-delegate endorsements
+// path. Used pre-fork (and as the legacy entry point); post-fork blocks are
+// finalized via FinalizeWithAggregate instead.
 func (b *Block) Finalize(endorsements []*endorsement.Endorsement, ts time.Time) error {
-	if len(b.endorsements) != 0 {
+	if b.isFinalized() {
 		return errors.New("the block has been finalized")
 	}
 	b.endorsements = endorsements
 	b.commitTime = ts
 
 	return nil
+}
+
+// FinalizeWithAggregate creates a footer for the block using the BLS12-381
+// aggregate signature path (IIP-52). The aggregate signature is a single
+// 96-byte BLS sig over the per-block COMMIT vote; signerBitmap identifies
+// which epoch delegates contributed (bit i = delegate i in the epoch's
+// delegate list, LSB-first within each byte).
+//
+// Same one-shot contract as Finalize: a block can only be finalized once.
+func (b *Block) FinalizeWithAggregate(aggregatedSignature, signerBitmap []byte, ts time.Time) error {
+	if b.isFinalized() {
+		return errors.New("the block has been finalized")
+	}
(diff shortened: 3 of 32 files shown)


envestcc:
Thanks for the back-and-forth on this. Three concrete updates from the discussion, all on commit `893cc171b`:

## 1. `[]byte` → narrow `block.Verifier {Bytes; Verify}` interface

Your push to preserve type safety was right; the answer turned out not to be "make BLS satisfy `crypto.PublicKey`" but "carve out a smaller interface that both schemes satisfy honestly."

Why not just extend `*crypto.BLS12381PublicKey` to implement `crypto.PublicKey`:

- `Address() address.Address` — `address.Address.Bytes()` is documented as "underlying 20-byte public key hash" and is consumed throughout the codebase under that assumption. A BLSAddress that returns 48 bytes silently truncates in `common.BytesToAddress` / `hash.BytesToHash160` consumers (EVM Coinbase, state-key derivation, ABI encoding, ~12 sites total — no compiler help to find them all). A BLSAddress that returns 20 bytes (hash160 of BLS pubkey) re-introduces the "phantom BLS account" the IIP audit explicitly rejected. Either way the abstraction lies.
- `Hash() []byte` — same trap: callers feed it into `BytesToHash160`.
- `EcdsaPublicKey() interface{}` — the method name itself says ECDSA.

So Header stores `block.Verifier {Bytes() []byte; Verify(msg, sig []byte) bool}`. Both `crypto.PublicKey` and `*crypto.BLS12381PublicKey` already satisfy it. Length-based dispatch now lives only at `LoadFromBlockHeaderProto` (the wire→typed boundary); everywhere else in `header.go` the code reads `h.pubkey.Verify(...)` / `h.pubkey.Bytes()` / type-switches when it actually needs the address form.

Concrete improvements over the prior `[]byte`-based version:

- `VerifySignature` no longer length-switches; it just calls `h.pubkey.Verify`
- `ProducerAddress` type-switches on the stored Verifier (`if blsPK, ok := ...` ) — intent is "I'm a BLS key, hex-encode me" rather than "I have a 96-byte sig, hex-encode me"
- `PublicKey()` accessor no longer re-parses bytes via `BytesToPublicKey` on every call; it type-asserts

## 2. EVM Coinbase → `Candidate.Reward` (Eth2 `fee_recipient` model)

Fully agree — `coinbase = blkCtx.Producer` is exactly the conflation Ethereum had pre-merge and split post-merge. Adopting the Eth2 model is the right call for iotex post-fork.

Plumbing (this lands in PR-Y4, not this PR — Y1 stays focused on Header decoupling):

| Layer | Change |
|---|---|
| `BlockCtx` schema | Add `FeeRecipient address.Address` (or repurpose `Producer` to fee-recipient semantics) + `ProducerPubKey []byte` |
| `blockchain.go` BlockCtx assembly | post-fork: state lookup `candidate := lookupByBLSPubKey(header.ProducerPubKey())`; `FeeRecipient = candidate.Reward` |
| `evm.go` | `Coinbase: common.BytesToAddress(blkCtx.FeeRecipient.Bytes())` |
| Reward/poll/slasher identity matching | `blkCtx.ProducerPubKey` (BLS pubkey direct compare) |

Caught as a side-effect of this discussion: the other `blk.PublicKey().Address()` consumers (`statedb.go`, `blockindexer.go`, `coreservice.go` API) each need an explicit post-fork answer (some want `Reward`, some want pubkey hex, some want both). This is the rest of Y4. The narrow `Verifier` interface forces these decisions to happen at the call site instead of being silently auto-resolved by a "BLS-flavored" `Address()` — which is the structural reason for not putting `Address` on the storage interface.

## 3. IIP draft

I've added a Section R "EVM Coinbase / Fee Recipient" capturing the Eth2 model, the two BlockCtx layout options (repurpose `Producer` vs add `FeeRecipient`), the state-lookup mechanics, and the per-callsite resolution table. Will land alongside the Y4 PR.

Ready for another look on the Header refactor whenever you have cycles.

PR #4852, opened 2026-06-07
fix(evm): only treat 3-topic logs as in-contract transfers (follow-up to #4844)

guo:
## Summary

Follow-up hardening to #4844. `StateDBAdapter.AddLog` still calls
`log.Panic("Invalid in contract transfer topics")` when a log's first topic
equals the in-contract-transfer sentinel but the topic count is not exactly 3.

`_inContractTransfer` is `hash.BytesToHash256([]byte{byte(IN_CONTRACT_TRANSFER)})`,
i.e. the all-zero topic, and log topics are EVM-controlled. #4844 added a
`len(topics) > 0` guard that removed the empty-topic (LOG0) index-out-of-range,
but a 1- or 2-topic log whose first topic is all-zero still enters the branch and
hits the `len != 3` panic.

Genuine in-contract transfers are always emitted by the protocol with exactly
three topics (type, from, to), so the correct guard is to enter the special
handling only when `len(topics) == 3`. Any other shape is an ordinary log and is
appended as such. This removes the remaining panic without changing behavior for
real in-contract-transfer logs or for any normal log (real event signatures are
non-zero keccak hashes, so honest traffic never carries an all-zero first topic).

## Change

- `AddLog`: gate the in-contract-transfer handling on
  `len(topics) == 3 && topics[0] == _inContractTransfer`; drop the now-dead
  `len != 3` panic. Malformed shapes fall through to the normal log append.

## Test

- Added `TestAddLogInContractTransferTopicShape`: asserts `AddLog` does not panic
  on a 1-topic log whose topic0 is the sentinel (recorded as a normal log), and
  that a well-formed 3-topic in-contract transfer is still recorded as a
  transaction log with the correct sender/recipient.
- `go test ./action/protocol/execution/evm/ -race` — full package passes.

### DIFF
--- a/action/protocol/execution/evm/evmstatedbadapter.go
+++ b/action/protocol/execution/evm/evmstatedbadapter.go
@@ -926,10 +926,12 @@ func (stateDB *StateDBAdapter) AddLog(evmLog *types.Log) {
 		copy(topic[:], evmTopic.Bytes())
 		topics = append(topics, topic)
 	}
-	if len(topics) > 0 && topics[0] == _inContractTransfer {
-		if len(topics) != 3 {
-			log.T(stateDB.ctx).Panic("Invalid in contract transfer topics")
-		}
+	// A genuine in-contract transfer is emitted by the protocol with exactly
+	// three topics (type, from, to). Log topics are EVM-controlled, so only
+	// treat the log as an in-contract transfer when the shape matches exactly;
+	// any other shape (including a forged sentinel topic on a 1- or 2-topic
+	// log) falls through to the normal log append below.
+	if len(topics) == 3 && topics[0] == _inContractTransfer {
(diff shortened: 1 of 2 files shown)


raullenchai:
Closing as a duplicate of #4871. Both fix the same `log.Panic("Invalid in contract transfer topics")` in `StateDBAdapter.AddLog` for the malformed (non-3-topic) reserved-topic case. #4871 takes the same fix but gates the drop-instead-of-panic behind a hardfork height (Zanzibar): because the pre-fix behavior is a panic (an un-upgraded node cannot process such a block), an ungated flip could let an upgraded producer commit a block that permanently bricks un-upgraded nodes, so coordinated activation is needed. Consolidating on #4871 to avoid the overlap — thanks @guo, your diagnosis here was spot on and fed directly into it.

PR #4854, opened 2026-06-11
fix(staking): BLS proof-of-possession at candidate register/update (CRITICAL, pre-IIP-52)

envestcc:
> **Security severity: CRITICAL, latent.** Replaces #4853 — this PR is based directly on master, NOT stacked on the IIP-52 PR chain. The vulnerable registration path (\`crypto.BLS12381PublicKeyFromBytes\` only, no PoP) is already live on master; IIP-52 turns the latent bug into a live exploit, but the un-attested BLS pubkeys can be collected starting now.
>
> Depends on **iotex-proto#173** (adds the \`blsPop\` field on \`CandidateBasicInfo\`).

## Why land BEFORE IIP-52

Master today has the full BLS registration infrastructure: \`Candidate.BlsPubKey\` proto field 5, \`candidateRegisterWithBLS\` ABI method, \`crypto.BLS12381PublicKeyFromBytes\` (format + subgroup check only, no PoP). Every BLS pubkey that gets registered between now and the IIP-52 activation block is a potential attack vector — once \`FastAggregateVerify\` goes live, all un-attested pubkeys become exploitable.

Activating \`EnforceBLSPoP\` BEFORE the IIP-52 fork means:

- **New registrations from this PR onward all carry verified PoPs**
- The pre-fork migration scope shrinks to existing pubkeys registered before this PR lands
- IIP-52 doesn't have to ship the PoP check — it just relies on the gate already being in place

## The bug

IIP-52's footer aggregation calls

\`\`\`go
crypto.BLSAggregateSignatureFromBytes(sig).Verify(pubKeys, msg)
// → blst.FastAggregateVerify(true, pubKeys, msg, dst)
\`\`\`

with DST \`BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_\`. The \`_POP_\` suffix names the IRTF Proof-of-Possession ciphersuite, sound only if every registered pubkey arrived with a possession proof.

Today \`action/candidate_register.go\` and \`action/candidate_update.go\` validate only:

\`\`\`go
_, err := crypto.BLS12381PublicKeyFromBytes(blsPubKey)  // format + subgroup
\`\`\`

No PoP. The proto has no PoP field.

## Attack (becomes exploitable when IIP-52 activates)

A single active delegate:
1. Reads other delegates' BLS pubkeys from chain
2. Picks any secret \`x\` they control
3. Computes \`pk_rogue = g^x − Σ(other pubkeys)\` in G1 (public information only)
4. Registers \`pk_rogue\` via \`candidateRegisterWithBLS\` — current handler accepts
5. After IIP-52 activates, submits a footer with all-ones bitmap and \`σ = sign(x, msg)\`
6. \`FastAggregateVerify\` collapses \`Σ pk_i + pk_rogue\` to \`g^x\`. \`σ\` verifies. \`isMajorityCount\` passes.

**One delegate forges a 2/3+ quorum certificate → finalizes arbitrary blocks → consensus safety break.**

## The fix (Eth2 \`process_deposit\` pattern)

\`\`\`go
// action/protocol/staking/bls_pop.go
const blsPopDomain = \"IOTEX_BLS_POP_v1\"

func BLSPopSigningRoot(blsPubKey []byte, ownerAddress address.Address) []byte {
    h := sha256.New()
    h.Write([]byte(blsPopDomain))
    h.Write(blsPubKey)
    h.Write(ownerAddress.Bytes())
    return h.Sum(nil)
}
\`\`\`

Three values bind into the signed message:

| Binding | Closes |
|---|---|
| \`blsPubKey\` | Rogue key — attacker without the secret cannot sign the canonical message |
| \`ownerAddress\` | Candidate substitution — PoP for owner A doesn't validate for owner B |
| \`IOTEX_BLS_POP_v1\` prefix | Cross-domain replay — PoP disjoint from any other BLS-signed iotex message; version suffix reserves room for a future fork to rotate |

## Implementation

- **Proto** (iotex-proto#173): \`bytes blsPop = 5\` on \`CandidateBasicInfo\`. go.mod replace updated to envestcc/iotex-proto@7a486f6.

- **FeatureCtx.EnforceBLSPoP**: \`g.IsToBeEnabled(height)\`. Standalone flag — does NOT share state with any IIP-52 flag, so this PR is fully independent.

- **action.CandidateRegister / CandidateUpdate**: carry \`blsPop\` through constructors, Proto/LoadProto, accessors. \`NewCandidateRegisterWithBLS\` and \`NewCandidateUpdateWithBLS\` now take a \`blsPop\` parameter; pre-fork callers may pass nil.

- **handlers.go**: \`handleCandidateRegister\`, \`handleCandidateUpdate\`, and \`handleCandidateUpdateByOperator\` call \`VerifyBLSPop\` before writing \`Candidate.BLSPubKey\` when \`EnforceBLSPoP\` is active. Failure → \`ReceiptStatus_ErrUnauthorizedOperator\`.

- **bls_pop_test.go**: 5 cases including \`TestBLSPop_RogueKeyAttackBlocked\` as the regression guard — models the attacker registering a pubkey for which they don't have the secret; demonstrates no PoP they can produce (under any secret they control) validates against that pubkey; control case confirms a legitimate registrant succeeds.

## Migration (out of scope for this PR)

Existing pre-PoP-merge candidate records have no PoP. Before \`EnforceBLSPoP\` activates, all active candidates must either:

- Submit a \`candidateUpdate\` carrying a valid \`blsPop\` (grace period), or
- Be dropped from \`ActiveCandidates\` at the fork height via a state migration

This is a coordination problem (talking to delegates, picking a grace window), not a code problem. IIP draft will spell it out.

## Out of scope for this PR (follow-ups)

- **ioctl / SDK wiring**: \`stake2register.go\` and \`stake2update.go\` pass nil for \`blsPop\` with TODO markers. They need a flag to ingest a BLS private key and derive the PoP. Pre-fork accepted; post-fork rejected.

- **Dedicated receipt status code**: reusing \`ErrUnauthorizedOperator\` (closest semantic match) to keep proto-change scope minimal. Adding \`ErrInvalidBLSPoP\` is another proto bump and not necessary for correctness.

## Test plan
- [x] \`go test ./action/protocol/staking/ ./action/\` — passing locally (470 cases including 5 new PoP tests)
- [x] Rogue-key regression guard: \`TestBLSPop_RogueKeyAttackBlocked\` ✓
- [ ] CI green on master + iotex-proto#173
- [ ] Coordinate migration plan with delegates before flipping fork height

### DIFF
--- a/action/protocol/staking/bls_pop.go
+++ b/action/protocol/staking/bls_pop.go
@@ -30,14 +30,14 @@ const blsPopDomain = "IOTEX_BLS_POP_v1"
 // must be computed over for the given candidate.
 //
 // Binding three values into the signed message — the domain tag, the BLS
-// public key itself, and the candidate owner address — closes the rogue
-// key attack and two related replays:
+// public key itself, and the candidate's identity address — closes the
+// rogue key attack and two related replays:
 //
 //   - blsPubKey: forces the signer to know the private key for THIS
 //     specific BLS pubkey. A rogue pubkey constructed as
 //     g^x − Σ(other pubkeys) cannot produce a valid PoP because the
 //     attacker does not know its discrete log.
-//   - ownerAddress: prevents two distinct candidates from sharing a
+//   - candidateID: prevents two distinct candidates from sharing a
 //     single BLS keypair (and thus a single PoP) without each owner
 //     independently re-attesting; also prevents a PoP submitted for
 //     candidate A from being replayed for candidate B by a man-in-the-
@@ -45,31 +45,47 @@ const blsPopDomain = "IOTEX_BLS_POP_v1"
 //   - blsPopDomain: keeps PoP signatures disjoint from consensus
 //     signatures, future PoP schemes, and any other BLS-signed iotex
 //     message that may exist or be added later.
-func BLSPopSigningRoot(blsPubKey []byte, ownerAddress address.Address) []byte {
+//
+// candidateID is the candidate's stable identity:
+//   - At register: the owner address declared in the action. For
+//     non-collision registrations this becomes c.Identifier verbatim
+//     (see generateCandidateID — it returns owner directly when free),
+//     so a PoP signed at register time matches the identifier the
+//     candidate will carry post-fork.
+//   - At update: c.GetIdentifier(), which returns the immutable
+//     Identifier for post-Xingu candidates and falls back to c.Owner
+//     for pre-Xingu records. This means a candidate that has been
+//     transferred to a new owner still uses its original identity for
+//     PoP, so the binding is stable across ownership transfers.
+func BLSPopSigningRoot(blsPubKey []byte, candidateID address.Address) []byte {
 	h := sha256.New()
 	h.Write([]byte(blsPopDomain))
 	h.Write(blsPubKey)
-	if ownerAddress != nil {
-		h.Write(ownerAddress.Bytes())
+	if candidateID != nil {
+		h.Write(candidateID.Bytes())
 	}
 	return h.Sum(nil)
 }
 
 // SignBLSPop produces a proof-of-possession for the given BLS private
-// key, binding it to the candidate owner address. Used by tooling
+// key, binding it to the candidate's identity. Used by tooling
 // (ioctl, SDK) to generate the bls_pop field on CandidateRegister /
 // CandidateUpdate transactions.
-func SignBLSPop(sk *crypto.BLS12381PrivateKey, ownerAddress address.Address) ([]byte, error) {
+//
+// At registration time pass the proposed owner address (which becomes
+// the candidate identifier); at update time pass the candidate's
+// existing identifier (c.GetIdentifier()).
+func SignBLSPop(sk *crypto.BLS12381PrivateKey, candidateID address.Address) ([]byte, error) {
 	if sk == nil {
 		return nil, errors.New("nil BLS private key")
 	}
 	pk := sk.PublicKey().Bytes()
-	return sk.Sign(BLSPopSigningRoot(pk, ownerAddress))
+	return sk.Sign(BLSPopSigningRoot(pk, candidateID))
 }
 
 // VerifyBLSPop verifies the proof-of-possession against the provided
-// pubkey and owner. Returns nil on success.
-func VerifyBLSPop(blsPubKey, blsPop []byte, ownerAddress address.Address) error {
+// pubkey and candidate identity. Returns nil on success.
+func VerifyBLSPop(blsPubKey, blsPop []byte, candidateID address.Address) error {
 	if len(blsPubKey) != crypto.BLSPubkeyLength {
 		return errors.Errorf("invalid BLS pubkey length: got %d, want %d", len(blsPubKey), crypto.BLSPubkeyLength)
 	}
@@ -80,7 +96,7 @@ func VerifyBLSPop(blsPubKey, blsPop []byte, ownerAddress address.Address) error 
 	if err != nil {
 		return errors.Wrap(err, "invalid BLS pubkey")
 	}
-	if !pk.Verify(BLSPopSigningRoot(blsPubKey, ownerAddress), blsPop) {
+	if !pk.Verify(BLSPopSigningRoot(blsPubKey, candidateID), blsPop) {
 		return errors.New("BLS proof-of-possession verification failed")
 	}
 	return nil
--- a/action/protocol/staking/bls_pop.go
+++ b/action/protocol/staking/bls_pop.go
@@ -1,0 +1,87 @@
+// Copyright (c) 2026 IoTeX Foundation
+// This source code is provided 'as is' and no warranties are given as to title or non-infringement, merchantability
+// or fitness for purpose and, to the extent permitted by law, all liability for your use of the code is disclaimed.
+// This source code is governed by Apache License 2.0 that can be found in the LICENSE file.
+
+package staking
+
+import (
+	"crypto/sha256"
+
+	"github.com/iotexproject/go-pkgs/crypto"
+	"github.com/iotexproject/iotex-address/address"
+	"github.com/pkg/errors"
+)
+
+// blsPopDomain is the iotex-specific domain separator for BLS
+// proof-of-possession signatures at candidate register / update time.
+//
+// Domain separation matters for two reasons:
+//   1. It prevents a PoP signature from being replayed as a consensus
+//      vote signature (or vice versa) — even though both schemes use
+//      the same BLS ciphersuite DST, the message they sign starts with
+//      this iotex-application-level tag and so the resulting signing
+//      root will never collide with a consensus signing root.
+//   2. The version suffix ("v1") reserves room for a future fork to
+//      rotate the PoP scheme without ambiguity.
+const blsPopDomain = "IOTEX_BLS_POP_v1"
+
+// BLSPopSigningRoot returns the bytes that a BLS proof-of-possession
+// must be computed over for the given candidate.
+//
+// Binding three values into the signed message — the domain tag, the BLS
+// public key itself, and the candidate owner address — closes the rogue
+// key attack and two related replays:
+//
+//   - blsPubKey: forces the signer to know the private key for THIS
+//     specific BLS pubkey. A rogue pubkey constructed as
+//     g^x − Σ(other pubkeys) cannot produce a valid PoP because the
+//     attacker does not know its discrete log.
+//   - ownerAddress: prevents two distinct candidates from sharing a
+//     single BLS keypair (and thus a single PoP) without each owner
+//     independently re-attesting; also prevents a PoP submitted for
+//     candidate A from being replayed for candidate B by a man-in-the-
+//     middle who repackages a CandidateRegister tx.
+//   - blsPopDomain: keeps PoP signatures disjoint from consensus
+//     signatures, future PoP schemes, and any other BLS-signed iotex
+//     message that may exist or be added later.
+func BLSPopSigningRoot(blsPubKey []byte, ownerAddress address.Address) []byte {
+	h := sha256.New()
+	h.Write([]byte(blsPopDomain))
+	h.Write(blsPubKey)
+	if ownerAddress != nil {
+		h.Write(ownerAddress.Bytes())
+	}
+	return h.Sum(nil)
+}
+
+// SignBLSPop produces a proof-of-possession for the given BLS private
+// key, binding it to the candidate owner address. Used by tooling
+// (ioctl, SDK) to generate the bls_pop field on CandidateRegister /
+// CandidateUpdate transactions.
+func SignBLSPop(sk *crypto.BLS12381PrivateKey, ownerAddress address.Address) ([]byte, error) {
+	if sk == nil {
+		return nil, errors.New("nil BLS private key")
+	}
+	pk := sk.PublicKey().Bytes()
+	return sk.Sign(BLSPopSigningRoot(pk, ownerAddress))
+}
+
+// VerifyBLSPop verifies the proof-of-possession against the provided
+// pubkey and owner. Returns nil on success.
+func VerifyBLSPop(blsPubKey, blsPop []byte, ownerAddress address.Address) error {
+	if len(blsPubKey) != crypto.BLSPubkeyLength {
+		return errors.Errorf("invalid BLS pubkey length: got %d, want %d", len(blsPubKey), crypto.BLSPubkeyLength)
+	}
+	if len(blsPop) != crypto.BLSAggregateSignatureLength {
+		return errors.Errorf("invalid BLS PoP length: got %d, want %d", len(blsPop), crypto.BLSAggregateSignatureLength)
+	}
+	pk, err := crypto.BLS12381PublicKeyFromBytes(blsPubKey)
+	if err != nil {
+		return errors.Wrap(err, "invalid BLS pubkey")
+	}
+	if !pk.Verify(BLSPopSigningRoot(blsPubKey, ownerAddress), blsPop) {
+		return errors.New("BLS proof-of-possession verification failed")
+	}
+	return nil
+}
--- a/action/protocol/staking/bls_pop.go
+++ b/action/protocol/staking/bls_pop.go
@@ -58,12 +58,18 @@ const blsPopDomain = "IOTEX_BLS_POP_v1"
 //     transferred to a new owner still uses its original identity for
 //     PoP, so the binding is stable across ownership transfers.
 func BLSPopSigningRoot(blsPubKey []byte, candidateID address.Address) []byte {
+	// Refuse to produce an "unbound" root. Allowing nil candidateID to
+	// silently fall through to a domain+pubkey-only digest would
+	// degrade the scheme to the weakest of its three bindings —
+	// exactly what an attacker reaching for a cross-candidate replay
+	// would hope for. Force callers to commit to a candidate identity.
+	if candidateID == nil {
+		return nil
+	}
 	h := sha256.New()
 	h.Write([]byte(blsPopDomain))
 	h.Write(blsPubKey)
-	if candidateID != nil {
-		h.Write(candidateID.Bytes())
-	}
+	h.Write(candidateID.Bytes())
 	return h.Sum(nil)
 }
 
@@ -79,6 +85,9 @@ func SignBLSPop(sk *crypto.BLS12381PrivateKey, candidateID address.Address) ([]b
 	if sk == nil {
 		return nil, errors.New("nil BLS private key")
 	}
+	if candidateID == nil {
+		return nil, errors.New("nil candidate ID; PoP must bind to a candidate identity")
+	}
 	pk := sk.PublicKey().Bytes()
 	return sk.Sign(BLSPopSigningRoot(pk, candidateID))
 }
@@ -91,6 +100,9 @@ func VerifyBLSPop(blsPubKey, blsPop []byte, candidateID address.Address) error {
 	}
 	if len(blsPop) != crypto.BLSAggregateSignatureLength {
 		return errors.Errorf("invalid BLS PoP length: got %d, want %d", len(blsPop), crypto.BLSAggregateSignatureLength)
+	}
+	if candidateID == nil {
+		return errors.New("nil candidate ID; PoP must bind to a candidate identity")
 	}
 	pk, err := crypto.BLS12381PublicKeyFromBytes(blsPubKey)
 	if err != nil {
--- a/action/protocol/staking/bls_pop.go
+++ b/action/protocol/staking/bls_pop.go
@@ -29,46 +29,52 @@ const blsPopDomain = "IOTEX_BLS_POP_v1"
 // BLSPopSigningRoot returns the bytes that a BLS proof-of-possession
 // must be computed over for the given candidate.
 //
-// Binding three values into the signed message — the domain tag, the BLS
-// public key itself, and the candidate's identity address — closes the
-// rogue key attack and two related replays:
+// The signed message is the domain tag plus the candidate's stable
+// identity. The BLS public key is intentionally NOT in the message:
 //
-//   - blsPubKey: forces the signer to know the private key for THIS
-//     specific BLS pubkey. A rogue pubkey constructed as
-//     g^x − Σ(other pubkeys) cannot produce a valid PoP because the
-//     attacker does not know its discrete log.
-//   - candidateID: prevents two distinct candidates from sharing a
-//     single BLS keypair (and thus a single PoP) without each owner
-//     independently re-attesting; also prevents a PoP submitted for
-//     candidate A from being replayed for candidate B by a man-in-the-
-//     middle who repackages a CandidateRegister tx.
-//   - blsPopDomain: keeps PoP signatures disjoint from consensus
-//     signatures, future PoP schemes, and any other BLS-signed iotex
-//     message that may exist or be added later.
+//   - The pairing verifier Verify(PK, msg, sig) already commits PK
+//     into the signature equation. An attacker who does not know
+//     sk_PK cannot produce a sig that verifies under PK over any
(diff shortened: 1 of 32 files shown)


envestcc:
Pushed a companion commit `4d5389792` adding BLS pubkey uniqueness at register/update.

## Why this belongs in the same PR

Even with PoP in place, two candidates registering the SAME blsPubKey breaks IIP-52's quorum-counting model. PoP only proves "I know the secret for this pubkey" — it does not prevent two delegates from independently producing valid PoPs for the same key (e.g. an actively shared keypair, or a delegate selling/leasing their secret).

When IIP-52 footer aggregation runs:

- The signer bitmap counts each delegate as 1 — committee size = N+1
- `FastAggregateVerify` sums pubkeys as a SET — committee size = N (duplicate dropped)
- `isMajorityCount(N+1)` may pass even when only N distinct keys actually signed → quieter cousin of the rogue-key attack: aggregation math still verifies, accounting is off, second delegate's stake-weight votes for free

Mitigation matches the existing collision-check pattern: `csm.ContainsBLSPubKey(blsPubKey, except)` returns true if any candidate **other than** \`except\` already holds the pubkey, parallel to \`ContainsName\` / \`ContainsOperator\`.

## Implementation

- \`CandidateCenter.ContainsBLSPubKey\` — linear scan over \`m.All()\`. Registration is sparse and active-set is bounded; not worth maintaining a 4th index map across the change/base commit flow.
- \`CandidateStateManager\` interface adds the new method; \`candSM\` delegates to \`candCenter\`.
- All 3 handler paths (\`handleCandidateRegister\`, \`handleCandidateUpdate\`, \`handleCandidateUpdateByOperator\`) call it after PoP verification, returning \`ErrCandidateConflict\` on collision.
- Gated under the same \`EnforceBLSPoP\` feature flag — pre-fork replay unchanged.
- Update paths exempt the candidate-being-updated (\`except = c.GetIdentifier()\`) so re-registering / updating with one's own pubkey is allowed (no-op pubkey case).

## Tests

- \`TestCandidateCenter_ContainsBLSPubKey\` covers nil / empty / no-match / match-with-nil-except / match-against-self (allowed) / match-against-other (rejected).
- Full staking + action suite: 472 passing.

envestcc:
Pushed \`3fbaa0dcd\` — bind update-path PoP to \`c.GetIdentifier()\` instead of \`c.Owner\`.

## Motivation

\`c.Owner\` shifts when ownership is transferred via \`CandidateTransferOwnership\`; \`c.GetIdentifier()\` is the candidate's stable handle. The previous owner-binding meant:

- a new owner doing BLS rotation right after receiving a transfer had to sign with the new owner address — not wrong, but conceptually awkward
- mid-flight concurrent transfers (\`Alice → Bob → Carol\` while Bob's update tx is in mempool) would invalidate Bob's PoP for a non-cryptographic reason

## Why this isn't a big change

\`generateCandidateID\` returns the owner address directly when free, only hash-deriving on collision. For the common case **identifier == registration-time owner**, so:

- **register**: PoP binds to \`act.OwnerAddress()\` (the future identifier in the non-collision case). Unchanged from before.
- **update**: PoP binds to \`c.GetIdentifier()\`:
  - post-Xingu non-collision: same value as the register-time owner — register and update PoPs use identical bindings
  - post-Xingu collision (edge case): identifier is a hash; updates consistently use it across the candidate's lifetime
  - pre-Xingu: \`GetIdentifier()\` falls back to \`c.Owner\` — identical to the previous behavior

The on-the-wire PoP bytes don't change. Only the conceptual binding shifts from "current owner" to "candidate identity."

## Implementation

- \`bls_pop.go\`: renamed the parameter from \`ownerAddress\` to \`candidateID\` throughout (\`BLSPopSigningRoot\`, \`SignBLSPop\`, \`VerifyBLSPop\`); docstring spells out what to pass at register vs update
- \`handlers.go\`: \`handleCandidateUpdate\` and \`handleCandidateUpdateByOperator\` switched from \`c.Owner\` to \`c.GetIdentifier()\` for the PoP binding (uniqueness check already used \`c.GetIdentifier()\`)
- \`bls_pop_test.go\`: \`TestBLSPop_RejectWrongOwner\` → \`TestBLSPop_RejectWrongCandidateID\`; added \`TestBLSPop_StableAcrossOwnershipTransfer\` locking in the stability property

## Test plan
- [x] 6 PoP tests pass including the new transfer-stability case
- [x] No on-the-wire changes — pre-fork replay unaffected

action/protocol/staking/candidate_statemanager.go
```diff
@@ -49,6 +49,13 @@ type (
 		ContainsName(string) bool
 		ContainsOwner(address.Address) bool
 		ContainsOperator(address.Address) bool
+		// ContainsBLSPubKey reports whether any candidate other than the
+		// one identified by `except` already holds the given BLS pubkey.
+		// `except` may be nil at registration time (no incumbent), in
+		// which case the check rejects on any match. Used to enforce one
+		// BLS pubkey per delegate, a hard requirement for IIP-52's
+		// FastAggregateVerify quorum-counting model.
+		ContainsBLSPubKey(blsPubKey []byte, except address.Address) bool
```
envestcc: The `ContainsBLSPubKey` should not need the `except` parameter.

action/protocol/staking/bls_pop.go
```diff
@@ -0,0 +1,103 @@
...
+//     Identifier for post-Xingu candidates and falls back to c.Owner
+//     for pre-Xingu records. This means a candidate that has been
+//     transferred to a new owner still uses its original identity for
+//     PoP, so the binding is stable across ownership transfers.
+func BLSPopSigningRoot(blsPubKey []byte, candidateID address.Address) []byte {
+	h := sha256.New()
+	h.Write([]byte(blsPopDomain))
+	h.Write(blsPubKey)
+	if candidateID != nil {
+		h.Write(candidateID.Bytes())
```
CoderZhi: return nil if candididateID == nil ?

action/protocol/staking/bls_pop.go
```diff
@@ -0,0 +1,103 @@
...
+//     non-collision registrations this becomes c.Identifier verbatim
+//     (see generateCandidateID — it returns owner directly when free),
+//     so a PoP signed at register time matches the identifier the
+//     candidate will carry post-fork.
+//   - At update: c.GetIdentifier(), which returns the immutable
+//     Identifier for post-Xingu candidates and falls back to c.Owner
+//     for pre-Xingu records. This means a candidate that has been
+//     transferred to a new owner still uses its original identity for
+//     PoP, so the binding is stable across ownership transfers.
+func BLSPopSigningRoot(blsPubKey []byte, candidateID address.Address) []byte {
```
CoderZhi: `blsPubKey` could be deleted. It will be needed during verification.

action/protocol/staking/bls_pop.go
```diff
@@ -0,0 +1,103 @@
...
+// SignBLSPop produces a proof-of-possession for the given BLS private
+// key, binding it to the candidate's identity. Used by tooling
+// (ioctl, SDK) to generate the bls_pop field on CandidateRegister /
+// CandidateUpdate transactions.
+//
+// At registration time pass the proposed owner address (which becomes
+// the candidate identifier); at update time pass the candidate's
+// existing identifier (c.GetIdentifier()).
+func SignBLSPop(sk *crypto.BLS12381PrivateKey, candidateID address.Address) ([]byte, error) {
+	if sk == nil {
```
CoderZhi, line 91: or return error if `candidateID == nil`

action/protocol/staking/candidate_statemanager.go
CoderZhi: BLSPubKeyOwner to return the owner address.

envestcc:
Pushed \`ab96bbc4b\` — closes the web3-path gap you flagged.

## What was broken

The PoP changes earlier in this PR live in the Go action layer + proto + handler. The **ABI layer** (which web3 callers actually use via \`eth_sendRawTransaction\`) had no \`blsPop\` slot:

- \`native_staking_contract_abi.json\` defined \`candidateRegisterWithBLS\` with 8 fields (no PoP) and \`candidateUpdateWithBLS\` with 4 fields (no PoP).
- \`CandidateRegister.EthData()\` / \`CandidateUpdate.EthData()\` packed only those 8 / 4 fields — \`cr.blsPop\` was never encoded into calldata.
- \`NewCandidateRegister/UpdateFromABIBinary\` unpacked only those 8 / 4 fields — even if a client somehow embedded PoP bytes, the decoder would have ignored them.
- e2etest \`SignedCandidateRegisterWithBLS\` / \`SignedCandidateUpdateWithBLS\` were called with \`blsPop = nil\` everywhere, so the missing codec coverage went undetected.

Post-fork impact would have been: every web3-routed register/update with BLS — wallets, SDKs, \`ioctl\` flows going through eth-tx — gets rejected by the handler for missing PoP, with no way to provide it from the client side.

## What changed

Added two new V2 ABI entries (your option B):

| Method | Inputs |
|---|---|
| \`candidateRegisterWithBLSAndPoP\` | name, operatorAddress, rewardAddress, ownerAddress, duration, autoStake, blsPubKey, **blsPop**, data |
| \`candidateUpdateWithBLSAndPoP\` | name, operatorAddress, rewardAddress, blsPubKey, **blsPop** |

Coexistence model:
- Legacy \`candidateRegisterWithBLS\` / \`candidateUpdateWithBLS\` keep their selector IDs. Tooling that hardcodes them keeps working pre-fork.
- Post-fork the legacy entries are still accepted at ABI decode time, but the handler rejects them for lacking PoP — clients must switch to the V2 selector to register.

\`EthData()\` routes by content:
- \`WithBLS() && len(blsPop) > 0\` → V2 selector
- \`WithBLS()\` → legacy selector (pre-fork OK, post-fork rejected at handler)
- otherwise → non-BLS \`candidateRegister\`

\`FromABIBinary\` adds the V2 selector recognition + decodes \`blsPop\`.

## Tests

- \`TestCandidateRegisterABIEncodeAndDecode\` gets a new "with public key and PoP" subcase that round-trips a non-empty PoP through Pack/Unpack via the V2 method.
- \`TestCandidateUpdate\` gets "ABI encode with PoP" doing the same on the update side.

Both fail without the EthData/FromABIBinary changes — they're regression guards for exactly the bug you spotted.

## Still out of scope

End-to-end web3 → handler test (proves the V2 selector flows blsPop all the way into \`Candidate.BLSPubKey\` post-fork, and that legacy selector + post-fork rejects properly). Will follow once we wire up the EnforceBLSPoP gate in a proper integration scenario — that's the same fork-activation work as the \`ioctl\` PoP-generation flag.

envestcc:
Good catch — pushed \`98bdaa969\` mirroring the V2 PoP methods in \`native_staking_contract_interface.sol\` and adding a sync-warning header to both files.

## Findings

- \`action/native_staking_contract_interface.sol\` is the human-readable source-of-truth (clients / docs consume it)
- \`action/native_staking_contract_abi.json\` is what iotex-core actually loads via \`NativeStakingContractABI()\` (\`go:embed\` + parse at init)
- **Makefile has no target regenerating one from the other** — confirmed by grepping for \`solc\` / \`abigen\` / staking-related targets. The two are kept in sync by hand.

## What's in the commit

1. \`candidateRegisterWithBLSAndPoP\` and \`candidateUpdateWithBLSAndPoP\` Solidity declarations added to the .sol interface, mirroring the JSON I added in \`ab96bbc4b\`.
2. Header comment at the top of \`native_staking_contract_interface.sol\` calling out the no-auto-gen invariant.
3. Header comment in \`native_staking_contract_abi.go\` (above the \`//go:embed\` directive) saying the same thing — so anyone touching the JSON via Go-tooling code sees it.

The warning text is the same in both: "Adding a new method, parameter, or event requires editing BOTH files by hand, otherwise the web3 path silently misencodes — clients embed one signature while the node decodes another. There is no Makefile target that catches the drift."

## On the lack of a Makefile target

It would be cleaner to wire one up (\`solc --abi\` → generated JSON, fail CI on drift). Out of scope for this PR — would need a solc version pin, an addition to the dev-deps target, and a CI step. Worth filing as a follow-up if the team wants. For now the comment is the cheapest guard.

envestcc:
Pushed \`9229438ac\` addressing 4 of the 5 review comments. The 5th (Comment 3 about \`blsPubKey\` in the signing root) I'm leaving alone and will reply on the thread separately.

## What changed

### Comments 1 + 5 → \`GetByBLSPubKey\` replaces \`ContainsBLSPubKey\`

\`ContainsBLSPubKey(pubkey, except) bool\` hid the "is this me?" decision in a helper. Replaced with \`GetByBLSPubKey(pubkey) *Candidate\` — callers get the (possibly nil) holder and compare identifiers themselves. The three register / update handler call sites now read:

\`\`\`go
if holder := csm.GetByBLSPubKey(act.BLSPubKey()); holder != nil &&
    holder.GetIdentifier().String() != c.GetIdentifier().String() {
    return ErrCandidateConflict
}
\`\`\`

Two side benefits:
- Unifies with the \`GetByBLSPubKey\` method added in #4857 (Y4a) so the two PRs don't introduce parallel BLS-pubkey-lookup APIs.
- Returns the full \`*Candidate\` so future consumers can read Owner / Operator / Reward without another API extension.

### Comments 2 + 4 → strict nil-candidateID

\`BLSPopSigningRoot\` / \`SignBLSPop\` / \`VerifyBLSPop\` used to silently accept nil candidateID by skipping the candidate-binding write — which degrades the scheme to domain+pubkey-only and is exactly the shape an attacker reaching for cross-candidate replay would hope for. The three entry points now refuse:

- \`BLSPopSigningRoot\` returns nil
- \`SignBLSPop\` returns \`error("nil candidate ID; PoP must bind to a candidate identity")\`
- \`VerifyBLSPop\` rejects before any cryptographic work

\`TestBLSPop_RejectNilCandidateID\` locks the contract.

## Test plan
- [x] \`TestBLSPop_RejectNilCandidateID\` covers all three entry points
- [x] \`TestCandidateCenter_GetByBLSPubKey\` covers the new lookup (nil / empty / unregistered / hit)
- [x] \`go test ./action/protocol/staking/ ./action/\` — 476 passing

action/protocol/staking/bls_pop.go
envestcc: Pushed \`fbfd7ae92\` — dropped \`blsPubKey\` from the signing root per our discussion.

The signing root is now \`H(blsPopDomain || candidateID)\`. Security model:

1. **Basic rogue-key registration**: blocked by the pairing verifier alone. \`Verify(pk_rogue, msg, sig)\` requires \`sig\` to be produced under \`sk_rogue\`, which the attacker doesn't have — regardless of what's in \`msg\`.
2. **Same-message aggregation** (the classical attack that pubkey-in-message defends against): requires two distinct honest signers to sign the *same* root. Protocol-enforced uniqueness of candidate identifiers (generateCandidateID + ContainsName/Operator/Owner checks at register time) rules that out — no two honest delegates ever produce PoPs over the same root.
3. **Cross-candidate replay**: still blocked by candidateID binding.
4. **Cross-domain replay**: still blocked by blsPopDomain.

\`SignBLSPop\` no longer derives \`pk\` from \`sk\` since the signing root doesn't need it. PoP bytes are still 96 B G2-compressed; DST unchanged.

Tests updated: \`TestBLSPop_RogueKeyAttackBlocked\` + \`TestBLSPop_RejectNilCandidateID\` use the new \`BLSPopSigningRoot(candidateID)\` signature.

envestcc:
Pushed \`b80f0f316\` — ioctl UX for the BLS PoP path.

## What lands

### \`ioctl stake2 register\`

Three-option matrix:
- **Option 1 (default)** — no \`--bls-*\` flags: tool derives BLS from the signer's ECDSA key, computes PoP, prompts the user to confirm the derived pubkey + the ECDSA↔BLS coupling warning. \`--yes\` for CI.
- **Option 2** — \`--bls-priv-key HEX\`: standalone BLS key.
- **Option 3** — \`--bls-pubkey HEX --bls-pop HEX\`: explicit, local \`VerifyBLSPop\` before submission.

BREAKING: \`BLS_PUBKEY\` moves out of positional args (was \`args[4]\`) into flags. Pre-fork pre-existing scripts need updating, but post-fork they need re-tooling anyway since they must supply PoP.

### \`ioctl stake2 update\`

Same three options PLUS Option 0:
- **Option 0** — no BLS flags: BLS untouched (the common case when changing only name / operator / reward).
- Option 1 for update requires explicit \`--bls-from-signer\` opt-in so a name-only update never silently rotates the producer key.
- Update path needs \`--candidate-id\` for any non-Option-0 mode (RPC fallback is a future improvement).

### \`ioctl account bls-sign-pop\` (new)

Offline PoP generator. Takes \`--bls-priv-key HEX\` OR \`--signer ADDRESS\` (the latter decrypts the ECDSA keystore + derives BLS), plus \`--candidate-id ADDRESS\`. Writes 96-byte PoP hex to stdout (informational fields to stderr so \`> pop.hex\` captures just the PoP). Air-gap flow:

\`\`\`bash
# cold machine
ioctl account bls-sign-pop --candidate-id io1owner... --bls-priv-key 0x... > pop.hex

# USB / QR to hot machine
ioctl stake2 register mycand io1op io1reward io1owner 10000 350 \\
  --bls-pubkey 0x... --bls-pop \$(cat pop.hex) \\
  --signer io1owner --auto-stake
\`\`\`

### Why hot/cold separation matters here

BLS sk derived from ECDSA sk means a single compromise blast both. The \`--yes\`-able prompt makes the coupling explicit so delegates choosing the easy path know what they signed up for; security-conscious delegates use Option 2 / 3 with separately-managed BLS material.

## Tests

- \`TestBlsPoPFlags_ClassifyForRegister\` — 6 cases: auto-derive default, explicit key, explicit PoP, partial input rejection, mutual-exclusion rejection
- \`TestBlsPoPFlags_ClassifyForUpdate\` — 8 cases: Option 0 + each of 1/2/3 + reject branches including \`--bls-from-signer + --bls-priv-key\` mutual exclusion
- \`TestLoadBLSPrivKey\` — hex parse: good / 0x-prefixed / empty / non-hex / wrong-length

## Deferred

- \`--bls-keystore PATH\` — placeholder, errors out pointing at \`--bls-priv-key\`. iotex doesn't have a BLS keystore format yet; separate RFC.
- RPC fallback to resolve \`--candidate-id\` from the signer on update. Requires a gRPC roundtrip; for v1 the user supplies it explicitly.
- e2e test exercising the whole chain (decrypt → derive → PoP → V2 ABI → handler accept) — depends on \`EnforceBLSPoP\` activation fixture.

## Out of scope (separate PR territory)

- Migration command for existing pre-PoP BLS registrations (just re-run \`stake2 update --bls-from-signer\` — no new command needed).
- JS / Go SDK exposure of the same helpers — that lands in the respective SDK repos.

envestcc:
Pushed \`143010dfe\` — handler integration + e2etest coverage for the PoP gate.

## Handler integration (\`handlers_blspop_test.go\`, new)

3 functions, 11 subcases covering EnforceBLSPoP at all three handler sites:

**TestHandleCandidateRegister_PoPGate** — gate on/off × {valid PoP, empty PoP, PoP signed by wrong key}. The "wrong key" subcase exercises the rogue-key blocking via the pairing verifier, separate from the unit-test-level proof in \`bls_pop_test.go\`.

**TestHandleCandidateRegister_BLSPubKeyUniqueness** — second candidate trying to register the same BLS pubkey with a fresh valid PoP under its own owner → ErrCandidateConflict. Locks the GetByBLSPubKey invariant in at the handler boundary.

**TestHandleCandidateUpdate_PoPGate** — rotation under gate on/off × {valid PoP, empty PoP, PoP for wrong candidateID}. The "wrong candidateID" subcase confirms the cross-candidate replay defence (the binding lives in BLSPopSigningRoot, not at the protocol layer above).

Fixture flips the gate via \`genesis.ToBeEnabledBlockHeight\` (0 vs \`math.MaxUint64\`) and forces XinguBlockHeight = 0 so the BLS register/update codepath is reachable.

## e2etest (\`native_staking_test.go\`)

**TestCandidateBLSPoP** — full chain pipeline encode → submit → mint → handler → receipt → state, with 5 subcases:

1. Pre-fork register without PoP → Success (backward compat anchor)
2. Post-fork register with valid PoP → Success, \`Candidate.BlsPubKey\` persisted in state
3. Post-fork register without PoP → ErrUnauthorizedOperator
4. Post-fork update rotates BLS pubkey with valid PoP → \`Candidate.BlsPubKey\` reflects the new key
5. Post-fork update without PoP → ErrUnauthorizedOperator, previous BLS pubkey untouched in state

Genesis wires \`ToBeEnabledBlockHeight == XinguBlockHeight\` so the BLS register path and the PoP gate activate together — same shape as the planned fork rollout. Each test asserts both the receipt status AND the resulting candidate state, so a silent state mutation under a failure receipt wouldn't slip past.

## Pre-existing flake

\`TestProtocol_FetchBucketAndValidate/validate_owner_and_selfstake\` flakes about 1 in 3 runs on this branch — and confirmed earlier it also flakes on master. Not introduced by this PR.

## Test plan
- [x] \`go test -count=1 -run "TestHandleCandidateRegister_PoPGate|TestHandleCandidateRegister_BLSPubKeyUniqueness|TestHandleCandidateUpdate_PoPGate" ./action/protocol/staking/\` — 11/11 passing
- [x] \`go test -count=1 -run "TestCandidateBLSPoP" ./e2etest/\` — 5/5 passing
- [x] \`go test -count=1 ./action/protocol/staking/ ./action/ ./ioctl/cmd/action/\` — clean except for the pre-existing flake noted above

guo (approved):
Reviewed exact head 1d6af81105f8c8141f8303f3dbebbb07eace09ec. Code review and isolated chain-level validation covered native and ABI paths, pre-fork/fork boundary behavior, reachable legacy duplicate-key state, same-block duplicate-key handling, binary-swap compatibility, and repeated full state replays from mainnet-derived candidate data. No blocking correctness, determinism, or consensus-safety issue found.

Issue #4813, opened 2026-03-20
bug: TestProtocol_FetchBucketAndValidate flaky in staking test suite


raullenchai:
## Description

`TestProtocol_FetchBucketAndValidate` fails intermittently when run as part of the full staking test suite (`go test ./action/protocol/staking/`) but passes consistently when run in isolation.

## Reproduction

```bash
# Fails intermittently:
go test ./action/protocol/staking/ -count=1

# Always passes:
go test ./action/protocol/staking/ -run TestProtocol_FetchBucketAndValidate -v
```

## Likely cause

Concurrent state DB access or shared test fixture race condition when other staking tests modify global state.

Observed on both macOS arm64 (M3 Pro, Go 1.26.1) and macOS arm64 (M3 Ultra, Go 1.24.6).

PR #4860, opened 2026-06-25
feat(staking): IIP-59 VoterWeightView infrastructure (PR 2/5)

envestcc:
## Summary

PR 2 of the IIP-59 series. Introduces the per-(candidate, voter) weighted-votes view data structure that the rest of the IIP-59 implementation hangs off of, integrates it into the staking protocol's standard view lifecycle (Fork/Snapshot/Revert/Commit), and persists its deterministic hash digest to anchor cross-node consensus on the view contents. **No runtime behavior changes** — the view is built when the flag activates but no callers consume it in this PR.

The actual handler hooks that keep the view in lock-step with each bucket lifecycle event are split into PR 2.5 to keep this PR scoped to the view's data structure and lifecycle wiring (low-risk; no production handler files touched).

## Series

| # | PR | Status |
|---|---|---|
| 0 | iotex-proto: SetCommissionRate + Candidate.commissionRate | https://github.com/iotexproject/iotex-proto/pull/174 (draft) |
| 1 | iotex-core: fields + feature flag | #4859 (draft, depends on #174) |
| **2** | **This PR — VoterWeightView infrastructure** | **here, depends on #4859** |
| 2.5 | wire handler hooks to keep view in sync | TODO |
| 3 | SetCommissionRate action + handler | TODO |
| 4 | distributeVoterReward + PutPollResult snapshot | TODO (depends on 2.5) |
| 5 | multi-node stress test + e2e harness | TODO |

## What's in this PR

**Data structure (`voter_weight_view.go`):**
- `VoterWeightView` keys candidates by Hash160 identifier; each entry is a sorted slice of `(voter, *big.Int)` plus an index map for O(log n) insert/remove.
- `Apply(candID, voter, delta)` is the single mutation entry point. Aggregates per-voter, removes the entry on drain-to-zero, removes the candidate when its last voter leaves, and is defensively no-op when a withdrawal targets an unknown (cand) or (voter).
- `VoterWeightsByCandidate(candID)` returns a sorted copy of the slice — **distribution iterates this slice, never the map**, which makes receipt-log order and state-write order identical across nodes (this is the bug class flagged as #2 in the original PoC review at iotex-core#4811).
- `Hash()` walks candidates in sorted Hash160 order and voters in their already-sorted slice order; insertion-order-independent (proven by tests).
- Tests: 14 covering add, aggregate, decrease, drain-to-zero, over-withdraw, unknown-key no-op, sortedness, clone deep-copy, hash determinism across insertion orders, hash sensitivity to weight change, and realistic-event-stream incremental-vs-rebuild equivalence. Benchmark: `Apply` at 73.8 ns/op on M1 Pro.

**viewData integration (`viewdata.go`):**
- `voterWeights *VoterWeightView` field plus matching field on `Snapshot`.
- `Fork` / `Snapshot` deep-clone (Apply mutates in place); `Revert` restores the cloned snapshot; `Commit` persists the digest when `IsDirty`.

**Persistence (`voter_weight_view.go`):**
- New 1-byte staking namespace tag `_voterWeights = 5` (next free tag; the existing 0-4 are reserved in `protocol.go`).
- `voterWeightDigest` serializes as exactly 32 bytes; corruption surfaces as a deserialization error rather than a silent zero.
- `viewData.commitVoterWeights` writes the digest only when dirty, so pre-flag chains are byte-identical to today's behavior.

**Initial population (`protocol.go` `CreatePreStates`):**
- When `EnableVoterRewardDistribution` is on and `viewData.voterWeights` is nil (first block after activation, or restart), `ensureVoterWeightView` scans all active native buckets, computes per-bucket weight via `CalculateVoteWeight` (with `ContractAddress == \"\" && Index == cand.SelfStakeBucketIdx` for the self-stake gate — fixes PoC review finding #5), and Apply()s each weight into a fresh view.
- After the build, the rebuilt hash is checked against the persisted digest. Mismatch is **fatal**; ErrStateNotExist (no record yet — first activation) is normal and the view is marked dirty so the next Commit writes the initial digest.

**Helper for PR 2.5 (`voter_weight_view.go`):**
- `applyVoterWeightDelta(csm, candIdentifier, voter, delta)` — the single entry point PR 2.5 will wire into every native + contract-staking handler that changes a bucket's weight contribution. No-op when the flag is off and when delta is zero, so callers can drop it next to any existing `candidate.AddVote` / `candidate.SubVote` without first checking the flag.

## What's **not** in this PR

- Handler hooks. PR 2.5 (`task #7` in our internal tracker) will instrument:
  - `handlers.go`: `handleCreateStake`, `handleUnstake`, `handleChangeCandidate`, `handleTransferStake` (asymmetric: −W on old voter, +W on new voter), `handleDepositToStake`, `handleRestake`
  - `handler_candidate_endorsement.go`, `handler_candidate_selfstake.go`
  - `handler_stake_migrate.go` (−W native, +W contract)
  - `nfteventhandler.go` (V1/V2/V3 contract staking events)
  - `vote_reviser.go`, `protocol.go`
- Contract-staking initial population — `ensureVoterWeightView` currently scans native buckets only. Contract buckets are read from each indexer via the existing `ContractStakeView.CandidateStakeVotes` boundary, which is per-candidate aggregated. To enumerate per-voter contract bucket contributions at startup, PR 2.5 (or its precursor) will extend the indexer interface with a `Buckets()` enumerator (or use the existing `BucketsByCandidate` over the candidate-center list).

## Why split this way

The data structure and lifecycle are **independently reviewable** — they compile, they pass tests, and they cannot change runtime behavior because no caller reads from the view. Splitting handler instrumentation into its own PR contains the production-handler-touching change to a separately-scoped review window, where most of the work is mechanical (delta wiring next to existing AddVote/SubVote sites) but each site needs careful attention to corner cases (transfer-style asymmetric updates, self-stake bonus changes, native↔contract migration).

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./action/...` — 922 tests pass; no regression
- [x] New `voter_weight_view_test.go` — 14 unit tests + 1 benchmark covering data-structure correctness, determinism, and clone safety
- [ ] `make lint` (run before un-drafting)

## Behavior gating

- Pre-flag: `EnableVoterRewardDistribution = false`. `voterWeights` stays nil. `Apply` paths are no-op. `Commit` is no-op. Persistence path is no-op. **Byte-identical to today.**
- Post-flag, no callers yet (this PR): the view is built and verified, but no caller reads it. State-trie footprint: one 32-byte digest per chain, rewritten only on dirty commits.


action/protocol/staking/protocol.go
```diff
@@ -578,6 +584,17 @@ func (p *Protocol) CreatePreStates(ctx context.Context, sm protocol.StateManager
 			vd.contractsStake.Revise(ctx)
 		}
 	}
+	// IIP-59: when EnableVoterRewardDistribution first activates, do a
+	// one-shot scan of all active buckets to populate the VoterWeightView.
+	// Subsequent blocks rely on the incremental Apply hooks in the
+	// per-action handlers. The build is idempotent — subsequent calls
+	// see a non-nil voterWeights and short-circuit.
+	if featureCtx.EnableVoterRewardDistribution {
+		if err := p.ensureVoterWeightView(ctx, sm); err != nil {
```
envestcc: I don't think it's necessary to load `VoterWeightView` every time `CreatePreState` is called; it only needs to be loaded once at startup. There should be no scenario where the `VoterWeightView` is missing here.

action/protocol/staking/viewdata.go
```diff
@@ -78,9 +90,11 @@ func (v *viewData) Fork() protocol.View {
 			amount:         new(big.Int).Set(v.snapshots[i].amount),
 			count:          v.snapshots[i].count,
 			contractsStake: v.snapshots[i].contractsStake,
+			voterWeights:   v.snapshots[i].voterWeights.Clone(),
```
envestcc: Cloning `voterWeights` during `viewData.Fork` is an inefficient implementation that will cause serious performance issues. You can refer to `contractStakeView` as a reference.

action/protocol/staking/viewdata.go
```diff
@@ -78,9 +90,11 @@ func (v *viewData) Fork() protocol.View {
...
 		}
 	}
 	fork.contractsStake = v.contractsStake.Fork()
+	fork.voterWeights = v.voterWeights.Clone()
```
envestcc: also inefficient

action/protocol/staking/viewdata.go
```diff
@@ -94,13 +108,43 @@ func (v *viewData) Commit(ctx context.Context, sm protocol.StateManager) error {
 	if err := v.contractsStake.Commit(ctx, sm); err != nil {
 		return err
 	}
+	if err := v.commitVoterWeights(sm); err != nil {
```
envestcc: can define a `Commit` method of the VoterWeightView, and call it instead

action/protocol/staking/viewdata.go
```diff
@@ -112,6 +156,7 @@ func (v *viewData) Snapshot() int {
 		amount:         new(big.Int).Set(v.bucketPool.total.amount),
 		count:          v.bucketPool.total.count,
 		contractsStake: v.contractsStake,
+		voterWeights:   v.voterWeights.Clone(),
```
envestcc: also inenfficient

action/protocol/staking/voter_weight_view.go
```diff
@@ -0,0 +1,415 @@
...
+// Hash returns a deterministic 32-byte digest of the entire view. It is
+// constructed by iterating candidates in sorted hash160 order and, within
+// each candidate, walking the already-sorted voter slice — so two nodes
+// that observed the same sequence of Apply calls (in any interleaving
+// permissible by the staking handlers, which are themselves deterministic)
+// produce identical hashes.
+//
+// The hash is persisted at the _voterWeights namespace tag on every block
+// commit and re-checked at restart against a rebuilt-from-buckets view.
+func (v *VoterWeightView) Hash() hash.Hash256 {
```
envestcc: We should run a benchmark for `Hash()`. Currently, there are approximately 100 candidates and 100,000 staking buckets on mainnet.

envestcc:
All six comments addressed in `2678a6eb6` (force-pushed). Refactor follows the `voteView` / `candidateVotesWraper` pattern from `systemcontractindex/stakingindex`.

**Comment 1 — `protocol.go:593` (don't load every CreatePreState)**

Removed the CreatePreStates branch. `Protocol.Start` now calls a new `loadVoterWeightView(view, sr)` once at node startup, after both `candCenter` and `contractsStake` have loaded. The startup-time digest verification is preserved; ErrStateNotExist (first activation, no digest yet) marks the view dirty so the next block commit writes the initial hash.

**Comments 2, 3, 5 — `Fork`/`Snapshot` deep-clone is inefficient**

`VoterWeightView` is now an interface with three implementations:

```text
voterWeightBase  — terminal layer, the sorted map.
voterWeightWrap  — Snapshot overlay. Wraps a parent and accumulates Apply
                   deltas locally; Commit replays into the parent's base
                   (no data copy at Snapshot time).
voterWeightFork  — commit-in-clone overlay for viewData.Fork. Base shared
                   until Commit actually flushes; then cloned so the
                   parent stays intact. Forks that never mutate the view
                   pay nothing.
```

Per-snapshot cost drops from O(N) (full map clone of ~100k voters) to O(k) where k is the number of (cand, voter) tuples touched between snapshot and revert — typically a handful per action. `viewData.Snapshot` now installs a `Wrap()` overlay; `Revert` restores the saved pointer; `Fork()` installs a `Fork()` overlay.

**Comment 4 — `Commit` method on `VoterWeightView`**

Moved both flatten and persist logic onto the interface: `Commit(sm protocol.StateManager) (VoterWeightView, error)`. `viewData.Commit` is now just `v.voterWeights = v.voterWeights.Commit(sm)`. The old free function `commitVoterWeights` and `MarkClean` helper are gone.

**Comment 6 — `Hash()` benchmark at mainnet scale**

Added `BenchmarkVoterWeightView_Hash` with 100 candidates and 100,000 voters distributed via a skewed normal across candidates (popular delegates with more voters):

```
goos: darwin
goarch: arm64
pkg: github.com/iotexproject/iotex-core/v2/action/protocol/staking
cpu: Apple M1 Pro
BenchmarkVoterWeightView_Hash-10    272    13177876 ns/op
```

**~13.2ms per call** at full mainnet scale, called at most once per block via the viewData commit path. Well within a 5-second block budget.

### Test sweep

- New: `TestVoterWeightView_ForkIsolation` and `TestVoterWeightView_WrapMergesIntoParent` exercise the wrap/fork overlay semantics directly (changes don't leak from fork to parent before commit; wrap merges into parent on commit).
- Existing tests retargeted to the interface API: `Hash() == ZeroHash256` instead of removed `IsEmpty()`; `Fork()`/`Wrap()` instead of removed `Clone()`.
- All 15 `TestVoterWeightView_*` tests pass; full `./action/protocol/staking/` package passes (modulo the pre-existing flaky `TestProtocol_FetchBucketAndValidate` #4813).

action/protocol/staking/protocol.go
```diff
@@ -380,9 +380,63 @@ func (p *Protocol) Start(ctx context.Context, sr protocol.StateReader) (protocol
...
 		}
 	}
+
+	// IIP-59: build the voter weight view once at protocol startup when the
+	// feature has activated. The persisted digest, if any, is verified
+	// against the rebuilt view; mismatch is fatal so a desync between
+	// bucket state and view state cannot silently pass into block
+	// production. This is a one-shot full scan — subsequent blocks rely on
+	// the incremental Apply hooks in the per-action handlers.
+	if !protocol.MustGetFeatureCtx(ctx).NoVoterRewardDistribution {
```
envestcc: should load voterWeightView even if NoVoterRewardDistribution

action/protocol/staking/protocol.go
```diff
@@ -380,9 +380,63 @@ func (p *Protocol) Start(ctx context.Context, sr protocol.StateReader) (protocol
...
 
+// loadVoterWeightView populates view.voterWeights via a full bucket scan and
+// verifies the rebuilt hash against the digest persisted at _voterWeightsKey.
+// Called once per node startup (from Start) — does not run on every block.
+//
+// Contract-staking buckets are not enumerated here yet; PR 2.5 will extend
+// the enumeration once its handler hooks are wired. Until then, the view is
+// built from native buckets only — the persisted hash still anchors
+// determinism within that scope.
+func (p *Protocol) loadVoterWeightView(c *viewData, sr protocol.StateReader) error {
```
envestcc: In addition to native buckets, those contract staking buckets should also be counted.

envestcc:
Both addressed in `a55e19465` (force-pushed onto the refactor commit).

### Comment 1 — `protocol.go:390`: load even when `NoVoterRewardDistribution`

Removed the flag check at the call site. `Protocol.Start` now invokes `loadVoterWeightView` unconditionally, so the view is built and digest-verified on every startup regardless of activation status. The pair to this change is in `viewData.Commit`:

```go
var persistSM protocol.StateManager
if !protocol.MustGetFeatureCtx(ctx).NoVoterRewardDistribution {
    persistSM = sm
}
updated, err := v.voterWeights.Commit(persistSM)
```

Pre-fork chains pass `sm=nil` into `VoterWeightView.Commit`, which flattens overlays and clears the dirty flag without touching the state trie — byte-identical to today's behavior. Post-fork, the real `sm` flows through and the digest is written whenever the view is dirty. This way the view is always in sync with the bucket state from the moment the node boots, but persistence stays gated on the flag.

`loadVoterWeightView` also got an updated dirty-on-empty-digest rule: it now only sets `dirty=true` on first-activation when `!NoVoterRewardDistribution`. Pre-fork startups with no persisted digest leave dirty=false so the next commit skips the write.

### Comment 2 — `protocol.go:406`: include contract staking buckets

Extended the enumeration to walk all three contract staking indexers (V1/V2/V3) in addition to native buckets:

```go
for _, indexer := range []ContractStakingIndexer{
    p.contractStakingIndexer,
    p.contractStakingIndexerV2,
    p.contractStakingIndexerV3,
} {
    if indexer == nil || indexer.StartHeight() > height {
        continue
    }
    contractBuckets, err := indexer.Buckets(height)
    if err != nil {
        return errors.Wrapf(err, "failed to read contract staking buckets at %s", indexer.ContractAddress())
    }
    allBuckets = append(allBuckets, contractBuckets...)
}
```

The same `ContractAddress == "" && Index == cand.SelfStakeBucketIdx` gate inside `buildVoterWeightView` ensures self-stake bonus only applies to native self-stake buckets, not contract buckets with `Index == 0` (the PoC review-finding #5 bug from #4811).

### Test-mock updates

Two existing tests had hardcoded `Times(N)` expectations on `sr.Height()` / `MockContractStakingIndexer.StartHeight()` that broke because `Start` now does the extra work. Updated:

- `TestVoteReviser` — `sm.Height()` count 4 → 5 (one extra `sr.Height` from `loadVoterWeightView`)
- `TestCreatePreStatesMigration` — switched `StartHeight` / `ContractAddress` to `AnyTimes()` and added a `Buckets(any)` expectation, so future load-path additions don't ping this test again

Full sweep: `923/923 tests pass in action/protocol/...`.

envestcc:
Closing as part of IIP-59 PR reorganization. The proposal was amended (iotexproject/iips#73) to cover both block reward + epoch reward and to base voter distribution on a per-epoch snapshot. The new PR split is:

- **PR 1** (new): commissionRate schema + `SetCommissionRate` action (merges the old #4859 + #4862)
- **PR 2**: snapshot layer at PutPollResult (commission rate + voter weight blob)
- **PR 3**: voter reward distribution (epoch path)
- **PR 4**: block reward folding (pending pool + orphan drain)
- **PR 5**: stress test + e2e determinism harness
- **PR 6**: mainnet fork activation

The incremental voter weight view (old #4860 + #4863) is dropped in favor of a per-PutPollResult full scan; ~40k buckets sorted+encoded once per epoch fits well inside the mint budget and eliminates the 13-handler-hook consensus-safety surface. Old #4864 (voter reward distribution) is replaced by new PR 3 which reads the snapshot instead of the live view.

Superseded by new PRs — will link once opened.

PR #4867, opened 2026-07-06
[evm] record in-contract transfer logs from MakeTransfer directly

envestcc:
## What

`IN_CONTRACT_TRANSFER` transaction logs are meant to reflect native token transfers the node performs during contract execution. Today `MakeTransfer` emits them by calling `AddLog` with a reserved topic, and `AddLog` converts **any** EVM log carrying that reserved topic into a transaction log.

Because the reserved-topic handling lives in the shared `AddLog` path, the transaction-log stream does not always correspond to actual balance movements: an `IN_CONTRACT_TRANSFER` entry can be present for a transfer that never occurred. Observed on mainnet tx [`0xadd4b0f5…4eec06`](https://iotexscan.io/tx/0xadd4b0f5930e7acc56a06ae585a91ef60da997d7c3f8075a1be7a3f0fe4eec06): the transaction log reports a 100 IOTX transfer, but the recipient's on-chain balance is unaffected. Consumers that reconstruct balances from transaction logs (indexers, wallets, exchanges) can be misled by such entries.

## Change

Record the transfer log directly where the transfer happens, keeping the node's transfer-log channel separate from contract-emitted EVM logs:

- `MakeTransfer` now calls a node-only `StateDBAdapter.AddInContractTransferLog` method, exposed through the `inContractTransferRecorder` interface (deliberately **not** part of `vm.StateDB`), instead of routing through `AddLog`.
- `AddLog` no longer special-cases the reserved topic. Such logs were never appended to `receipt.Logs`, so dropping the special case leaves `receipt.Logs` — and therefore the receipt root — unchanged.

## Consensus impact

None. Transaction logs are not part of the receipt root or state root; `receipt.Logs` is byte-identical before and after. Balances are unaffected — they are always moved by `SubBalance`/`AddBalance` inside `MakeTransfer`, independent of the log. The change is intentionally not height-gated.

## Test

Adds `TestInContractTransferLogNoForgery`, covering both a log that must be ignored and a genuine `MakeTransfer` that records the log and moves balance. The full `action/protocol/execution/evm` package suite passes.

### DIFF
--- a/action/protocol/execution/evm/evm.go
+++ b/action/protocol/execution/evm/evm.go
@@ -63,19 +63,26 @@ func CanTransfer(db vm.StateDB, fromHash common.Address, balance *uint256.Int) b
 	return db.GetBalance(fromHash).Cmp(balance) >= 0
 }
 
+// inContractTransferRecorder records a native token transfer that happens inside
+// contract execution. It is intentionally NOT part of vm.StateDB: only the node,
+// via MakeTransfer, can reach it. A contract therefore cannot forge an
+// IN_CONTRACT_TRANSFER transaction log by emitting a crafted LOG opcode.
+type inContractTransferRecorder interface {
+	AddInContractTransferLog(from, to common.Address, amount *big.Int)
+}
+
 // MakeTransfer transfers account
 func MakeTransfer(db vm.StateDB, fromHash, toHash common.Address, amount *uint256.Int) {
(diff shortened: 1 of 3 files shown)


envestcc:
Superseded by https://github.com/iotexproject/iotex-core/pull/4868, which uses a branch on this repository directly.

PR #4869, opened 2026-07-06
fix(evm): GetCommittedState must not return post-mutation value for prestate-absent keys

envestcc:
## Summary

`contract.GetState` propagates `trie.ErrNotExist` for a storage key that is absent in the pre-tx trie without recording the observation. If the same key is later written (via `SetState`) and re-read within the same tx, the follow-up `GetState` succeeds and populates `committed[key]` with the **just-written** value. `GetCommittedState` then returns that post-mutation value as if it were the tx-start prestate.

EIP-2200's SSTORE dynamic gas uses `committed` to decide whether a write is a dirty in-place update (SLOAD_GAS, 100) or a full slot reset (SSTORE_RESET, 2900). With the polluted cache, the second write to a previously-absent slot is charged 2800 extra gas per SSTORE. In an ERC-721 mint immediately followed by a transfer within the same tx — for example an ERC-6551 token-bound-account setup that touches `_owners[tokenId]` twice — the accumulated overcharge is large enough to OOG-revert a transaction that would otherwise succeed under the intended gas semantics.

## Fix

Track prestate-absent keys in a new `missing` map alongside `committed`:

- `GetCommittedState` short-circuits missing keys to `trie.ErrNotExist` without touching the live trie.
- `GetState` records `ErrNotExist` observations in `missing`, and does not overwrite `committed` with a live-trie value once a key has been observed missing.
- `SetState` skips the priming `GetState` call for keys already in `missing` (they cannot resolve to a real prestate later in the tx).
- `Snapshot` propagates the shared `missing` map, `Commit` resets it alongside `committed`, and `newContract` initializes it.

## Hardfork gate

Because this changes state-transition semantics for the class of transactions that hit the pattern above, the new behavior is gated on a new feature flag `CorrectPrestateForAbsentKeys`, currently wired to `g.IsToBeEnabled(height)` — staged, no scheduled fork height yet. Pre-fork execution keeps the historical (buggy) code path byte-for-byte so mainnet catch-up from state prior to the fork height replays identically. Post-fork execution gets the fix.

Plumbing follows the existing pattern used by `AsyncContractTrieOption`, `FixSnapshotOrderOption`, etc.:

  - `FeatureCtx.CorrectPrestateForAbsentKeys` in `action/protocol/context.go`
  - `CorrectPrestateForAbsentKeysOption` on `StateDBAdapter`
  - `newContract` / `newContractAdapter` take an explicit `trackAbsent` bool, propagated through `Snapshot`
  - `evm.go` `prepareStateDBAdapter` toggles the option from the feature ctx

## Regression test

`TestGetCommittedStateAbsentKey` covers both fork sides:

  - **post-fork (`trackAbsent=true`)**: the repro pattern must return `trie.ErrNotExist` for `GetCommittedState` on a prestate-absent key.
  - **pre-fork (`trackAbsent=false`)**: the buggy behavior is preserved — `GetCommittedState` returns the post-mutation value. This guards mainnet catch-up determinism.

Both async and sync modes are exercised (4 subtests total).

## Verification against real mainnet state

Applied the post-fork behavior to a debug build of v2.4.2, restarted a full node's mainnet catch-up against a stopped-at-target-1 snapshot, let it execute the block containing failing tx `0xfe792b0c…7ec54a`:

|                       | pre-fork (mainnet reality) | post-fork (with this PR) |
| --------------------- | -------------------------- | ------------------------ |
| SSTORE at `pc=7929 depth=6` | cost `2900` (SSTORE_RESET) | cost **`100`** (dirty in-place) |
| out-of-gas events during tx | ≥1 (`pc=5788 depth=4`)      | **0**                    |
| `ExecuteContract EVM done statusCode` | 106 (revert) | **1** (success)   |
| gas consumed          | 826,574 (reverted)         | 768,498                  |

At the target block's tip, the fixed node's post-execution state digest diverges from the mined block header (which was produced by the still-buggy production build) — this is the expected outcome and confirms the fix is state-affecting, which is precisely why it needs the fork gate.

## Test plan

- [x] `go test ./action/protocol/execution/evm/...` (all tests pass)
- [x] Fork-gated regression test asserts both pre- and post-fork behavior
- [ ] Reviewer input on the target hardfork height once release planning picks one

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -173,6 +173,13 @@ type (
 		AlwaysWriteCachedContract     bool
 		NoCandidateExitQueue          bool
 		FixInContractTransferLogTopic bool
+		// CorrectPrestateForAbsentKeys, when true, makes contract.GetCommittedState
+		// return the true tx-start prestate value (zero, via trie.ErrNotExist) for
+		// storage slots that were absent in the pre-tx trie. Prior to this gate the
+		// contract-level committed[] cache was populated with post-mutation values
+		// for such slots, causing EIP-2200 SSTORE dynamic gas to misclassify
+		// dirty in-place writes as SSTORE_RESET (100 → 2900 gas overcharge per hit).
+		CorrectPrestateForAbsentKeys bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
@@ -348,6 +355,7 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			AlwaysWriteCachedContract:               !g.IsYap(height),
 			NoCandidateExitQueue:                    !g.IsYap(height),
 			FixInContractTransferLogTopic:           g.IsToBeEnabled(height),
+			CorrectPrestateForAbsentKeys:            g.IsToBeEnabled(height),
 		},
 	)
 }
--- a/action/protocol/execution/evm/contract.go
+++ b/action/protocol/execution/evm/contract.go
@@ -51,8 +51,15 @@ type (
 		code       protocol.SerializableBytes // contract byte-code
 		root       hash.Hash256
 		committed  map[hash.Hash256][]byte
-		sm         protocol.StateManager
-		trie       trie.Trie // storage trie of the contract
+		// trackAbsent gates the EIP-2200-correct prestate tracking below. When
+		// false (pre-fork), the buggy pre-existing behavior is preserved: keys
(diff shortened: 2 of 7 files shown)


guo (approved):
Reviewed the rebased head. The hardfork gate preserves pre-fork behavior, the post-fork absent-prestate handling is consistent across snapshots and adapters, and the focused tests plus required CI pass.

PR #4871, opened 2026-07-06
fix(evm): gate malformed in-contract-transfer log drop behind Zanzibar fork

raullenchai:
## Summary

Removes a **contract-triggerable node panic** in `StateDBAdapter.AddLog` that is a **chain-halt / liveness DoS** (P0) — now **gated behind the `Zanzibar` hardfork** instead of activating unconditionally.

`AddLog` panics (`"Invalid in contract transfer topics"`) whenever it sees a log whose `topic0` equals the reserved in-contract-transfer marker but whose topic count is not exactly 3. That marker is the **all-zero hash256** (`TransactionLogType_IN_CONTRACT_TRANSFER == 0`), so **any contract** can execute a `LOG1`/`LOG2`/`LOG4` opcode with `topic0 = 0x00..00` and trip the panic. There is **no recover around EVM execution on the block-apply / validation / sync path**, so a Byzantine proposer can include such a tx and crash every validating/applying node.

## Why this is now hardfork-gated (changed from the previous revision)

The previous revision dropped these logs **unconditionally**. That is unsafe: the pre-fork behavior is a **CRASH**, not a different state — an un-upgraded node **cannot process** a block carrying such a log. If the drop were ungated, an upgraded producer could commit a block that old nodes can **never** process → a **permanent poison block** and a network split.

So the crash→drop switch must be **coordinated at a fork height**:

- **pre-`Zanzibar`**: preserve the historical behavior **byte-for-byte** — panic on a reserved-topic log with topic count != 3 (drop on exactly 3). Required for deterministic replay of historical blocks.
- **at/after `Zanzibar`**: drop the reserved-topic log for **any** topic count and **never panic**.

Note this fork does **not** change any state: these reserved-topic logs are **never appended to `stateDB.logs`** on either side, so `receipt.Logs` (and the receipt root) is **byte-identical** across the fork. The gate exists purely to coordinate the *liveness* switch (crash → graceful drop) across the network, mirroring how transaction-log behavior is gated elsewhere.

## Implementation

- New `FeatureCtx` flag **`DropMalformedInContractTransferLog`**, set from `genesis.IsZanzibar(height)` in `WithFeatureCtx` (`action/protocol/context.go`). It is a **separate** flag from #4872's `CorrectSelfDestructTransferLog` — same fork height, different behavior.
- Wired into `StateDBAdapter` via a construction option `DropMalformedInContractTransferLogOption()` in `prepareStateDBAdapter` (`action/protocol/execution/evm/evm.go`), mirroring the existing `SuicideTxLogMismatchPanicOption` / `CorrectSelfDestructTransferLogOption` pattern.
- `AddLog` reads the option field: `if !stateDB.dropMalformedInContractTransferLog && len(topics) != 3 { panic(...) }; return`.

## Dependencies & required merge order ⚠️

This change requires **both**:

1. **#4868** (`pr/in-contract-transfer-log-forgery`) — moves genuine in-contract transfers out of `AddLog` into `AddInContractTransferLog`, so any reserved-topic log reaching `AddLog` is necessarily contract-emitted (untrusted) and safe to drop.
2. **#4872** (`fix/evm-selfdestruct-transferlog`) — introduces the `Zanzibar` hardfork (`genesis.ZanzibarBlockHeight` / `IsZanzibar`, wired through `FeatureCtx`) that this gate reuses.

**#4868 and #4872 must merge before #4871.** This branch is stacked on #4868 (its base) and additionally **merges #4872** so it compiles and CI is green; the merge commit clearly provenances #4872's content in the diff. **After #4868 and #4872 land, retarget this PR's base to `master` and rebase** — the two merge commits collapse and only the gating change remains.

The `Zanzibar` fork height is a **placeholder** (`math.MaxUint64`, set by #4872) pending maintainer sign-off; no real mainnet/testnet height is chosen here.

## Tests (both sides of the fork)

- `action/protocol/context_test.go` — `TestDropMalformedInContractTransferLogFeature`: the flag is off at `Zanzibar-1` and on at/after `Zanzibar`.
- `action/protocol/execution/evm/evmstatedbadapter_test.go` — `TestAddLogMalformedInContractTransfer`: **pre-fork** a 1/2/4-topic reserved log **panics** (and 3-topic drops), **post-fork** every malformed count is **dropped** (no panic, nothing in `stateDB.logs`, no forged transaction log).
- `e2etest/incontract_addlog_panic_test.go` — full block pipeline for init code emitting `LOG1` with `topic0 = 0`: **pre-fork** (`TestInContractTransferAddLogPanicPreZanzibar`) mint fails via the draft recover (the doomed block is never produced); **post-fork** (`TestInContractTransferAddLogPanicPostZanzibar`) the block is minted/applied/committed, receipt is `Success`, no forged `IN_CONTRACT_TRANSFER` log, and `receipt.Logs` is empty.
- No regression: `TestInContractTransferLogNoForgery`, `TestInContractTransferForgeryDeploy`, `TestSelfDestructTransferLog` still pass.

`gofmt -l` clean, `go build ./...`, `go vet ./action/... ./e2etest/...`, and the evm + `action/protocol` suites all green on the integration base (`master` + #4868 + #4872). Independent adversarial review with `codex` returned **CONVERGED** with no substantive findings.


envestcc:
fixed by https://github.com/iotexproject/iotex-core/pull/4868

PR #4872, opened 2026-07-06
fix(evm): derive SELFDESTRUCT transfer logs from actual balance movement

raullenchai:
## Problem

SELFDESTRUCT `IN_CONTRACT_TRANSFER` transaction logs are derived from a fragile heuristic in `evmstatedbadapter.go`: *"the last `AddBalance` call before `SelfDestruct`"* (`lastAddBalanceAddr` / `lastAddBalanceAmount`). This misrecords the beneficiary/amount in several cases and, in particular, emits a **phantom `self -> self` `IN_CONTRACT_TRANSFER`** for EIP-6780 self-beneficiary selfdestructs where no native token actually moves between accounts.

An internal audit attributed **~753 historical bogus `IN_CONTRACT_TRANSFER` records** (transfers with no corresponding real native-token movement) to this path. This is a separate root cause from the contract-forgery issue.

### Why the heuristic is wrong

go-ethereum's `opSelfdestruct` / `opSelfdestruct6780` call
`AddBalance(beneficiary, contractBalance, tracing.BalanceIncreaseSelfdestruct)`
immediately before `StateDB.SelfDestruct[6780]`. The heuristic reconstructs the transfer *after the fact* and cannot faithfully distinguish:
- **self-beneficiary** (EVM adds then removes the same amount on the same account — net-zero, nothing leaves the account) → phantom `self->self` log;
- **zero-value** selfdestructs;
- **EIP-6780 no-op** (transfer/deletion semantics).

The legacy path additionally **panics** on the self-beneficiary case once `SuicideTxLogMismatchPanic` (Upernavik) is active.

## Fix

Capture the **exact beneficiary and amount at the point the EVM performs the transfer** — the `AddBalance` call tagged `tracing.BalanceIncreaseSelfdestruct`. The capture is established only *after* the credit is successfully persisted, and invalidated on `AddBalance` entry so a failed / zero / stale credit cannot leak a phantom log.

The new derivation emits an `IN_CONTRACT_TRANSFER` **iff a real transfer of native token between two distinct accounts occurred** (`amount > 0` and `beneficiary != contract`). It is skipped for zero-value and self-beneficiary selfdestructs (including the EIP-6780 no-op).

## Hardfork gate — needs maintainer sign-off

The corrected derivation is gated behind a new place-named hardfork, **Zanzibar** (`FeatureCtx.CorrectSelfDestructTransferLog`, `g.IsZanzibar(height)`). The pre-fork heuristic path is preserved **byte-identical** for historical replay.

⚠️ **`ZanzibarBlockHeight` is a PLACEHOLDER sentinel (`math.MaxUint64`)** — the real mainnet/testnet fork height requires maintainer sign-off before release. See `// TODO(maintainer): set fork height` in `blockchain/genesis/genesis.go`.

## Non-consensus rationale

Transaction logs are **not** part of the receipt root or state root (`Receipt.ConvertToReceiptPb` / `Serialize` do not include `transactionLogs`), so this does not affect block validity. But because it changes historical-replay output for a given height, it is still fork-gated for determinism/consistency.

## Historical cleanup (out of scope here)

The ~753 pre-existing bogus records are corrected out of band via the `tools/txlogpatch` generator in **#4870** (extend its input CSV). **This PR is the forward fix only.** The exact fork height and the 753-record CSV require the maintainer / analyser output.

## Tests

- **Reproduction (pre-fork):** the current heuristic emits a phantom `self -> self` `IN_CONTRACT_TRANSFER` for an EIP-6780 self-beneficiary selfdestruct (asserted `Sender == Recipient`).
- **Post-fork:** correct beneficiary/amount for external beneficiary (legacy + 6780); **no log** for EIP-6780 self-beneficiary, zero-value, or legacy self-beneficiary; stale-capture invalidation guard.
- **FeatureCtx boundary:** `CorrectSelfDestructTransferLog` is off at `Zanzibar-1` and on at `Zanzibar` / `Zanzibar+1`.

Verified: `gofmt` clean, `go build ./...`, `go vet` on evm/protocol/e2etest, `go test ./action/protocol/execution/... ./action/protocol/ ./blockchain/genesis/... ./blockchain/integrity/...` green. Reviewed adversarially with Codex to convergence (one finding — capture-before-persist — addressed; Erigon-backend and fork-widening concerns verified as non-issues).

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -172,6 +172,11 @@ type (
 		// contracts are committed and written back
 		AlwaysWriteCachedContract bool
 		NoCandidateExitQueue      bool
+		// CorrectSelfDestructTransferLog derives SELFDESTRUCT IN_CONTRACT_TRANSFER
+		// transaction logs from the actual EVM balance movement (exact beneficiary
+		// and amount; no log when no native token moves) instead of the fragile
+		// "last AddBalance" heuristic.
+		CorrectSelfDestructTransferLog bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
@@ -346,6 +351,7 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			PrePectraEVM:                            !g.IsYap(height),
 			AlwaysWriteCachedContract:               !g.IsYap(height),
 			NoCandidateExitQueue:                    !g.IsYap(height),
+			CorrectSelfDestructTransferLog:          g.IsZanzibar(height),
 		},
 	)
 }
--- a/action/protocol/context_test.go
+++ b/action/protocol/context_test.go
@@ -15,6 +15,8 @@ import (
 	"github.com/iotexproject/go-pkgs/hash"
 	"github.com/iotexproject/iotex-address/address"
 	"github.com/stretchr/testify/require"
+
+	"github.com/iotexproject/iotex-core/v2/blockchain/genesis"
 )
 
 func TestRegistryCtx(t *testing.T) {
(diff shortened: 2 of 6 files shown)


envestcc:
fixed by https://github.com/iotexproject/iotex-core/pull/4910

PR #4874, opened 2026-07-07
[blockdao] add TransactionLogIndexer to patch transaction logs at startup

envestcc:
## What

Adds a startup-loaded, read-only patch channel for block transaction (system) logs, parallel to the existing `ReceiptIndexer`. A node can load a small BoltDB that overrides the transaction logs served for a sparse set of block heights.

This is needed because `iotextypes.Receipt` protobuf has no transaction-log field, so the existing `ReceiptIndexer` (which patches receipts) cannot carry corrected transaction logs.

The same change is already merged to `rc_2.4.2` (#4870); this brings it to `master`.

## Why

Some historical blocks contain `IN_CONTRACT_TRANSFER` transaction-log records that do not correspond to any real native-token transfer. These records are returned by `GetTransactionLogByActionHash` / `GetTransactionLogByBlockHeight` and can mislead consumers that reconstruct balances from transaction logs. They are not part of any receipt or state root, so they can be corrected by a startup-loaded patch without a re-sync or hard fork.

## Change

- `blockchain/blockdao/transactionlogindexer.go`: `TransactionLogIndexer`, a read-only BoltDB keyed by block height storing corrected `iotextypes.TransactionLogs`. The set of patched heights is sparse (non-contiguous). An optional `WithPreloadHeights` builds the patched-height set in memory at `Start`, so a query for the common unpatched-height case is answered without a DB lookup.
- `blockchain/blockdao/blockdao.go`: `WithTransactionLogIndexer` option + lifecycle registration; `TransactionLogs(height)` and `GetReceipts` consult the indexer first and fall back to the block store on not-exist / bucket-not-exist.
- `blockchain/config.go`, `chainservice/builder.go`: new optional `patchTransactionLogPath`, loaded read-only (with preloading enabled) at startup — mirrors `patchReceiptIndexPath`.
- `tools/txlogpatch`: generator that fetches each block's transaction logs from a node API, strips the target records, and writes the patch BoltDB.

## Consensus impact

None. Transaction logs are not part of any receipt or state root; only the transaction logs served for the patched heights change. Nodes that do not configure the patch are unaffected (no fork).

## Config

```yaml
chain:
  patchTransactionLogPath: /var/data/txlog.db.patch
```

Only set `patchTransactionLogPath` if the file exists at that path — a node configured with a missing patch file will fail to start. Leave it unset (the default) to run without the patch. Generate the file with `go run ./tools/txlogpatch -csv <records.csv> -endpoint <api:443> -out txlog.db.patch`.

## Test

`TestTransactionLogIndexer` and `TestTransactionLogIndexerPreloadEmpty` unit tests cover the sparse read path and the preload fast path; the blockdao `TransactionLogs` / `GetReceipts` read paths fall through to the block store for unpatched heights. `go build ./...` and the blockdao tests pass on `master`.


tools/txlogpatch/main.go
```diff
@@ -0,0 +1,199 @@
+// Copyright (c) 2024 IoTeX Foundation
```
raullenchai, line 1: 2026

raullenchai (approved):

PR #4877, opened 2026-07-08
feat(metrics): expose on-disk size of each db file

### LINKED ISSUE #4160: Add metrics for db size
### What would you like to be added
Add metrics for size of every db file

### Why is this needed
As transactions increase, especially after increase the block gas limit, it is hoped to observe the actual growth of the database in order to have a rough idea of the disk requirements.

### How important you think this is for IoTeX
- [x] nice to have

raullenchai:
## What

Adds a Prometheus gauge exposing the on-disk size of each node database file, so operators can observe DB growth (especially after block gas limit bumps) and estimate disk requirements.

## Details

- New metric `iotex_db_file_size_bytes{db, path}` — one series per configured DB file (bounded cardinality).
- Covers: chain, trie, index, bloomfilter/candidate/staking/contractstaking indexes, blob store, patch receipt index, history index, gravity chain, and consensus DBs. Empty/disabled paths are skipped.
- Populated by a lightweight background `RecurringTask` (5m interval) started/stopped via the existing `ChainService` lifecycle — no goroutine leak.
- Handles pebbledb-style **directory** DBs by recursively summing contained files; single-file boltdb DBs are stat'd directly.
- Missing/rotated files are skipped gracefully (stat error → skip, no crash).
- The `ChainDBPath` URL form is resolved the same way as `buildBlockDAO` (`file://`/plain → local path; `grpc://` remote → skipped).
- Metric is registered once per process via `init()`, avoiding duplicate-register panics when multiple `ChainService` instances are constructed.

Observability-only: no consensus or state-path impact.

## Test

- Unit tests for file/directory/missing stat, empty-path skipping, remote (grpc) chain-db skipping, and gauge population.
- `gofmt` clean, `go build ./...`, `go vet ./chainservice/...`, `go test ./chainservice/` pass.

Closes #4160

### DIFF
--- a/chainservice/builder.go
+++ b/chainservice/builder.go
@@ -886,10 +886,20 @@ func (builder *Builder) build(forSubChain, forTest bool) (*ChainService, error) 
 	if err := builder.buildActionSyncer(); err != nil {
 		return nil, err
 	}
+	builder.buildDBFileSizeCollector()
 	cs := builder.cs
 	builder.cs = nil
 
 	return cs, nil
+}
+
+// buildDBFileSizeCollector registers a background task that periodically exports the on-disk
+// size of each configured DB file as a Prometheus gauge. It is observability-only and has no
+// consensus/state impact. Skipped when no DB files are configured (e.g. some test builds).
+func (builder *Builder) buildDBFileSizeCollector() {
+	if collector := newDBFileSizeCollector(builder.cfg); collector != nil {
+		builder.cs.lifecycle.Add(collector)
+	}
 }
 
 // estimateTipHeight estimates the height of the block at the given time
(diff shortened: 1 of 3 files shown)


guo (approved):
Approving — observability-only, no consensus/state impact.

Verified:
- `init()` registers the gauge once per process → no duplicate-register panic across multiple ChainService instances.
- Bounded label cardinality (static db names + static paths); empty/remote (grpc://) paths correctly skipped.
- Directory (pebbledb) sizes summed recursively; single-file (boltdb) stat'd directly; file rotated mid-walk is skipped, not fatal.
- Collector runs as a lifecycle-managed RecurringTask (started/stopped with ChainService) → no goroutine leak.
- Good unit coverage. CI green.

Minor (non-blocking): series are never deleted and the startup sample walks synchronously — both negligible.

PR #4878, opened 2026-07-08
fix(api): use a fresh tracer per transaction in debug_traceBlock*

### LINKED ISSUE #4825: debug_traceBlock*: shared tracer across transactions causes incorrect per-tx results
## Summary

`debug_traceBlockByNumber` / `debug_traceBlockByHash` use a **single tracer instance shared across all transactions** in the block. This diverges from geth, where `traceTx` creates a fresh tracer per transaction.

## Current behavior

`coreService.traceBlock` creates one tracer and passes it to `WorkingSetAtTransaction`, which processes all block actions sequentially. The tracer's internal state accumulates across transactions:

- **`prestateTracer`**: `t.pre` grows across txs, so the result for tx N contains prestates from tx 0…N combined, not just tx N alone.
- **`callTracer`**: the callstack is never reset between txs. After tx 0's `OnEnter(0)` adds a frame, tx 1's `OnEnter(0)` pushes a second frame. `GetResult()` requires `len(callstack) == 1` and returns `"incorrect number of top-level calls"`, causing that tx's result to be silently dropped from the output.

## Expected behavior (geth)

In geth's `traceBlock` ([`eth/tracers/api.go`](https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/api.go)):

```go
for i, tx := range txs {  // txs = block.Transactions() — user txs only
    res, err := api.traceTx(ctx, tx, msg, txctx, blockCtx, statedb, config, nil)
    results[i] = &txTraceResult{TxHash: tx.Hash(), Result: res}
}
```

`traceTx` calls `DefaultDirectory.New(...)` to create a **fresh tracer for each transaction**. Each tx gets an independent trace with no state leaking from previous txs.

Additionally, system-level calls (EIP-4788 beacon root, EIP-2935 history storage) are executed before the loop **without a tracer**, so they never appear in the results array.

## Root cause

`coreService.traceBlock` creates one tracer upfront and `WorkingSetAtTransaction` processes all actions (user + system) in a single `Process()` call. There is no mechanism to reset or replace the tracer between transactions mid-flight.

## What needs to change

- `traceBlock` should create a fresh tracer per user transaction.
- System actions (e.g. `GrantReward`) should not appear in trace results, consistent with geth excluding system txs from `block.Transactions()`.

The current architecture — one tracer + `WorkingSetAtTransaction` with all actions — makes per-tx tracer isolation difficult. One approach: instantiate a new tracer inside `CaptureTx` for the next tx and switch the `VMConfigCtx` hook via a proxy `*tracing.Hooks`. Another approach: call `WorkingSetAtTransaction` once per user tx.

## Related

- Fix for `systemCallContext` stripping the tracer to prevent system EVM calls from polluting the tracer callstack (already merged).
- Fix for nil-checking `OnExit`/`OnTxEnd`/`OnTxStart`/`OnEnter` in `TraceStart`/`TraceEnd` so tracers that don't register all hooks (e.g. `prestateTracer`) don't panic (already merged).

raullenchai:
## Problem

`debug_traceBlockByNumber` / `debug_traceBlockByHash` created a **single tracer instance** and shared it across every transaction in the block, diverging from geth (which instantiates a fresh tracer per tx via `DefaultDirectory.New`). The shared tracer accumulated per-tx state:

- **prestateTracer**: `t.pre` grew across txs, so tx N's result contained the combined prestates of tx 0…N instead of only tx N's touched accounts.
- **callTracer**: the callstack was never reset between txs. After tx0's `OnEnter` added a frame, tx1's `OnEnter` pushed a second; `GetResult()` requires `len(callstack) == 1` and returned `"incorrect number of top-level calls"`, **silently dropping that tx's result** from the output array.

## Root cause

`coreService.traceBlock` builds one tracer up front and passes it to `WorkingSetAtTransaction`, which processes all block actions in a single pass. There was no mechanism to reset/replace the tracer between transactions.

## Fix

The block is processed in a single pass over a working set that must carry state forward from one tx to the next, so we cannot simply re-invoke the trace entrypoint per tx. Instead:

- Wire a **stable `*tracing.Hooks`** into the EVM via `VMConfigCtx`.
- Between transactions, swap in a **freshly instantiated tracer** (built exactly like the single-tx `debug_traceTransaction` path, via `parseTracer`). `CaptureTx` — invoked once per traced user tx after its hooks have fired — collects the current tracer's result and rearms a new tracer for the next tx.

Per-tx traces are now fully independent and preserve execution order. Single-tx tracing is unchanged. System actions (e.g. `GrantReward`) already skip `TraceStart`/`TraceEnd`, so they never appear in results.

This is a **read-only tracing (debug API) path** and does not affect state or receipt roots — non-consensus.

## Test

Adds `TestTraceBlockPerTxIsolation` (e2e): traces a block containing two independent transfers (two disjoint senders) and asserts:
- **callTracer** returns one valid top-level call result per tx (none dropped),
- each **prestateTracer** entry contains only its own tx's accounts (no single entry leaks both senders).

The test fails on `master` (callTracer returns 1 entry instead of 2) and passes with this fix.

## Verification

- `go build ./...`, `gofmt -l` clean, `go vet ./api/...` clean
- `go test ./api/... -run 'Trace|trace'` and `go test ./e2etest/ -run 'Trace'` pass
- Independent Codex adversarial review: CONVERGED (isolation/ordering/nil-safety correct; only note was a pre-existing `parseTracer` timeout no-op shared with the single-tx path, out of scope)

Closes #4825

### DIFF
--- a/api/coreservice.go
+++ b/api/coreservice.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/tracing"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
 	"github.com/ethereum/go-ethereum/eth/tracers"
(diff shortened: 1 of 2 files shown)


guo (approved):
Approving — correct per-transaction tracer isolation fix.

Verified:
- The stable `*tracing.Hooks` is re-armed between txs in `CaptureTx`, and a fresh EVM is constructed per execution reading the updated pointer, so prestateTracer no longer accumulates touched accounts and callTracer's callstack resets per tx (fixing the silently-dropped later txs).
- `parseTracer` is a pure parse (builds a fresh tracer, does not mutate ctx); VMConfigCtx is set explicitly with the shared hooks — clean separation.
- Error handling keeps results/retvals/receipts aligned, and a GetResult error on one tx no longer aborts the rest.
- Regression test exercises both failure modes (callTracer drop + prestate accumulation) with two disjoint txs. CI green.

Pre-existing / orthogonal (not this PR): `parseTracer`'s timeout goroutine is canceled immediately by its `defer cancel()`, so the per-trace timeout isn't actually enforced — worth a separate look.

PR #4879, opened 2026-07-08
test(staking): fix flaky TestProtocol_FetchBucketAndValidate (shared-state isolation)

### LINKED ISSUE #4813: bug: TestProtocol_FetchBucketAndValidate flaky in staking test suite

raullenchai:
## Summary

`TestProtocol_FetchBucketAndValidate` fails intermittently in the full staking
suite (and under `-count`/`-shuffle`) but passes in isolation — issue #4813.

## Root cause

The test patched the **process-global** `(*candSM).NativeBucket` machine code
(and the `isSelfStakeBucket` function) *separately in each subtest* with
gomonkey, resetting between subtests. Repeatedly re-patching the same global
function code is unreliable: the second and later re-patches intermittently did
not take effect, so the real `NativeBucket` ran and returned
`ErrStateNotExist`. The failing assertion was:

```
Error "failed to fetch bucket by index 1: state does not exist"
  does not contain "bucket owner does not match"
```

This shared mutable state (the patched global code) is what made the test
order-dependent and flaky.

## Fix

Patch each shared target **exactly once** for the whole test and drive
per-subtest behavior through closure variables (`bucket`/`bucketErr`,
`isSelfStake`/`isSelfStakeErr`), instead of re-patching/resetting global
function code across subtests. Also bind assertions to each subtest's own `*T`
so a failing subtest no longer calls `FailNow` on the parent (the
"subtest may have called FailNow on a parent test" panic).

No production/consensus code changes. All scenarios and assertions are
unchanged — no retries, skips, sleeps, or hidden reordering.

## Verification

- `go test ./action/protocol/staking/ -count=100 -run TestProtocol_FetchBucketAndValidate` — green (was ~100% flaky before on Go 1.26.1 darwin/arm64)
- `go test ./action/protocol/staking/ -count=1 -shuffle=on` full suite x20 — green
- `go test ./action/protocol/staking/ -race -count=20 -run TestProtocol_FetchBucketAndValidate` — green
- `gofmt -l` clean, `go build ./...`, `go vet ./action/protocol/staking/...` clean

Closes #4813

### DIFF
--- a/action/protocol/staking/handlers_test.go
+++ b/action/protocol/staking/handlers_test.go
@@ -2884,78 +2884,68 @@ func TestProtocol_HandleDepositToStake(t *testing.T) {
 }
 
 func TestProtocol_FetchBucketAndValidate(t *testing.T) {
-	require := require.New(t)
+	r := require.New(t)
 	ctrl := gomock.NewController(t)
 	sm, p, _, _ := initAll(t, ctrl)
+	csm, err := NewCandidateStateManager(sm)
+	r.NoError(err)
+
+	// The patched targets — the process-global `(*candSM).NativeBucket` machine
+	// code and the `isSelfStakeBucket` function — are shared mutable state. Patch
+	// each of them exactly once here and vary behavior per subtest through these
+	// closure variables, instead of repeatedly patching/resetting the same global
+	// function code across subtests. Repeated re-patching of the same target was
+	// unreliable (issue #4813: the second and later re-patches intermittently did
+	// not take effect, so the real NativeBucket ran and returned ErrStateNotExist),
+	// which made this test order-dependent and flaky in the full staking suite.
+	var (
+		bucket         *VoteBucket
+		bucketErr      error
+		isSelfStake    bool
+		isSelfStakeErr error
+	)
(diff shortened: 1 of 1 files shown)


guo (approved):
Approving — test-only.

Removes the anti-pattern of repeatedly re-patching the same process-global function code across subtests (the flakiness root cause per #4813: later re-patches intermittently didn't take effect). Now patches `NativeBucket`/`isSelfStakeBucket` exactly once and varies behavior via closure variables across the sequential (non-parallel) subtests — no data race, deterministic. CI green.

PR #4880, opened 2026-07-08
feat(rewarding,staking): distribute epoch voter reward at GrantEpochReward (IIP-59 PR 3/6)

envestcc:
## Summary

Third PR of the IIP-59 protocol-native voter reward distribution series (spec: iotexproject/iips#73). **Stacked on #4866** — PR 2 freezes the per-epoch commission rate and per-voter weight snapshot; this PR reads that frozen state and splits the epoch reward across the delegate (commission) and their voters, replacing the "credit-all-to-reward-address, let Hermes redistribute" leg for the epoch stream.

- **No behavior change pre-flag.** Gated on `FeatureCtx.NoVoterRewardDistribution` (introduced in PR 1). Legacy path — `grantToAccount(rewardAddr, epochReward)` + `EPOCH_REWARD` log — still runs when the flag is set, when the candidate is missing/unresolvable, when commission rate is zero, or when `Identity` is empty.
- **What's new**:
  - `distributeVoterReward(ctx, sm, cand, rewardAddr, total, blkHeight, actionHash) → (logs, handled, err)` in `action/protocol/rewarding/voter_reward.go`. Called once per delegate from the existing `GrantEpochReward` loop; `handled=false` means "not eligible, run the legacy path."
  - `RewardLog` proto grows two enums: `VOTER_REWARD = 5` and `DELEGATE_COMMISSION = 6`. Legacy paths keep emitting `EPOCH_REWARD`.
  - Exported test helper `staking.WriteVoterWeightSnapshotForTest` so rewarding-package tests can seed a per-candidate snapshot without spinning up bucket state.

## Reward math

Given `total = epochReward_for_this_delegate`, snapshot `[(voter_i, weight_i)]` sorted by voter address bytes, and `rate ∈ [0, 10000]` basis points:

```
commission = total * rate / commissionRateDenominator          // 10000 bps
voterPool  = total - commission
voter_i_amt = voterPool * weight_i / Σ weight_j                 // floor
dust        = voterPool - Σ voter_i_amt                         // integer-division remainder
delegate    = commission + dust                                 // dust folds into commission line
```

Conservation: `Σ voter_i_amt + delegate == total`, exact to the wei. Emits one `VOTER_REWARD` log per positive voter share (skips zero-weight voters entirely) and one `DELEGATE_COMMISSION` log per delegate that took the IIP-59 path.

Degenerate cases:
- `rate == 10000` (100% commission) or `Σ weight == 0` → single `DELEGATE_COMMISSION` log for the full amount, no voter logs. This matches the "no voters snapshotted" reality.
- Snapshot missing (`ErrStateNotExist`) → same degenerate branch, delegate keeps everything.

## Legacy fallback triggers

`distributeVoterReward` returns `handled=false` (caller runs the existing `grantToAccount` + `EPOCH_REWARD` log path) when any of:
- `fCtx.NoVoterRewardDistribution` is set (pre-flag).
- `cand == nil` (no candidate object passed).
- `cand.CommissionRate == 0`.
- `cand.Identity == ""` (unresolvable).

Everything else — `rate > 10000`, `Identity` fails to parse, `rewardAddr == nil`, snapshot read error other than `ErrStateNotExist` — is a hard error and propagates. Silent legacy fallback would mask a real misconfiguration.

## `GrantEpochReward` wiring

`splitEpochReward` return signature widens from 3-tuple → 4-tuple (add `[]*state.Candidate` aligned by index with `addrs`/`amounts`) so the reward loop has the full `state.Candidate` for the commission-rate lookup. The loop:

```go
for i, addr := range addrs {
    logs, handled, err := p.distributeVoterReward(ctx, sm, candidates[i], addr, amounts[i], blkHeight, actionHash)
    if err != nil { return err }
    if handled {
        rewardLogs = append(rewardLogs, logs...)
        continue
    }
    // legacy: grantToAccount + EPOCH_REWARD log
}
```

**Fix vs. PoC**: the PoC passed `state.Candidate.Address` (operator) to `CandidateByIdentifier`, which expects an `Identity`. Now the loop passes the whole `*state.Candidate` pointer, no field confusion.

## Tests

`voter_reward_test.go` — 16 cases across:
- Legacy fallback: flag off, `nil cand`, `rate=0`, empty identity → assert no state change, `handled=false`.
- Invalid input: `rate > 10000`, unparseable identity, `nil rewardAddr` → assert error propagates.
- No-voter degenerate: no snapshot / snapshot with all-zero weights → single `DELEGATE_COMMISSION` log, delegate credited full amount.
- 100% commission → same degenerate path via the rate branch.
- Even split: 3 equal-weight voters → assert 4 logs (3 `VOTER_REWARD` + 1 `DELEGATE_COMMISSION`), ordering, exact per-voter amount, delegate = commission + 0 dust.
- Dust folding: weights `1:2:4` (sum 7), 1000 IOTX at 5% → `commission=50`, per-voter `{135, 271, 542}` sum 948, dust 2, delegate = 52. Every field asserted.
- Zero-weight voter skipped: 3 voters weighted `w:0:w` → 2 `VOTER_REWARD` logs, middle voter no log and no state credit.

Plus `staking.WriteVoterWeightSnapshotForTest` — thin exported wrapper over the internal snapshot writer (inlines `sort.Slice(entries, ...)` since the internal `sortSnapshotEntries` helper isn't exported and PR 2 didn't need it externally). Used only by tests; zero production callers.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./action/protocol/rewarding/... ./action/protocol/staking/...`
- [x] `go test ./action/protocol/rewarding/... -count=1` — full suite green (50/50 rewarding + new voter_reward suite).
- [x] `go test ./action/protocol/staking/... -count=1` — no regression. Known darwin/arm64 `gomonkey` flake in `TestProtocol_FetchBucketAndValidate/validate_owner` when the full package runs; the same test passes 3× when run alone. Unrelated to this PR (reproduces on `master`).
- [ ] Cross-flag matrix at real block heights — deferred to PR 5's e2e determinism harness.

## Series roadmap

- PR 2: #4866 — incremental voter weight view + snapshot @ `PutPollResult`
- **PR 3 (this PR)**: voter reward distribution (epoch path)
- PR 4: block reward folding — pending pool credited at block, drained at epoch via this PR's helper


envestcc:

PR #4910, opened 2026-07-10
[evm] fix corrupted SELFDESTRUCT transaction log amount

envestcc:
## What

Fixes a bug where the `IN_CONTRACT_TRANSFER` transaction log recorded for a `SELFDESTRUCT` carries the wrong amount.

`generateSelfDestructTransferLog` stored `stateDB.lastAddBalanceAmount` — a `*big.Int` — directly into the transaction log:

```go
Amount: stateDB.lastAddBalanceAmount,
```

That `big.Int` is mutated in place (`SetBytes`/`SetUint64`) by **every subsequent `AddBalance`**. In particular, right after the EVM returns, `ExecuteContract` refunds the gas deposit via `stateDB.AddBalance(origin, depositGas*gasPrice, …)`. Since `SELFDESTRUCT` is the last transfer inside the EVM frame, that gas-refund `AddBalance` runs *after* the self-destruct log was recorded and overwrites the aliased `big.Int` — so the already-recorded log ends up reporting the **gas-refund amount** instead of the value actually transferred to the beneficiary.

## Impact

Transaction logs are not part of any receipt root or state root, so **balances, receipts, and block hashes are unaffected** and nodes do not fork. Only the transaction logs served for such actions are wrong; systems that reconstruct balances from transaction logs (indexers, wallets, exchanges) can be misled.

On mainnet this surfaced on MEV/arbitrage bots that make a value-carrying sub-call which reverts and then self-destruct (post-Upernavik). Because a given bot uses a fixed `gasLimit`/`gasPrice`, the corrupting gas-refund is a constant, so the same wrong amount is stamped onto every such log in a block (e.g. `0.274835 IOTX`), while the real self-destruct amounts differ per action.

## Fix

Copy the amount instead of aliasing the mutable pointer:

```go
Amount: new(big.Int).Set(stateDB.lastAddBalanceAmount),
```

(This matches how the in-contract-transfer log records its amount.)

## Test

`e2etest/TestSelfDestructTransactionLogAmount` reproduces the on-chain pattern — a contract makes a value-carrying sub-call that reverts, then self-destructs, batched in one minted block — and asserts the self-destruct log reports the transferred amount, not the gas refund. It fails on the old code (log amount = gas refund) and passes with the fix.

## Correcting already-recorded logs

`tools/txlogpatch` gains a `-correct` mode so the transaction logs already written for affected heights can be corrected without a re-sync (via the existing `chain.patchTransactionLogPath` / `TransactionLogIndexer` channel). Unlike the default strip mode (which removes a record), correct mode rewrites only the amount of a matched `IN_CONTRACT_TRANSFER` record — CSV: `block_height,tx_hash,sender,recipient,wrong_amount_rau,correct_amount_rau`. The correct amount for each record is the value actually transferred by the self-destruct (the `SELFDESTRUCT` opcode value, obtainable from `debug_traceTransaction`).

### DIFF
--- a/action/protocol/execution/evm/evmstatedbadapter.go
+++ b/action/protocol/execution/evm/evmstatedbadapter.go
@@ -1076,7 +1076,11 @@ func (stateDB *StateDBAdapter) generateSelfDestructTransferLog(sender string, am
 					Type:      iotextypes.TransactionLogType_IN_CONTRACT_TRANSFER,
 					Sender:    sender,
 					Recipient: stateDB.lastAddBalanceAddr,
-					Amount:    stateDB.lastAddBalanceAmount,
+					// Copy the amount: lastAddBalanceAmount is a mutable *big.Int that is
+					// modified in place by every subsequent AddBalance (e.g. the gas-deposit
+					// refund done right after the EVM returns). Aliasing it here retroactively
+					// corrupts this already-recorded log with a later value.
+					Amount: new(big.Int).Set(stateDB.lastAddBalanceAmount),
 				})
 			}
 		} else {
--- a/e2etest/selfdestruct_txlog_test.go
+++ b/e2etest/selfdestruct_txlog_test.go
@@ -1,0 +1,94 @@
+package e2etest
+
+import (
+	"context"
+	"encoding/hex"
+	"math/big"
+	"testing"
+
+	"github.com/iotexproject/iotex-address/address"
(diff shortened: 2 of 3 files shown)


guo (approved):
Reviewed the rebased head. The defensive big.Int copy fixes the SELFDESTRUCT transaction-log alias without affecting consensus receipts, the historical correction tool remains fail-closed, and the focused real-node regression test plus required CI pass.

PR #4912, opened 2026-07-10
feat(rewarding): DelegateProfile contract-read bridge for IIP-59

envestcc:
## Summary

- Add read-only bridge over the on-chain DelegateProfile contract (mainnet `io1lfl4ppn2c3wcft04f0rk0jy9lyn4pcjcm7638u`) so `PutPollResult` can snapshot per-delegate commission rates from the amended IIP-59 design.
- Bridge fetches `blockRewardPortion` and `epochRewardPortion` per delegate via `getProfileByField`, inverts voter-take portions into commission basis points, and returns a per-delegate map keyed by identity string.
- Package has no dependency on `protocol.StateManager`; exercised by 14 unit tests using an ABI-round-tripping in-process fake reader.

## Design notes

- **Read API**: `getProfileByField(_delegate, _field) returns bytes` invoked twice per delegate. Avoids reverse-engineering `getEncodedProfile`'s undocumented composite return format.
- **Registered semantics**: empty bytes for either portion field marks the delegate `Registered=false`, and the caller MUST fall back to the legacy Hermes path. A partial profile (one field set, other empty) is treated as unregistered on purpose — the IIP relies on the "either fully opted-in or fully legacy" invariant so a delegate cannot end up with one reward stream migrated mid-fork.
- **Explicit zero**: a non-empty single `0x00` byte encodes an explicit 0% voter-take and yields `Registered=true` with a 100% commission split. This is distinguishable from "field never set".
- **Range check**: values exceeding `uint64` or `10000` basis points are rejected rather than silently truncated via `SetBytes → Uint64`.
- **Determinism**: `Snapshot` iterates the caller-supplied delegate list in order (2N sequential reads); errors abort with the offending delegate address named.

## Sequencing

Refs iotexproject/iips#74.

- Depends on: none (isolated read-only package).
- Follow-up **PR 2'**: `PutPollResult` calls `Snapshot`, freezes commission rates into the poll snapshot per delegate. That PR wires the concrete `ContractReader` around `evm.SimulateExecution`.
- Follow-ups **PR 3'/4'/4.6/4.7**: consume the snapshotted rates at epoch close, block-reward folding, compound bridge, and batched delegate-distributed log emitter respectively.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./action/protocol/rewarding/delegateprofile/...`
- [x] `go test ./action/protocol/rewarding/delegateprofile/` — 14/14 pass
- [ ] Reviewer: confirm `getProfileByField` return-encoding assumption (big-endian uint256 bytes, empty ⇒ unset) matches the deployed contract behaviour.
- [ ] Reviewer: confirm mainnet contract address is the intended source of truth for this fork (governance handover is a separate prereq).

Draft — subsequent PRs in the IIP-59 stack build on this bridge.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/action/protocol/rewarding/delegateprofile/abi.go
+++ b/action/protocol/rewarding/delegateprofile/abi.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2026 IoTeX Foundation
+// This source code is provided 'as is' and no warranties are given as to title or non-infringement, merchantability
+// or fitness for purpose and, to the extent permitted by law, all liability for your use of the code is disclaimed.
+// This source code is governed by Apache License 2.0 that can be found in the LICENSE file.
+
+package delegateprofile
+
+// abiJSON is the minimal ABI subset needed to read commission portions from
+// the existing DelegateProfile contract deployed at:
+//
+//	mainnet:  0xfa7f50866ac45d84adf54bc767c885f92750e258
+//	testnet:  0xd19ffB48a5C18B77c541D32c1B1ac2440c287774
+//
+// The bridge only ever performs read-only view calls, so nothing beyond
+// getProfileByField(address,string) is included. The full contract ABI lives
+// off-chain; expanding this subset requires a matching audit of the caller.
+const abiJSON = `[
+	{
+		"constant": true,
+		"inputs": [
+			{ "name": "_delegate", "type": "address" },
+			{ "name": "_field", "type": "string" }
+		],
+		"name": "getProfileByField",
+		"outputs": [
+			{ "name": "", "type": "bytes" }
+		],
+		"payable": false,
+		"stateMutability": "view",
+		"type": "function"
+	}
+]`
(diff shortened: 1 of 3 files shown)


envestcc:
Force-pushed one additional commit (`49d2c6245`) folding in a bridge semantic change previously landed on PR #4915:

- `Bridge.Snapshot` now absorbs per-delegate read errors (RPC hiccup, out-of-range portion, ABI mismatch) and returns a `Registered=false` entry for the affected delegate instead of erroring out. Nil-reader / nil-address still hard-error — those are wiring bugs, not on-chain data issues.
- Rationale: the previous behaviour let one malformed on-chain `DelegateProfile` field deterministically halt block production at every subsequent epoch boundary. Deterministic reward-path fallback (route bad delegate through legacy Hermes) beats deterministic block-production failure when the fallback exists.
- Errors are logged via `log.L().Warn` with delegate + contract for observability.

Test updates in the same commit rename the affected assertions (`*_Rejected` → `*_DegradesToUnregistered`) and add `TestSnapshot_PerDelegateErrorIsolated` to confirm one bad delegate does not poison batched successful reads.

PR #4915 rebased on top; its follow-up commit now only touches the staking-side docstring + one test rename.

envestcc:

Issue #4745, opened 2025-11-03
debug_traceBlockByNumber returning various errors


gflow33:
### What version of iotex-core image (or code branch) are you using?
`v2.3.0-d3c7dca`

### What operating system and processor architecture are you using?
Ubuntu 22.04 amd64

### What did you do? If possible, provide a recipe for reproducing the error.
When calling `debug_traceBlockByNumber` on certain block numbers it returns errors in the logs and response.

### What did you expect to see?
No errors 

### What did you see instead?
`'content-type: application/json' -d '{"jsonrpc":"2.0","method":"debug_traceBlockByNumber","params":["0x17f2a83",{"tracer":"callTracer","timeout":"30s"}],"id":0}
`

```
[...]

PR #4914, opened 2026-07-10
fix(evm): don't panic in SetNonce on the simulation/trace path

raullenchai:
## Problem

`StateDBAdapter.SetNonce` (`action/protocol/execution/evm/evmstatedbadapter.go`) had two direct `log.T(ctx).Panic` calls that were **not** gated by `panicUnrecoverableError`. Most error sites in this file go through `assertError`, which only panics when `panicUnrecoverableError == true` (the block-processing/consensus path, enabled post-Upernavik via `featureCtx.PanicUnrecoverableError = IsUpernavik(height)`) and otherwise logs an `Error` + records the error. These two `SetNonce` panics bypassed that gate, so `debug_traceBlock*` / `eth_call` replaying certain historical blocks **crashed the node** instead of returning a soft error.

This is the exact stack in #4745: `SetPendingNonce` returns `"nonce overflow"` while replaying a pre-Upernavik block (height 25,111,171, `0x17f2a83`), where `panicUnrecoverableError` is `false`, yet the ungated direct panic fired anyway — then got recovered in `WorkingSetAtTransaction` and cascaded into a downstream nil-pointer deref in `traceBlock`.

## Fix

Add an `assertMessage` helper that mirrors `assertError`'s gating for panics not tied to an `err` value, and route the ungated nonce panics through gating:

- **`SetNonce` "invalid nonce zero"** (pending-nonce mode, `nonce == 0`) → `assertMessage` + early return.
- **`SetNonce` "Failed to set nonce."** (`SetPendingNonce` returns an error — the #4745 site) → routed through the existing `assertError` (it already carries an `err`), then early return.
- **`GetNonce` "invalid pending nonce"** → `assertMessage` for consistency with the sibling nonce guards.

On the consensus path (`panicUnrecoverableError == true`) all three still `Panic` with the **same message and zap fields** — poison-block protection during block production is preserved. On the simulation/trace/API path (`false`) they now log an `Error`, record the error via `logError`, and return gracefully, exactly like every other `assertError` site.

## Consensus-safety

- **Not a hardfork.** Behavior is byte-for-byte identical when `panicUnrecoverableError == true`; the change only affects the `false` path (simulation/trace/`eth_call`), which is non-consensus.
- The early `return` before `StoreAccount` on a `SetPendingNonce` error is state-neutral: `state.Account.SetPendingNonce`'s error paths do not advance the nonce, so nothing partial is written.
- For any block that does **not** hit these guards, control flow is unchanged.

## Scope / other panic sites audited

Other direct `log.T(ctx).Panic` sites in this file were reviewed and deliberately left unchanged:

- **Already gated** by `panicUnrecoverableError`: `GetNonce` "Failed to get nonce.", `RevertToSnapshot`.
- **Pure EVM-internal invariants** (only an implementation bug, not replay data, can trigger them) with a dedicated security guard/test: `SubRefund` "Refund counter not enough!" (asserted by `TestRefundBounds`), `AddBalance` touch-account non-zero amount. Left as hard guards.
- **Hardfork-flag gated / documented as a separate change**: `Snapshot` duplicate version (`fixSnapshotOrder`), `AddLog` `len(topics)!=3` (explicit "separate change" comment), self-destruct balance mismatch (`suicideTxLogMismatchPanic`).

## The other two #4745 failure modes (out of scope)

The issue reports three block traces. This PR fixes only the hard **panic** (block `0x17f2a83`), which is the crash/DoS. The other two are already returned as graceful JSON-RPC errors, not panics, and are **separate replay-data correctness issues** in how `traceBlock` re-executes historical blocks:

- `"nonce too low"` (`0x248c462`) — trace re-execution computes a different nonce than the original block.
- `"failed to deposit gas: ... state does not exist"` (`0x5281e1`) — historical account state not available during replay.

These need separate investigation and are **not** addressed here.

## Testing

- New `TestSetNoncePanicGating` drives both `SetNonce` panic sites with `panicUnrecoverableError=false` (asserts no panic + `Error()` recorded) and `=true` (asserts `require.Panics`). Confirmed the `false` case panics on master before the fix.
- `go build ./...`, `go test ./action/protocol/execution/evm/ -count=1` pass.

Partially addresses #4745 (fixes the SetNonce panic mode only; the two replay-data modes — "nonce too low" and "state does not exist" — remain open for separate investigation, so this PR intentionally does NOT auto-close the issue)

### DIFF
--- a/action/protocol/execution/evm/evmstatedbadapter.go
+++ b/action/protocol/execution/evm/evmstatedbadapter.go
@@ -299,6 +299,21 @@ func (stateDB *StateDBAdapter) assertError(err error, msg string, fields ...zap.
 	return true
 }
 
+// assertMessage mirrors assertError for invariant violations that are not tied to an
+// error value. On the block-processing/consensus path (panicUnrecoverableError == true)
+// it panics, halting the node on a truly unrecoverable state error (poison-block
+// protection). On the simulation/trace/API path (panicUnrecoverableError == false) it
+// degrades to an Error log and records the error, so the caller returns a soft error
+// instead of crashing the process. This keeps these direct panics consistent with
+// assertError, which already never panics when panicUnrecoverableError is false.
+func (stateDB *StateDBAdapter) assertMessage(msg string, fields ...zap.Field) {
+	if stateDB.panicUnrecoverableError {
+		log.T(stateDB.ctx).Panic(msg, fields...)
+	}
+	log.T(stateDB.ctx).Error(msg, fields...)
+	stateDB.logError(errors.New(msg))
+}
+
 // Error returns the first stored error during evm contract execution
 func (stateDB *StateDBAdapter) Error() error {
 	return stateDB.err
@@ -472,7 +487,8 @@ func (stateDB *StateDBAdapter) GetNonce(evmAddr common.Address) uint64 {
 	}
 	if stateDB.useConfirmedNonce {
 		if pendingNonce == 0 {
-			log.T(stateDB.ctx).Panic("invalid pending nonce")
+			stateDB.assertMessage("invalid pending nonce")
+			return pendingNonce
 		}
 		pendingNonce--
 	}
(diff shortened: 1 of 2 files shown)


envestcc:
the issue https://github.com/iotexproject/iotex-core/issues/4745 should already be fixed

PR #4911, opened 2026-07-10
feat(staking): add Candidate.VoterRewardOnchainOptIn for IIP-59 (amended)

envestcc:
## Summary

Adds a persistent per-delegate opt-in flag on `staking.Candidate` that gates
protocol-native voter reward distribution.

This is the first code PR against the **amended** IIP-59 design (see
[iotexproject/iips#74](https://github.com/iotexproject/iips/pull/74)). The
earlier native-only design and its implementation PRs
(#4865 / #4866 / #4880 / #4881) were superseded and closed on 2026-07-10.

Draft status intentional — will promote to Ready after the IIP amendment
lands and the downstream PRs in the series (PR 4.5 contract-read bridge,
PR 2' snapshot rework, PR 3'/4' rewarding rework) are ready to stack.

## What changed

- `stakingpb.Candidate` gains field 11: `bool voterRewardOnchainOptIn`.
- `staking.Candidate` Go struct gains `VoterRewardOnchainOptIn bool`;
  `Clone`, `Equal`, `toProto`, `fromProto` all thread the new field.
- New test `TestSerWithVoterRewardOnchainOptIn` covers:
  - proto round-trip with the flag both false and true,
  - `Equal` sensitivity to the flag,
  - `Clone` independence (flipping one copy doesn't affect the other).

Default zero-value keeps every existing candidate opting **out** on decode,
so behavior is unchanged pre-fork and, post-fork, unchanged for any delegate
that hasn't sent the (later) opt-in action.

## Deferred to follow-up PRs

Per the amended IIP-59:

- **`SetVoterRewardOptIn` native action + handler**: mutates the flag with
  a one-epoch delay via the `PutPollResult` snapshot. Requires an external
  `iotex-proto` proto change (new action type). Separated to keep this PR
  self-contained.
- **`BlockCommissionRate` / `EpochCommissionRate`**: per-epoch snapshot
  values populated from the DelegateProfile contract at `PutPollResult`;
  they live on the poll snapshot (PR 2'), not on the persistent Candidate.
  See the amended proposal for the flow.

## Test plan

- [x] `make fmt` — clean
- [x] `go build ./...` — clean
- [x] `go test ./action/protocol/staking/` — 264 tests pass, includes new roundtrip test
- [ ] `make lint` — pending review
- [ ] Reviewer: confirm field 11 doesn't collide with any WIP proto branch

## Refs

- Amended IIP: iotexproject/iips#74
- Superseded PRs (closed 2026-07-10): #4865, #4866, #4880, #4881

PR #4915, opened 2026-07-10
feat(staking,poll): freeze DelegateProfile commission rates + opt-in at PutPollResult (IIP-59)

envestcc:
## Summary

- Adds a per-candidate `CandidatePollSnapshot` frozen at every `PutPollResult`, capturing block/epoch commission basis points from the DelegateProfile contract, a `Registered` bit, and the delegate's `VoterRewardOnchainOptIn` flag from PR #4911.
- Wires a config- and fork-gated call into `poll.setCandidates`; downstream rewarding (PR 3', follow-up) will read via `staking.PollSnapshotFor` instead of live sources so mid-epoch mutations do not retroactively re-split rewards.
- Introduces `Blockchain.DelegateProfileContractAddress` (per-network YAML) and `FeatureCtx.NoVoterRewardDistribution` (bound to `!g.IsToBeEnabled(height)`) as the two independent gates.

## Stack

Stacks on top of both:
- #4911 — `iip-59/pr1-candidate-schema-optin` (Candidate.VoterRewardOnchainOptIn field)
- #4912 — `iip-59/pr4.5-delegateprofile-bridge` (DelegateProfile read bridge)

Diff will look larger than the actual delta until those two land.

## Scope explicitly deferred

- Voter-weight source. \`Entries\` is written empty for now; PR 3' has a degenerate branch that treats an empty list as full-amount-to-delegate. A separate follow-up fills in the actual weight computation.
- Rewarding consumption of the snapshot (PR 3').
- \`SetVoterRewardOptIn\` native action (separate PR).

## Feature-flag matrix

| \`NoVoterRewardDistribution\` | \`DelegateProfileContractAddress\` | Behaviour |
|---|---|---|
| true (pre-fork) | any | Snapshot NOT written; \`PollSnapshotFor\` returns \`ErrStateNotExist\` for every candidate. |
| false (post-fork) | empty | Snapshot written with \`Registered=false\`, rates zero, \`OptIn\` from live Candidate. Rewarding falls back to legacy path. |
| false (post-fork) | non-empty | Bridge called; snapshot carries frozen rates + registration bit + opt-in flag. |

## Test plan

- [x] \`go test ./action/protocol/staking/...\` — 14 new tests in \`poll_snapshot_test.go\` cover serialize roundtrip, key layout, nil-bridge / happy-path / partial-profile / bridge-error / invalid-address, reader semantics, opt-in read from live Candidate.
- [x] \`go test ./action/protocol/poll/...\` — no regressions in the poll layer; setCandidates writes snapshots when the fork gate is active.
- [x] \`go build ./...\`, \`go vet ./action/protocol/... ./blockchain/genesis/...\`, \`gofmt -l\` — clean.
- [ ] Determinism spot-check under e2e (deferred to PR 5' harness).

Refs iotexproject/iips#74


envestcc:
Force-pushed a rebase on top of the updated PR #4912.

Change: the bridge-semantic fix (absorb per-delegate read errors as `Registered=false` instead of erroring out) has been folded into PR #4912 where it belongs. The only remaining commit in this PR on top of the writer itself is a docstring + test rename (`FreezePollSnapshot_BridgeErrorPropagates` → `FreezePollSnapshot_BridgeErrorDegradesToLegacy`) that follows the new bridge contract.

No behavioural change relative to the prior HEAD of this branch — same three-commit diff, cleaner split of responsibilities across the stack.

envestcc:

PR #4919, opened 2026-07-12
fix(evm): drop reserved-topic logs with unexpected topic count

guo:
After #4868 the reserved in-contract-transfer topic only appears on contract-emitted logs, which are already dropped from the receipt (never appended to `stateDB.logs`, so `receipt.Logs` and the receipt root are unaffected).

A malformed one (topic count != 3) was still special-cased separately. This drops it uniformly with the rest so the reserved-topic handling is consistent.

Gated on the next upgrade (`ToBeEnabled`) so block replay stays byte-identical before activation. Adds the post-upgrade regression test alongside the existing pre-upgrade invariant.

### DIFF
--- a/action/protocol/execution/evm/evmstatedbadapter.go
+++ b/action/protocol/execution/evm/evmstatedbadapter.go
@@ -91,6 +91,7 @@ type (
 		panicUnrecoverableError    bool
 		enableCancun               bool
 		fixRevertSnapshot          bool
+		fixInContractTransferTopic bool
 		// ignoreBalanceChangeTouchAccount indicates whether to ignore balance change touch account
 		ignoreBalanceChangeTouchAccount bool
 		// skipWriteCleanContract indicates whether to skip writing back read-only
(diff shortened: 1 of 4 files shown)


action/protocol/execution/evm/evmstatedbadapter.go
```diff
@@ -941,9 +951,10 @@ func (stateDB *StateDBAdapter) AddLog(evmLog *types.Log) {
 		// log. It is dropped here, exactly as before it was never appended to
 		// stateDB.logs, so receipt.Logs (and thus the receipt root) is unchanged.
 		//
-		// NOTE: the historical len(topics) != 3 panic is preserved to keep behavior
-		// identical for malformed inputs; hardening that DoS is a separate change.
-		if len(topics) != 3 {
+		// A malformed one (topic count != 3) used to be special-cased; it is now
+		// dropped uniformly with the rest. The legacy special case is kept verbatim
+		// before the upgrade so block replay stays byte-identical, and removed after.
+		if !stateDB.fixInContractTransferTopic && len(topics) != 3 {
```
envestcc: After the hardfork, we can add it to stateDB.logs as a regular log.

envestcc (approved):

PR #4921, opened 2026-07-13
fix(factory): evict sender when an action panics during minting

envestcc:
#4840 recovers panics on the mint path so a doomed draft cannot crash the process, and #4920 extends the same recover to `ValidateBlock`. Neither removes the offending action from the pool, though: today a panic while `runAction` executes a single pending action unwinds past the sender-eviction logic in `validateAndRun` and is only caught by the mint goroutine's top-level recover, discarding the whole draft. The same action is then picked up again on the next mint attempt.

This adds a recover scoped to the single-action `runAction` call during minting, converting the panic into an ordinary error so it flows through the existing default-error handling in `validateAndRun` — which already evicts the sender from the pool before the draft is abandoned for any other unexpected error.

- Still discards this draft (unchanged behavior), but the sender no longer stays in the pool to be retried on the next mint attempt — a repeated failure loop becomes a one-time lost draft.
- No apply-semantics change, so this needs no fork gate.
- Adds a `iotex_mint_action_panics_total` metric and a regression test.

### DIFF
--- a/state/factory/blockpreparer.go
+++ b/state/factory/blockpreparer.go
@@ -26,9 +26,15 @@ var _validateBlockPanicMtc = prometheus.NewCounter(prometheus.CounterOpts{
 	Help: "Number of panics recovered while validating a proposed block (the block was rejected; process kept alive).",
 })
 
+var _mintActionPanicMtc = prometheus.NewCounter(prometheus.CounterOpts{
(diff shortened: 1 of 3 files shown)


guo (approved):
Reviewed the rebased head. The mint-path panic recovery and sender eviction are correctly scoped, and the focused regression coverage passes.

PR #4922, opened 2026-07-13
feat(rewarding): add AutoDeposit compound bridge for IIP-59 (PR 4.6)

envestcc:
## Summary

Adds a read-only bridge to the on-chain AutoDeposit contract so the (upcoming) PR 3' rework of `distributeVoterReward` can decide per-voter compound-vs-credit routing at epoch reward distribution time, per IIP-59 §3.6.

Structural clone of PR 4.5's `delegateprofile.Bridge` — same shape, same ContractReader interface, same per-item fallback discipline — so both on-chain reads present a uniform surface.

## What lands

**New package `action/protocol/rewarding/autodeposit`**

- `Bridge` wraps a pinned contract deployment; `LookupBucket(ctx, reader, voter)` returns `(bucketID, present, err)`.
- `ContractReader` is package-local so tests can supply a fake without pulling the EVM simulator into scope.
- `Route` / `Decision` types match IIP-59 §3.5's `DelegateDistributed.routings[]` wire encoding, giving PR 4.7 a single source of truth.
- `IsBucketEligibleForCompound` applies §3.6 preconditions 2–4 (bucket exists, is native, `AutoStake==true`, currently active, `Owner==voter`). Precondition 1 (positive bucket ID) is enforced inside `LookupBucket`.

**Semantics**

Per IIP-59 §3.6 the drain path must not halt on a single voter's malformed registration:

| Contract returns | `LookupBucket` result |
|---|---|
| positive int256 fitting in uint64 | `(id, true, nil)` |
| zero, negative, or > 2⁶³−1 | `(0, false, nil)` — silent fallback to credit |
| ABI drift / RPC error / nil reader / nil voter | `(0, false, err)` — wiring bug, hard error |

**Genesis**

- New `Blockchain.AutoDepositContractAddress` field, default empty. Empty means compound routing is inactive and every voter share is credited to `unclaimedBalance` for pull-claim — functionally identical to legacy behaviour, so no fork gate churn.

**Staking**

- Exported `VoteBucket.IsUnstaked()` and `IsNative()` wrappers on top of the existing unexported predicates, so the `autodeposit` eligibility helper doesn't have to duplicate staking-internal invariants.

## Explicit non-goals

- No `AddDeposit` invocation. Applying the compound (mutating bucket state) is PR 3''s job via staking's native `handler_stake_add_deposit.go` path.
- No poll integration. Unlike DelegateProfile, AutoDeposit is read **live at drain time**, not frozen at `PutPollResult`.
[...]

PR #45, opened 2018-08-24
Try again

Frankonly:
#42

PR #4923, opened 2026-07-13
feat(rewarding): add DelegateDistributed batched log encoder (IIP-59 PR 4.7)

envestcc:
## Summary

Adds the encoder primitives for IIP-59 §3.2's per-delegate batched `DelegateDistributed` receipt log. `distributeToVoters` (PR 3', upcoming) will call `Pack` once per delegate at epoch close and wrap the returned `(topics, data)` into an `*action.Log`.

Batching keeps total epoch-last-block logs bounded by `|topDelegates| + |orphanDrain|` (~200) regardless of voter count, avoiding receipt-trie bloat under load (top 100 delegates × ~2k voters each).

Structural sibling of PR 4.5 (\`delegateprofile\`) and PR 4.6 (\`autodeposit\`) — same minimal-ABI package shape, same discipline about where the seam lands.

## What lands

**New package \`action/protocol/rewarding/distributedlog\`**

- \`Pack(EventArgs) → (action.Topics, []byte, error)\` produces the exact wire layout an EVM-emitted \`DelegateDistributed\` event would produce, so \`eth_getLogs\` clients (and PR #45's off-chain verifier) decode with stock \`ethers.js\` / \`web3.py\` against the same ABI.
- \`EventArgs\` mirrors §3.2's Solidity signature field-for-field: \`Epoch(indexed)\` + \`Delegate(indexed)\` + \`RewardAddr\` + \`TotalCommission\` + \`TotalVoterPool\` + \`SnapshotHash\` + \`Voters[]\` + \`Amounts[]\` + \`Routings[]\`.
- \`Routings[]\` reuses \`autodeposit.Route\` verbatim — PR 4.6's enum is the single source of truth for the credit/compound wire encoding.
- \`SnapshotHash(voters, weights)\` computes the \`bytes32\` digest of a delegate's frozen voter list, domain-separated (\`iip59.delegatedistributed.snapshot.v1\`) so verifiers can pin the exact bytes and it cannot collide with same-layout hashes from other contexts.

## Semantics

Hard-error on caller-side wiring bugs:

| Condition | Error |
|---|---|
| nil \`Delegate\` / \`RewardAddr\` / any \`Voters[i]\` | \`ErrNilAddress\` |
| nil \`TotalCommission\` / \`TotalVoterPool\` / any \`Amounts[i]\` | \`ErrNilBigInt\` |
| \`len(Voters) != len(Amounts) != len(Routings)\` | \`ErrParallelArrayLengthMismatch\` |

The encoder is not a consensus path, so there is no per-item to degrade — loud failure surfaces PR 3' bugs. **Empty voter list is NOT an error**: a delegate with zero voters still emits a log for observability, per §3.2's \"one log per delegate\".

## Explicit non-goals

- **No \`*action.Log\` construction.** Block context (height, action hash, protocol address) belongs to PR 3'. Mirrors PR 4.5 / 4.6 seams.
- **No event dispatch / subscription.** Encoder, not bus.
- **No proto message.** The event is ABI-encoded so decode-time behaviour is identical to a Solidity-emitted event.
- **No orphan-drain log variant.** §3.3's orphan drain emits the same shape; one encoder covers both call sites.
- **No \`foundationBonus\` field.** §3.5 leaves foundation bonus out of scope.

## Test plan

- [x] \`gofmt -l\` clean on touched files
- [x] \`go vet ./action/protocol/rewarding/distributedlog/...\` clean
- [x] \`go test ./action/protocol/rewarding/distributedlog/... -count=1 -v\` — 15 tests, all green
- [x] Regression \`go test ./action/protocol/rewarding/... ./action/...\` — 1104 tests across 31 packages, all green

Test coverage: happy path, zero voters, both parallel-length mismatch variants, nil-field variants (delegate/reward/commission/pool/per-voter address/per-voter amount), **selector pin** (golden 32 bytes so external verifiers can hard-code the same signature hash), full byte-level **round-trip** via the same ABI, and \`SnapshotHash\` determinism + empty-list constant + truncation semantics.

## Stack

Branched on top of \`iip-59/pr4.6-autodeposit-bridge\` (#4922) because this PR imports \`autodeposit.Route\` at the Go-module level. Once PR 4.6 merges to \`master\`, the base of this PR can be flipped to \`master\` without conflict.

- Depended on by: PR 3' (rework \`distributeVoterReward\`), PR #45 (off-chain verifier).
- Depends on: PR 4.6 (#4922).

### DIFF
--- a/action/protocol/rewarding/distributedlog/abi.go
+++ b/action/protocol/rewarding/distributedlog/abi.go
@@ -1,0 +1,35 @@
+// Copyright (c) 2026 IoTeX Foundation
+// This source code is provided 'as is' and no warranties are given as to title or non-infringement, merchantability
+// or fitness for purpose and, to the extent permitted by law, all liability for your use of the code is disclaimed.
+// This source code is governed by Apache License 2.0 that can be found in the LICENSE file.
+
+package distributedlog
+
+// abiJSON declares the single event this package encodes. Field ordering
+// and indexed flags mirror IIP-59 §3.2 exactly; off-chain verifiers pin
+// the selector derived from this signature.
+const abiJSON = `[
+	{
+		"anonymous": false,
+		"inputs": [
+			{"indexed": true,  "name": "epoch",           "type": "uint64"},
+			{"indexed": true,  "name": "delegate",        "type": "address"},
+			{"indexed": false, "name": "rewardAddr",      "type": "address"},
+			{"indexed": false, "name": "totalCommission", "type": "uint256"},
+			{"indexed": false, "name": "totalVoterPool",  "type": "uint256"},
+			{"indexed": false, "name": "snapshotHash",    "type": "bytes32"},
+			{"indexed": false, "name": "voters",          "type": "address[]"},
+			{"indexed": false, "name": "amounts",         "type": "uint256[]"},
+			{"indexed": false, "name": "routings",        "type": "uint8[]"}
+		],
+		"name": "DelegateDistributed",
+		"type": "event"
+	}
+]`
+
+// eventName is the event's ABI name; used for method lookup and error text.
+const eventName = "DelegateDistributed"
+
+// eventSignature is the canonical Solidity signature. keccak256 of this
+// string is Topics[0]. Kept as a const for the golden-selector test.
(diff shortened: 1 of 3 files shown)


envestcc:

PR #4924, opened 2026-07-13
feat(rewarding,staking): rework distributeVoterReward for IIP-59 (PR 3')

envestcc:
## Summary

PR 3' of the IIP-59 amended stack: wires the three read-side bridges landed
upstream into `GrantEpochReward` so opt-in delegates get their epoch share
split on-chain per IIP-59 §3.2.

For each opt-in delegate, the per-delegate loop now:

- Splits the epoch pool by the frozen commission basis points from the
  poll snapshot (PR 2').
- Allocates the voter pool proportionally by frozen voter weight in
  canonical (snapshot-order) sequence. Modular-division dust attaches to
  the last non-zero-weight voter so `sum(shares) == voterPool` exactly.
- Routes each per-voter share via the AutoDeposit bridge (PR 4.6): if the
  voter has a registered bucket and it passes `IsBucketEligibleForCompound`,
  the share is compounded via `staking.AddDepositForCompound` (new seam);
  otherwise credited to the voter's rewarding `unclaimedBalance`.
- Credits the delegate's commission to its reward address.
- Emits exactly one batched `DelegateDistributed` log per delegate via
  `distributedlog.Pack` (PR 4.7).

## Feature-flag matrix (honoured)

| `NoVoterRewardDistribution` | Opted-in? | AutoDeposit configured? | Behavior |
|---|---|---|---|
| true (pre-fork) | any | any | Legacy `grantToAccount(delegate.RewardAddr, full)` |
| false | false | any | Legacy (delegate opted out) |
| false | true | empty | Split runs; every voter → credit |
| false | true | non-empty | Split runs; per-voter bridge decides route |

## Consensus fallback

Per [feedback-consensus-fallback-vs-halt], malformed on-chain data (bridge
RPC error, bucket read error, ineligible bucket) downgrades the affected
voter to credit rather than halting the block — pure function of on-chain
state ⇒ same fallback on every validator ⇒ no fork. Wiring bugs (nil
staking protocol, log-encoder failure) still hard-fail.

## New

- `action/protocol/rewarding/voter_reward.go` — `distributeVoterReward` +
  `splitCommission` basis-points helper + `resolveAutoDepositReader`.
- `action/protocol/rewarding/voter_reward_test.go` — 14 tests covering
  `splitCommission` edge cases, fork gate, nil-input guards,
  missing-snapshot fallback, bridge-nil default, Option wiring.
- `action/protocol/staking/add_deposit_compound.go` — cross-protocol seam.
  Package-exported entry point for the rewarding-side compound path;
  skips the `handleDepositToStake` action-plumbing checks that PR 3' has
  already enforced. Does NOT emit a staking receipt log — the batched
  `DelegateDistributed` log is the single source of truth per delegate.

## Modified

- `action/protocol/rewarding/reward.go` — one call site + one branch
  inside `GrantEpochReward`'s per-delegate loop. `splitEpochReward` now
  also returns the filtered candidate list so callers see the same
  index domain as `addrs`/`amounts`.
- `action/protocol/rewarding/protocol.go` — `Option` type +
  `WithAutoDepositBridge` / `WithAutoDepositReader`. `NewProtocol`
  accepts `opts ...Option`; 11 existing call sites remain unchanged.
- `chainservice/builder.go` — construct `autodeposit.Bridge` from
  `Blockchain.AutoDepositContractAddress` at
  `registerRewardingProtocol`; empty string ⇒ nil bridge (compound
  routing inactive).

## Stack (four unmerged dependencies)

- iip-59/pr4.5-delegateprofile-bridge (#4912)
- iip-59/pr2-snapshot-writer (#4915)
- iip-59/pr4.6-autodeposit-bridge (#4922)
- iip-59/pr4.7-delegatedistributed-log (#4923) — this PR's merge base

`d7fe741f7` brings PR 2' into scope via a merge commit (no history rewrite).
Reviewers should focus on commits above the merge.

## Out of scope

- Block-reward folding + orphan drain — that lives in PR 4'.
- Voter-weight source population — snapshot `Entries` is still empty per
  PR 2' skeleton; the downstream degenerate branch pays the full amount
  as commission until the weight-source follow-up lands. Once weights
  are populated, distributeVoterReward already does the correct thing.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./action/protocol/rewarding/... ./chainservice/...`
- [x] `go test ./action/protocol/rewarding/... -count=1` (91 passed)
- [x] `go test ./action/protocol/staking/... -count=1` (385 passed)
- [x] `go test ./blockchain/integrity/... -count=1` (16 passed)
- [ ] End-to-end integration test with populated voter snapshot
      (deferred to PR 5; requires the weight-source follow-up).

Refs iotexproject/iips#74


envestcc:

PR #4928, opened 2026-07-15
feat(rewarding,evm): direct-slot AutoDeposit bucket reader (IIP-59 mitigation 2)

envestcc:
## Summary

Fixes the 5s block-budget breach identified by #4927 (perf bench). Replaces the `evm.SimulateExecution`-per-voter path in `distributeCombinedReward` with a direct storage-slot reader that amortises `StateDBAdapter` construction across the whole drain. Byte-for-byte identical semantics — no consensus change, no fork gate needed.

**Base**: #4927 (perf bench). Merge that first.

## Motivation (from `docs/iip-59-perf-report.md`)

At mainnet scale (107,200 opted-in voters/epoch):

- **Baseline** `evm.SimulateExecution`: 65μs × 107,200 = **7.13s inside a 5s block → jams the chain on the first opted-in epoch**.
- **This PR** (adapter reuse via direct SLOAD): 2.5μs × 107,200 = **270ms (5.4% of block budget)** — a ~26× speedup.

Coupling risk is bounded: `AutoDeposit is Pausable, Ownable`. No proxy, no `upgradeTo` — storage layout is frozen for the lifetime of the deployment. Slot constants (`buckets=1`, `registrants=2`) verified empirically against mainnet runtime bytecode; guarded by a sanity test (see below).

## Design — three-layer split

**1. `evm.SlotReader`** (`action/protocol/execution/evm/evm.go`)
Read-only handle over the working set. Encapsulates the private `prepareStateDBAdapter` + Erigon-dryrun wrap (same setup as `ReadContractStorage`) so downstream callers never touch go-ethereum types. Single new exported surface on `evm`:

```go
func NewSlotReader(ctx context.Context, sm protocol.StateManager) (*SlotReader, error)
func (r *SlotReader) GetSlot(contract address.Address, key []byte) []byte
```

**2. `autodeposit.SlotBucketReader`** (`autodeposit/slot_reader.go`)
Consumes a `SlotReader`-shaped dependency (interface autodeposit defines, `*evm.SlotReader` satisfies). Computes `keccak256(pad32(voter) || pad32(slotNumber))` per Solidity's mapping rule, reads `registrants[voter]` then `buckets[voter]`, returns `(bucketID, present, err)`. Malformed on-chain values (registered voter with `bucket ≤ 0` or `!IsUint64`) degrade silently to `(0, false, nil)` per the module's fallback-vs-halt rule — same semantics as the retired `Bridge.LookupBucket`.

**3. Rewire `distributeCombinedReward`** (`voter_reward.go`)
Build one `SlotReader` per drain, feed it into `Bridge.NewSlotBucketReader`, pass the `BucketReader` through `allocateAndRouteVoters`. The per-voter loop calls `bucketReader.LookupBucket(voter)` instead of `SimulateExecution` — the only behavioural difference is throughput. Deleted the retired path: `autoDepositContractReader`, `ContractReader` interface, `Bridge.LookupBucket`, `Bridge.abi`, `WithAutoDepositReader`, `abi.go`.

## Sanity guard

`action/protocol/execution/autodeposit_slot_reader_sanity_test.go` (untagged, always runs). Deploys mainnet AutoDeposit runtime bytecode, seeds voters 1 and 2 with matching bucket ids via real `register(int256)` txs, then asserts `SlotBucketReader.LookupBucket(voter)` returns the exact ids the contract wrote. Also verifies an unregistered voter falls through to `(0, false, nil)`.

A failure means either the slot constants have drifted (post-mainnet-upgrade the contract changed) or `evm.NewSlotReader`'s setup no longer matches the EVM's working set — either would silently corrupt production drain routing without this test.

## Files

**New**: `execution/evm/evm.go` (+SlotReader), `autodeposit/slot_reader.go`, `autodeposit/slot_reader_test.go`, `execution/autodeposit_slot_reader_sanity_test.go`

**Modified**: `autodeposit/bridge.go` (stripped to `{contract}` + factory), `autodeposit/bridge_test.go` (kept only `TestNew`), `rewarding/protocol.go` (`WithAutoDepositReader` → `WithAutoDepositBucketReader`), `rewarding/voter_reward.go` (call-site rewire), `rewarding/voter_reward_test.go` (fake BucketReader seam)

**Deleted**: `autodeposit/abi.go` (no more runtime ABI parse), `~200 LOC` of `TestLookupBucket_*` in `bridge_test.go`

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean (with and without `-tags=iip59bench`)
- [x] `go test ./action/protocol/rewarding/...` — 102 tests pass
- [x] `go test -run TestSlotBucketReader_MatchesContract ./action/protocol/execution/...` — sanity guard passes on real bytecode
- [ ] Reviewer verification: `go test -tags=iip59bench -bench=BenchmarkAutoDeposit_bucket_AdapterReuse -count=3 ./action/protocol/execution/` reproduces ~2.5μs/op


envestcc:
Superseded by #4953 (consolidated IIP-59 stack). The slot-reader mitigation commit `fde3599a → e3d851fc` was cherry-picked onto `iip-59/consolidated-pr-5-through-5.5b` (conflicts against 5.5b's windowed payout resolved: caller passes the new `bucketReader` alongside `startVoter/endVoter`; early-return arity fixed to match `distributeVoterOnly`'s 6-value contract). Please move review to #4953.

PR #4938, opened 2026-07-16
fix(api): populate block-level gasUsed from fetched receipts (#4937)

### LINKED ISSUE #4937: eth_getBlockByNumber returns gasUsed: 0x0 for most blocks (≠ sum of receipt gasUsed)
### Summary

`eth_getBlockByNumber` (and `eth_getBlockByHash`) returns `gasUsed: "0x0"` at the **block level** for essentially all blocks past a short recent window, even when the block's transactions clearly burned gas. Per the Ethereum JSON-RPC spec, a block's `gasUsed` is the total gas used by all transactions in the block, i.e. it should equal the sum of the per-transaction `eth_getTransactionReceipt.gasUsed`. It does not.

The per-transaction receipts themselves are correct, and `hash` / `stateRoot` / `logsBloom` are correct — only the block-level `gasUsed` aggregate is wrong. This is a JSON-RPC correctness issue, not a consensus/state issue (the value is not committed by the block hash and does not affect balances or receipts). It does break consumers that read `block.gasUsed` — block explorers, gas/utilization analytics, and generic Ethereum tooling that assumes spec compliance.

### Environment

- `web3_clientVersion`: `v2.4.4/go version go1.24.13 linux/amd64`
- Public mainnet endpoints, observed 2026-07-15:
  - Babel: `https://babel-api.mainnet.iotex.io`
  - Archive: `https://archive-mainnet.iotex.io`

### Two distinct behaviors

- **Archive nodes: always `0x0`** — block-level `gasUsed` is never populated, even for recent blocks.
- **Babel (non-archive): correct only for a very recent window, then decays to `0x0`** — a just-processed block reports the right value; once it is no longer recent, the same block reports `0x0`. This strongly suggests `gasUsed` is served from a non-persisted in-memory aggregate rather than recomputed from receipts on read, so historical reads (and archive nodes, which serve from storage) return `0x0`.

### Reproduction

**A recent block — Babel correct, Archive wrong (tier inconsistency):**

```bash
# block 50259305 (0x2fee569), 2 txs, receipts sum to 0x2f74d
curl -s -X POST https://babel-api.mainnet.iotex.io -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x2fee569",false]}' \
  | python3 -c 'import sys,json;print("babel   block.gasUsed =", json.load(sys.stdin)["result"]["gasUsed"])'
curl -s -X POST https://archive-mainnet.iotex.io -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x2fee569",false]}' \
  | python3 -c 'import sys,json;print("archive block.gasUsed =", json.load(sys.stdin)["result"]["gasUsed"])'
# authoritative sum from the two receipts:
curl -s -X POST https://babel-api.mainnet.iotex.io -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"eth_getTransactionReceipt","params":["0x01ec0a1e8697f249eb1a732c7210961ab97d3114b83c2ba91ef7f81c384990e1"]}' \
  | python3 -c 'import sys,json;print("receipt[0].gasUsed =", json.load(sys.stdin)["result"]["gasUsed"])'
```

Observed:
```
babel   block.gasUsed = 0x2f74d   # correct (== Σ receipt gasUsed) while recent
archive block.gasUsed = 0x0       # wrong
```

**An older block — both wrong:**

```bash
# block 40000000 (0x2625a00), 2 txs, receipts sum to 0x328f5
curl -s -X POST https://babel-api.mainnet.iotex.io -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x2625a00",false]}' \
  | python3 -c 'import sys,json;print("babel   block.gasUsed =", json.load(sys.stdin)["result"]["gasUsed"])'
curl -s -X POST https://archive-mainnet.iotex.io -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x2625a00",false]}' \
  | python3 -c 'import sys,json;print("archive block.gasUsed =", json.load(sys.stdin)["result"]["gasUsed"])'
```

Observed:
```
babel   block.gasUsed = 0x0   # wrong (Σ receipt gasUsed = 0x328f5)
archive block.gasUsed = 0x0   # wrong
```

### Expected

`block.gasUsed == Σ eth_getTransactionReceipt(tx).gasUsed` over all transactions in the block, for every block, on every node type (archive and non-archive alike), independent of how recent the block is.

### Notes / scope

- **Not** a consensus or state issue: block hash, `stateRoot`, per-tx receipts (`gasUsed`/`status`/`logs`) and block `logsBloom` are all correct and consistent across Babel and Archive. Only the synthesized block-level `gasUsed` aggregate is affected.
- Likely fix direction: populate `block.gasUsed` from the block's receipts when assembling the eth-API block response (so it is correct on read regardless of recency / node type), rather than from a transient in-memory value.

envestcc:
## Problem

Fixes #4937.

`eth_getBlockByNumber` / `eth_getBlockByHash` return `gasUsed: "0x0"` at the **block level** for essentially all blocks past a short recent window, even though the block's transactions clearly burned gas. Per the Ethereum JSON-RPC spec, a block's `gasUsed` should equal the sum of the per-transaction receipt `gasUsed`.

Observed behavior (from the issue):
- **Archive nodes: always `0x0`** — they always serve blocks from storage.
- **Babel (non-archive): correct only for a very recent window, then decays to `0x0`.**

## Root cause

`getBlockResult.MarshalJSON` (`api/web3server_marshal.go`) summed block-level `gasUsed` from `obj.blk.Receipts`:

```go
for _, r := range obj.blk.Receipts {
    gasUsed += r.GasConsumed
}
```

But blocks loaded from storage via the DAO do **not** carry their receipts on the `block.Block` object — receipts are persisted and read separately (`dao.GetReceipts`). So `blk.Receipts` is empty for any storage-served block, yielding `0x0`. This exactly explains the two behaviors: archive nodes always serve from storage; non-archive nodes are correct only while the freshly-committed block is still in the DAO block LRU cache (receipts populated in memory), then decay to `0x0` once evicted and reloaded.

The correct receipts are **already fetched** on this path — `BlockByHeight` / `BlockByHash` call `dao.GetReceipts` unconditionally and pass them into `getBlockWithTransactions` — but they were dropped instead of stored on the result.

## Fix

- Carry the explicitly-fetched `receipts` on `getBlockResult` and sum `gasUsed` from them.
- Fall back to `obj.blk.Receipts` only when no receipts were supplied, which preserves the websocket `newHeads` streaming path (`blocklistener.go`) that uses an in-memory block with receipts already attached.

This adds **no extra I/O** — it reuses receipts that were already read from storage on every block query; previously they were read and then discarded for the aggregate.

### Note vs go-ethereum

Unlike go-ethereum, IoTeX's block header does not carry `gasUsed` (it is not committed by the block hash), so `internal/ethapi` reading `head.GasUsed` directly is not an option here — the block-level aggregate must be recomputed from receipts on read. Since receipts are already loaded on this path, there is no performance regression.

## Test

Added `TestBlockObjectMarshal/BlockGasUsedFromFetchedReceipts`, which simulates a storage-loaded block (`blk.Receipts = nil`) with receipts supplied separately and asserts `gasUsed == 0x5208`. Verified this test reproduces the bug (`0x0`) without the fix and passes with it.

### DIFF
--- a/api/web3server_marshal.go
+++ b/api/web3server_marshal.go
@@ -48,6 +48,7 @@ type (
 
 	getBlockResult struct {
 		blk          *block.Block
+		receipts     []*action.Receipt
 		transactions []interface{}
 	}
 
@@ -191,7 +192,15 @@ func (obj *getBlockResult) MarshalJSON() ([]byte, error) {
 	for _, tx := range obj.blk.Actions {
 		gasLimit += tx.Gas()
 	}
-	for _, r := range obj.blk.Receipts {
+	// Prefer the receipts explicitly fetched from storage so that the block-level
+	// gasUsed aggregate is correct on read regardless of recency or node type. The
+	// in-memory block loaded from the DAO does not carry its receipts, so falling
+	// back to obj.blk.Receipts would yield 0x0 for historical/archive reads.
+	receipts := obj.receipts
+	if len(receipts) == 0 {
+		receipts = obj.blk.Receipts
+	}
+	for _, r := range receipts {
 		gasUsed += r.GasConsumed
 	}
 	if logsBloom := obj.blk.Header.LogsBloomfilter(); logsBloom != nil {
--- a/api/web3server_marshal_test.go
+++ b/api/web3server_marshal_test.go
@@ -227,6 +227,32 @@ func TestBlockObjectMarshal(t *testing.T) {
 		 }
 		`, string(res))
 	})
+
+	t.Run("BlockGasUsedFromFetchedReceipts", func(t *testing.T) {
+		// Blocks loaded from storage (historical reads / archive nodes) do not carry
+		// their receipts on the block object; the receipts are fetched separately and
+		// supplied via the receipts field. The block-level gasUsed must still be the
+		// sum of those receipts, not 0x0. See iotexproject/iotex-core#4937.
+		storageBlk := blk
+		storageBlk.Receipts = nil
+		res, err := json.Marshal(&getBlockResult{
+			blk: &storageBlk,
+			receipts: []*action.Receipt{{
+				Status:      1,
+				BlockHeight: 1,
+				ActionHash:  txHash,
+				GasConsumed: 21000,
+				TxIndex:     1,
+			}},
+			transactions: []interface{}{string("0x2133ee7ff4562535166e3f16fd7407c19e5ed1acd036f78d3528a5a40e40ad42")},
+		})
+		require.NoError(err)
+		var obj struct {
+			GasUsed string `json:"gasUsed"`
+		}
+		require.NoError(json.Unmarshal(res, &obj))
+		require.Equal("0x5208", obj.GasUsed)
+	})
 }
 
 func TestTransactionObjectMarshal(t *testing.T) {
--- a/api/web3server_utils.go
+++ b/api/web3server_utils.go
@@ -120,6 +120,7 @@ func (svr *web3Handler) getBlockWithTransactions(blk *block.Block, receipts []*a
 	}
 	return &getBlockResult{
 		blk:          blk,
+		receipts:     receipts,
 		transactions: transactions,
 	}, nil
 }


guo (approved):
Reviewed the rebased head. The fetched-receipt aggregation is correctly wired for stored blocks, focused tests and CI pass, and the cold-cache devnet check matches block gasUsed to the receipt sum.

Issue #4942, opened 2026-07-16
web3: support `finalized`/`safe` block tags in eth_getBlockByNumber (+ minor EVM-RPC compat gaps)


guo:
## Summary

IoTeX's EVM JSON-RPC (`api/web3server.go`) does not recognize the standard `finalized` / `safe` block tags, plus a few smaller EVM-compatibility gaps in the same area. The primary ask is `finalized`/`safe`; the rest are bundled because they live in the same handler.

Tested against the public endpoint `https://babel-api.mainnet.iotex.io` (node `v2.4.4`).

## 1. `finalized` / `safe` block tags (primary)

`eth_getBlockByNumber` with `finalized` or `safe` fails today:

```
$ curl -s $RPC -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["finalized",false]}'
{"code":-32603,"message":"strconv.ParseUint: parsing \"finalized\": invalid syntax"}
$ curl -s $RPC -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["safe",false]}'
{"code":-32603,"message":"strconv.ParseUint: parsing \"safe\": invalid syntax"}
```

`parseBlockNumber` only recognizes the three tags defined at `api/web3server.go:93-95` (`pending` / `latest` / `earliest`); any other string falls through to `strconv.ParseUint` and errors.

These tags are standard on Ethereum (since The Merge, 2022), BSC (BEP-126 fast finality), and Polygon PoS (Heimdall v2 milestones). IoTeX is the outlier among modern EVM chains in not exposing them — even though it is the easiest case: Roll-DPoS gives instant/deterministic finality, so both `safe` and `finalized` can map directly to the latest committed (irreversible) block.

Motivation: cross-chain bridges, indexers, and wallets increasingly gate on the `finalized` tag for reorg-safe confirmation logic. Against IoTeX today those tools get a hard error and must special-case the chain.

**Proposed fix:** recognize `safe` and `finalized` in `parseBlockNumber` (next to the constants at `api/web3server.go:93-95`) and resolve both to the current tip / last irreversible block.

## 2. `eth_feeHistory` parses `blockCount` as decimal (bug)

```
$ curl -s $RPC -d '{"jsonrpc":"2.0","id":1,"method":"eth_feeHistory","params":["0x5","latest",[25,50,75]]}'
[...]

PR #4943, opened 2026-07-16
feat(api): support finalized/safe block tags & fix eth_feeHistory blockCount parsing

envestcc:
## Summary

Addresses the two concrete items in #4942 for IoTeX's EVM JSON-RPC (`api/web3server.go`):

1. **`finalized` / `safe` block tags (primary).** `eth_getBlockByNumber` with `finalized`/`safe` previously failed with `strconv.ParseUint: parsing "finalized": invalid syntax`. The string-based `parseBlockNumber` only recognized `pending`/`latest`/`earliest`. It now also resolves `safe` and `finalized`. Roll-DPoS gives deterministic instant finality, so the committed tip is already irreversible and both tags map to `TipHeight()` — consistent with the existing `rpc.BlockNumberOrHash` path (`blockNumberOrHashToHeight`), which already handled these tags.

2. **`eth_feeHistory` `blockCount` hex bug.** `blockCount` was parsed with base-10 `strconv.ParseUint`, so spec-compliant callers sending a hex quantity (`"0x5"`, e.g. MetaMask EIP-1559 fee estimation) got `strconv.ParseUint: parsing "0x5": invalid syntax`. It is now parsed as a hex quantity via `hexStringToNumber`. Backward compatible (`0x4 == 4`, existing test unaffected).

## Out of scope

The optional parity items in #4942 — `eth_getProof` (EIP-1186), `eth_createAccessList` (EIP-2930), and the `txpool_*` namespace — are **not** included here. They require new state-trie / txpool plumbing that doesn't exist today and warrant separate PRs. This PR keeps to the concrete request + the small bug fix.

## Tests

- `TestParseBlockNumber`: added `safe` and `finalized` cases.
- `eth_feeHistory` integrity test: added a hex-string `blockCount` (`"0x4"`) case.
- `go build ./api/...` and the web3 test suites pass.

Refs #4942

### DIFF
--- a/api/web3server.go
+++ b/api/web3server.go
@@ -90,9 +90,11 @@ var (
 	errHTTPNotSupported  = errors.New("http not supported")
 	errPanic             = errors.New("panic")
 
-	_pendingBlockNumber  = "pending"
-	_latestBlockNumber   = "latest"
-	_earliestBlockNumber = "earliest"
+	_pendingBlockNumber   = "pending"
+	_latestBlockNumber    = "latest"
+	_earliestBlockNumber  = "earliest"
+	_safeBlockNumber      = "safe"
+	_finalizedBlockNumber = "finalized"
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


guo (approved):
Reviewed at exact head 06fe54dc0762174581e1c17ac5e0ffe5eebea77f. CI and focused API tests pass. An isolated A/B run against a real mainnet fork at height 50,000,000 reproduced the pre-fix parsing failures and verified safe/finalized handling plus hex/decimal eth_feeHistory parity. No blocking findings.

PR #4944, opened 2026-07-16
feat(rewarding): IIP-59 chunked drain — dedicated action + e2e bench + perf report

envestcc:
## Summary

Consolidated PR covering the IIP-59 chunked-drain machinery end-to-end.
Absorbs what were previously four stacked PRs (#4945, #4946, #4947, #4948)
plus the C2 multi-block fix:

| Sub-scope | Commit(s) |
|---|---|
| **PR C2**: chunk `GrantEpochReward` across blocks via `EpochDrainCursor` | `e2eb83056` |
| **C2 fix**: allow the drain to span epoch boundaries (era-scoped cursor) | `65a368dd0` |
| **PR C4**: e2e integration bench `TestIIP59EpochGrantPerf` (tier: small / medium / mainnet) | `1232720fa` |
| **PR C2.1**: dedicated `GrantVoterRewardChunk` system action + Phase-A-only `GrantEpochReward` + pre-fork legacy path split | `d85f9d4a0` + `5a06eb106` + `466617b9a` |
| **PR C3** (#4948): block-time commission split in `GrantBlockReward`; collapse `GrantEpochReward` into a single body; cursor field rename `PoolAmountFrozen` → `VoterAmountFrozen`; `decrementPendingBlockRewardPool` helper; delete `delegateAllocation` / `prepareDelegateAllocation` / `runPhaseADelegateGrants` / `grantLegacyEpochReward` | `8441d62bd` (merge of `303f977ae`) |
| **Perf report**: `docs/iip-59-perf-report.md` with real 2.5s-block-budget numbers | (folded) |

## Semantic model after this PR

Three system actions, three purposes:

| Action | When emitted | What runs |
|---|---|---|
| `BlockReward` | every block | for opt-in delegates, split `blockReward` by frozen `snap.BlockCommissionBasisPoints` — commission paid to `rewardAddr` immediately (visible on the `BLOCK_REWARD` log); voter share folded into per-delegate pending pool |
| `EpochReward` | epoch-last block | slashing → per-delegate epoch split via `splitDelegateEpochReward` (commission grant + `EPOCH_REWARD` log + credit voter share to pool) → foundation bonus → sentinel `updateRewardHistory` → cursor persist iff any delegate has voter share to drain |
| `VoterRewardChunk` | non-epoch-last block, while cursor is live | next chunk of the frozen work list. Coda on the last chunk: orphan sweep + `updateAvailableBalance` + cursor delete. |

[...]

PR #4946, opened 2026-07-16
refactor(rewarding): dedicated GrantVoterRewardChunk system action for IIP-59 drain

envestcc:
## Summary

Stacks on #4945 (C4). Follow-up refactor to #4944 (C2).

C2 landed the era-boundary chunked drain by extending `GrantEpochReward` with cursor-demux continuation: on every non-epoch block a live cursor caused `CreatePostSystemActions` to emit an extra `GrantReward{EpochReward}` that skipped Phase A and jumped straight into the drain loop. Semantically that meant *an action literally named epoch reward* ran on non-epoch blocks. That mismatch:

- misled reviewers into believing \`EpochReward\` only runs at epoch boundaries,
- was the direct cause of the C2 continuation-guard bug (\`!=\` vs \`>\`) fixed in 65a368dd0, and
- put a landmine under any future change to \`GrantEpochReward\`'s pre-Phase-A checks.

Promote the drain continuation to its own system action. Post-refactor:

| Action              | When emitted                          | Runs                                |
|---------------------|---------------------------------------|-------------------------------------|
| \`BlockReward\`       | every block                           | fold block reward into pool         |
| \`EpochReward\`       | epoch-last block                      | Phase A (freeze) + first chunk      |
| \`VoterRewardChunk\`  | non-epoch-last, cursor live           | next chunk (last chunk runs coda)   |

### Changes

- **iotex-proto v0.6.8** (iotexproject/iotex-proto#175, merged): \`RewardType.VoterRewardChunk = 2\` enum extension. \`go.mod\` bumped.
- **\`action.GrantReward\`**: reuses the polymorphic envelope; adds \`VoterRewardChunk = 2\` const and \`Proto\`/\`LoadProto\` switch arms. Wire format and receipt handling unchanged beyond the one enum value.
- **\`rewarding.Protocol\`**: monolithic \`GrantEpochReward\` split into:
  - \`GrantEpochReward\` — Phase A only. ANY cursor at entry is now unambiguous corrupt state and fails loud (semantics tightened from \"future-epoch cursors are corrupt\" to \"any live cursor is corrupt\").
  - \`GrantVoterRewardChunk\` — reads cursor, loads epoch-scoped state from **\`cursor.TargetEra\`** (not the block's epoch), delegates to the shared runner. This is the fix that the C2 \`>\` guard patched, now handled naturally because the handler never reads \`rp.GetEpochNum(blockHeight)\`.
  - \`loadEpochDistributionInputs\` — pure helper; deterministic re-load of admin / exempt / candidates / \`splitEpochReward\` output.
  - \`runVoterDistributionChunk\` — shared Phase B loop + Phase C coda. Called by both entry points.
- **\`CreatePostSystemActions\`**: emits \`VoterRewardChunk\` on continuation.
- **\`Validate\`**: rejects \`VoterRewardChunk\` when \`NoVoterRewardDistribution\` is true. Defense in depth — normal execution can't reach this pre-fork.
- **\`Handle\`**: dispatches \`VoterRewardChunk\` with the same settle path as \`EpochReward\`.

Compound routing stays inside \`runVoterDistributionChunk\` per prior design agreement — splitting it into a third action would force either a redundant O(voter) walk or a second cursor state machine.

### Test migration

- \`TestGrantEpochReward_RejectsStaleCursor\` → \`RejectsAnyLiveCursor\`. Phase A now rejects *any* live cursor, not just future-epoch cursors.
- New \`voter_reward_chunk_test.go\`:
  - \`HappyPath\` — mid-drain chunk advances DelegateIndex by chunkSize; cursor persists; sentinel not written.
  - \`LastChunkRunsCoda\` — terminal chunk deletes cursor and writes the epoch reward sentinel.
  - \`CrossEraContinuation\` — cursor with \`TargetEra=1\` completes cleanly on a block many epochs later (the scenario the C2 \`>\` guard patched).
  - \`MissingCursorErrors\` — dispatcher invariant: no live cursor → loud error.
  - \`PreForkRejects\` — \`Validate\` rejects + handler defense-in-depth rejects.

E2E bench (\`e2etest/iip59_perf_test.go\`) needs no changes — it observes the cursor via \`TestOnlyEpochDrainSnapshot\` and mints blocks in a loop; which action-type carries the continuation is invisible to it.

## Test plan

- [x] \`go build ./... && go vet ./...\` clean
- [x] \`go test ./action/... ./action/protocol/rewarding/...\` — 1188 passed
- [x] Small-tier \`go test -run TestIIP59EpochGrantPerf -tags e2e ./e2etest/\` — passed
- [ ] Medium/mainnet e2e tier (deferred; unchanged perf profile since this is a refactor)

## Deliberately out of scope

- Cursor payload expansion (self-contained \`RewardAddress\`/\`EpochAmount\`) — follow-up optimization, not needed for the semantic-clarity goal here.
- Splitting compound routing into its own action.
- Renaming or resemanticating \`EpochReward\`.
- Any change to \`GrantBlockReward\`.

### DIFF
--- a/action/grantreward.go
+++ b/action/grantreward.go
@@ -25,6 +25,10 @@ const (
 	BlockReward = iota
 	// EpochReward indicates that the action is to grant epoch reward
 	EpochReward
+	// VoterRewardChunk indicates that the action is to advance one chunk
+	// of an in-progress IIP-59 era-boundary voter-reward drain. Emitted
+	// only on non-epoch-boundary blocks while a drain cursor is live.
+	VoterRewardChunk
 
 	_grantrewardInterfaceABI = `[
 		{
@@ -99,6 +103,8 @@ func (g *GrantReward) Proto() *iotextypes.GrantReward {
 		gProto.Type = iotextypes.RewardType_BlockReward
 	case EpochReward:
 		gProto.Type = iotextypes.RewardType_EpochReward
+	case VoterRewardChunk:
+		gProto.Type = iotextypes.RewardType_VoterRewardChunk
 	}
 	return &gProto
 }
@@ -113,6 +119,8 @@ func (g *GrantReward) LoadProto(gProto *iotextypes.GrantReward) error {
 		g.rewardType = BlockReward
 	case iotextypes.RewardType_EpochReward:
 		g.rewardType = EpochReward
+	case iotextypes.RewardType_VoterRewardChunk:
+		g.rewardType = VoterRewardChunk
 	}
 	return nil
 }
--- a/action/protocol/rewarding/chunked_drain_test.go
+++ b/action/protocol/rewarding/chunked_drain_test.go
@@ -83,36 +83,38 @@ func TestBuildEpochDrainCursor_FreezesPoolBalances(t *testing.T) {
 		"cursor value must decouple from post-freeze pool credits")
 }
 
-// TestGrantEpochReward_RejectsStaleCursor confirms the corruption guard.
-// A cursor pinned to a FUTURE epoch is corrupt state — Phase A can only
-// pin to the current epoch. GrantEpochReward must reject this loud
-// instead of silently proceeding. (A cursor pinned to a PRIOR epoch,
-// by contrast, is a legitimate multi-block drain still in progress.)
-func TestGrantEpochReward_RejectsStaleCursor(t *testing.T) {
+// TestGrantEpochReward_RejectsAnyLiveCursor confirms Phase A's tightened
+// corruption guard. After C2.1 split the continuation path into
+// GrantVoterRewardChunk, GrantEpochReward runs Phase A only — and Phase A
+// can never coexist with a live cursor. Any cursor at Phase A entry is
+// unambiguous corrupt state (mid-drain continuation was supposed to run
+// GrantVoterRewardChunk, not GrantEpochReward), so the guard must fire
+// regardless of which era the cursor pins.
+func TestGrantEpochReward_RejectsAnyLiveCursor(t *testing.T) {
 	testProtocol(t, func(t *testing.T, ctx context.Context, sm protocol.StateManager, p *Protocol) {
 		r := require.New(t)
-		// Flip the fork gate on so cursorEnabled=true and the stale-cursor
-		// check runs. Height is already the last block of epoch 1, which is
-		// past ToBeEnabledBlockHeight=1.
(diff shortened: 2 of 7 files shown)


envestcc:
Reviewer flagged that `GrantEpochReward` still ran the first distribution chunk inline via the shared `runVoterDistributionChunk`, which contradicted the whole "distribution belongs in `VoterRewardChunk`" premise of C2.1. Pushed 5a06eb106 to finish the split:

- Post-fork `GrantEpochReward` now runs slashing + freeze + persist cursor, then returns. Receipt carries only Phase A output.
- All distribution (Phase B chunks + Phase C coda: orphan drain, foundation bonus, sentinel, cursor delete) lives entirely in `GrantVoterRewardChunk` on subsequent non-boundary blocks.
- Pre-fork legacy path (`NoVoterRewardDistribution=true`) unchanged — still routes through `runVoterDistributionChunk` with `chunkSize=0`.

Trade-off: +1 block of drain wall-clock. Small tier bench confirms `drain_blocks` 2 → 3; mainnet will grow 6 → 7. At 2.5s/block against ~8,640-block era window, negligible.

New test `TestGrantEpochReward_PostForkOnlyPhaseA` locks the invariant: post-fork Phase A must persist a cursor and must NOT write the sentinel. All 64 rewarding-package tests + e2e small tier still green.

envestcc:
Follow-up amend `466617b9a` per review discussion:

**Route pre-fork legacy path around the chunk machinery entirely.**
Previously `GrantEpochReward` on pre-fork blocks called
`runVoterDistributionChunk(chunkSize=0)`, relying on
`distributeCombinedReward`'s internal fork gate to fall through to legacy
`grantToAccount`. Correct semantically, but the function name suggested
IIP-59 logic was running on pre-fork chains — misleading and a landmine
for future edits.

**Now split cleanly:**

- Pre-fork → new `grantLegacyEpochReward` (no cursor / pool / compound
  bridge touched, only the classic single-block grant loop + foundation
  bonus + sentinel).
- Post-fork → Phase A only (slashing + freeze cursor + persist), coda
  deferred entirely to `GrantVoterRewardChunk`.
- `runVoterDistributionChunk` becomes post-fork-only; dropped its
  `cursorEnabled` guard and the fork-off branch of the `chunkSize`
  comment.

Behavior unchanged on both sides; verified with `./action/protocol/rewarding/...`
unit suite + small-tier e2e drain bench.

PR #4953, opened 2026-07-20
feat(rewarding,staking): IIP-59 on-chain voter reward distribution

envestcc:
## Summary

Protocol-native, on-chain implementation of IIP-59 voter reward
distribution. Delegate commission is paid at block time; per-delegate
voter payouts run once per era (24 epochs, ≈ 1 day) and are chunked
across multiple blocks so a single 30k-bucket delegate cannot exceed
the 2.5s Dardanelles block budget.

Fully gated behind `NoVoterRewardDistribution` (bound to
`!g.IsToBeEnabled(height)`) — until the fork height is set in a
follow-up PR, every path in this change is a no-op and the chain
stays on the legacy Hermes distribution.

## Motivation

Voter rewards are currently distributed off-chain via Hermes. This
change moves the accounting on-chain so:

- delegate/voter splits are enforced by the protocol, not by a trusted
  operator;
- per-voter payouts are deterministic and independently verifiable
  from block state;
- the fund cannot silently drift from the ledger.

The design fits four constraints from the IIP-59 proposal
(`iip_proposals/iip-59.md`): (1) rates come from an on-chain
`DelegateProfile` contract, (2) compound routing comes from an
on-chain `AutoDeposit` contract, (3) opt-in is per-delegate, (4)
distribution work must fit inside the 2.5s block budget at 30k
buckets.

### Dual-stream commission, era-based voter payout

- **Block reward** — commission portion paid to the delegate at block
  time via `GrantBlockReward`; voter portion credited to
  `PendingBlockRewardPool[candidateID]`.
- **Epoch reward** — Phase A pays epoch commission + credits epoch
  voter portion to the same per-delegate pool. Runs every epoch.
- **Voter payout** — at the era boundary
  (`epochNum % EpochsPerRewardEra == 0`), `FreezePollSnapshot`
  writes per-voter weights into `CandidatePollSnapshot.Entries`; a
  cursor kicks off, and subsequent blocks drain the accumulated pool
  across all opted-in delegates until the era's frozen amounts are
  fully paid.

### Chunked drain with two caps

`runVoterDistributionChunk` iterates delegates in a deterministic order
and is bounded by two independent genesis-configured caps:

- `CompoundBatchSize` — max delegates processed per block. Prevents
  a chain with many small delegates from spending its whole block on
  per-delegate overhead.
- `VoterBudgetPerBlock` — max voters paid per block. Lets one large
  delegate stop mid-list and resume in a later block.

Either cap can terminate a chunk; both zero = single-block full drain
(pre-fork behavior).

### `EpochDrainCursor` — mid-delegate resumable

Persisted per era. Two fields: `DelegateIndex` (next delegate) and
`VoterIndex` (next voter within `DelegateIndex`). Deterministic
allocation over the full frozen `snap.Entries` is recomputed every
chunk; only the payout window `[VoterIndex, VoterIndex+budget)`
changes. Splitting one delegate across K blocks produces
byte-identical per-voter amounts to the un-split single-block run.

### `VoterWeightView` — per-voter weight source

Reward distribution needs each voter's total weight against a
delegate. Rather than enumerate all buckets at freeze time (benched
~7 s at the 30k ceiling — ≈ 3× the 2.5 s block budget), a
`VoterWeightView` is maintained incrementally: hooked into 13
mutation sites in the staking handlers, it applies deltas per block
in O(1) amortized. `FreezePollSnapshot` reads a materialized
`snap.Entries` slice from it in O(voters).

The view is rebuilt from all buckets at `CreateBaseView`; its digest
is persisted and validated at boot so any hook drift fails loudly.

### Contract bridges (read-only)

- `rewarding/delegateprofile/` — reads
  `(commissionRateBlock, commissionRateEpoch, optIn)` at
  `PutPollResult`; frozen into the poll snapshot. Per-delegate read
  failures degrade the delegate back to the legacy path rather than
  halting the block (per-item fallback, not deterministic halt).
- `rewarding/autodeposit/` — queried at voter payout to route
  compound-eligible amounts. Valid registered bucket ⇒
  `AddDepositForCompound` (adds to the existing bucket).
  Invalid/unregistered/error ⇒ direct credit to the voter.
  A direct-slot reader path (`SlotBucketReader`) avoids reentrant EVM
  calls in the reward handler.

### Batched `DelegateDistributed` log

One log per delegate per chunk, encoding `Voters[]`, `Amounts[]`,
`Routings[]`, `TotalVoterPool`, `SnapshotHash`. `SnapshotHash` covers
the full frozen entries and is identical across a delegate's chunks.

**Off-chain semantic to be aware of:** `TotalVoterPool` equals the sum
of `Amounts[]` in *this* chunk, not the era-wide frozen amount. Any
off-chain consumer that reassembles distributions must aggregate
partial logs by `(SnapshotHash, delegate, epoch)`. Hermes-patch and
the off-chain verifier are being updated separately for this.

## Genesis knobs

All new; safe defaults for the pre-fork branch.

| Field | Default | Purpose |
|---|---|---|
| `EpochsPerRewardEra` | `24` | Era length in epochs (≈ 1 day) |
| `EpochDrainChunkSize` | `50` | `CompoundBatchSize`: delegates/block |
| `VoterBudgetPerBlock` | `2000` | Voters/block cap |
| `AutoDepositContractAddress` | `""` | Filled with mainnet address in the fork-activation PR |
| `DelegateProfileContractAddress` | `""` | Filled with mainnet address in the fork-activation PR |

## Feature-flag semantics

Bound to `NoVoterRewardDistribution` in `protocol.FeatureCtx`:

- **Pre-fork** (`NoVoterRewardDistribution == true`): `PutPollResult`
  does not freeze the snapshot; `runVoterDistributionChunk` is not
  invoked; `voterBudgetPerBlock(ctx)` and `epochDrainChunkSize(ctx)`
  both return 0. Every code path added here is inert.
- **Post-fork** (`NoVoterRewardDistribution == false`): the snapshot
  freezes at every `PutPollResult`; era boundaries kick off chunked
  drains; both caps read from genesis.

No new named fork; the existing `ToBeEnabled` gate is reused because
IIP-59 activation is planned as a scheduled upgrade rather than a
long-lived hard fork.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./action/protocol/rewarding/... ./action/protocol/staking/...`
- [x] `go test -tags e2e -run 'TestIIP59ChunkedDrainStress_' ./e2etest/`
      (`SmallTier`, `MultiEra`, and `SingleDelegateLargeVoter` all pass)

Coverage highlights:

- **Determinism**: chunked drain at `chunkSize ∈ {1, 2, all}` produces
  byte-identical per-voter amounts and fund invariants.
- **Cross-era continuation**: pool amounts accrued after the drain
  starts land in the next era, not the current one.
- **Mid-delegate resume**: cursor pauses with
  `DelegateIndex=k, VoterIndex>0`, resumes on the next block, ends
  identically to a single-block run.
- **VoterWeightView consistency**: hook-driven mutation across all 13
  handler sites keeps the view in sync with the bucket store;
  digest-mismatch on boot is an error.
- **AutoDeposit slot reader**: sanity test deploys the real mainnet
  runtime bytecode and asserts direct-slot lookups match what the
  contract itself wrote.

## Reviewer sanity checks

- [ ] Run the Design-B bench for the `VoterWeightView` cost
      justification: `go test -tags iip59bench -bench BenchmarkFreezeSnapshot -benchtime=5x ./action/protocol/staking/`.
- [ ] Confirm a boot-time digest-mismatch error fires when a
      `VoterWeightView` hook is intentionally removed.
- [ ] Skim `runVoterDistributionChunk` for the two `break` paths
      (delegate-cap and voter-cap) and confirm the cursor persists in
      both.
- [ ] Confirm `NoVoterRewardDistribution == true` short-circuits every
      new call site (this is the pre-fork guarantee).

## Follow-ups (separate PRs)

- **Mainnet fork activation** (task #17) — fill the two contract
  address genesis fields, set `ToBeEnabledBlockHeight`. Governance
  prerequisites: `DelegateProfile` ownership handover (#43) and
  the hermes-patch service filter for opted-in delegates (#44).
- **Off-chain verifier** (task #45) — recompute `DelegateDistributed`
  from voter snapshots; must aggregate partial chunks by
  `(SnapshotHash, delegate, epoch)`.


envestcc (commented):
## Storage-structure review (self-review addendum)

Went over this from two angles — **storage-structure soundness** and **long-term archive/history retention**. No correctness blockers. Actionable optimization list below. Overall the design matches the IIP-59 spec and existing repo conventions (V1 hashed-key / V2 raw-key dual write, Erigon `Encode/Decode` dual storage, big.Int `.Bytes()` big-endian).

### Correctness checks (pass)
- `staking.proto` `Candidate` adds `voterRewardOnchainOptIn = 11` / `rewardAddressUpdated = 12`: field numbers 11/12/13 in the base belong to `message Bucket`, while `Candidate` previously stopped at `deactivatedAt = 10`, so there is **no field-number reuse**; old records decode with the new bools defaulting to false. Compatible.
- New key prefixes don't collide with existing rewarding keys (`adm/fnd/acc/xpt`).

### Archive-mode bloat assessment
Archive keeps history per (key, change), so growth ≈ write frequency × value size. Two hotspots are the worst-case combo of "large blob + high re-versioning frequency + mostly-unchanged content":
- **`CandidatePollSnapshot`**: a full per-candidate voter snapshot is written every era for each on-chain delegate. 30k voters ≈ 1.5MB/era → ~440MB/year for a single such delegate, and the voter set barely changes era-to-era yet the whole blob is re-versioned. **Largest bloat source.**
- **`EpochDrainCursor`**: a singleton blob holding all delegates' frozen work items, yet it is **rewritten in full on every continuation block** just to advance a few-byte cursor pointer.

---

### Optimization list

**P0 · doable in this PR, low-risk / high-value**

1. **Split `EpochDrainCursor` into a frozen work list vs. a progress pointer.** Write `delegates[]` + metadata once per `targetEra` (immutable); keep `{delegate_index, voter_index, distributed for the active delegate}` in a small separate singleton that re-versions only a few dozen bytes per block. ~1000× reduction in cursor history, no consensus-semantics change.
2. **Clean up snapshots for delegates that leave the poll list.** Currently `FreezePollSnapshot` only overwrites candidates still in the list; a candidate that drops out leaves its (potentially large) `CandidatePollSnapshot` blob lingering in current state forever (no `deleteState`). Delete/clear its snapshot key when a candidate drops out or switches to legacy.

**P1 · suggested in this PR, clear latent issues**

3. **`pbrp` is a prefix of `pbrpx`.** Safe today via exact-match lookups, but a future prefix scan over `pbrp` would wrongly pick up the index key. Add a delimiter or use non-prefix-overlapping keys.
4. **The pool index reuses `rewardingpb.Exempt` as a "list of bytes" container.** Works, but couples the index wire format to a semantically-unrelated message; prefer a dedicated message.

**P2 · follow-up (archive-bloat track)**

5. **Make `CandidatePollSnapshot.entries[]` content-addressed / trie-based.** Two tiers:
   - **5a — whole-blob content addressing by `snapshotHash`**: skip the write when the key already exists (check-before-write). Low risk; pruned nodes need GC/refcounting. Moderate gain (dedups only when the entire set is byte-identical).
   - **5b — order-statistics Merkle tree** (each node carries `subtreeCount` + `subtreeWeightSum`): shares at **node** granularity, so a partial change re-writes only ~log(N) path nodes → best archive gain; the root is incrementally computable and **replaces the hand-rolled `snapshotHash`**; `select/rank/prefixSum` are native, which exactly covers the positional semantics the cursor relies on (`R % count` circular start, dust to `lastWeightedIndex`, logical `voter_index`). Cost: deterministic node hashing across native + Erigon (new consensus surface), GC, audit effort.
   - Key point: a plain address-keyed MPT is **not** enough — the hard part isn't plain iteration, it's **rank/select + prefix sums**. Only an aggregate-carrying Merkle tree gives both archive structural sharing and the positional semantics. With copy-on-write, "freeze the root while the live pool keeps accruing" falls out naturally, unifying today's separate incremental view / snapshotHash / frozen entries.
   - Middle path: store entries as individual leaves in the staking namespace (raw, order-preserving keys) to get node sharing with zero new tree code — but rank/select still needs a separate index, so it doesn't buy the positional semantics.

### Already correct, keep as is
- `VoterWeightView` persists only a 32-byte digest and rebuilds the full view from buckets at boot (verified against the digest) — the right pattern; 5b is essentially bringing `entries` in line with it.
- `VoterRewardDestination`: sparse overrides, reset deletes the record (no tombstone), strict 20-byte validation.

**Sequencing**: do 1 and 2 (and 3, 4 if convenient) in this PR; treat 5 as a separate follow-up — 5a as a safety net first, 5b only if archive growth becomes a real operational concern.

envestcc (commented):
## Liveness risk: `VoterWeightView` divergence is non-self-healing and can halt the network

Flagging this as the highest-severity concern in the change. It is not a "wrong rewards" bug — it is an **availability/liveness** risk: under a specific (deterministic) failure the whole network can become **unable to restart**, with **no self-healing path**.

### The setup: two derivations of the same consensus-committed value

`VoterWeightView` is maintained two independent ways:

1. **Live / incremental** — 13 staking hooks apply per-bucket deltas each block; `Commit` writes `Hash()` of the accumulated view as a **32-byte digest into the staking namespace, i.e. into the state root**, every dirty block.
2. **Full rebuild** — at boot, `buildVoterWeightView` reconstructs the entire view by scanning **all** buckets, then compares its `Hash()` against the persisted digest; on mismatch it **refuses to start** (`"voter weight digest mismatch"`).

So the incrementally-maintained aggregate is consensus-committed, and a from-scratch recomputation must reproduce it byte-for-byte at every restart.

### Why a divergence cannot self-heal

Because the digest is **in the state root**, it is consensus-agreed. If the two derivations ever disagree, neither recovery is safe:

- **fail-closed (current behavior):** the node refuses to boot → it stops.
- **auto-heal to the rebuild:** the node rewrites a state value that the network already committed at that height → its state root diverges → it **forks off** the canonical chain.

There is no third arbiter to decide which derivation is "correct," and the authoritative chain state already encodes one of them. `Apply` even contains an explicit swallow-path for an inconsistent delta, with the comment that *"the view-hash check at restart catches any real divergence and surfaces it loudly"* — i.e. the live path's correctness argument is deliberately deferred to the restart check, which is exactly the check that has no safe outcome on mismatch.

### Why this escalates from one node to the whole network

The failure is **deterministic**: every node runs the same hook code, so every node commits the **same** (possibly wrong) incremental digest, and every node's rebuild produces the **same** divergent value. This is not a per-node corruption that peers can out-vote — it is uniform across the fleet. Consequently:

- A coordinated restart (a routine binary upgrade) or a wave of crash-recoveries would cause **simultaneous boot failure across validators** → block production stops.
- The divergence is **temporally decoupled** from its cause: a bad digest can be committed at height H and lie dormant until the next restart, which may be days later, so the outage appears unrelated to whatever change introduced it.
- Recovery is not automatic: it requires a **code hotfix** (and, if the buggy digest was already committed, likely **state surgery**), during which the chain is down.

### Where a divergence can realistically originate

Mitigating fact: `CalculateVoteWeight` reads the stored `StakedDuration` (not wall-clock), so weight is **time-invariant** — the scariest "weight silently decays between mutations" divergence is off the table, and both paths share `Apply`/`Hash`. The **residual** risk is therefore bounded but real, and all of it is "did the incremental path enumerate every weight-changing event correctly":

- a weight-mutating site with no hook, or a wrong delta (today 13 sites line up; a future staking change adding a 14th is a standing hazard — this is a permanent dual-implementation coupling, not a one-time correctness check);
- asymmetric zero/unstake handling (rebuild skips `w==0` and `isUnstaked()`; the live path removes on `≤0` — the paths to zero must land on identical map state);
- bucket-set coverage drift between what the hooks track and what the rebuild scans (native vs contract-staking vs NFT).

### What would reduce it (brief — happy to expand separately)

- **Don't fail silently or brick blindly:** emit per-candidate sub-digests to localize *which* `(candidate, voter)` diverged; add a counter/alert on the `Apply` swallow-path so a latent miss is observable **before** a restart; run the rebuild-vs-incremental cross-check at **each era boundary before freezing** (near the cause, and before a wrong weight is baked into a reward snapshot), not only at boot.
- **Root fix — remove the second derivation:** make the aggregated weights **first-class committed state** (loaded at boot, not rebuilt-and-compared) so there is only one derivation and nothing to disagree with. That is a larger change; the point here is that as long as a *separately-derived* digest sits in the state root, fail-closed is the only "correct" option and it is a network-halt trigger.

Given the blast radius (network-wide, non-self-healing), I think this warrants an explicit mitigation before the fork is activated, even though every path is inert pre-fork.

guo (commented):
Blocking: the superseded-window retirement is still reverted on the real  path.\n\n snapshots the state manager before calling . The latter writes , but  then calls ;  reverts to that pre-abandon snapshot for every Failure receipt. The committed cursor therefore remains non-terminal, and  continues to emit the same failing  every non-boundary block until the era rollover, producing repeated  logs.\n\nThe new unit test still calls  directly, so it bypasses both  and the Failure-receipt revert; it cannot establish the production behavior. The previous exact-head devbox/e2e probe reproduced 20 abandonment logs and 34 repeated chunks from one superseded window.\n\nPlease make the retirement survive the failure rollback (for example, perform it after the revert or settle against a post-retirement snapshot), and add a real Handle-path test that asserts committed terminal state plus no chunk on the following block.

guo:
Correction to my immediately previous review: its code identifiers were lost by a client-side quoting error. This comment is the authoritative finding.\n\nBlocking: the superseded-window retirement is still reverted on the real Handle path.\n\nHandle snapshots the state manager before calling abandonVoterRewardDistribution. That helper writes ScanPhase = voterScanAbandoned, but Handle then calls settleSystemAction with a Failure receipt and the original snapshot. settleAction reverts to that pre-abandon snapshot for every Failure receipt. The committed cursor therefore remains non-terminal, and CreatePostSystemActions continues to emit the same failing VoterRewardChunk every non-boundary block until the era rollover, producing repeated DRAIN_ABANDONED logs.\n\nThe new unit test still calls abandonVoterRewardDistribution directly, so it bypasses both Handle and the Failure-receipt revert; it cannot establish the production behavior. The prior exact-head devbox/e2e probe reproduced 20 abandonment logs and 34 repeated chunks from one superseded window.\n\nPlease make the retirement survive the failure rollback, for example by performing it after the revert or settling against a post-retirement snapshot. Add a real Handle-path test that asserts committed terminal state and no chunk on the following block.

guo (approved):
Re-reviewed at `dd56a161691493645333959e20864cfc2abfdde8`.

The previously reported rollback-retirement blocker is resolved: the failure path now reverts the failed chunk before retiring the cursor, then settles against a post-retirement snapshot. Exact-head adversarial tests through `Handle`, committed working-set state, and next-block `CreatePostSystemActions` all pass. Reversing only this fix reproduces the prior repeated Failure receipt / `DRAIN_ABANDONED` behavior, confirming the regression coverage is effective.

Relevant unit and e2e suites pass, and CI plus SonarCloud are green. Approved for this exact head.

GitHub currently reports the branch as behind `master`; please update the branch before merge and re-check the resulting head if it changes.

PR #4973, opened 2026-08-17
fix(api): serialize tracer Stop against GetResult

envestcc:
## The race

`parseTracer` arms a timeout watchdog that calls `tracer.Stop` from its own goroutine, while the goroutine running the trace calls `GetResult`. geth's `StructLogger` stores the interruption reason on a plain field:

```go
func (l *StructLogger) Stop(err error) {
	l.reason = err            // plain write
	l.interrupt.Store(true)   // atomic, but *after*
}

func (l *StructLogger) GetResult() (json.RawMessage, error) {
	if l.reason != nil {      // plain read, never consults l.interrupt
		return nil, l.reason
	}
	...
```

`reason` is written and read without synchronisation, and the atomic `interrupt` establishes no happens-before edge for it because `GetResult` never reads it. Any `debug_trace*` request that hits its deadline can trip this.

## The fix

Wrap the tracer so `Stop` and `GetResult` hold a common mutex. `Hooks` are passed through untouched — they run on the tracing goroutine and coordinate through `l.interrupt`, which is already atomic. Nothing in the geth fork is modified.

The guard is applied to every branch of `parseTracer` (default struct logger, configured struct logger, and named JS/native tracers), since the watchdog covers all of them.

## Reproduction

`TestParseTracerDefaultLoggerTimeoutFires` already exercises the path; the race detector is what surfaces it, so no new test is needed.

On `master` (85fd452c6):

```
$ go test -count=1 -gcflags="all=-N -l" -short -race \
    -run TestParseTracerDefaultLoggerTimeoutFires ./api/
exit=1   WARNING: DATA RACE x2
```

With this commit:

```
exit=0   WARNING: DATA RACE x0
```

Full `./api/` package under the same flags: `ok`, 0 races.

## Notes

Found while integrating this fix into `rc_2.5.0`; opening it separately against `master` so the branches don't diverge on it.

### DIFF
--- a/api/web3server_utils.go
+++ b/api/web3server_utils.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/big"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


guo (approved):
Reviewed the tracer wrapper. Stop and GetResult are serialized without changing Hooks, and repeated race validation passed on the combined tree.

PR #4976, opened 2026-08-18
action: return error instead of panicking in eth-tx conversion

guo:
## Summary

`toEthTx` (the internal helper that converts an action to its go-ethereum
transaction form) could panic on a few malformed or edge-case inputs. This
makes those paths return an error instead.

- a blob tx proto with no blob data left a `nil` blob that was later dereferenced
- set-code / blob txs dereferenced a `nil` destination for the contract-creation
  case — these tx types have no creation form
- an out-of-range amount / fee cap tripped `uint256.MustFromBig`

## Changes

- `toEthTx` now returns `(*types.Transaction, error)`
- reject missing blob data and nil destination up front
- use the checked `uint256.FromBig` so an oversized value returns an error
- add regression tests

`go test ./action/` passes; `api/ actpool/ chainservice/ state/` build.


envestcc (approved):
Approve, with one thing I would like done before merge (the mock, item 1 below).

## What I verified

**These panics are reachable pre-signature-verification, from an unauthenticated caller.** `envelope.ToEthTx()` is called from `sealedenvelope.loadProto` for the `ETHEREUM_EIP155` / `ETHEREUM_UNPROTECTED` encodings, i.e. inside `Deserializer.ActionToSealedEnvelope` — the same entry point used by gRPC `SendAction`, p2p action gossip, and block deserialization. The `generic_validator` checks that would otherwise catch these (`ErrSetCodeTxCreate`, the blob checks) all run downstream and cannot help. Only `api/web3server.go`'s JSON-RPC handler has a `recover()`; the gRPC and p2p dispatcher paths do not, so the panic takes the process down.

I wrote a throwaway repro against the pre-change code; all cases hit, including two end-to-end ones through the real deserializer:

```
--- PASS: TestZZReproPanics/e2e-deserializer-setcode-create
    PANIC => runtime error: invalid memory address or nil pointer dereference
--- PASS: TestZZReproPanics/e2e-deserializer-blob-no-blobdata
    PANIC => runtime error: invalid memory address or nil pointer dereference
```

<details>
<summary>Repro (feel free to lift this into the PR as an e2e test — see item 6)</summary>

```go
// SetCodeTx in contract-creation form: EthTo() returns nil, *to derefs.
core := &iotextypes.ActionCore{
    TxType: SetCodeTxType, Nonce: 1, GasLimit: 100000, ChainID: 1,
    GasFeeCap: "1000000000000", GasTipCap: "1000000000000",
    Action: &iotextypes.ActionCore_Execution{
        Execution: &iotextypes.Execution{Contract: "", Amount: "1"},
    },
}
pb := &iotextypes.Action{
    Core:         core,
    SenderPubKey: identityset.PrivateKey(0).PublicKey().Bytes(),
    Signature:    make([]byte, 65), // never checked before the panic
    Encoding:     iotextypes.Encoding_ETHEREUM_EIP155,
}
ad := (&Deserializer{}).SetEvmNetworkID(1)
r.Panics(func() { ad.ActionToSealedEnvelope(pb) })

// Same shape with TxType: BlobTxType and no BlobTxData panics too.
```
</details>

The bar for an attacker is: a well-formed public key plus 65 arbitrary signature bytes. That is it.

**One correction to the PR description.** `uint256.MustFromBig(nil)` does **not** panic in v1.3.2 — it has an explicit nil check and returns `nil`, and the geth fork's `SetCodeTx.copy()` / `BlobTx.copy()` both guard with `if tx.Value != nil`. So the third bug is the *overflow* case only, not nil. That is reachable: `Execution.amount` is parsed from a proto decimal string with no width bound and `Execution.SanityCheck` only rejects negatives, so a value ≥ 2²⁵⁶ gets all the way to `MustFromBig`. Same for `gasFeeCap`/`gasTipCap` via `DynamicFeeTx.fromProto`. The fix is right either way; only the rationale needs the tweak.

**Consensus safety — no fork gate needed, and replay is byte-identical.** Every newly-rejected input previously **panicked** rather than being accepted. For the nil-blob case I checked all three paths a blob tx must survive to be committed, and it panics on each: `toEthTx`, `SanityCheck()` (via `tx.blob.SanityCheck()` on a nil receiver), and `ProtoForHash()` → `toProto()` → `tx.blob.toProto()`. So such an action can never have been hashed, validated, or included — it cannot exist in chain history. And `uint256FromBig(nil) → 0` is observationally identical to the old `MustFromBig(nil) → nil`, since geth normalizes nil to zero in `copy()`; the tx hash is unchanged.

**Regression check:** requiring `BlobTxData` to be present does not break blob txs propagated without their sidecar — `Block.ProtoWithoutSidecar()` and `BlobTx.ProtoForRawHash()` both only nil out `BlobTxSidecar` and keep `BlobTxData`.

## Comments

**1. Please regenerate the mock before merge.** `misc/scripts/mockgen.sh` generates `test/mock/mock_envelope/mock_envelope.go` with `-source=./action/envelope.go`, which covers `TxCommonInternal`. `MockTxCommonInternal.toEthTx` still has the old `*types.Transaction` signature. It does not break the build — an unexported method in another package can never satisfy `action.TxCommonInternal` anyway, so that mock is unusable regardless — but it is now inconsistent with the interface and anyone re-running mockgen gets a diff.

**2. The nil-blob invariant is only enforced at one entrance.** After this change the proto path is closed, but `SanityCheck()`, `toProto()` and `ProtoForRawHash()` still nil-deref when `blob == nil`. The remaining producers are the exported `NewBlobTx(..., nil)` (unguarded) and `AbstractAction.convertToTx()` (guarded by `validateTx()` in `builder.go`, though that one `panic`s rather than returning). A nil check in `SanityCheck()` returning `ErrMissingBlobData` is one line and makes the invariant hold at every entrance instead of one.

**3. Two sentinels for one condition.** `actctx.go`'s `validateTx()` already returns `errors.Wrap(ErrInvalidAct, "blob tx with empty blob")` for exactly "blob tx without blob data". The new `ErrMissingBlobData` is a second, unrelated sentinel for the same condition. actpool and api classify errors by sentinel, so this invites divergent downstream handling — consider reusing the existing one, or having the new one wrap `ErrInvalidAct`.

**4.** `uint256FromBig(tx.BlobGasFeeCap())` can never return an error: `BlobGasFeeCap()` is `tx.blob.gasFeeCap().ToBig()`, round-tripped from a `uint256.Int`, so overflow is impossible by construction. Harmless, but as written it reads as though there is a real overflow path there. A short comment would clarify if it is deliberate defence in depth.

**5.** One `MustFromBig` of the same family survives: the `BlobTxType` arm of `AbstractAction.convertToTx()` in `actctx.go`. Only reachable from the builder path where values come from a real eth tx, so much lower severity — a follow-up rather than this PR.

**6. Test coverage.** The six subtests are all unit-level; none of them exercises `Deserializer.ActionToSealedEnvelope`, which is the path that actually demonstrates the vulnerability is closed. One e2e case there would also guard against a future reordering inside `loadProto` silently regressing this. The repro above is ready to lift. Minor: the new file's copyright header says 2024.

**7. PR description.** Two suggestions. First, per AGENTS.md ("confirm with a maintainer before altering action serialization paths"), it would help to state the replay-safety argument explicitly — every newly-rejected input previously panicked, therefore it cannot exist in chain history, therefore no hardfork gate — so a maintainer does not have to re-derive it. Second, the description does not mention that this is an anonymously triggerable node crash, which is the thing that determines the release cadence.

PR #4977, opened 2026-08-18
action: bound blob sidecar element counts before allocating

guo:
## Summary

`FromProtoBlobTxSideCar` allocated its `Blobs` / `Commitments` / `Proofs`
backing arrays sized directly from the proto-supplied element counts,
before the per-element length check runs. A `kzg4844.Blob` is 128 KiB, so
a proto declaring a large blob count allocates a large slice from a small
message.

This rejects element counts above the existing per-tx blob limit
(`MaxBlobGasPerBlock / BlobTxBlobGasPerBlob`) up front, so the allocation
is bounded regardless of the declared count. The same numeric limit is
already enforced later in `BlobTxData.SanityCheck`; this just applies it
before the allocation.

## Changes

- bound `len(Blobs)/len(Commitments)/len(Proofs)` before `make`
- add regression tests

`go test ./action/` passes.


envestcc (approved):
Approve. Small, correctly scoped, and it closes the nastiest of the three deserialization bugs in this series.

## What I verified

**Reachability.** `FromProtoBlobTxSideCar` is reached from `Deserializer.ActionToSealedEnvelope` → `sealedenvelope.loadProto` → `BlobTx.fromProto` → `fromProtoBlobTxData`, i.e. **before signature verification**, on the same entry point that serves gRPC `SendAction`, p2p action gossip, and block deserialization. No permission or valid signature required.

**Amplification.** An empty `bytes` entry in `repeated bytes blobs` costs 2 bytes on the wire and allocates `len(kzg4844.Blob)` = 131072 bytes — **65536×**. Measured locally against the pre-change code:

```
entries=1000     wire=2000 B      -> make() allocates 0.12 GiB   (65536x)
entries=100000   wire=200000 B    -> make() allocates 12.21 GiB  (65536x)
entries=2097152  wire=4194304 B   -> make() allocates 256.00 GiB (65536x)
```

Driving the real entry point (darwin/arm64):

```
FromProtoBlobTxSideCar(100k empty blobs) -> err="invalid blob length"
    took=2.86s   HeapSys=12.35 GiB
```

So a **200 KB message costs ~12 GiB of heap and ~2.9 s of CPU** before it is rejected. Repeatable and concurrent. gRPC's default `MaxRecvMsgSize` is 4 MiB (not overridden in this repo) and `api/websocket.go` sets a 15 MiB read limit, so there is ample headroom to OOM a node.

Worth noting for triage: this one is harder to contain than a nil-deref. Go's large-allocation failure path is `throw("out of memory")`, which `recover()` cannot catch — the `recover()` in `api/web3server.go` does not help here.

**Consensus safety — no behaviour change for any valid input.** For a valid blob tx the sidecar array lengths must equal `len(blobHashes)` (enforced by `verifySidecar`), and `BlobTxData.SanityCheck` already caps `len(blobHashes)` at the same `MaxBlobGasPerBlock / params.BlobTxBlobGasPerBlob` = 6. So nothing legitimate is newly rejected, and oversized inputs previously OOM'd rather than being accepted, so they cannot exist in chain history. Sidecars are excluded from the tx hash (`BlobTx.ProtoForRawHash`) and stripped from block serialization (`Block.ProtoWithoutSidecar`), so there is no hash or replay impact.

**Regression check I ran specifically:** does requiring `len(pb.Blobs) <= 6` break blob txs propagated with their sidecar stripped? No — both `Block.ProtoWithoutSidecar()` and `BlobTx.ProtoForRawHash()` only set `BlobTxSidecar = nil` and keep `BlobTxData`. Blob tx gossip and block sync are unaffected. The second caller, `blockchain/blockdao/blob_store.go`, reads back sidecars that were written from already-validated txs, so it is also within the bound.

## Non-blocking comments

**1. This is one instance of a class; worth a follow-up sweep.** The same `make(..., len(<proto field>))`-before-validation pattern exists elsewhere on the untrusted path:

- `fromProtoBlobTxData`: `make([]common.Hash, len(bh))` — 32 B per 2 B ≈ **16×**
- `fromAccessListProto`: `make(types.AccessList, len(list))` ≈ **24×**, plus `make([]common.Hash, numKey)` ≈ **16×**. Note this one is on the `fromProto` path of *all four* tx types (accessList / dynamicFee / blob / setCode).

Three orders of magnitude below 65536×, so a 4 MiB message buys roughly 100 MiB rather than 256 GiB — not fatal on its own, but a real memory-pressure surface under concurrency. I would not hold this PR for it; a separate issue establishing a "bound the count before allocating" convention seems like the right home.

**2. Forward compatibility: the `Proofs` bound will silently become wrong.** The current geth fork's `BlobTxSidecar` has only `Blobs/Commitments/Proofs` and no `Version` field, so `len(Proofs) == len(Blobs)` holds today and the bound is correct. Once geth is bumped to an EIP-7594 / PeerDAS-capable version, a v1 sidecar's `Proofs` are **cell proofs** with length `CELLS_PER_EXT_BLOB (128) × len(Blobs)`, and this flat bound would start rejecting valid blob txs — a failure that would be quite hard to diagnose. A one-line comment saying the bound assumes `len(Proofs) == len(Blobs)` and must be revisited when upgrading geth would pay for itself.

**3.** The error message carries only `permitted`. The neighbouring `SanityCheck` uses `"too many blobs in transaction: have %d, permitted %d"` — including `have` is friendlier to debug and more consistent.

**4.** The test covers `Blobs` and `Commitments` but not the `Proofs` arm of the same condition.

PR #4978, opened 2026-08-18
actpool: use write lock in ResetAccount

guo:
`queueWorker.ResetAccount` mutates the account pool: it calls `accountPool.PopAccount`, which removes the account from the `accounts` map and from the priority queue. It was holding `worker.mu.RLock()`.

A read lock allows multiple concurrent holders, and the other `RLock` paths on the same `worker.mu` read that same map from other goroutines (`PendingActions`, `Reset`, `getConfirmedState`, `AllActions`, `PendingNonce`). These are not serialized onto one goroutine — `PendingActionMap` and `reset` both fan out one goroutine per worker — so a removal can overlap a read of the same shard.

Take the write lock in `ResetAccount` instead, matching `putAction` and the other writers. No other `RLock` path in the worker removes from the map, so this is the only place that needed changing.

Also adds a concurrency regression test that ranges the account map (`PendingActions`) while `ResetAccount` removes and re-adds entries on the same worker. It reports a data race under `-race` before the change and passes after:

```
go test -race ./actpool/ -run TestQueueWorker_ResetAccountConcurrentWithPendingActions
```

`go test -race ./actpool/...` passes.


envestcc (approved):
Approve — the two-line change is correct and I confirmed it fixes what it claims. But there is a second, worse writer-under-`RLock` on the same mutex that this PR does not touch, and the PR description currently asserts there isn't one. Details below; I do not think it should block merging, but the description should be corrected and the remaining race tracked.

## What I verified

I wrote three `-race` tests, ran them against the unpatched code, then applied this PR's two-line change and re-ran:

| test | scenario | before | after |
|---|---|---|---|
| A | `PendingActions` (RLock) ∥ `ResetAccount` (RLock→Lock) | 🔴 DATA RACE | ✅ PASS |
| B | 4× concurrent `PendingActions` | ✅ PASS | ✅ PASS |
| C | `PendingActions` (RLock) ∥ **`Reset` (RLock)** | 🔴 DATA RACE | 🔴 **still DATA RACE** |

A confirms the bug is real and that this PR fixes it. C is the problem.

**Deadlock check:** the only caller, `actPool.DeleteAction`, holds no lock when it calls `ResetAccount`, and `popLowestPriorityAcrossWorkers` acquires every worker mu in index order relying on "no other code path holds more than one worker mu at a time" — this upgrade preserves that invariant. No deadlock risk.

**The comment change is right, and the old comment was wrong.** `emptyAccounts` is a `go-pkgs/cache/ttl.Cache` with its own `sync.RWMutex`, so "thread safe on its own" is accurate. The old comment attributed its safety to `worker.mu.RLock`, which had already been released at that point — good riddance.

## 🔴 `accountPool.Range` writes the heap under `RLock` — not fixed here

```go
func (ap *accountPool) Range(callback func(addr string, acct ActQueue)) {
	for addr, account := range ap.accounts {
		callback(addr, account.actQueue)
	}
	heap.Init(&ap.priorityQueue)   // <- this is a write
}
```

`heap.Init` calls `Swap`, and `Swap` does more than reorder the slice — it rewrites every `accountItem.index`:

```go
func (aq accountPriorityQueue) Swap(i, j int) {
	aq[i], aq[j] = aq[j], aq[i]
	aq[i].index = i
	aq[j].index = j
}
```

So `Range` is a writer wearing a reader's clothes, and it is called from two `RLock` paths: `queueWorker.Reset` and `queueWorker.PendingActions`. An `RWMutex` lets both in at once.

<details>
<summary>Race report (still reproduces with this PR applied)</summary>

```
WARNING: DATA RACE
Read at 0x00c0009eef38 by goroutine 87:
  actpool.accountPriorityQueue.Less()        accountpool.go:124
  container/heap.Init()
  actpool.(*accountPool).Range()             accountpool.go:107
  actpool.(*queueWorker).Reset()             queueworker.go:257   <- RLock

Previous write at 0x00c0009eef38 by goroutine 86:
  actpool.accountPriorityQueue.Swap()        accountpool.go:142
  container/heap.Init()
  actpool.(*accountPool).Range()             accountpool.go:107
  actpool.(*queueWorker).PendingActions()    queueworker.go:289   <- RLock
```
</details>

**It is reachable on a normal node.** `actPool.reset()` and `actPool.PendingActionMap()` each fan out one goroutine per worker, and nothing at the actPool level serializes them against each other:

- `ReceiveBlock` → `ap.reset()` → goroutine per worker → `worker.Reset(ctx)` → RLock → `Range`
- proposer / API → `PendingActionMap()` → goroutine per worker → `worker.PendingActions(ctx)` → RLock → `Range`

"A block lands while the proposer or the API is collecting pending actions" is routine operation.

**The consequences are worse than the race being fixed.** Once `accountItem.index` has been concurrently rewritten, `PopAccount` and `DeleteIfEmpty` call `heap.Remove(&ap.priorityQueue, account.index)` with a stale index: at best the wrong account is removed, at worst the index is out of range and it panics. The priority ordering also becomes meaningless, and `popLowestPriorityAcrossWorkers` picks its eviction victim from `priorityQueue[0]`.

Two ways forward, and the choice is really yours rather than mine:

- **(a) Fix it here**: also take `Lock()` in `Reset` and `PendingActions`. The cost is that both are on the block-production path, so this serializes reset against pending-collection per worker. That trade-off deserves a deliberate call from you or a maintainer.
- **(b) Land this PR as-is and track the rest separately.** Fine by me — but then please correct the description.

## 🟡 The completeness claim in the description is not accurate

> "No other `RLock` path in the worker removes from the map, so this is the only place that needed changing."

"Does not remove from the map" is not the same as "does not write". `Range` writes the heap. My worry is less the bug than the sentence: left in the commit message, it tells the next person that `worker.mu` has already been audited end to end.

## Minor

The new test is a pure race harness with no assertions, which is fine — `make test` runs `-race` on the non-e2e packages, so it has real CI value. 200 rounds × 24 accounts reproduced reliably for me.

PR #4979, opened 2026-08-18
fix(factory): drop an action that panics during block production

guo:
### Description

`pickAndRunActions` runs each action picked from the action pool without any per-action `recover`. If one action panics, the panic unwinds the whole mint and is only caught by the block preparer goroutine, which throws the draft away. The action is also never removed from the pool, because the panic happens before `DeleteAction` is reached, so the same action is picked again on every following round and the proposer keeps producing nothing.

This wraps the per-action call in a `recover` and turns the panic into a normal error:

- the offending sender's queue is deleted from the action pool
- the draft is abandoned exactly as it is today, so the next round starts from a clean working set without that action

The draft is deliberately not continued on the same working set: it still holds the partial writes of the action that panicked, and `runAction`'s deferred `ResetSnapshots` only clears the snapshot list, it does not revert them.

Also adds a `iotex_mint_action_panics_total` counter, next to the existing mint/validate panic counters.

### Test

`TestWorkingSet_Mint_SkipsPanickingAction` in `state/factory/workingset_test.go`: an action whose handler panics aborts the first mint and gets its sender dropped from the pool; the next mint then produces a block containing the remaining action. Fails before the change (the panic escapes `Mint`), passes after.

`go test ./state/factory/... ./actpool/...` passes.

### DIFF
--- a/state/factory/workingset.go
+++ b/state/factory/workingset.go
@@ -55,6 +55,10 @@ var (
 		},
 		[]string{"type"},
 	)
+	_runActionPanicMtc = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "iotex_mint_action_panics_total",
+		Help: "Number of panics recovered while running an action during block production (the action is dropped from the pool and the draft is discarded).",
+	})
 
 	errInvalidSystemActionLayout = errors.New("system action layout is invalid")
 	errUnfoldTxContainer         = errors.New("failed to unfold tx container")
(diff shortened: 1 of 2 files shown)


guo:
Closing in favor of #4921, which already implements the same per-action recover + sender eviction during minting (opened earlier, same `iotex_mint_action_panics_total` metric). Let's land #4921 instead.

PR #4980, opened 2026-08-18
staking: guard the non-stop alias fix with the alias-era feature height

guo:
### What was wrong

`Protocol.CreatePreStates` rebuilds the candidate center's name/operator maps
(`loadNameOperatorMapOwnerList`) when the block height equals
`FixAliasForNonStopHeight`, but it did not check whether the candidate center
still keeps aliases at that height.

Every other place that touches the same state pairs its height check with
`CandCenterHasAlias` — see `needToReadCandsMap` / `needToWriteCandsMap` in the
same file. This one branch was missing it.

### What the fix does

Adds the same `CandCenterHasAlias(blkCtx.BlockHeight)` guard to the branch, so
the rebuild only runs while the candidate center is still alias-based.

No behaviour change on the configured height (the guard is true there); past
Okhotsk the branch is now consistently skipped.

### Test

`TestCreatePreStatesFixAliasForNonStop` covers both sides: the rebuild runs at a
height in the alias era and is skipped once past Okhotsk. It fails without the
change.

`go test ./action/protocol/staking/...` passes.


envestcc (dismissed):
Approve. One question I would like answered before merge, but I do not consider it blocking.

## What I verified

**The numbers work out to a no-op on mainnet.**

| | value | source |
|---|---|---|
| `FixAliasForNonStopHeight` | **19,778,036** | `blockchain/config.go`, `DefaultConfig` |
| mainnet `OkhotskBlockHeight` | **21,542,761** | `blockchain/genesis/genesis.go`, `defaultConfig()` |
| `CandCenterHasAlias(h)` | `!IsOkhotsk(h)`, i.e. `h < 21,542,761` | `action/protocol/context.go` |

The branch is an `==` test, so it can only fire at one height, and at that height `CandCenterHasAlias` is unconditionally true. The added conjunct is therefore a tautology on mainnet. Mainnet is long past both heights, so this code only executes during sync/replay from genesis — and block 19,778,036 replays exactly as before. **Byte-identical replay.**

**The precedent cited in the description checks out.** In the same file:

```go
func (p *Protocol) needToReadCandsMap(ctx, height) bool {
	return height > p.config.PersistStakingPatchBlock && fCtx.CandCenterHasAlias(height)
}
func (p *Protocol) needToWriteCandsMap(ctx, height) bool {
	return height >= p.config.PersistStakingPatchBlock && fCtx.CandCenterHasAlias(height)
}
```

Both pair a config height with `CandCenterHasAlias`. The `FixAliasForNonStopHeight` branch really was the only one missing it, and the file is now internally consistent.

**The branch is not cosmetic**, which is why the question below matters. `loadNameOperatorMapOwnerList(owners, owners, nil)` passes `nil` as the third argument, so it also **clears `cb.owners`**, and that field has two real downstream readers:

- `vote_reviser.correctAliasCands()` builds an owner map from `base.owners` to backfill each candidate's `Operator` / `Reward` / `Name`
- `CandidateCenter.WriteToStateDB()` sorts `ownersList()` and `PutState`s it into `CandsMapNS`

So if this ever flipped on some network, it would be a real divergence, not a no-op.

## Question (non-blocking, but please confirm)

`FixAliasForNonStopHeight` is **chain config** (node-local YAML) while `CandCenterHasAlias` derives from **genesis** (`okhotskHeight`). Different sources. On any network where `fixAliasForNonStopHeight >= okhotskHeight`, this change flips behaviour at that height — and per the paragraph above, that is a state divergence.

Mainnet defaults are safe; I checked the arithmetic. But testnet/devnet genesis is not in the repo, so I cannot verify those. **Could you confirm that `fixAliasForNonStopHeight < okhotskHeight` holds on testnet and on any network with a custom genesis?** If some network does not satisfy it, that case needs its own handling.

## Suggestion on the description

I would sell this harder than "keep it consistent". Today a **YAML-only knob can reach into a state-transition path** (`CreatePreStates`) and mutate the candidate center, which means two nodes with different YAML can diverge. Adding a genesis-derived guard means that after Okhotsk the knob is inert no matter what the YAML says — the blast radius of a misconfiguration shrinks. That is the same problem #4981 is solving for the sibling heights, and it is worth saying out loud.

## Minor

- `MustGetFeatureWithHeightCtx` panics (`log.S().Panic`) when the key is absent. I checked reachability and the risk looks low: all three `CreatePreStates` call sites in `workingset.go` (`Process`, `runActionsLegacy`, `pickAndRunActions`) get their ctx from `blockchain.go`, which adds `WithFeatureWithHeightCtx`; the API path in `coreservice.go` does likewise; and most directly, `needToWriteCandsMap` in this same file already uses the `Must` form in the same block-processing flow. Consistent with the file's own convention, so no objection — just noting that `CandidateCenter.Commit()` uses the non-panicking `GetFeatureWithHeightCtx` with an `ok` check, so both styles coexist in the tree.
- The PR is `BEHIND`; needs a rebase.

## Test

Good — it covers both sides (`OkhotskBlockHeight-1` runs the rebuild and the owner list is cleared to nil; `OkhotskBlockHeight` skips it and the entry survives), which is exactly the AGENTS.md convention of testing `height = fork - 1` and `height = fork`. I confirmed `genesis.TestDefault()` inherits `defaultConfig()`'s Okhotsk = 21,542,761, so `-1` does not underflow and the test is meaningful.

For the record, not a complaint: the test sets `FixAliasForNonStopHeight` to the height under test, including past Okhotsk. That is the right way to exercise the new guard, but it proves the guard works rather than that the combination is reachable in production — which is what the question above is about.

envestcc (approved):

PR #4981, opened 2026-08-18
genesis: source staking patch heights from genesis instead of chain config

guo:
### What was wrong

`persistStakingPatchBlock` and `fixAliasForNonStopHeight` decide when the
candidate center's name/operator maps get persisted to the staking patch file
and when they are rebuilt. Those maps back `ContainsName` / `ContainsOperator`,
so the heights are network-wide behaviour — but they were read from
`cfg.Chain`, i.e. node-local YAML, where two nodes can disagree. Every other
height of that kind lives in genesis.

### What this does

- Adds `PersistStakingPatchBlock` / `FixAliasForNonStopHeight` to
  `genesis.Blockchain`, with the exact values `blockchain.DefaultConfig`
  defaulted to (`19778037` / `19778036`). `Genesis.Hash()` covers a fixed field
  set, so it is unaffected.
- `chainservice/builder.go` now takes both from `cfg.Genesis`.
- The two `chain.*` keys are marked deprecated and default to 0. They are still
  parsed, so existing config files keep loading, but nothing is wired from them.
  A leftover key logs a warning at start-up naming the ignored value and the
  genesis value in effect.

No behaviour change on mainnet or testnet: the genesis defaults are the values
mainnet nodes already ran with, and on testnet the published
`persistStakingPatchBlock` (`MaxUint64`) and the genesis default are equivalent —
`needToReadCandsMap` / `needToWriteCandsMap` also require the height to be below
Okhotsk, which on testnet is well below both.

### Test

`TestStakingPatchHeights` pins the moved defaults, checks the deprecated keys
are unset by default, and checks that a config carrying them — including the
testnet `MaxUint64` — passes every validator.
`TestDeprecatedStakingPatchHeightsFromYAML` loads a config file with both
deprecated keys set and checks it loads without error, with the genesis values
in effect.

`go build ./...`, `go test ./config/... ./chainservice/... ./blockchain/genesis/...`
pass. `./blockchain/...` passes except `TestBlockIndexerChecker_CheckIndexer`,
which fails the same way on master.


envestcc (dismissed):
Approve on the code. I would hold the merge until item 1 is answered, and I think #4980 should land first — see below.

## What I verified

**1. `Genesis.Hash()` really is unaffected — this was the scariest part and it checks out.**

`Hash()` builds `iotextypes.GenesisBlockchain` from an explicitly enumerated field list: `Timestamp`, `BlockGasLimit`, `ActionGasLimit`, `BlockInterval`, `NumSubEpochs`, `NumDelegates`, `NumCandidateDelegates`, `TimeBasedRotation`. **Not a single fork height**, let alone the two new fields. Adding fields to the Go `Blockchain` struct cannot change the genesis hash. The claim in the description holds.

**2. Keeping the deprecated keys parseable is not a courtesy — it is mandatory. I tested it.**

Config loading goes through `go.uber.org/config`, which is **strict**:

```
existing key  -> err=<nil>
unknown  key  -> err=failed to unmarshal YAML config to struct: yaml: unmarshal errors: ...
```

So had `chain.persistStakingPatchBlock` been deleted from the struct, **every node whose config file still carries that key would fail to start** — including the published testnet config. Retaining the keys, ignoring their value, and warning is the only shape that works, and `TestDeprecatedStakingPatchHeightsFromYAML` is the right regression guard for it. This also matches AGENTS.md red line 4 ("Deprecated fields stay; never remove them"). Good call.

**3. The rewiring is complete.** `cfg.Chain.PersistStakingPatchBlock` / `FixAliasForNonStopHeight` are read in exactly one place in the tree — `chainservice/builder.go` — and those are the two lines this PR changes. No stragglers, no split brain.

## 1. Merge-order dependency on #4980 (please confirm)

The `FixAliasForNonStopHeight` half of the migration is only a no-op on testnet if #4980's `CandCenterHasAlias` guard is already in the tree.

Suppose the published testnet config sets `fixAliasForNonStopHeight` to `MaxUint64` — plausible, since this PR's own table includes `{MaxUint64, MaxUint64}` and the YAML test writes **both** keys as `MaxUint64`. Then:

- before: the branch fires only at `h == MaxUint64`, i.e. never
- after: the branch fires at `h == 19,778,036`
- **with #4980**: the guard is `19,778,036 < testnetOkhotsk`, which is false, so it still never fires — equivalent ✅
- **without #4980**: the branch genuinely executes at testnet height 19,778,036, rebuilding the name/operator maps and **clearing `cb.owners`**, which feeds `vote_reviser.correctAliasCands` and `CandidateCenter.WriteToStateDB` → **state divergence at that height when replaying from genesis**

Testnet is well past 19.78M, so running nodes would not notice; only nodes **re-syncing from scratch** would break. That is the unpleasant kind: silent, and it only bites whoever resyncs.

Conversely, if the published testnet config does not set the key at all, its old effective value was `DefaultConfig`'s 19,778,036 — identical to the new genesis default — and there is no dependency.

**So: what does the published testnet config actually set `fixAliasForNonStopHeight` to?** If it is anything other than 19778036, **#4980 is a hard prerequisite** and the ordering should be stated in the description.

## 2. The "no behaviour change on testnet" argument rests on a number I cannot check

```
needToReadCandsMap(h)  = h >  PersistStakingPatchBlock && CandCenterHasAlias(h)
needToWriteCandsMap(h) = h >= PersistStakingPatchBlock && CandCenterHasAlias(h)
```

Testnet before: `MaxUint64`, so both are identically false. Testnet after: `19,778,037`, so they are true for `19,778,037 <= h < testnetOkhotsk`. The two are equivalent **iff `testnetOkhotsk <= 19,778,037`**.

The description says "well below both", but testnet genesis is not in the repo, so I cannot verify it — and the entire testnet-safety conclusion rests on that one number. Could you put the actual testnet `okhotskHeight` value in the PR body, so reviewers (and whoever reads this in a year) can redo the arithmetic themselves?

## 3. The warning never fires when the staking protocol is disabled

Both `warnDeprecatedStakingPatchHeight` calls sit inside `registerStakingProtocol`, which opens with:

```go
if !builder.cfg.Chain.EnableStakingProtocol {
	return nil
}
```

So a node with staking disabled (API nodes, some tooling nodes) that still carries the deprecated keys gets no signal at all. Since this warning is the *only* user-facing signal for the whole migration, that is a shame. It depends purely on config, not on whether the protocol got registered, so `config.Validates` — or at least above the early return — would be a better home.

## Minor

- A network with a custom genesis that does not explicitly set `persistStakingPatchBlock` silently inherits `defaultConfig()`'s 19,778,037, a mainnet-specific number. It is currently saved by the Okhotsk guard, but "safe because another condition happens to be false" is a fragile arrangement. Worth setting both values explicitly in the published testnet genesis.
- Warning rather than erroring on a conflicting value is the right call — erroring would put us back in "node won't start" territory. I would just make the log line hard to miss, and/or emit a metric so this can be swept fleet-wide rather than found by reading startup logs.
- `BEHIND`; needs a rebase.

## Test

Both tests earn their place, particularly `TestDeprecatedStakingPatchHeightsFromYAML`, which is the regression guard for the startup-blocking failure mode above. One small gap: nothing asserts that the warning actually fires — had there been such an assertion, it would probably have surfaced item 3.

envestcc (approved):

PR #4986, opened 2026-08-24
fix(actpool): reject actions the protocols will certainly reject

envestcc:
## Problem

The action pool applies only `GenericValidator` — nonce, balance, signature. An action that a protocol's own feature gate makes **impossible to include** is accepted anyway: it takes a nonce and holds it until `ActionExpiry` (10 min). Because nonces must be consecutive, that one action stalls every later action from the same account.

Nothing anywhere says so:

- the RPC returns a transaction hash
- `eth_getTransactionCount` counts it
- resubmitting reports `known transaction`, confirming it really is in the pool
- **no receipt, and no warn/error log**
- after expiry the nonce quietly drops back

The user sees a transaction go out and nothing happen.

## The official CLI reaches it

Before Zanzibar, `ioctl stake2 update NAME OPERATOR REWARD` with no `--bls-*` flags builds a BLS-less `candidateUpdate` — which `action/protocol/staking/validations.go` rejects for the entire Xingu..Zanzibar range. No malformed input required.

Observed on TestNet at ~46,864,000 (Zanzibar height 46,880,641, so Xingu rules):

```
nonce 1218  candidateUpdate(name,operator,reward)   entered pool, expired 10 min later, no receipt, no log
nonce 1219  candidateUpdateWithBLS(...)             stuck behind 1218
nonce 1220  plain transfer                          stuck behind 1218

replace 1218 with a higher-priced transfer  -> mined at 46,864,006
1219 then drains on its own                 -> mined at 46,864,041, status success
```

Diagnosing that took hours — gas price, transaction type, submission endpoint, single-node submission, and a restart of all six nodes were all tried, because the chain emits no signal pointing at the real cause.

## Fix

Run the registered protocols' own `ActionValidator.Validate` at admission, mirroring the `ActionCtx` the proposer builds so the pool and the proposer judge an action identically. `actpool.context` already supplies a tip+1 `FeatureCtx`, which is the same height the proposer will evaluate at.

### Why this cannot affect consensus

Registration goes through a new `WithPrivateValidator` option, landing on the **Add-only** list — deliberately *not* `AddActionEnvelopeValidators`.

That distinction is the whole safety argument. `actionEnvelopeValidators` also run in `actPool.Validate`, which block validation reaches through the bundle pool. Rejecting there would turn *"included in a block with a failure receipt"* into *"block is invalid"*. The comment above `validateBLSPairing` makes the same point about gas semantics: **mempool leniency can never fork the chain, but validation strictness can.**

Protocol validation already runs independently at proposal (`workingset.go`, `validateAndRun`) and at block validation (`process` / `processLegacy`). Both still do; neither is touched.

### Registry is resolved lazily

`chainservice` builds the pool (`buildActionPool`) **before** it registers any protocol. The registry is therefore passed by pointer and read at `Validate` time — resolving eagerly would capture an empty registry and silently validate nothing. There is a regression test for exactly this.

System actions are skipped: they are proposer-produced and several carry no sender to build an `ActionCtx` from.

## Tests

4 new tests covering rejection-at-admission, lazy registry resolution, the system-action bypass, and the private-vs-envelope validator list placement.

```
actpool       126 passed
chainservice   31 passed
e2etest        130 passed  (-run 'Staking|ActPool|Native')
```

## Not included

Item 3 of the original report — logging the action hash and reason when `pickAndRunActions` skips an action — is left out; `workingset.go` already logs `failed to validate tx` with the height and error there. Happy to extend it with the action hash if you'd like it in this PR.

### DIFF
--- a/actpool/options.go
+++ b/actpool/options.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/facebookgo/clock"
+
+	"github.com/iotexproject/iotex-core/v2/action"
 )
 
 // ActQueueOption is the option for actQueue.
@@ -42,6 +44,22 @@ func WithBundlePool(bp *BundlePool) func(*actPool) error {
 			return errors.New("bundle pool cannot be nil")
 		}
 		ap.bundlePool = bp
+		return nil
+	}
+}
+
+// WithPrivateValidator adds a validator that runs on the Add path only.
+//
+// Private validators are deliberately excluded from actPool.Validate, which
+// block validation reaches via the bundle pool. Anything whose rejection would
+// change a block's validity -- rather than merely keeping an action out of this
+// node's pool -- belongs here rather than in AddActionEnvelopeValidators.
+func WithPrivateValidator(v action.SealedEnvelopeValidator) func(*actPool) error {
+	return func(ap *actPool) error {
+		if v == nil {
+			return errors.New("private validator cannot be nil")
+		}
+		ap.privateValidators = append(ap.privateValidators, v)
 		return nil
 	}
 }
--- a/actpool/protocol_validator.go
+++ b/actpool/protocol_validator.go
@@ -1,0 +1,94 @@
+// Copyright (c) 2026 IoTeX Foundation
+// This source code is provided 'as is' and no warranties are given as to title or non-infringement, merchantability
+// or fitness for purpose and, to the extent permitted by law, all liability for your use of the code is disclaimed.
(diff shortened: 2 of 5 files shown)


guo (approved):
Reviewed at 57abd10e. Fable high review and devbox2 fork-boundary adversarial validation passed: the admission gate matches tip+1 fork activation, rejects only protocol-certain actions without consuming nonce, removes the reproduced queue stall, and leaves normal/state-dependent traffic admitted. CI is green.

PR #4987, opened 2026-08-24
fix(api): stop the 21000 call floor from inflating native action estimates

envestcc:
## Problem

`eth_estimateGas` applies a flat 21000 floor to every result. That is the floor eth clients expect for a call that reaches the EVM — but a native protocol action never does. It is charged its own intrinsic gas, and **every native staking action's base intrinsic gas is 10000**:

```
CandidateActivate / CandidateRegister / CandidateUpdate / CandidateEndorsement
CandidateTransferOwnership / CandidateDeactivate / DepositToStake / MoveStake
CreateStake / MigrateStake / ReclaimStake / Restake        ... all = 10000
```

So the reported number cannot match the receipt for any of them.

## Measured

Eight submissions on a live network: **21000 estimated against 10000 consumed, every time** — a constant gap that does not move with the arguments. MetaMask reads that gap as a failed interaction, so the action shows up as failed in the wallet while the chain reports success.

This affects **every native staking operation initiated from a wallet**, not one feature.

## This is not the `codeSize == 0` short-circuit

It looks like one from the outside, but `ethTxToEnvelope` dispatches on the destination address **before** it ever asks whether the target holds code:

```go
if to == address.StakingProtocolAddr  { return elpBuilder.BuildStakingAction(tx) }
if to == address.RewardingProtocol    { return elpBuilder.BuildRewardingAction(tx) }
isContract, err := svr.checkContractAddr(to)   // ← never reached for the two protocol addresses
```

The native action *is* decoded and `IntrinsicGas()` *is* consulted, returning 10000 correctly. The floor then overwrites it. The fix belongs at the floor, not at the dispatch.

## Fix

Skip the floor when the transaction is addressed to a native protocol pseudo-address.

`Transfer` and `Execution` keep it. `TransferBaseIntrinsicGas` is *also* 10000, so a bare transfer is precisely the case the floor exists for, and eth tooling expects 21000 there — this change deliberately does not disturb that.

The address set in the new helper is kept identical to the dispatch in `ethTxToEnvelope` and to `coreService.checkContract`. Widening one without the others would make `eth_estimateGas` and `eth_sendRawTransaction` disagree about what an action even is, so this change widens none of them.

## Tests

Two subtests pinning both sides of the boundary:

- a staking-protocol action (`candidateActivate(uint64)`) reports its intrinsic gas — 10000, not 21000
- a plain transfer still reports 21000

## Pre-existing failure, unrelated

`TestEstimateExecutionGasConsumption/FailedToAccountState` panics on a nil dereference. Verified identical on an unmodified `master` checkout — not introduced here, but worth a separate look.

### DIFF
--- a/api/web3server.go
+++ b/api/web3server.go
@@ -611,7 +611,14 @@ func (svr *web3Handler) estimateGas(ctx context.Context, in *gjson.Result) (inte
 	if err != nil {
 		return "0x" + hex.EncodeToString(retval), err
 	}
-	if estimatedGas < 21000 {
+	// 21000 is the floor eth clients expect for a call that reaches the EVM. A
+	// native protocol action never does: it is charged its own intrinsic gas,
+	// which for most staking and rewarding actions is below 21000. Applying the
+	// floor there reports a number the receipt will never match -- 21000
+	// estimated against 10000 actually consumed for setVoterRewardDestination,
+	// measured on TestNet -- and wallets read that gap as a failed interaction
+	// even though the action succeeded and its view returned the new value.
+	if estimatedGas < 21000 && !isNativeProtocolTx(tx) {
 		estimatedGas = 21000
 	}
 	return uint64ToHex(estimatedGas), nil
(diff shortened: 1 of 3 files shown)


guo (approved):
Reviewed at d8be3e4e. Fable high review and devbox2 head/base validation passed: native staking and rewarding estimates below 21000 now equal intrinsic gas and minted receipt gas, while transfer, EVM, adjacent pseudo-address, malformed-request, and gRPC control behavior remains unchanged. CI is green.

PR #4988, opened 2026-08-24
fix(chainservice): reject a Zanzibar height that activates past the first freeze

envestcc:
## Problem

An era settles at the last block of a boundary epoch, but the window it settles against opens **~1.5 epochs earlier**: `PutPollResult` is created at the midpoint of the *preceding* epoch, and that is where `FreezeCandidateRewardSnapshots` and `BeginEraCOWWindow` run.

Both sides are gated on the same `IsZanzibar` predicate, and `GrantEpochReward` treats *"this is a boundary epoch"* as equivalent to *"a window was opened for this era"* — `reward.go` says so explicitly, and notes there is no second condition.

Activate between those two points and the equivalence breaks:

1. the freeze no-ops, because the feature is not live yet
2. the settlement fires anyway, looking for a window nobody opened
3. **that era's entire epoch reward is lost**

Observed as a silent zero for epoch 6 on a local chain, with nothing in the logs to explain it.

## The intuitive rule is wrong

"Activate at the start of an era" — meaning the epoch where `epochNum % epochsPerRewardEra == 0` — produces *exactly* the broken case. That epoch settles at its own end and needs its freeze 1.5 epochs earlier, which by then has already passed. Worth stating in the error message, which this PR does.

## Why the exact rule, not a rounded margin

The original writeup proposed "leave at least two whole epochs". This implements the precise rule instead — compare against the freeze height the runtime actually computes.

Two reasons, both about blast radius:

- This runs at **startup against the operator's real genesis**. A rule stricter than the mechanism would take a correctly-configured fleet down on restart.
- `Genesis.validate()` is skipped for `defaultConfig()` and test literals, so **no test suite would catch that before the fleet did.**

There is a test asserting the live testnet height stays valid.

## Why `chainservice` and not `Genesis.validate()`

`blockchain/genesis` cannot import `rolldpos` — `genesis → rolldpos → action/protocol → genesis` is a cycle. `rolldpos.Protocol` is a pure value type and exactly the epoch calculator this rule needs; duplicating its three-regime arithmetic (`numSubEpochs` / Dardanelles / Wake) would leave two copies of consensus-adjacent math to keep in lockstep.

The tradeoff: the check no longer fires for non-chainservice consumers of `genesis.New`. That seemed better than the duplication.

## Calibration

The arithmetic is checked against settlements that actually happened, not against its own derivation:

| era | freeze | settlement |
|---|---|---|
| 54888 | 46,892,881 | 46,895,040 |
| 54912 | 46,927,441 | — |
| 54936 | 46,962,001 | — |

All read off TestNet. All five values asserted. `GetEpochNum(46880641) == 54879` and `GetEpochHeight(54888) == 46893601` are asserted too — if `eraFreezeHeight` disagrees with any of them the rule is measuring the wrong thing and every other case is meaningless.

## Tests

```
chainservice   35 passed
e2etest       132 passed   (-run 'Staking|Native|Local')
```

Covers: observed-value calibration, live configurations pass, activation exactly at the freeze passes, one block past it fails, the first block of a boundary epoch fails, just after the boundary epoch passes, and degenerate inputs (nil protocol, era length 0, height 0) do not turn a startup check into a startup crash.

### DIFF
--- a/chainservice/builder.go
+++ b/chainservice/builder.go
@@ -927,6 +927,11 @@ func (builder *Builder) build(forSubChain, forTest bool) (*ChainService, error) 
 	if err := builder.registerRollDPoSProtocol(); err != nil {
 		return nil, errors.Wrap(err, "failed to register roll dpos related protocols")
 	}
+	// Needs the epoch arithmetic rolldpos was just registered with, which is
+	// why this is not in Genesis.validate().
+	if err := checkZanzibarEraMargin(builder.cfg.Genesis, builder.cs.registry); err != nil {
+		return nil, err
+	}
 	if err := builder.registerExecutionProtocol(); err != nil {
 		return nil, errors.Wrap(err, "failed to register execution protocol")
 	}
--- a/chainservice/zanzibar_era_margin.go
+++ b/chainservice/zanzibar_era_margin.go
@@ -1,0 +1,111 @@
+// Copyright (c) 2026 IoTeX Foundation
+// This source code is provided 'as is' and no warranties are given as to title or non-infringement, merchantability
+// or fitness for purpose and, to the extent permitted by law, all liability for your use of the code is disclaimed.
+// This source code is governed by Apache License 2.0 that can be found in the LICENSE file.
+
+package chainservice
+
+import (
+	"math"
+
+	"github.com/pkg/errors"
+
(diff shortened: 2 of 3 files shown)


envestcc:

PR #4989, opened 2026-08-24
feat(genesis): add zanzibarBetaHeight

envestcc:
Plumbing only — **no behaviour is gated on it yet.** This is the base the two Zanzibar corrections land on.

## Why a beta height is needed

Zanzibar has already activated on testnet (46,880,641), and its one-shot work runs in the **single block** at `ZanzibarBlockHeight` — `migrateHermesRewardOptIn` and `backfillOwnerIndex`, both from `CreatePreStates`.

Correcting either in place would make a node replaying history compute a different receipt root for that block. **A resync would fork off the live chain.**

A beta height is how this codebase has handled the situation before. A chain that has not activated yet sets `zanzibarBetaHeight == zanzibarHeight` and gets the corrected behaviour at activation — the same arrangement mainnet already uses for XinguBeta (`41648761` for both) and YapBeta.

## What's here

The five sites the existing pattern requires:

| file | change |
|---|---|
| `blockchain/genesis/genesis.go` | field + `yaml:"zanzibarBetaHeight"` |
| " | `math.MaxUint64` default |
| " | `IsZanzibarBeta` predicate |
| `config/config.go` | ordering case in `ValidateForkHeights` |
| `testutil/genesis.go` | entry in `NormalizeGenesisHeights`' ordered slice |

That last one matters more than it looks: the slice is walked pairwise, so a height **not** listed in it is silently skipped by the clamp and tests can construct orderings that `ValidateForkHeights` would reject.

**No `FeatureCtx` flag.** A flag with no consumer is dead code; it lands with the change that reads it.

## Test coverage gap closed along the way

The ordering table gained the `Zanzibar`/`ZanzibarBeta` pair — and also the `YapBeta`/`Zanzibar` pair, which `ValidateForkHeights` has implemented since Zanzibar went in but the table never covered.

Both new cases pin Zanzibar to a real height first. It defaults to `MaxUint64`, so the table's usual "successor + 1" wraps to **0** and trips an earlier case instead of the one under test — which is exactly what happened on the first attempt here.

## Hash neutrality

Fork heights are not part of the genesis hash. `Genesis.Hash()` covers eight consensus parameters plus accounts and delegates. Adding one is hash-neutral; existing chains are unaffected.

## Tests

```
config + blockchain/genesis + testutil    41 passed
e2etest                                  128 passed   (-run 'Staking|Native')
```

## Deployment note

`iotex-bootstrap/genesis_{mainnet,testnet}.yaml` will need the key once the behaviour PRs land — mainnet at `zanzibarHeight`, testnet at a height past activation. Not part of this PR.

### DIFF
--- a/blockchain/genesis/genesis.go
+++ b/blockchain/genesis/genesis.go
@@ -83,6 +83,7 @@ func defaultConfig() Genesis {
 			YapBlockHeight:            48985561,
 			YapBetaBlockHeight:        48985561,
 			ZanzibarBlockHeight:       math.MaxUint64,
+			ZanzibarBetaBlockHeight:   math.MaxUint64,
 			ToBeEnabledBlockHeight:    math.MaxUint64,
 		},
 		Account: Account{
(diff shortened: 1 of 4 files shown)


envestcc:
作废。

`zanzibarBetaHeight` 存在的唯一目的是保护测试网上已经产生的 Zanzibar 历史。master 线上这个功能从未在任何链上激活过,没有需要保护的已提交区块,所以那一层高度是多余的复杂度。

改为按仓库既有惯例给每个修正单独一个 feature flag(见 #4953 新增的 `RequireProfileForHermesMigration` 和 `EmitEraFreezeLog`),均暂挂 `IsToBeEnabled`,实际激活高度后续统一设置。

PR #4990, opened 2026-08-24
fix(staking): require a complete DelegateProfile before Hermes migration

envestcc:
> Stacked on #4989 (`feat-zanzibar-beta-height`). Review that one first; this PR's base will retarget to `rc_2.5.0` once it merges.

## Problem

Being **migrated** and having a **usable commission configuration** are independent facts.

`migrateHermesRewardOptIn` looks only at the reward address. A candidate whose portions were never published in DelegateProfile is opted in anyway — and `poll_snapshot` then writes `_fullCommissionBasisPoints` into its era snapshot, because `CommissionConfigured` is false.

The result is **100% commission**. Its voters receive nothing, and nothing on chain says so: no error, no event, and every view returns successfully.

These candidates never chose any of it. They were migrated, so they never pass through the opt-in flow where a client could have prompted them to publish portions — which is exactly why the mitigation shipped in the wallet does not cover them.

## Measured

TestNet's first era freeze (46,892,881), all four migrated delegates:

```
delegate       blockBps  epochBps  configured  weight       voters get
uuu (rank 1)     10000     10000     false     14,898,927      0%
robotex          10000     10000     false      2,384,539      0%
faucet           10000     10000     false      1,635,067      0%
iotextest12       1500      1400     true             303    85/86%
```

**Three of four at 100%**, one of them the highest-weighted delegate on the network. `iotextest12` confirms the commission arithmetic itself is fine (`10000−8500`, `10000−8600`) — the defect is that a *missing* configuration is silently read as *100% commission* rather than as *not configured*.

**Mainnet exposure: 90 candidates** whose reward address points at a Hermes vault, with their portions living in the Hermes service rather than in DelegateProfile.

## Fix

From ZanzibarBeta on, migrate only candidates whose profile holds **both** portions. The rest stay opted out and keep the Hermes off-chain payout they are already using — the outcome they expect — instead of moving into an on-chain path that pays their voters zero.

**Half a configuration counts as none.** TestNet's `uuu` had published `blockRewardPortion` and not `epochRewardPortion`, and was frozen at 100% exactly like the candidates that had published neither. `readOne` already collapses that case, so this reads `Configured` and nothing else.

## Why ZanzibarBeta

The migration runs in the single block at `ZanzibarBlockHeight`, which testnet has passed. Changing what that block did would alter its receipt root and **fork any node that resynced**. Mainnet sets both heights to the same value and gets this at activation.

There is a test asserting pre-Beta behaviour is byte-for-byte unchanged, with a reader that fails the test if it is consulted at all.

## Two deliberate choices

**Filtering is skipped** when the chain has no DelegateProfile contract or no reader was injected. Treating an unreadable contract as *"nobody is configured"* would opt out **every** candidate rather than the misconfigured ones — strictly worse than today.

**The reader is injected, not constructed in staking.** It runs a simulated view call, which lives in `evm`. staking cannot reach poll's copy (poll imports staking), and importing `evm` would add a dependency edge to a package that otherwise only touches state. `chainservice` imports both and wires them together.

## Determinism

Candidates are already sorted by identifier bytes before the reads, and `Bridge.Snapshot` reads in the order given — so every node issues the same calls in the same order, and collapses a failed read to `Configured=false` identically.

## Tests

```
staking + poll + protocol + chainservice   514 passed
e2etest                                    134 passed   (-run 'Staking|Native|IIP59')
```

New coverage: complete profile migrates; block-without-epoch does not; no profile does not; pre-Beta unchanged (reader must not be touched); and both no-contract / no-reader paths migrate everyone rather than nobody.

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -206,6 +206,18 @@ type (
 		// for such slots, causing EIP-2200 SSTORE dynamic gas to misclassify
 		// dirty in-place writes as SSTORE_RESET (100 → 2900 gas overcharge per hit).
 		CorrectPrestateForAbsentKeys bool
+		// RequireProfileForHermesMigration, when true, makes the fork-block
+		// Hermes opt-in migration skip candidates whose DelegateProfile portions
+		// are missing or only half set, leaving them on the off-chain Hermes
+		// payout instead of moving them into an on-chain path that would freeze
+		// them at 100% commission and pay their voters nothing.
+		//
+		// Gated on ZanzibarBeta rather than Zanzibar because the migration runs
+		// in the single block at ZanzibarBlockHeight, which testnet has already
+		// passed; changing what that block did would alter its receipt root and
+		// fork any node that resynced. A chain that has not activated sets both
+		// heights to the same value and gets this from the start.
+		RequireProfileForHermesMigration bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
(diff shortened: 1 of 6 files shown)


envestcc:

PR #4991, opened 2026-08-24
feat(staking): emit a DelegateRewardFrozen log at the era freeze

envestcc:
> Stacked on #4990, which is stacked on #4989. Base retargets as those merge.

## Problem

Freezing an era is **completely silent** on chain. The freeze block and its neighbours carry the same single untopiced block-reward log, so an event-driven indexer cannot tell that an era was frozen, which delegates are in it, or what commission each was frozen at.

Measured on TestNet:

```
chain state                            analyser observed
uuu/robotex/faucet/iotextest12         delegate_reward_config      0 rows
all optedIn=true                       voter_reward_era            0 rows
all with freezeHeight=46892881         voter_reward_distribution   0 rows
                                       voter_reward_destination    0 rows
```

The indexer ran across both the freeze and the settlement heights **without a single error**. It followed the three IIP-59 events faithfully; none of them had occurred.

## Fix

One log per frozen delegate: `era`, `delegate`, `freezeHeight`, both commission rates, `totalWeight`, `selfStakeBucketIdx` — and `commissionConfigured`.

That last field is the one that matters most. A delegate that published no portions is frozen at `10000/10000`, which **by value is indistinguishable** from one that deliberately takes everything — and the two mean opposite things to a voter deciding where to stake. The bool is the only discriminator, and it is exactly what an indexer needs in order to warn.

## Why a separate ABI package

Rather than an entry in `action/native_staking_contract_abi.json`. That file is hand-kept in sync with `native_staking_contract_interface.sol` with no Makefile target to catch drift — and the two have **already** diverged: `setVoterRewardOptIn` and `VoterRewardOptInSet` are in the JSON and absent from the `.sol`. This follows the `distributedlog` precedent instead.

## Plumbing

`FreezeCandidateRewardSnapshots` now returns logs, threaded up through `freezeIIP59RewardState` → `setCandidates` → `handle`, which already builds a receipt and simply never populated `Logs`. The four `CreateGenesisStates` callers of `setCandidates` discard them — a freeze cannot occur at height 1.

## Determinism

Logs are emitted **inside the loop over `frozen`**, which is already sorted by identifier bytes. The candidate center enumerates from a Go map, so without that ordering the log sequence would differ between nodes and the receipt root with it.

## Why ZanzibarBeta

Receipt logs are part of the receipt root. Testnet has already produced freeze blocks without these logs; adding them to those heights retroactively would change their receipt roots and fork any node that resynced.

The feature read is the non-panicking `GetFeatureCtx`. This function is called directly by tests with a bare context, and "no feature context" can only mean "not gated on", never "emit" — production always arrives through poll's `handle`, which builds one.

## Not included

An event for the **fork-block Hermes migration**. `CreatePreStates` returns only an error and runs before `withActionCtx`, so there is no log channel and `MustGetActionCtx` would panic there. Giving it one means widening `PreStatesCreator` across 8 implementations and 3 workingset call sites, or moving the migration into a system action. Both are worth a separate discussion; neither belongs here.

Practically, testnet's migration has already happened and cannot be backfilled either way. Mainnet's can be reconstructed by a one-off state sweep at activation.

## Tests

```
staking + poll + protocol + chainservice   660 passed
e2etest                                    134 passed   (-run 'Staking|Native|IIP59|Poll')
```

New: signature/ABI agreement (the topic0 consumers filter on is derived from `EventSignature`, so the constant and the ABI must not drift), pack/unpack round-trip, the unconfigured-vs-100% distinction, and nil handling.

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -218,6 +218,22 @@ type (
 		// fork any node that resynced. A chain that has not activated sets both
 		// heights to the same value and gets this from the start.
 		RequireProfileForHermesMigration bool
+		// EmitEraFreezeLog, when true, makes the era freeze emit one
+		// DelegateRewardFrozen log per frozen delegate.
+		//
+		// Freezing is otherwise completely silent: the freeze block and its
+		// neighbours carry the same single untopiced block-reward log, so an
+		// event-driven indexer cannot tell that an era was frozen, which
+		// delegates are in it, or what commission each was frozen at. On TestNet
+		// the analyser ran without error across the freeze and settlement heights
+		// and left all four of its tables empty, because none of the three IIP-59
+		// events it follows had occurred.
+		//
+		// Receipt logs are part of the receipt root, so this is gated on
+		// ZanzibarBeta rather than Zanzibar: testnet has already produced freeze
+		// blocks without it, and adding logs to those heights retroactively would
+		// change their receipt roots.
+		EmitEraFreezeLog bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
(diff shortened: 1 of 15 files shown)


envestcc:

PR #4996, opened 2026-08-26
genesis,protocol: introduce Zanzibar and Zanzibar Beta hardfork heights

envestcc:
Moves the eight feature flags off the shared WIP gate onto two named heights. Both default to `math.MaxUint64`, so nothing activates until the real heights are filled in.

## Which flag goes where

Not a judgement call. `v2.5.0-rc0` already cut Zanzibar and testnet has run it, so **Zanzibar is fixed by what that release activated**. rc0 gated exactly five (`v2.5.0-rc0:action/protocol/context.go:383-387`):

| flag | fork |
|---|---|
| `FixInContractTransferLogTopic` | Zanzibar |
| `CorrectPrestateForAbsentKeys` | Zanzibar |
| `NoVoterRewardDistribution` (inverted) | Zanzibar |
| `EnforceBLSPoP` | Zanzibar |
| `OptionalCandidateBLSPublicKey` | Zanzibar |
| `FixEpochSettlementFaultHandling` | **Zanzibar Beta** |
| `RequireProfileForHermesMigration` | **Zanzibar Beta** |
| `EmitEraFreezeLog` | **Zanzibar Beta** |

The three on Beta all landed after rc0 and each corrects behaviour Zanzibar already turns on. Selecting them with Zanzibar would rewrite the semantics of blocks testnet has committed — a node with the code and one without would disagree at the same height with no fork boundary between them. Their own doc comments already said they needed a height of their own.

## Also moved

These keyed on the same WIP gate and would otherwise start answering for whatever WIP feature reuses it next:

- `staking.CreatePreStates` — the one-shot owner-index backfill and Hermes opt-in migration, run in the block IIP-59 activates in
- `chainservice.validateIIP59EraMargin` — rejects an activation height past the first freeze
- `Genesis.validate` — the IIP-59 constraints (Greenland/Xingu ordering, `epochsPerRewardEra >= 2`, `delegateProfileContractAddress` required)

`validate` gains one more rule: Zanzibar Beta may not precede Zanzibar.

Test fixtures set both heights to the value they used to give the WIP gate, which reproduces the old behaviour exactly — all eight flags flipped together before, and they still do at those fixtures' heights.

## Height assignment rule

**On any chain that has not activated Zanzibar yet — mainnet — set `zanzibarBetaHeight` equal to `zanzibarHeight`.**

A later Beta height is not a free choice there. It costs two things:

1. It schedules a window in which the chain knowingly runs behaviour already found to be wrong: a node-local fault picking a voter's payout destination, an era boundary with no window taking the whole epoch grant down with it, and an epoch-grant fault settling a `Failure` receipt that other nodes do not settle.
2. **One correction cannot take effect at all.** `RequireProfileForHermesMigration` guards `migrateHermesRewardOptIn`, which runs in exactly one block — the one at `ZanzibarBlockHeight` — and never again. A Beta height later than that block means the guard reads `false` there, the migration runs unguarded, and no later height brings it back. The flag is then permanently inert.

Testnet is the exception, and it is forced: it activated Zanzibar from rc0, whose code had no such flag, so that migration has already run unguarded and cannot be re-run. Beta must be later there because Zanzibar is already behind it. That is also why `validate()` only requires `zanzibarBetaHeight >= zanzibarHeight` rather than equality.

The rule is stated on the `ZanzibarBetaBlockHeight` field and on the `RequireProfileForHermesMigration` flag, so it is in front of whoever assigns the heights rather than only here.

## Known gap — needs real testnet state, not a guess

Two post-rc0 fixes in the BLS PoP work still ride `EnforceBLSPoP` and therefore land on Zanzibar:

1. **`HasBLSPubKeyOtherThan`** (the order-independent duplicate-key check). Gating it on Beta means restoring the map-iteration-order lookup it replaced, so it is worth first confirming whether any two testnet candidates ever shared a BLS pubkey across a post-Zanzibar register/update. If none did, old and new agree on that history and no gate is needed.
2. **`rejectPrePoPForkPop`**, which fires only while `EnforceBLSPoP` is off. A Beta gate cannot help — Beta implies Zanzibar implies the flag is on, so the branch would be dead code. It changes pre-Zanzibar validation, which on testnet is history; whether that matters depends on whether any pre-Zanzibar block carries a non-empty `blsPop`.

## Verification

`go vet ./...` clean, `gofmt` clean, 31 unit packages pass, full e2e suite passes.

One pre-existing failure unrelated to this change and reproducible on `master`: `blockchain/blockdao` / `TestBlockIndexerChecker_CheckIndexer`.


guo (dismissed):
Reviewed the current head. The hardfork height definitions and compatibility behavior are correct. Approved; conflict resolution and final CI remain required before merge.

guo (approved):
Reviewed the current aggregate head, including the conflict resolution and merged fixes. The implementation and integration checks pass.

PR #4997, opened 2026-08-27
feat(genesis): testnet-only balance grant at a scheduled height

envestcc:
## Problem

A test network that has lost the owner keys of its delegate set cannot recover on its own. The old delegates' stake is unreachable — every path that could move it (`CandidateUpdate`, `CandidateTransferOwnership`, unstake, `CandidateActivate`) requires an owner signature — and no remaining account holds enough balance to register replacements. Registering a full 24-delegate set needs roughly 28.8M IOTX of self-stake.

Neither existing mechanism helps:

- **`genesis.initBalances`** runs only at height 0 (`account.Protocol.CreateGenesisStates` asserts it, and `createAccount` rejects addresses that already exist). Changing it changes the genesis state *and* `Genesis.Hash()`, which is the p2p network identity — that is starting a new chain, not continuing one.
- **The staking patch store** (`stakingPatchDir/{height}.patch`) only feeds the candidate name/operator/owner maps into the in-memory view at `Protocol.Start()`, on the legacy pre-Greenland path where `candsMap` is not in the state DB. It writes no state and touches no balances.

`chain.trieDBPatchFile` *can* do it — it PUTs arbitrary bytes into the `Account` namespace at a given height — but it writes past the state manager (so the Erigon secondary store never sees it, and archive nodes diverge), it overwrites the whole account object rather than adding to it, and it is an out-of-band file every operator has to place by hand.

## What this adds

`genesis.Account.TestnetGrants`: a list of `{height, recipients}` applied by a new `account.Protocol.CreatePreStates`, before any action in that block.

```yaml
account:
  testnetGrants:
    - height: 44500000
      recipients:
        - address: io1...
          amount: "1200100000000000000000000"   # Rau
        - address: io1...
          amount: "1200100000000000000000000"
```

Credits go through the state manager (`LoadOrCreateAccount` → `AddBalance` → `StoreAccount`), so they reach the Erigon secondary store as well as the trie, and they **add** to an existing account rather than overwriting it — nonce and prior balance are preserved. Delegate registration then proceeds through the ordinary `CandidateRegister` path with real keys, real receipts and real transaction logs.

`Recipients` is a list rather than a map because the apply order is part of consensus.

## Mainnet is gated twice

1. **Genesis hash.** `ValidateTestnetGrants` refuses a config whose `Hash()` equals the pinned mainnet genesis hash. The grant list is *deliberately* kept out of the fields `Hash()` covers — that is what makes the gate work, because a mainnet genesis file still hashes to that value after grants are appended to it and is still recognised. Putting grants into the hashed set would let a doctored mainnet file hash to something else and slip past. `TestMainnetGenesisHashIsPinned` asserts `New("")` still hashes to the constant, so the gate cannot silently stop guarding.
2. **Chain ID.** `chainservice` refuses to build a node whose chain ID is `1` or whose EVM network ID is `4689`. This catches what the first gate misses: a config derived from mainnet's with a hashed field edited.

The build-time check re-runs the genesis validation, because a `Genesis` built as a literal never reaches `genesis.New()` — `server/main.go` assigns `cfg.Genesis` *after* `config.New` has run its validators.

`ValidateTestnetGrants` also rejects height 0, non-increasing heights, empty recipient lists, duplicate addresses, and unparsable or non-positive amounts.

There is no cap on the amount. An earlier revision had one at 1e9 IOTX; it was removed as an invented policy number that bought nothing — a grant lands through a coordinated hard fork with a reviewed genesis file, and a wrong-but-under-cap amount is just as bad as one over it. What is left is the amount's **value domain**, which is not a policy question: the Erigon secondary store holds balance as a uint256 and writes it with `uint256.MustFromBig`, which *panics* past 2^256. Nothing guarded that before, because every other path that moves balance conserves supply, so no account could approach the bound — a grant mints, so it is the first way a configured number can cross it, and a panic in the account write takes down every node at the activation height rather than rejecting one block. So:

- `ValidateTestnetGrants` rejects an amount wider than 256 bits — a startup error.
- `CreatePreStates` rejects a credit whose *resulting* balance is over the bound. The balance a grant lands on is only known at the activation height, so config validation cannot see this case; here it degrades to a rejected block instead of a crash.

Guarding against a wrong-but-representable amount is left to review of the genesis file, which is where it belongs.

## Rollout is a hard fork

A grant is part of the delta state digest, so every node needs the same genesis file before the activation height. Keeping the grant out of the genesis hash is what makes the failure **loud rather than silent**: a node that missed the update stays on the same p2p network, follows the chain normally, and then stops at the activation height with `ErrDeltaStateMismatch`. It does not fork quietly. `TestTestnetGrant/node_missing_the_grant_rejects_the_block` pins exactly that.

## Known limitation

There is no transaction log — nothing spends, so there is no sender to record. Indexers that derive balances from transfer logs alone will not see the credit. This is noted in the code and is the one externally visible cost of the mechanism.

## Tests

- `blockchain/genesis/testnet_grant_test.go` — pinned mainnet hash, mainnet refusal, 12 validation cases, configured order preserved, **YAML round-trip** (the tags actually parse, and a bad file is rejected at `New()` rather than at the activation height), and that grants do not change the genesis hash.
- `action/protocol/account/testnet_grant_test.go` — fresh account, existing account (balance added, nonce preserved), no-op at non-grant heights, malformed config returns an error instead of panicking mid-block, a credit that overflows the balance bound is rejected while one landing exactly on it is not, and that the protocol is actually picked up as a `PreStatesCreator`.
- `chainservice/testnet_grant_test.go` — the two gates across 6 combinations.
- `e2etest/testnet_grant_test.go` — **two nodes**. One mints, the other independently `ValidateBlock`s and commits, recomputing the delta state digest from its own working set. Plus the divergence case above, and that a recipient can spend the credited balance in a later block.

Verified the e2e test is load-bearing: stubbing `CreatePreStates` to a no-op makes it fail.

### DIFF
--- a/action/protocol/account/protocol.go
+++ b/action/protocol/account/protocol.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/iotexproject/iotex-core/v2/action"
 	"github.com/iotexproject/iotex-core/v2/action/protocol"
+	accountutil "github.com/iotexproject/iotex-core/v2/action/protocol/account/util"
 	"github.com/iotexproject/iotex-core/v2/blockchain/genesis"
 	"github.com/iotexproject/iotex-core/v2/pkg/log"
 	"github.com/iotexproject/iotex-core/v2/state"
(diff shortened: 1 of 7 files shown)


guo (approved):
Reviewed the uint256 boundary checks and runtime overflow handling. The implementation and tests look good.

PR #5003, opened 2026-09-01
fix(blockchain): validate header gas fields against receipts

envestcc:
## What this changes

The block header carries `gasUsed` and `blobGasUsed`. The proposer derives both from the execution receipts in `CreateBuilder`, but validators accept whatever the header says without recomputing them. This adds the validator-side check.

`workingSet.ValidateBlock` now recomputes both values from `ws.receipts` and compares them against the header, immediately after the existing receipt-root check. It reuses `calculateGasUsed` and `calculateBlobGasUsed`, the same helpers the proposer uses, so the two sides cannot drift.

Two sentinel errors are added alongside `ErrReceiptRootMismatch`: `ErrGasUsedMismatch` and `ErrBlobGasUsedMismatch`, with `Header.VerifyGasUsed` / `Header.VerifyBlobGasUsed` mirroring the existing `VerifyReceiptRoot` helper.

## Fork gating

Rejecting a block that would previously have been accepted changes block acceptance, so the check is gated on a new `FeatureCtx.ValidateHeaderGasUsed` flag wired to `IsZanzibarGamma(height)`. It was found after Zanzibar Beta had been scheduled, so folding it into Beta would rewrite the semantics of blocks a chain that activated Beta has already committed -- the same argument that gave Beta a height of its own, one level up. Gamma is `math.MaxUint64` everywhere today, so this is inert: below the gate the validator behaves exactly as before and historical blocks replay unchanged.

The inner conditions mirror `CreateBuilder` exactly. The `gasUsed` comparison runs under `EnableDynamicFeeTx` and the `blobGasUsed` comparison under `EnableBlobTransaction`, because the proposer only writes those fields under the same flags. The validator therefore never demands a field the proposer never wrote.

## Tests

**`TestWorkingSet_ValidateBlock_HeaderGasFields`** -- table driven, using a real mint/validate round trip rather than synthetic receipts. A block is minted through `stateDB.Mint`, rebuilt and re-signed with tampered header gas fields via `block.NewBuilder`, then validated on a separate factory instance so the working-set cache cannot short-circuit it. Pre-gate and post-gate configurations differ only in the Zanzibar heights.

- pre-fork gas used mismatch accepted
- pre-fork blob gas used mismatch accepted
- post-fork matching header accepted
- post-fork gas used mismatch rejected
- post-fork blob gas used mismatch rejected

**`TestWorkingSet_ValidateBlock_HeaderGasFieldsForkBoundary`** -- the case above compares a gate that is off against one that has been on since height one, which never crosses an activation height. This one mints at `fork - 1` and at `fork` against a single config and a single pair of factories, applying the same tampering to both, so the height the block lands on is the only thing that differs between accepted and rejected. Reaching the second height means committing the first block, so the fixture runs on a real key-value store (the in-memory one cannot range scan) and threads the tip through: `Mint` derives the height it is asked for from the tip it is handed, and EIP-1559 header verification derives the next base fee from the parent, so the tip carries the whole gas picture rather than just a height and a hash.

**`TestHeaderVerifyGasUsed`** (in `blockchain/block`) -- direct unit tests for the two comparisons, including that a header which never carried either field agrees with a zero recomputation. The proposer only writes those fields under their own feature gates, so that is the case a validator meets first.

**`TestCalculateGasUsedFromReceipts`** (in `state/factory`) -- unit tests for the two sums the validator re-derives, covering the non-zero blob sum that a header carrying blob gas has to agree with. The end-to-end cases only reach the blob comparison in the mismatching direction, since minting a block that genuinely carries blob gas needs a blob transaction with a sidecar; this covers the positive direction of the same arithmetic.

**`TestHeaderGasFieldsThroughBlockchainValidate`** (in `e2etest`) -- drives the mismatch through `blockchain.ValidateBlock`, the entry point a node uses for a block it received, rather than through the state factory alone. It is the end of the path the error has to travel: `workingSet.ValidateBlock` -> `stateDB.Validate` -> `block.Validator` -> `blockchain.ValidateBlock`. The untampered block is validated first, so that a gas-field verdict means something.

Each case was checked by neutering the gate and confirming the post-fork assertions flip to failing while the pre-fork ones keep passing.

`go build ./...`, `go vet`, and `go test` pass on `state/factory`, `action/protocol`, `blockchain/block`, `blockchain`, `chainservice` and `e2etest`.

## For reviewers

**Fork height.** Gamma is unscheduled. Per AGENTS.md the height needs maintainer sign-off, and assigning it is what lets a fork-at-height devnet run attribute behaviour to this batch rather than to the eight unrelated WIP features that used to share a gate with it.

**Lockstep requirement.** This makes proposer and validator permanently coupled on how `header.gasUsed` is derived. Both call `calculateGasUsed(receipts)` today. Any future change to the proposer side must update the validator in the same fork, where previously only the proposer mattered.

**Related.** A sibling branch changes how stake migration reports receipt gas. Both sides here sum the same `receipt.GasConsumed`, so the two move together, but the branches touch the same two files and will need a trivial merge.

---

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -464,9 +464,10 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			FixEpochSettlementFaultHandling:  g.IsZanzibarBeta(height),
 			RequireProfileForHermesMigration: g.IsZanzibarBeta(height),
 			EmitEraFreezeLog:                 g.IsZanzibarBeta(height),
-			// Same batch: a correction that did not exist in rc0, so it rides
-			// Beta for the same reason the three above do.
-			ValidateHeaderGasUsed: g.IsZanzibarBeta(height),
+			// Zanzibar Gamma: found after Beta was scheduled, so folding it into
+			// Beta would rewrite blocks a chain that activated Beta has
+			// already committed.
+			ValidateHeaderGasUsed: g.IsZanzibarGamma(height),
 		},
 	)
 }
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -260,6 +260,22 @@ type (
 		// An update that omits both leaves any previously registered key
 		// untouched.
 		OptionalCandidateBLSPublicKey bool
+		// ValidateHeaderGasUsed makes a validating node recompute the header's
+		// gasUsed and blobGasUsed from the receipts it produced while executing
+		// the block, and reject the block when either disagrees with the value
+		// the proposer put in the header.
+		//
+		// The proposer fills both fields from its own receipts (see
+		// workingSet.CreateBuilder), but until now nothing on the validating
+		// side re-derived them, so the two header fields were accepted as
+		// given. gasUsed feeds the EIP-1559 base fee of the next block and
+		// blobGasUsed feeds its excessBlobGas, and both are served over the
+		// API, so they need to agree with the executed block.
+		//
+		// Needs its own height: a chain that has already committed blocks whose
+		// header gas fields do not match their receipts must keep accepting
+		// them on replay, so the check may only start at a fork boundary.
+		ValidateHeaderGasUsed bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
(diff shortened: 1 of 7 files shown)


guo (approved):
Reviewed the current head. The header gas-field validation and fork gating are correct. Approved; hardfork scheduling, stack cleanup, and final CI remain required before merge.

PR #5004, opened 2026-09-01
fix(staking): correct migrate-stake receipt gas accounting

envestcc:
## What this changes

Three related corrections in how stake migration reports gas, how an underfunded migration fails, and how the block gas budget consumes a receipt.

**1. Inner call budget.** A `MigrateStake` charges its own intrinsic gas and then runs a staking contract call. The contract call was budgeted with the full gas limit of the action, and the receipt then reported the sum of both, so the reported total could exceed the limit the action itself declared. The block gas budget is reserved against that declared limit, so such a receipt reports gas the block never had. The inner call is now budgeted with `elp.Gas() - insGas`, which bounds the reported total by the declared limit and matches what `estimateMigrateStakeGasConsumptionAt` has always quoted (contract gas + intrinsic gas).

**2. A call that never starts.** The EVM refuses a call whose gas does not cover what the call costs just to begin, and reports that as a plain error rather than a failed receipt. `ExecuteContract` propagates it, a plain error out of a staking handler is not a `ReceiptError`, so it leaves `handle()` as an error and `validateAndRun` routes it to its `default` case, which abandons the block draft.

There are two such refusals, at different gas limits. The `stake0(uint256,address)` call packs 68 bytes, so it needs `10,000 + 68*100 = 16,800` intrinsic gas (`evm.go:706`) and, once Prague is active, an EIP-7623 data floor of `10,000 + 68*250 = 27,000` checked after it (`evm.go:722`). `MigrateStake` intrinsic gas is 10,000, and admission only requires the action to cover that.

So `[10000, 27000)` already costs a block draft on the current code. Change 1 hands the call the declared limit less the migration's own 10,000, which moves that band up to `[10000, 37000)` rather than creating it.

Rather than recompute either bound in the handler, `handleStakeMigrate` reads the verdict off the error the call returns and settles an `ErrOutOfGas` receipt for it. The arithmetic behind both bounds lives in the EVM, is gated on its own fork, and has grown once already; a handler that recomputed it would go stale without anything failing loudly. `ErrInsufficientFunds` is in the set alongside `ErrIntrinsicGas` and `ErrFloorDataGas` because the EVM currently reports the intrinsic shortfall with it as well as a genuine balance shortfall -- either way the call never ran, and neither should cost a draft. #5007 narrows that first one. Gated with change 1, so the pre-gate bands keep their current behaviour.

**3. Budget subtraction.** The remaining block gas budget is an unsigned counter and the per-receipt subtraction was unchecked, so a receipt reporting more gas than was left wrapped the counter and the rest of the block ran as if the budget were untouched. The four subtraction sites in `workingset.go` now route through a `deductBlockGas` helper that saturates at zero. Every node lands on the same remaining budget, and the next action is then rejected for want of gas rather than admitted.

## Fork gating

All of it is consensus data: the EVM budget, the receipt gas and status, and which actions a block may carry. Changes 1 and 2 are gated on `CorrectStakeMigrationGas`, change 3 on `CheckedBlockGasDeduction`, both wired to `IsZanzibarGamma(height)`. They were found after Zanzibar Beta had been scheduled, so folding them into Beta would rewrite the semantics of blocks a chain that activated Beta has already committed -- the same argument that gave Beta a height of its own, one level up. Below the gate the arithmetic and the error handling are byte-for-byte what they were, so historical blocks replay identically.

## Tests

`TestHandleStakeMigrateGasAccounting` pins the inner call budget and the reported receipt gas, pre- and post-gate.

`TestHandleStakeMigrateUnderfundedContractCall` covers change 2 against both refusals, since they sit at different gas limits: 12,000 for the intrinsic shortfall and 20,000 (pre-gate) / 30,000 (post-gate) for the data floor. Pre-gate the handler reports a plain error wrapping the matching sentinel; post-gate it settles an `ErrOutOfGas` receipt bounded by the declared limit. The execution stub mirrors both real checks in the order the EVM applies them, so the pre-gate paths exercise the same error classes the EVM produces.

`TestWorkingSet_RemainingBlockGas` covers change 3, with a `gasReportingProtocol` post-action-handler stub. Note this test must set `VanuatuBlockHeight = 1`: with the default test genesis, `Process` dispatches to `processLegacy`, which carries no per-action gas budget, and the test would silently exercise nothing.

Each change was verified in both directions, by neutering it and confirming the relevant subtest flips back to failing while the pre-gate subtest keeps passing.

`gofmt` clean. `go build ./...` and `go vet` pass. `go test -count=1` passes across `action/protocol`, `action/protocol/staking/...` and `state/factory/...` (669 tests, 14 packages).

## For reviewers

**Fork heights.** Both flags ride Zanzibar Gamma, which is unscheduled. They are logically independent, and if they are ever split, `CheckedBlockGasDeduction` has to activate at or before `CorrectStakeMigrationGas` so the guard is in place when the reporting change lands. On one shared height that ordering holds trivially. Per AGENTS.md the height needs maintainer sign-off.

**Receipt status choice.** Change 2 settles `ReceiptStatus_ErrOutOfGas`. It is the closest existing status for a gas shortfall; there is no dedicated intrinsic-gas status. Worth confirming that reads correctly for integrators.

**`ErrInsufficientFunds` at `evm.go:706` is a misnomer** and is left alone here. That line checks `remainingGas < intriGas` and touches no balance, while the branch below it does the real funds check. #5007 renames it and decides where both refusals belong in the `validateAndRun` switch; this PR only stops them costing a draft on the migration path, and does not depend on that one landing.

**Saturate or reject.** Change 3 saturates at zero rather than returning an error at the subtraction. It is deterministic and uniform, and the next action then fails with `ErrGasLimit`. The one gap: if the over-reporting action is the last in the block, the block is accepted with a zero remaining budget. If the intended policy is to always fail the block, that is a different call and should be stated explicitly.

**Fee change on the failure path.** On the contract-failure branch `gasToBeDeducted` becomes the full `gasConsumed`, so post-gate a failing migration is charged at most the declared limit rather than potentially more. Intended, but user visible, and worth a release note.

**Related.** A sibling branch adds validator-side recomputation of the header gas fields. `calculateGasUsed` sums `receipt.GasConsumed`, and change 1 changes that sum, so post-gate a block containing a `MigrateStake` reports a smaller `header.gasUsed`. Both branches touch `action/protocol/context.go` and `state/factory/workingset.go` and will need a trivial merge.

---

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -260,6 +260,38 @@ type (
 		// An update that omits both leaves any previously registered key
 		// untouched.
 		OptionalCandidateBLSPublicKey bool
+		// CorrectStakeMigrationGas budgets the contract call a stake migration
+		// makes with the action's gas limit minus the migration's own intrinsic
+		// gas, instead of the full gas limit.
+		//
+		// The migration receipt reports the intrinsic gas plus whatever the
+		// contract call consumed. While the call is budgeted with the whole gas
+		// limit, that sum can exceed the gas limit the action declared -- and
+		// the gas limit is what the block gas budget was reserved against, so a
+		// receipt reporting beyond it reports gas the block never had. Taking
+		// the intrinsic gas out of the call's budget bounds the reported total
+		// by the declared limit, which is also what the gas estimator has always
+		// assumed (it quotes contract gas + intrinsic gas).
+		//
+		// Both the budget handed to the EVM and the gas on the receipt are
+		// consensus data, so this needs its own height: a chain that has already
+		// executed migrations under the old budget would recompute different
+		// receipts for those blocks. Wired to IsToBeEnabled until the named
+		// height is assigned.
+		CorrectStakeMigrationGas bool
+		// CheckedBlockGasDeduction subtracts a receipt's gas from the remaining
+		// block gas budget without letting the unsigned counter wrap: a receipt
+		// reporting more gas than is left saturates the budget at zero rather
+		// than rolling it over to near 2^64, which would let the rest of the
+		// block run as if the budget were untouched.
+		//
+		// Saturating is deterministic -- every node lands on the same remaining
+		// budget -- and the following action is then rejected for want of gas,
+		// so the condition is never absorbed silently. Changing the remaining
+		// budget changes which actions a block may carry, so this is gated at
+		// its own height. Wired to IsToBeEnabled until the named height is
+		// assigned.
+		CheckedBlockGasDeduction bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
@@ -448,6 +480,11 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			FixEpochSettlementFaultHandling:  g.IsZanzibarBeta(height),
 			RequireProfileForHermesMigration: g.IsZanzibarBeta(height),
 			EmitEraFreezeLog:                 g.IsZanzibarBeta(height),
+			// Zanzibar Gamma: both were found after Beta was scheduled, so
+			// folding them into Beta would rewrite blocks a chain that
+			// activated Beta has already committed.
+			CorrectStakeMigrationGas: g.IsZanzibarGamma(height),
+			CheckedBlockGasDeduction: g.IsZanzibarGamma(height),
 		},
 	)
 }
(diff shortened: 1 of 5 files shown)


guo (approved):
Reviewed the current head. The migrate-stake gas accounting and gas-shortfall settlement are correct, including floor-data-gas handling. Approved; hardfork scheduling, stack cleanup, and final CI remain required before merge.

PR #5005, opened 2026-09-01
fix(staking): revert state manager writes on failed receipts

envestcc:
## What this changes

`Protocol.Handle` snapshotted only the protocol view. A staking handler that writes to the state manager and then returns a `ReceiptError` settles a **failure receipt**, not an error, so those writes stayed in place under a receipt saying the action did not take effect.

Concrete shape: `handleWithdrawStake` calls `csm.delBucketAndIndex(...)`, a real `DelState`, and only then credits the bucket pool. A rejected credit returns `*handleError{ErrWriteAccount}`, which settles a failure receipt. Nothing undid the delete.

`handle()` now takes `csm.SM().Snapshot()` at its top, past the gate, and reverts to it on the failure paths. A working set snapshots every protocol view alongside its own store, so one snapshot covers both. `Handle` skips its own view snapshot past the gate rather than taking a second one, so the number of view snapshots per action is unchanged. The signature of `handle()` is untouched.

Two rollbacks that were reached only partly are fixed alongside it.

**Bundle abort.** A bundle that gives up part way through is skipped and block building carries on with the same working set, so its rollback is load bearing. The snapshot was taken on the store alone (`ws.store.Snapshot`), so the protocol views kept whatever the dropped bundle had written and the block the proposer went on to build reflected actions it did not include. Such a block computes a state root no validator reproduces, so it is rejected and the slot is lost. The snapshot and the revert now both go through the working set, which covers the store and every view. The pairing matters: view snapshot ids are recorded by `ws.Snapshot()`, so reverting a store-only snapshot through `ws.Revert()` cannot work. Mint is the only caller of `pickAndRunActions` -- validation runs the actions a block already contains -- so this changes what a proposer builds, not what a node accepts, and needs no fork gate.

**settleAction on the success path.** `settleAction` moves the gas fee before it touches the nonce, so a failure inside it can land between the two and leave a half-settled action behind. On the success path the snapshot is still untouched, so the whole action is put back through it. The receipt-failure path deliberately keeps no equivalent: that snapshot has been spent, reverting to the same index twice is rejected (`viewData.Revert` truncates the snapshot list to the index it restores), and covering it would mean taking a fresh snapshot on restored state and settling against that -- the shape that has gone wrong in this codebase before, added to a reachable path for a failure that is not reachable at all. All five of settleAction failure points are internal-failure guards.

## Ordering

The revert runs before `settleAction`, which then deposits gas and bumps the nonce on the restored state, so both survive a failed action. This is asserted in both the pre- and post-gate subtests. Nothing writes state between the handler returning and `settleAction` (only `rLog.Build`, which rides the receipt).

The non-receipt error path reverts too, and that rollback is not merely housekeeping. It abandons the block in the ordinary case, but not inside a bundle: `pickAndRunActions` skips a failed bundle and carries on building with the same working set, so anything left behind there reaches the block the proposer produces. The same commit fixes that path to revert through the working set rather than the store alone (see below).

The nested case was checked: `handleStakeMigrate` reverts to its own inner snapshot and this then reverts to the outer one. `cachedBatch.RevertSnapshot` sets `tag = snapshot+1`, `views.Revert` and `viewData.Revert` both accept a lower index afterwards, and `workingSetStoreWithSecondary.snMap` retains the outer entry.

## Fork gating

What state a failed action leaves behind is part of the state root, so this is gated on a new `FeatureCtx.RevertStakingStateOnFailedReceipt` wired to `IsZanzibarGamma(height)`. It was found after Zanzibar Beta had been scheduled, so folding it into Beta would rewrite the semantics of blocks a chain that activated Beta has already committed -- the same argument that gave Beta a height of its own, one level up. Below the gate the view-only snapshot and revert are kept exactly as they are, so historical blocks replay to identical roots.

## Tests

`handler_failure_revert_test.go` is new, an external `staking_test` package driving a real state factory and working set rather than a mock, which keeps no undo log and so cannot tell a rollback that happens from one that does not.

- **`TestWithdrawStakeFailureLeavesNoPartialState`** -- the reproduction. Pre-gate the withdraw settles a failure receipt and the bucket stays deleted; post-gate the bucket survives untouched while the gas charge and the nonce bump still land.
- **`TestCandidateRegisterPoPFailureLeavesNoBucket`** -- the ordering that makes this matter beyond the withdraw. The register writes the self-stake bucket and its indexes and only then verifies the BLS proof of possession, so a rejected proof would otherwise leave an orphan bucket owned by an address with no candidate record. Enforcement of the proof rides the same height as this rollback, so there is no pre-gate half to compare against and the case is post-gate only.
- **`TestStakeMigrateFailureRevertsNested`** -- the one handler that snapshots on its own. `handleStakeMigrate` reverts to its own inner snapshot when the contract call fails, so the outer snapshot has to compose with it. It runs and passes on both sides of the gate, which is the point: the inner revert already covered this path and adding an outer snapshot must not change that. It is also the plain-error path, so no receipt is settled and neither gas nor nonce moves.
- **`TestWorkingSet_Mint_BundleAbortRevertsView`** (in `state/factory`) covers the bundle rollback, using a protocol view whose snapshot and revert calls are traced through the fork the working set makes, so the test can watch a view that only ever exists inside `Mint`. Without the change the trace shows the two mutations of the dropped bundle and no revert at all.
- **`TestWithdrawStakeSuccessUnaffected`** and **`TestSnapshotCoversStakingView`** pin the success path and the assumption that a state-manager snapshot covers the protocol view, not only the store.

Each case was checked by removing the rollback: the withdraw and register cases then fail, the migrate case is unaffected either way, as described above.

**One gap, deliberately not filled.** There is no isolated test for the plain-error path on a handler that does *not* carry its own inner snapshot. Every plain-error return that sits after a write is an internal-failure guard -- `AddBalance`, `StoreAccount`, ABI packing, a partially deleted bucket index -- and none is reachable through the public action surface. Reproducing one means hand-building internal state keys from an external test package, which would encode the current key layout into a test and rot silently when it changes. The migrate case above exercises the same branch of `handle()`. Happy to add one if a reviewer sees a reachable trigger I missed.

Two small mock adjustments: `initAll` returns the concrete mock type so `testdb.AllowRevert` can be applied to it, and `testdb.AllowRevert(sm)` is added at four sites in the BLS tests that enable the gate and settle failure receipts, since the mock declares no `Revert` expectation by default.

`go build ./...`, `go vet`, and `go test` pass on `action/protocol`, `action/protocol/staking`, `state/factory` and `e2etest`. A full `go test ./...` shows 97 packages ok and 2 failures, both of which reproduce on the clean base tree with the changes stashed (`api.TestEstimateExecutionGasConsumption/FailedToAccountState` and `blockchain/blockdao.TestBlockIndexerChecker_CheckIndexer`); neither touches staking.

## For reviewers

**Scope.** Past the gate this changes what every staking failure receipt leaves behind, not only the one case reproduced above. Seven handlers share the write-then-failure-receipt shape: `handleCreateStake`, `handleUnstake`, `handleWithdrawStake`, `handleChangeCandidate`, `handleDepositToStake`, `handleRestake` and `handleCandidateRegister`. That breadth is the intent, but it is wider than the single reproduction and worth confirming.

Note `handleUnstake` already moved its `updateBucket` from before the self-stake check to after it, gated on `UnstakedButNotClearSelfStakeAmount`. That is the same class of fix applied once, locally, to one handler; this generalizes it.

**Coverage boundary.** The snapshot covers the store and every protocol view. It does not cover writes that bypass both, of which there is one in this package: `recordOwner` writes into `candCenter.base` rather than the dirty change set, and `viewData.Snapshot` records only `size`, `changes`, the pool totals and `contractsStake`. Both call sites sit behind `needToWriteCandsMap`, which requires `CandCenterHasAlias(height)`, i.e. `!IsOkhotsk(height)`. That window closed long ago, and this gate is far above it, so the two are mutually exclusive and no `recordOwner` write can occur where this rollback applies. Worth a source comment so the next reader does not have to re-derive it.

**Activation ordering, and a window that cannot be closed.** `handleCandidateRegister` writes the self-stake bucket and its indexes before it verifies the BLS proof of possession. Once `EnforceBLSPoP` is active, a failed proof leaves that bucket behind, owned by an address with no candidate record but still counted in `totalBucketCount`.

`EnforceBLSPoP` rides Zanzibar and this rollback rides Gamma, so on any chain where Gamma trails Zanzibar the proof is enforced for a while before the rollback covers it. Testnet is already in that position, which means the window is not something scheduling can avoid there -- worth a scan for buckets already left behind. On a chain that has activated none of the family the heights are equal and the window does not exist.

**Fork height.** Gamma is unscheduled and needs maintainer sign-off. Assigning it is also what lets a fork-at-height devnet run attribute behaviour to this batch, rather than to the eight unrelated WIP features that used to share a gate with it.

**Two adjacent issues, not addressed here.** `handler_stake_migrate.go` wraps a `ReceiptError` from `createNFTBucket` in `errors.Wrap`, which destroys the `ReceiptStatus()` method, so a contract failure surfaces as a hard error rather than a failure receipt. And `handleCandidateActivate` returns `SubVote`/`AddVote` errors raw rather than through `csmErrorToHandleError`, making them hard errors too, inconsistent with every sibling handler. Both look like candidates for the same fork.

**Not gated, deliberately.** The bundle rollback is the one change here that carries no feature flag. It only affects what a proposer assembles, and a block built on a polluted view was never acceptable to anyone else, so no historical block can change meaning. Worth a second opinion on that reasoning.

---

### DIFF
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -260,6 +260,26 @@ type (
 		// An update that omits both leaves any previously registered key
 		// untouched.
 		OptionalCandidateBLSPublicKey bool
+		// RevertStakingStateOnFailedReceipt makes a staking action that ends in
+		// a failure receipt discard every state-manager write it made before
+		// the failure, instead of only rolling back the protocol view.
+		//
+		// The handlers mutate the state manager as they go -- a bucket is
+		// written, then the bucket pool is debited, then the caller's account
+		// is stored -- and a verdict raised part way through is reported as a
+		// failure receipt rather than an error. Without this, whatever the
+		// handler had already written stays in committed state under a receipt
+		// that says the action did nothing.
+		//
+		// The rollback runs before gas is deposited and the nonce is bumped, so
+		// a failed action still charges its caller exactly as it does today;
+		// what it no longer keeps is the half-applied state.
+		//
+		// Needs its own height: what state a failed action leaves behind is
+		// part of the state root, so a chain that has already committed such
+		// blocks would recompute different roots for them on replay. Wired to
+		// IsToBeEnabled until the named height is assigned.
+		RevertStakingStateOnFailedReceipt bool
 	}
 
 	// FeatureWithHeightCtx provides feature check functions.
@@ -448,6 +468,10 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			FixEpochSettlementFaultHandling:  g.IsZanzibarBeta(height),
 			RequireProfileForHermesMigration: g.IsZanzibarBeta(height),
 			EmitEraFreezeLog:                 g.IsZanzibarBeta(height),
+			// Same batch: a correction that did not exist in rc0. Note this
+			// lands later than EnforceBLSPoP on any chain where Beta trails
+			// Zanzibar, which is the case testnet is already in.
+			RevertStakingStateOnFailedReceipt: g.IsZanzibarBeta(height),
 		},
 	)
 }
--- a/action/protocol/context.go
+++ b/action/protocol/context.go
@@ -468,10 +468,10 @@ func WithFeatureCtx(ctx context.Context) context.Context {
 			FixEpochSettlementFaultHandling:  g.IsZanzibarBeta(height),
 			RequireProfileForHermesMigration: g.IsZanzibarBeta(height),
 			EmitEraFreezeLog:                 g.IsZanzibarBeta(height),
-			// Same batch: a correction that did not exist in rc0. Note this
-			// lands later than EnforceBLSPoP on any chain where Beta trails
+			// Zanzibar Gamma: found after Beta was scheduled. Note this lands
+			// later than EnforceBLSPoP on any chain where Gamma trails
 			// Zanzibar, which is the case testnet is already in.
-			RevertStakingStateOnFailedReceipt: g.IsZanzibarBeta(height),
+			RevertStakingStateOnFailedReceipt: g.IsZanzibarGamma(height),
 		},
 	)
 }
--- a/action/protocol/staking/handler_failure_revert_test.go
+++ b/action/protocol/staking/handler_failure_revert_test.go
@@ -81,17 +81,24 @@ func testDepositGas(ctx context.Context, sm protocol.StateManager, gasFee *big.I
 // newStakingEnv boots a state factory whose genesis turns the corrected
 // rollback on at toBeEnabledHeight. math.MaxUint64 leaves it off, which is what
 // mainnet runs today.
-func newStakingEnv(t *testing.T, toBeEnabledHeight uint64) *stakingEnv {
-	return newStakingEnvWith(t, toBeEnabledHeight)
+func newStakingEnv(t *testing.T, gateHeight uint64) *stakingEnv {
+	return newStakingEnvWith(t, gateHeight)
 }
 
 // newStakingEnvWith is newStakingEnv with room to adjust the genesis before the
 // factory is built, for cases that need a feature the default heights leave off.
-func newStakingEnvWith(t *testing.T, toBeEnabledHeight uint64, tweaks ...func(*genesis.Genesis)) *stakingEnv {
+func newStakingEnvWith(t *testing.T, gateHeight uint64, tweaks ...func(*genesis.Genesis)) *stakingEnv {
 	r := require.New(t)
 
 	g := genesis.TestDefault()
-	g.ToBeEnabledBlockHeight = toBeEnabledHeight
+	// The rollback rides Zanzibar Gamma. All three heights are set together
+	// because a chain that has activated none of the family carries them
+	// equal; leaving the earlier ones off would be a partial-family genesis,
+	// and the register case below also needs EnforceBLSPoP, which Zanzibar
+	// carries.
+	g.ZanzibarBlockHeight = gateHeight
+	g.ZanzibarBetaBlockHeight = gateHeight
+	g.ZanzibarGammaBlockHeight = gateHeight
 	for _, tweak := range tweaks {
 		tweak(&g)
 	}
--- a/action/protocol/staking/handler_failure_revert_test.go
+++ b/action/protocol/staking/handler_failure_revert_test.go
@@ -15,6 +15,7 @@ package staking_test
 
 import (
 	"context"
+	"crypto/sha256"
 	"math"
 	"math/big"
 	"testing"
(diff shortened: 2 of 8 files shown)


envestcc:
## Why the existing view rollback does not cover this path

Worth spelling out, since the two failure paths in `handle()` behave very differently and only one of them is covered today.

**Hard error.** When a handler returns an error that is not a `ReceiptError`, `handle()` falls through to `return nil, err`. `Handle` rolls the view back and propagates, `runAction` returns the error (`workingset.go:238`), and the block is abandoned either way:

- validating: the block is rejected
- minting: `validateAndRun` hits its `default` case (`workingset.go:1076`) and `pickAndRunActions` returns the error (`workingset.go:967`), abandoning the draft

The mint branches that skip one action and keep building with the same working set (`ErrGasLimit`, `ErrChainID`, `errUnfoldTxContainer`, `errDeployerNotWhitelisted`) all trigger in `runAction` at lines 189-206, before any handler runs, so nothing has been written at that point.

On this path the rollback is therefore not what keeps the caller safe. The working set is discarded regardless.

**Receipt failure.** When the handler returns a `ReceiptError`, `handle()` converts it into a settled receipt and returns `(receipt, nil)`. Because the error is nil:

- `Handle` skips its `view.Revert` entirely
- `runAction` takes the `receipt != nil` branch and breaks out of the handler loop (`workingset.go:245`)
- the action is included in the block with a failure status, and the working set is finalized and committed like any other

Nothing is abandoned here. A failure receipt is an ordinary included action.

`handleWithdrawStake` shows the shape (`handlers.go:272-283`). `delBucketAndIndex` performs three real deletes and returns a plain error on failure, which is the safe path above. The `CreditBucketPool` call immediately after returns `&handleError{failureStatus: ErrWriteAccount}`, which is a `ReceiptError`, so the deletes remain in place under a receipt reporting that the withdraw did nothing.

Two things that follow, both relevant to reviewing this change:

1. The existing `view.Snapshot()` / `view.Revert()` pair fires only on the path where the block is discarded anyway, and does not fire on the path where state is committed. This PR moves the rollback to where it is load-bearing, and leaves the pre-gate pairing untouched.
2. The conversion is deterministic, so every node executes it identically and arrives at the same state and receipt roots. There is no divergence for the receipt-root or state-root checks to detect, which is why this does not surface as a validation failure anywhere today.

guo (approved):
Reviewed the current head. The failed-receipt state rollback and dropped-bundle view rollback are correct. Approved; hardfork scheduling, stack cleanup, and final CI remain required before merge.

PR #5007, opened 2026-09-02
fix(evm): report a gas shortfall as a gas shortfall

envestcc:
## What this changes

`executeInEVM` checks the gas limit handed to a call against the intrinsic cost of that call, and when it falls short returns `action.ErrInsufficientFunds` -- "insufficient funds for gas * price + value". No balance is read on that line; the branch immediately below it is the one that checks whether the sender can pay:

```go
if remainingGas < intriGas {
    return ..., action.ErrInsufficientFunds     // gas, not funds
}
...
if stateDB.GetBalance(ps.txCtx.Origin).Cmp(gasConsumed) < 0 {
    return action.ErrInsufficientFunds          // funds
}
```

`action.ErrIntrinsicGas` ("intrinsic gas too low") already exists and is what `GenericValidator` reports for the same condition at admission. This uses it in the EVM as well.

## The one place that depends on it

`estimateGas` binary searches by simulating with progressively smaller limits, and tolerates the failures that mean the limit was too small rather than reporting them (`coreservice.go:1872` and `:1882`). The intrinsic case was reaching that tolerance through `ErrInsufficientFunds`, so renaming the sentinel without touching the helper would make `eth_estimateGas` fail with an Internal error partway through its own search. `ErrIntrinsicGas` is added to the set, and the helper is renamed from `isInsufficientFundsError` to `isGasTooLowError` to say what the set actually is.

## No behaviour change

That is the whole of it. Before this PR the intrinsic shortfall was tolerated by gas estimation and abandoned the draft in block assembly; after it, the same, by name rather than by accident.

## What was tried and taken back out

Earlier revisions of this branch also moved `ErrIntrinsicGas`, and then `ErrFloorDataGas`, out of the `default` case of `validateAndRun` and in with the permanently-invalid actions that are skipped while the rest of the block is built. The reasoning was that the gas limit is signed into the action, so the call can never start on a retry, and losing a whole draft to it is disproportionate.

That was wrong, and the review that caught it was right. Both refusals are raised *after* `securityDeposit` has already debited the sender: the deposit is taken at `evm.go:677`, the intrinsic check at `:706`, the data floor at `:722`. The `default` case abandons the draft, so that debit went away with the working set. Skipping the action leaves it behind, and the proposer then builds a block whose state digest reflects a debit no action in it explains -- which every other node recomputes without, and rejects. A lost draft traded for an invalid block is not a trade worth making.

Making the skip safe needs the working set rolled back to a pre-action snapshot first. The main mint loop takes none per action today, and correctly so: every case it currently skips (`ErrGasLimit`, `ErrChainID`, `errUnfoldTxContainer`, `errDeployerNotWhitelisted`) is raised in `runAction` at lines 189-206, before any handler runs, so there is nothing to roll back. Adding a per-action snapshot is a change to the hot path -- it costs a store snapshot plus every protocol view, per action -- and deserves its own review and its own mint-plus-second-node test rather than riding along here.

The migration path that motivated the whole thing is handled in #5004 instead, at the handler level, where the rollback is already in hand: `handleStakeMigrate` snapshots before it calls the contract and reverts before settling the receipt.

## Tests

No new ones. The behaviour is unchanged, and the test that covered the skip went out with the skip. The existing gas-estimation cases cover the sentinel by exercising the search path.

`go build ./...`, `go vet` and `go test -count=1` pass across `state/factory`, `action/protocol/execution/...`, `action/protocol` and `actpool` (523 tests, 6 packages).

`api` has two failures, `TestEstimateExecutionGasConsumption` and its `FailedToAccountState` subtest, which reproduce unchanged on a clean `upstream/master` worktree -- a nil dereference in the mock state manager. Because the panic aborts the parent test on both, the sibling subtests never run in a normal invocation, so each was run individually against this branch: `FailedToCheckGasLimitEnough`, `GasLimitNotEnough`, `ExecutionRevertMsgIsNil`, `ExecutionRevertMsgIsNotNil` and `EstimateExecutionGasConsumptionSuccess` all pass. The middle two are the ones that exercise the search path touched here.

## For reviewers

`isGasTooLowError` now covers `ErrInsufficientFunds`, `ErrFloorDataGas` and `ErrIntrinsicGas` -- three different reasons a simulation never got off the ground. If you would rather keep the old name, the behaviour is unaffected; the set has not been only about funds for a while.


guo (dismissed):
Reviewed the gas-shortfall classification and its mint, validation, and API paths. The change is correct and safe to merge.

guo (approved):
Reviewed the current head and revalidated the previously blocked mint-state path. The fix is correct and the focused adversarial tests pass.

PR #5008, opened 2026-09-02
genesis: introduce the Zanzibar Gamma hardfork height

envestcc:
## Why

`ZanzibarBetaBlockHeight` exists because a chain that has already activated Zanzibar is running the pre-correction behaviour, so folding corrections into Zanzibar would rewrite the semantics of blocks that chain has committed. The field says so itself.

Once Beta is activated somewhere, the same argument applies to Beta, and a correction found afterwards needs a height of its own. This adds it.

## What

`ZanzibarGammaBlockHeight`, defaulting to `math.MaxUint64`, with `IsZanzibarGamma`, an ordering rule (Gamma must not precede Beta) alongside the one the family already carries, and a slot in `testutil.NormalizeGenesisHeights`.

A chain that has activated none of them sets all three equal. A later Gamma is only for a chain that has, and it buys the same thing Beta buys at the same price: a window in which the chain knowingly runs behaviour already found to be wrong.

## One window that cannot be closed

Recorded on the field itself, because it is the kind of thing that is expensive to rediscover:

`EnforceBLSPoP` rides Zanzibar, and `handleCandidateRegister` writes its self-stake bucket before it verifies the proof. Wherever Gamma trails Zanzibar, a rejected proof leaves that bucket behind -- owned by an address with no candidate record, still counted in the bucket total -- until Gamma activates. On a chain carrying the heights equal the window does not exist.

## Tests

Neither ordering rule had a test. Both do now, plus the default and the unscheduled case. They sit above the Greenland, Xingu and contract-address requirements in `validate`, so a case that trips one needs nothing else set up.

`go build ./...`, `go vet` and `go test -count=1` pass on `blockchain/genesis`, `action/protocol` and `blockchain/integrity`.

## For reviewers

The name is the open question. Gamma follows Beta, but if the family is meant to stop at two and this batch should carry a name of its own, that is a governance call and the field is trivial to rename while it is unscheduled.

### DIFF
--- a/blockchain/genesis/genesis.go
+++ b/blockchain/genesis/genesis.go
@@ -84,6 +84,7 @@ func defaultConfig() Genesis {
 			YapBetaBlockHeight:        48985561,
 			ZanzibarBlockHeight:       math.MaxUint64,
 			ZanzibarBetaBlockHeight:   math.MaxUint64,
+			ZanzibarGammaBlockHeight:  math.MaxUint64,
 			ToBeEnabledBlockHeight:    math.MaxUint64,
 			PersistStakingPatchBlock:  19778037,
 			FixAliasForNonStopHeight:  19778036,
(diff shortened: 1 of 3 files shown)


guo (approved):
Reviewed the hardfork-height placeholder, defaults, ordering, and compatibility behavior. The change is safe to merge while unscheduled. Please preserve the commit when merging because dependent branches share it.

PR #5014, opened 2026-09-07
fix(consensus): skip a height CommitBlock reports as already committed

### LINKED ISSUE #5010: consensus: CommitBlock on an already-committed height is treated as an error instead of a skip
## Summary

`blockchain/filedao/filedao.go:43` and `blockchain/blockchain.go:51` each declare their own `ErrInvalidTipHeight` variable with the same message. `blockchain.CommitBlock` propagates the `filedao` one when the block store refuses a block whose height is not `tip + 1`, but the callers switch on the `blockchain` one.

## Where it bites

- `consensus/scheme/rolldpos/rolldposctx.go:618-629`: if block sync already committed height H and the consensus round then calls `CommitBlock` for its own block at H, the switch does not hit `case blockchain.ErrInvalidTipHeight` (which returns `true, nil`), falls through to `default`, logs `error when committing the block`, and returns an error. The FSM stays in its pre-commit state until the round TTL expires instead of recognising the height as done.
- `chainservice/builder.go:690` has the same comparison but is masked there because the preceding `ValidateBlock` returns the `blockchain` variable first.

No finality or state impact: the block store guard still refuses the second block. This is a robustness issue that turns an expected "already committed" outcome into an error path.

## Suggested fix

Make `filedao` return `blockchain.ErrInvalidTipHeight` (or have `blockchain` re-export the `filedao` one), or wrap the storage error so `errors.Is` matches both. Add a test that commits H through the block-sync path and then through `rolldposCtx.Commit` and asserts the skip branch.

Found during the CORE-22 finality-reversal review at `30714c91e`.

envestcc:
## Description

`blockchain/filedao/filedao.go` and `blockchain/blockchain.go` each declared their own `ErrInvalidTipHeight` sentinel with the same message. The block store raises the `filedao` one when it refuses a block whose height is not `tip + 1`, and `blockchain.CommitBlock` propagates it verbatim — but callers compare against the `blockchain` one, so the comparison never matches.

In `rolldposCtx.Commit` that turns an expected outcome into an error path: if block sync has already committed height H and the consensus round then calls `CommitBlock` for its own block at H, the switch misses `blockchain.ErrInvalidTipHeight` (which returns `true, nil`), falls through to `default`, logs `error when committing the block`, and returns an error. The FSM then stays in its pre-commit state until the round TTL expires instead of recognising the height as done.

`chainservice/builder.go:690` has the same comparison, but it is masked there because the preceding `ValidateBlock` returns the `blockchain` variable first.

There is no finality or state impact — the block store guard still refuses the second block. This is a robustness fix.

Fixes #5010

## Changes

- `blockchain.ErrInvalidTipHeight` now aliases `filedao.ErrInvalidTipHeight`, so both names refer to a single sentinel and a height rejection from the block store matches what callers switch on. (`blockchain` already imports `filedao`; the reverse direction would be an import cycle.)
- `rolldposCtx.Commit` matches with `errors.Is` instead of `errors.Cause` equality, so the skip branch keeps working if the storage error is ever wrapped, and the skip intent is documented in a comment.

## Tests

- `TestCommitAlreadyCommittedHeight` (`consensus/scheme/rolldpos`): drives the reported scenario end to end on a real chain — the round registers its own block at H, block sync commits a *different* block at H, then the round collects a majority of COMMIT endorsements and calls `Commit`. Asserts `(true, nil)` and that the block-sync block is still the tip. Fails on `master` with `error when committing a block` / `invalid tip height`.
- `TestCommitBlockInvalidTipHeight` (`blockchain`): pins the contract at the layer where it broke — a height rejection from the block store must satisfy `errors.Is(err, blockchain.ErrInvalidTipHeight)`. Fails on `master`.

Full `./blockchain/... ./consensus/... ./chainservice/...` run is green apart from `TestBlockIndexerChecker_CheckIndexer`, which fails identically on unmodified `master`.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -47,8 +47,10 @@ var (
 		},
 		[]string{"type"},
 	)
-	// ErrInvalidTipHeight is the error returned when the block height is not valid
-	ErrInvalidTipHeight = errors.New("invalid tip height")
+	// ErrInvalidTipHeight is the error returned when the block height is not valid.
+	// It aliases filedao.ErrInvalidTipHeight so that a height rejection raised by the
+	// block store on CommitBlock still matches the sentinel callers compare against.
+	ErrInvalidTipHeight = filedao.ErrInvalidTipHeight
 	// ErrInvalidBlock is the error returned when the block is not valid
 	ErrInvalidBlock = errors.New("failed to validate the block")
 	// ErrActionNonce is the error when the nonce of the action is wrong
--- a/blockchain/blockchain_extra_test.go
+++ b/blockchain/blockchain_extra_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/iotexproject/iotex-core/v2/blockchain"
 	"github.com/iotexproject/iotex-core/v2/blockchain/block"
+	"github.com/iotexproject/iotex-core/v2/blockchain/filedao"
 	"github.com/iotexproject/iotex-core/v2/blockchain/genesis"
 	"github.com/iotexproject/iotex-core/v2/test/identityset"
 	"github.com/iotexproject/iotex-core/v2/test/mock/mock_blockchain"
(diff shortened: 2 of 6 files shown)


envestcc:
Thanks for the review. Addressed the two actionable points in 31e12d01a.

**`codecov/patch` (was 40%, 6 lines missing in `rolldposctx.go`)** — real. The paused and generic-error branches of the commit switch were untouched by the one test. Factored the three cases onto a shared fixture and added `TestCommitPausedChain` and `TestCommitBlockError`; every block in the changed range 618–628 now reports a non-zero count in the coverage profile.

**Block-sync callback behavior change lacks direct coverage** — also real, and it was the more interesting gap. The alias unmasks the `blockchain.ErrInvalidTipHeight` case in `chainservice/builder.go`: a height rejection raised by the block store on `CommitBlock` (rather than caught earlier by `ValidateBlock`) now skips instead of returning an error. That path sat in an anonymous closure inside `buildBlockSyncer` with no way to reach it from a test. Extracted it as `commitSyncedBlock`, following the same pattern as `estimateTipHeight`/`blockDistance` in that file, and covered it — including the case that changes behavior, which fails if the alias is reverted.

The extraction is a relocation only. One thing worth a look during review: the original returned `nil` *from the closure* on the skip, so it deliberately did not log "Successfully committed block" or call `consens.Calibrate`. That is preserved by returning `committed=false` and short-circuiting before the `Calibrate` call, rather than a bare `nil`. Classification (`errors.Cause`), retry counts, and the `ErrRemoteHeightTooLow` retry bump are unchanged.

**`codecov/project`** — I don't think this one is on the PR. The report is 343 commits behind head and compares against base `436e4d1`, producing a +124 files / +18486 lines diff and a head coverage of 62.63% against the configured 85% target. It should go green once the base report catches up; happy to rebase if that helps.

Full `./blockchain/... ./consensus/... ./chainservice/...` remains green apart from `TestBlockIndexerChecker_CheckIndexer`, which fails identically on unmodified `master`.

PR #5017, opened 2026-09-14
actpool: fix a hash handling edge case

envestcc:
Tightens the cached-action short-circuit in `actpool.validate` to also match sender. Single-file change in `actpool/actpool.go`; non-hardfork; behavior unchanged for normal traffic.


guo:
Superseded: the identical change landed on master in 213ac0177 via the security-advisory workflow. Thanks for the PR and the review.

PR #4861, opened 2026-06-26
feat(actpool): support removing addresses from blacklist at a future height

envestcc:
## Summary
- Add `BlackListRemoval` + `BlackListRemovalHeight` to `actpool.Config` so specific addresses can be dropped from the blacklist at a designated height without mutating the original `BlackList` (preserves historical SetCode authorization validation in EVM).
- Populate `BlackListRemoval` with the 13 mainnet addresses queued for removal at the next hardfork; `BlackListRemovalHeight` is left at `math.MaxUint64` as a feature gate — flip it to the real height in a follow-up once the next hardfork height is finalized.
- `IsBlackListedFunc` now enforces removal: in `[activeHeight, removalHeight)` the address still hits the blacklist; from `removalHeight` onward the listed addresses are no longer blacklisted.

## Test plan
- [x] `go test ./actpool/...`
- [x] `go test ./action/protocol/execution/... ./blocksync/... ./e2etest/...`
- [x] New unit tests cover: address still blacklisted at `removalHeight-1`, dropped at `removalHeight`, non-removal addresses unaffected, and `removalHeight=MaxUint64` behaves identically to current production config.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5020, opened 2026-09-18
feat(actpool): activate the blacklist removal at Zanzibar

envestcc:
## Problem

#4861 added `BlackListRemoval` together with a node-configurable `BlackListRemovalHeight`, defaulting to `MaxUint64`. As a result, the removal list has never activated by default.

The blacklist predicate is also passed to the execution protocol and used by the EIP-7702 authorization check during block execution. Letting operators configure the removal height can therefore make nodes validate the same authorization differently.

## Change

- remove `BlackListRemovalHeight` from node config
- source the removal height from genesis `ZanzibarBlockHeight`
- pass the genesis height to both actpool admission and execution protocol registration
- cover behavior immediately before and at the fork height
- preserve the unscheduled-fork behavior with an explicit `MaxUint64` test case

TestNet behavior is unchanged because its blacklist is empty. On MainNet, the 13 removal entries remain blacklisted until Zanzibar and are removed starting at block 53155801.

## Testing

- `CGO_ENABLED=1 go test ./actpool ./chainservice -count=1`
- `make lint`
- `make fmt` (the branch itself is clean; the command also reports five pre-existing formatting differences on master)
- `make test`: the changed `actpool` package passes under the race detector; the previously unrelated API test failures have since been addressed by #4973 and #5021


guo (approved):
Reviewed the Zanzibar-gated blacklist removal wiring and fork-boundary behavior. Both actpool admission and execution use the genesis fork height; CI is green.

PR #5021, opened 2026-09-18
test(api): wait for asynchronous log streaming

envestcc:
### Async log-listener assertions

The gRPC stream backpressure change made `gRPCLogListener.Respond` asynchronous: it now enqueues matched logs and returns before the stream handler runs.

`TestGRPCLogListener` still assumed the old synchronous behavior. Its match case appended from the sender goroutine and immediately read the same slice, so CI could observe an empty slice; under the race detector the read and append also race. The stream-error case likewise expected `Respond` to return the handler error even though asynchronous errors are reported through `errChan`.

This was the failure in the first `ci flow` run for #5020: https://github.com/iotexproject/iotex-core/actions/runs/35327207564/job/105542943878

### Duplicate e2e ports

After fixing that assertion, CI reached e2etest and exposed an independent port-allocation race. Consecutive `testutil.RandomPort()` calls could return the same currently-unbound port before either service started listening. In the failed run both gRPC and HTTP received port `30409`; gRPC bound first and HTTP terminated the test process with `address already in use`.

Failed run: https://github.com/iotexproject/iotex-core/actions/runs/35329343181/job/105549798201

## Changes

- receive streamed log responses through a buffered channel
- wait for asynchronous handler completion and assert handler failures through `errChan`
- stop each listener during subtest cleanup
- serialize test-port allocation and remember ports already claimed by this process
- test 512 concurrent port allocations for uniqueness and range

Node production behavior is unchanged; the port allocator is test utility code.

## Testing

- `CGO_ENABLED=1 go test -race ./api -run '^TestGRPCLogListener$' -count=1000`
- `CGO_ENABLED=1 go test -gcflags='all=-N -l' -short -coverprofile=coverage.out -covermode=count ./api`
- `CGO_ENABLED=1 go test -race ./testutil -run '^TestRandomPort$' -count=20`
- `CGO_ENABLED=1 go test -gcflags='all=-N -l' -short ./e2etest -count=1`
- `make lint`
- `make fmt` (the PR branch is clean; the command also reports five pre-existing formatting differences on master)


guo (approved):
Reviewed the asynchronous listener test fix and process-local port allocation changes. Combined validation passed, including repeated race tests.

